PR #2, opened 2021-09-21
mesonpy: add heuristics to try mapping unknown install paths to wheel

FFY00:
This will map files that do not use {py_purelib}/{py_platlib}
placeholders, which is the case of targets that receive the install
location as a string in the Meson build files (eg. configure_file).


FFY00:
Blocked by #9288, as Meson will still apply the hardcoded Debian handling on the Github Actions Python, which does not have the Debian modifications. Meson will try to install to `dist-packages`.

FFY00:
Verified that the logic works when we disable the Debian Python check.

PR #18, opened 2021-12-02
mesonpy: select the correct ABI tag

### LINKED ISSUE #16: Selecting the correct ABI tag
Selecting the ABI tag is not as straight forward as I hoped it could be.

Firstly, there are two main possible ABIs, the interpreter specific ABI, and the stable ABI. The stable ABI uses the `abi3` tag and is compatible across Python minor versions.

Secondly, there are many variants of the interpreter specific ABI, we need to detect that. I think the easier way is to do this is by looking at the extensions of the native modules.

And lastly, we will need to detect the ABI via the file extensions, but this is only standardized on Posix, via [PEP 3149](https://www.python.org/dev/peps/pep-3149). PEP 3149 is optional on Windows, but from what I have seen everyone seems to use it on Windows.

So, supporting this will consist on the following:
1) Identify and parse the ABI from native modules
2) Check incompatible that
3) Identify and select the more specific tag (interpreter tag over the stable ABI)

FFY00:
Fixes #16

### DIFF
--- a/mesonpy.py
+++ b/mesonpy.py
@@ -430,7 +430,7 @@ class _WheelBuilder():
                 import distutils.command.install
             except ModuleNotFoundError:
                 raise ModuleNotFoundError('Unable to import distutils, please install python3-distutils')
-            return 'deb_system' in distutils.command.install.INSTALL_SCHEMES  # type: ignore[attr-defined]
+            return 'deb_system' in distutils.command.install.INSTALL_SCHEMES
         except ModuleNotFoundError:
             return False
 
--- a/mesonpy.py
+++ b/mesonpy.py
@@ -10,6 +10,7 @@ Implements PEP 517 hooks.
 
 from __future__ import annotations
 
+import abc
 import collections
 import contextlib
 import functools
@@ -19,6 +20,7 @@ import json
 import os
 import os.path
 import pathlib
+import re
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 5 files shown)


FFY00:
Okay, the code is done, we have tests, and the CI is finally passing!

rgommers:
Nice!

Naive question: is there anything Meson-specific here? Should this live in `packaging` or some such library in the future?

FFY00:
Not really, there isn't anything Meson specific, we get a file list from Meson and look at the ABI tags. I don't really think this would be useful to a lot of people, but maybe some, as is it very uncommon for a build backend to not be in charge of building the native Python modules.

rgommers:
> as is it very uncommon for a build backend to not be in charge of building the native Python modules.

I'd expect this to be(come) fairly common actually, because build backends are thin shims that are not _that_ much work to write, while a build system dealing with compiled code is quite complicated. Scikit-build is in the same situation I guess, with CMake doing the heavy lifting. Same for enscons/scons, bento/waf, etc. - I'd expect the `setuptools` design to be the exception, not the rule.

FFY00:
Hum, that makes sense, though it would still be quite a limited use-case, so I doubt something like this would be accepted in pypa/packaging. We could, however, extract it to a new package with the goal of helping the development of Python build backends.

rgommers:
Sounds good, but maybe whenever someone needs it rather than right now? Maybe @henryiii has a use for this in skbuild soon?

FFY00:
Yes, that was my plan. It might be useful for https://github.com/FFY00/trampolim too.

PR #22, opened 2022-01-07
mesonpy: force Meson to use the current Python executable

### LINKED ISSUE #23: Meson does not run with the expected Python executable
One of the problems of Meson's handling of Python is that `find_program`, which is also used by Meson Python module, will find whatever Python executable is available, which is not necessarily the one we are using. Because of this, Meson might actually build native modules for a different Python version than we are using, and also there may be some mismatch in the paths returned by Meson and the ones we were expecting.

Currently, our heuristics should be able to work around the path mismatch issue, but that is not a long-term solution. That said, Meson building native modules for a different version is very much an issue and should be fixed.

We need to make Meson use the current Python interpreter.

FFY00:
Fixes #23

### DIFF
--- a/mesonpy.py
+++ b/mesonpy.py
@@ -573,6 +573,7 @@ class Project():
         self._working_dir = pathlib.Path(working_dir).absolute()
         self._build_dir = pathlib.Path(build_dir).absolute() if build_dir else (self._working_dir / 'build')
         self._install_dir = self._working_dir / 'install'
+        self._meson_native_file = self._source_dir / '.mesonpy-native-file.ini'
 
         # load config -- PEP 621 support is optional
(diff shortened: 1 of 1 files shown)


FFY00:
This is broken as specifying a cross file will cause Meson to fail to find pkgconfig, which I don't think should happen.

rgommers:
Could you link to the issues this is working around? Then we'll be able to understand this design decision in the future.

FFY00:
I don't have an issue for that, I should have opened one, sorry!

FFY00:
Opened #23.

FFY00:
It seems like we should be using `--native-file` instead of `--cross-file` to select the Python binary. I have updated the code and now everything works as expected.

Issue #20, opened 2022-01-01
Please add install and basic usage instructions in README


rgommers:
`pip install mesonpy` gives:
```
$ pip install mesonpy 
Collecting mesonpy
  Using cached MesonPy-0.0.1.tar.gz (14 kB)
  Preparing metadata (setup.py) ... error
  ERROR: Command errored out with exit status 1:
   command: /home/rgommers/anaconda3/envs/scipy-dev/bin/python3.9 -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-m_0m1w7a/mesonpy_f4910b5198f8497aa8f6fa1184381e91/setup.py'"'"'; __file__='"'"'/tmp/pip-install-m_0m1w7a/mesonpy_f4910b5198f8497aa8f6fa1184381e91/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-4xq7gbqy
       cwd: /tmp/pip-install-m_0m1w7a/mesonpy_f4910b5198f8497aa8f6fa1184381e91/
  Complete output (5 lines):
  Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "/tmp/pip-install-m_0m1w7a/mesonpy_f4910b5198f8497aa8f6fa1184381e91/setup.py", line 4, in <module>
      with open('requirements.txt') as f:
  FileNotFoundError: [Errno 2] No such file or directory: 'requirements.txt'
  ----------------------------------------
WARNING: Discarding https://files.pythonhosted.org/packages/e5/b6/ff8146b0332f23fc2c242a9e20204ef217249764c06eebac301609b21426/MesonPy-0.0.1.tar.gz#sha256=27a81c9af4f5bdea934bca0ac698a85f44ca6b26a9eaa096dc0f1fbf22108739 (from https://pypi.org/simple/mesonpy/). Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
ERROR: Could not find a version that satisfies the requirement mesonpy (from versions: 0.0.1)
ERROR: No matching distribution found for mesonpy
```

rgommers:
[...]

PR #26, opened 2022-01-10
mesonpy: replace auditwheel with an in-house vendoring mechanism

### LINKED ISSUE #4: Add tests for modules linking to distribution libraries

### LINKED ISSUE #6: Add `skip-external-bundling` option
See https://github.com/pypa/auditwheel/issues/339.

A workaround could be removing the external libraries from the wheel after repairing.

FFY00:
Fixes #4 
Closes #6

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -10,6 +10,7 @@ endif
 py.install_sources(
   'mesonpy/__init__.py',
   'mesonpy/_compat.py',
+  'mesonpy/_elf.py',
   'mesonpy/_tags.py',
   'mesonpy/_util.py',
   subdir: 'mesonpy',
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -117,7 +117,11 @@ class _WheelBuilder():
     _SCHEME_MAP: ClassVar[Dict[str, Tuple[str, ...]]] = {
         'scripts': ('{bindir}',),
         'purelib': ('{py_purelib}',),
-        'platlib': ('{py_platlib}', '{moduledir_shared}'),
+        # {moduledir_shared} should be listed in platlib here too, but currently
+        # there is a Meson bug preventing us from using, so we will have to let
+        # that fallback on heuristics.
+        # see https://github.com/mesonbuild/meson/pull/9474
+        'platlib': ('{py_platlib}',),
         'headers': ('{include}',),
         'data': ('{datadir}',),
         # our custom location
@@ -208,11 +212,16 @@ class _WheelBuilder():
                     return 'platlib', calculated_path
         # purelib or platlib -- go to wheel root
         for scheme in ('purelib', 'platlib'):
+            # try to match the install path on the system to one of the known schemes
+            scheme_path = pathlib.Path(sys_paths[scheme])
+            destdir_scheme_path = self._project._install_dir / scheme_path.relative_to(scheme_path.root)
             try:
-                wheel_path = destination.relative_to(sys_paths[scheme])
+                wheel_path = pathlib.Path(origin).relative_to(destdir_scheme_path)
             except ValueError:
                 continue
-            if sys_paths['purelib'] == sys_paths['platlib']:
+            # {moduledir_shared} is currently handled in heuristics due to a Meson bug,
+            # but we know that files that go there are supposed to go to platlib
+            if sys_paths['purelib'] == sys_paths['platlib'] and not origin.startswith('{moduledir_shared}'):
                 self._warn_unsure_platlib(origin, destination)
             return 'platlib', wheel_path
         return None
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -35,6 +35,7 @@ from typing import (
 import tomli
 
 import mesonpy._compat
+import mesonpy._elf
 import mesonpy._tags
 import mesonpy._util
 
@@ -43,16 +44,17 @@ from mesonpy._compat import Collection, Iterator, Mapping, Path
 
 if typing.TYPE_CHECKING:  # pragma: no cover
     import pep621 as _pep621  # noqa: F401
+    import wheel.wheelfile  # noqa: F401
 
 
 __version__ = '0.1.2'
 
 
 class _depstr:
-    auditwheel = 'auditwheel >= 5.0.0'
     ninja = 'ninja >= 1.10.0'
+    patchelf_wrapper = 'patchelf-wrapper'
     pep621 = 'pep621 >= 0.3.0'
-    wheel = 'wheel >= 0.36.0'
+    wheel = 'wheel >= 0.36.0'  # noqa: F811
 
 
 _COLORS = {
@@ -118,12 +120,14 @@ class _WheelBuilder():
         'platlib': ('{py_platlib}', '{moduledir_shared}'),
         'headers': ('{include}',),
         'data': ('{datadir}',),
+        # our custom location
+        'mesonpy-libs': ('{libdir}', '{libdir_shared}')
     }
-    # XXX: libdir - no match in wheel, we optionally bundle them via auditwheel
 
     """Helper class to build wheels from projects."""
     def __init__(self, project: Project) -> None:
         self._project = project
+        self._libs_build_dir = project._build_dir / 'mesonpy-wheel-libs'
 
     @property
     def basename(self) -> str:
@@ -150,7 +154,7 @@ class _WheelBuilder():
         return f'{self.basename}.data'
 
     @property
-    def wheel(self) -> bytes:
+    def wheel(self) -> bytes:  # noqa: F811
         '''dist-info WHEEL.'''
         return textwrap.dedent('''
             Wheel-Version: 1.0
@@ -253,6 +257,33 @@ class _WheelBuilder():
 
         return wheel_files
 
+    def _install_file(
+        self,
+        wheel_file: wheel.wheelfile.WheelFile,
+        counter: mesonpy._util.CLICounter,
+        origin: Path,
+        destination: pathlib.Path,
+    ) -> None:
+        location = os.fspath(destination).replace(os.path.sep, '/')
+        counter.update(location)
+
+        # fix file
+        if os.name == 'posix':
+            # add .mesonpy.libs to the RPATH of ELF files
+            if self._is_elf(os.fspath(origin)):
+                # copy ELF to our working directory to avoid Meson having to regenerate the file
+                new_origin = self._libs_build_dir / pathlib.Path(origin).relative_to(self._project._build_dir)
+                os.makedirs(new_origin.parent, exist_ok=True)
+                shutil.copy2(origin, new_origin)
+                origin = new_origin
+                # add our in-wheel libs folder to the RPATH
+                elf = mesonpy._elf.ELF(origin)
+                libdir_path = f'$ORIGIN/{os.path.relpath(f".{self._project.name}.mesonpy.libs", destination.parent)}'
+                if libdir_path not in elf.rpath:
+                    elf.rpath = [*elf.rpath, libdir_path]
+
+        wheel_file.write(origin, location)
+
     def build(
         self,
         sources: Dict[str, Dict[str, Any]],
@@ -264,13 +295,12 @@ class _WheelBuilder():
         self._project.build()  # ensure project is built
 
         wheel_file = pathlib.Path(directory, f'{self.name}.whl')
+        wheel_files = self._map_to_wheel(sources, copy_files)
 
         with wheel.wheelfile.WheelFile(wheel_file, 'w') as whl:
             # add metadata
             whl.writestr(f'{self.distinfo_dir}/METADATA', self._project.metadata)
             whl.writestr(f'{self.distinfo_dir}/WHEEL', self.wheel)
-
-            wheel_files = self._map_to_wheel(sources, copy_files)
 
             print('{light_blue}{bold}Copying files to wheel...{reset}'.format(**_STYLES))
             with mesonpy._util.cli_counter(
@@ -279,18 +309,21 @@ class _WheelBuilder():
                 # install root scheme files
                 root_scheme = 'purelib' if self._project.is_pure else 'platlib'
                 for destination, origin in wheel_files[root_scheme]:
-                    location = os.fspath(destination).replace(os.path.sep, '/')
-                    counter.update(location)
-                    whl.write(origin, location)
+                    self._install_file(whl, counter, origin, destination)
+
+                # install bundled libraries
+                for destination, origin in wheel_files['mesonpy-libs']:
+                    assert os.name == 'posix', 'Bundling libraries in wheel is currently only supported in POSIX!'
+                    destination = pathlib.Path(f'.{self._project.name}.mesonpy.libs', destination)
+                    self._install_file(whl, counter, origin, destination)
+
                 # install the other schemes
                 for scheme in self._SCHEME_MAP:
-                    if root_scheme == scheme:
+                    if scheme in (root_scheme, 'mesonpy-libs'):
                         continue
                     for destination, origin in wheel_files[scheme]:
-                        wheel_path = pathlib.Path(f'{self.data_dir}/{scheme}') / destination
-                        location = os.fspath(wheel_path).replace(os.path.sep, '/')
-                        counter.update(location)
-                        whl.write(origin, location)
+                        destination = pathlib.Path(self.data_dir, scheme, destination)
+                        self._install_file(whl, counter, origin, destination)
 
         return wheel_file
 
@@ -531,8 +564,8 @@ class Project():
     def platform_tag(self) -> str:
         if self.is_pure:
             return 'any'
-        # Choose the sysconfig platform here and let auditwheel fix it later if
-        # there are system dependencies (eg. replace it with a manylinux tag)
+        # XXX: Choose the sysconfig platform here and let something like auditwheel
+        #      fix it later if there are system dependencies (eg. replace it with a manylinux tag)
         return sysconfig.get_platform().replace('-', '_').replace('.', '_')
 
     def _calculate_file_abi_tag_heuristic_windows(self, filename: str) -> Optional[mesonpy._tags.Tag]:
@@ -679,39 +712,12 @@ class Project():
 
         return sdist
 
-    def wheel(self, directory: Path, skip_bundling: bool = True) -> pathlib.Path:
-        """Generates a wheel (binary distribution) in the specified directory.
-
-        Bundles the external binary dependencies by default, but can be skiped
-        via the ``skip_bundling`` parameter. This results in wheels that are
-        only garanteed to work on the current system as it may have external
-        dependencies.
-        This option is useful for users like Python distributions, which want
-        the artifacts to link the system libraries, unlike someone who is
-        distributing wheels on PyPI.
-        """
+    def wheel(self, directory: Path) -> pathlib.Path:  # noqa: F811
+        """Generates a wheel (binary distribution) in the specified directory."""
         wheel = _WheelBuilder(self).build(self._install_plan, self._copy_files, self._build_dir)
 
-        # return the wheel directly if pure or the user wants to skip the lib bundling step
-        if self.is_pure or skip_bundling:
-            final_wheel = pathlib.Path(directory, wheel.name)
-            shutil.move(os.fspath(wheel), final_wheel)
-            return final_wheel
-
-        # use auditwheel to select the correct platform tag based on the external dependencies
-        auditwheel_out = self._build_dir / 'auditwheel-output'
-        with mesonpy._util.cd(self._build_dir), mesonpy._util.add_ld_path(self._lib_paths):
-            self._proc('auditwheel', 'repair', '-w', os.fspath(auditwheel_out), os.fspath(wheel))
-
-        # get built wheel
-        out = list(auditwheel_out.glob('*.whl'))
-        assert len(out) == 1
-        repaired_wheel = out[0]
-
-        # move to the output directory
-        final_wheel = pathlib.Path(directory, repaired_wheel.name)
-        shutil.move(os.fspath(repaired_wheel), final_wheel)
-
+        final_wheel = pathlib.Path(directory, wheel.name)
+        shutil.move(os.fspath(wheel), final_wheel)
         return final_wheel
 
 
@@ -753,7 +759,7 @@ def get_requires_for_build_wheel(
     dependencies = [_depstr.wheel, _depstr.ninja]
     with _project(config_settings) as project:
         if not project.is_pure:
-            dependencies.append(_depstr.auditwheel)
+            dependencies.append(_depstr.patchelf_wrapper)
         if project.pep621:
             dependencies.append(_depstr.pep621)
     return dependencies
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -268,7 +268,7 @@ class _WheelBuilder():
 
     def _install_file(
         self,
-        wheel_file: wheel.wheelfile.WheelFile,
+        wheel_file: wheel.wheelfile.WheelFile,  # type: ignore[name-defined]
         counter: mesonpy._util.CLICounter,
         origin: Path,
         destination: pathlib.Path,
--- a/mesonpy/_elf.py
+++ b/mesonpy/_elf.py
@@ -1,0 +1,48 @@
+# SPDX-License-Identifier: EUPL-1.2
+# SPDX-FileCopyrightText: 2021 Quansight, LLC
+# SPDX-FileCopyrightText: 2021 Filipe Laíns <lains@riseup.net>
+
+import os
+import subprocess
+
+from typing import Optional
+
+from mesonpy._compat import Collection, Path
+
+
+class ELF:
+    def __init__(self, path: Path) -> None:
+        self._path = os.fspath(path)
+        self._rpath: Optional[Collection[str]] = None
+        self._needed: Optional[Collection[str]] = None
+
+    def _patchelf(self, *args: str) -> str:
+        return subprocess.check_output(['patchelf', *args, self._path]).decode()
+
+    @property
+    def rpath(self) -> Collection[str]:
+        if not self._rpath:
+            self._rpath = self._patchelf('--print-rpath').strip().split(';')
+        return self._rpath
+
+    @rpath.setter
+    def rpath(self, value: Collection[str]) -> None:
+        self._patchelf('--set-rpath', ':'.join(value))
+        self._rpath = value
+
+    @property
+    def needed(self) -> Collection[str]:
+        if self._needed is None:
+            self._needed = frozenset(self._patchelf('--print-needed').splitlines())
+        return self._needed
+
+    @needed.setter
+    def needed(self, value: Collection[str]) -> None:
+        value = frozenset(value)
+        for entry in self.needed:
+            if entry not in value:
+                self._patchelf('--remove-needed', entry)
+        for entry in value:
+            if entry not in self.needed:
+                self._patchelf('--add-needed', entry)
+        self._needed = value
--- a/mesonpy/_tags.py
+++ b/mesonpy/_tags.py
@@ -103,7 +103,8 @@ class LinuxInterpreterTag(Tag):
     @property
     def python(self) -> str:
         if self.implementation == 'cpython':
-            return f'cp{self.interpreter_version}'
+            # The Python tag for CPython does not seem to include the flags suffixes.
+            return f'cp{self.interpreter_version}'.rstrip('dmu')
         elif self.implementation in ('pypy', 'pypy3'):
             interpreter_version = f'{sys.version_info[0]}{sys.version_info[1]}'
             return f'pp{interpreter_version}'
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,6 +6,7 @@ import os.path
 import pathlib
 import shutil
 import tempfile
+import venv
 
 import git
 import pytest
@@ -52,6 +53,21 @@ def tmp_dir_session():
 
     try:
         yield pathlib.Path(path)
+    finally:
+        try:
+            shutil.rmtree(path)
+        except PermissionError:  # pragma: no cover
+            pass  # this sometimes fails on windows :/
+
+
+@pytest.fixture()
+def virtual_env():
+    path = pathlib.Path(tempfile.mkdtemp(prefix='mesonpy-test-venv-'))
+
+    venv.create(path, with_pip=True)
+    try:
+        # FIXME: windows
+        yield path / 'bin' / 'python'
     finally:
         try:
             shutil.rmtree(path)
--- a/tests/packages/link-against-local-lib/examplelib.c
+++ b/tests/packages/link-against-local-lib/examplelib.c
@@ -1,0 +1,3 @@
+int sum(int a, int b) {
+	return a + b;
+}
--- a/tests/packages/link-against-local-lib/examplelib.h
+++ b/tests/packages/link-against-local-lib/examplelib.h
@@ -1,0 +1,1 @@
+int sum(int a, int b);
--- a/tests/packages/link-against-local-lib/examplemod.c
+++ b/tests/packages/link-against-local-lib/examplemod.c
@@ -1,0 +1,33 @@
+#include <Python.h>
+
+#include "examplelib.h"
+
+static PyObject* example_sum(PyObject* self, PyObject *args)
+{
+    int a, b;
+    if (!PyArg_ParseTuple(args, "ii", &a, &b)) {
+        return NULL;
+    }
+
+    long result = sum(a, b);
+
+    return PyLong_FromLong(result);
+}
+
+static PyMethodDef methods[] = {
+    {"example_sum", (PyCFunction)example_sum, METH_VARARGS, NULL},
+    {NULL, NULL, 0, NULL},
+};
+
+static struct PyModuleDef module = {
+    PyModuleDef_HEAD_INIT,
+    "example",
+    NULL,
+    -1,
+    methods,
+};
+
+PyMODINIT_FUNC PyInit_example(void)
+{
+    return PyModule_Create(&module);
+}
--- a/tests/packages/link-against-local-lib/meson.build
+++ b/tests/packages/link-against-local-lib/meson.build
@@ -1,0 +1,20 @@
+project(
+    'link-against-local-lib', 'c',
+    version: '1.0.0',
+)
+
+example_lib = shared_library(
+    'example', 'examplelib.c',
+    install: true,
+)
+
+py_dep = dependency('python3')
+py_mod = import('python')
+py = py_mod.find_installation()
+
+py.extension_module(
+    'example', 'examplemod.c',
+    dependencies: py_dep,
+    link_with: example_lib,
+    install: true,
+)
--- a/tests/packages/link-against-local-lib/pyproject.toml
+++ b/tests/packages/link-against-local-lib/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+build-backend = 'mesonpy'
+requires = ['mesonpy']
--- a/tests/test_pep517.py
+++ b/tests/test_pep517.py
@@ -24,8 +24,8 @@ def test_get_requires_for_build_sdist(package, expected):
     [
         ('pure', set()),  # pure and no PEP 621
         ('full-metadata', {mesonpy._depstr.pep621}),  # pure and PEP 621
-        ('library', {mesonpy._depstr.auditwheel}),  # not pure and not PEP 621
-        ('library-pep621', {mesonpy._depstr.auditwheel, mesonpy._depstr.pep621}),  # not pure and PEP 621
+        ('library', {mesonpy._depstr.patchelf_wrapper}),  # not pure and not PEP 621
+        ('library-pep621', {mesonpy._depstr.patchelf_wrapper, mesonpy._depstr.pep621}),  # not pure and PEP 621
     ]
 )
 def test_get_requires_for_build_wheel(package, expected):
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -2,6 +2,7 @@
 
 import platform
 import re
+import subprocess
 import sys
 import sysconfig
 
@@ -50,6 +51,7 @@ def test_contents(package_library, wheel_library):
     artifact = wheel.wheelfile.WheelFile(wheel_library)
 
     for name, regex in zip(sorted(wheel_contents(artifact)), [
+        re.escape('.library.mesonpy.libs/libexample.so'),
         re.escape('library-1.0.0.data/scripts/example'),
         re.escape('library-1.0.0.dist-info/METADATA'),
         re.escape('library-1.0.0.dist-info/RECORD'),
@@ -96,3 +98,10 @@ def test_configure_data(wheel_configure_data):
 def test_interpreter_abi_tag(wheel_purelib_and_platlib):
     expected = f'purelib_and_platlib-1.0.0-{PYTHON_TAG}-{INTERPRETER_TAG}-{PLATFORM_TAG}.whl'
     assert wheel_purelib_and_platlib.name == expected
+
+
+def test_local_lib(virtual_env, wheel_link_against_local_lib):
+    subprocess.check_call([virtual_env, '-m', 'pip', 'install', wheel_link_against_local_lib])
+    subprocess.check_output([
+        virtual_env, '-c', 'import example; print(example.example_sum(1, 2))'
+    ]).decode() == '3'
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -15,6 +15,7 @@ INTERPRETER_VERSION = f'{sys.version_info[0]}{sys.version_info[1]}'
 
 if platform.python_implementation() == 'CPython':
     INTERPRETER_TAG = f'cp{INTERPRETER_VERSION}'
+    PYTHON_TAG = INTERPRETER_TAG
     # Py_UNICODE_SIZE has been a runtime option since Python 3.3,
     # so the u suffix no longer exists
     if sysconfig.get_config_var('Py_DEBUG'):
@@ -24,8 +25,6 @@ if platform.python_implementation() == 'CPython':
         pymalloc = sysconfig.get_config_var('WITH_PYMALLOC')
         if pymalloc or pymalloc is None:  # none is the default value, which is enable
             INTERPRETER_TAG += 'm'
-
-    PYTHON_TAG = INTERPRETER_TAG
 elif platform.python_implementation() == 'PyPy':
     INTERPRETER_TAG = f'pypy3_{INTERPRETER_VERSION}'
     PYTHON_TAG = f'pp{INTERPRETER_VERSION}'


FFY00:
@rgommers the tests are failing on 3.7 due to weirdness in the ABI flags, which fully goes away in 3.8 due to several improvements to the build system. Since SciPy only seems to be supporting 3.8+, do you think it makes sense to try to fix it, or should we just drop 3.7 for now? The benefit seems minimal to me, and we can always bring it back in the future if needed.

FFY00:
Anyway, after merging this, we should be ready for a release.

rgommers:
> Since SciPy only seems to be supporting 3.8+, do you think it makes sense to try to fix it, or should we just drop 3.7 for now? The benefit of working on this seems minimal to me, and we can always bring it back in the future if needed.

Agreed, it seems fine to drop 3.7; SciPy and other scientific projects all won't have a need for that (due to [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)).

FFY00:
> Have you thought about how to make this configurable? We established that there are valid reasons to do this bundling, and also valid reasons to skip it (xref [#20 (comment)](https://github.com/FFY00/mesonpy/issues/20#issuecomment-1004346950)).

Yes. I think that definitely makes sense if we run auditwheel by default, but I am not sure if I feel the same way if we do not.

IMO, I think we should probably be running it by default, so that user installs via sdist are less prone to breaking, but that does have drawbacks.

What do you want to do?

rgommers:
I did a bit of testing. First, a trivial fix for this PR:
```diff
diff --git a/meson.build b/meson.build
index 0514e2b..53a6d90 100644
--- a/meson.build
+++ b/meson.build
@@ -10,6 +10,7 @@ endif
 py.install_sources(
   'mesonpy/__init__.py',
   'mesonpy/_compat.py',
+  'mesonpy/_elf.py',
   'mesonpy/_tags.py',
   'mesonpy/_util.py',
   subdir: 'mesonpy',
diff --git a/pyproject.toml b/pyproject.toml
index 7030ec2..7ffdc78 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ classifiers = [
 dependencies = [
   'meson>=0.60.0',
   'ninja',
+  'pep621',
   'tomli>=1.0.0',
   'typing-extensions>=3.7.4; python_version<"3.8"',
 ]
```

Then I can run `python -m build --no-isolation --skip-dependency-check` in my [test-mesonpy branch](https://github.com/rgommers/scipy/tree/test-mesonpy).

The `sdist` gets created (two warnings; the first one is unexpected/new for me, we're making the second one a hard error in SciPy's current `setup.py`):
```
+ meson dist --no-tests --formats gztar
WARNING: Repository has uncommitted changes that will not be included in the dist tarball
WARNING: Submodule 'doc/sphinxext/doc/scipy-sphinx-theme' is not checked out and cannot be added to the dist
Created /tmp/mesonpy-aenppslx/build/meson-dist/SciPy-1.9.0.dev0.tar.gz
```
I'd recommend to either pick up uncommitted changes, or make it a hard error - this seems a little confusing. My changes were to `pyproject.toml`, so they seemed to be picked up - but when building the sdist from the wheel, it ended up switching back to `setup.py` because I hadn't committed my `build-backend` change. `sdist` creation is also fairly slow (I guess it does a fresh git checkout under the hood)

It then goes on to build a `wheel`, showing the build output (this is good, no need to add `-v` like for `pip`). The build itself succeeds, but then there's some warnings before the `_elf` part has a problem:

<details>

```
WARNING Using heuristics to map files to wheel, this may result in incorrect locations
WARNING Could not tell if file was meant for purelib or platlib, so it was mapped to platlib: /tmp/mesonpy-2qdx8tuh/install/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/__config__.py (/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/__config__.py)
WARNING Could not tell if file was meant for purelib or platlib, so it was mapped to platlib: /tmp/mesonpy-2qdx8tuh/install/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/version.py (/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/version.py)
WARNING Could not tell if file was meant for purelib or platlib, so it was mapped to platlib: /tmp/mesonpy-2qdx8tuh/install/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/special/tests/data/boost.npz (/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/special/tests/data/boost.npz)
WARNING Could not tell if file was meant for purelib or platlib, so it was mapped to platlib: /tmp/mesonpy-2qdx8tuh/install/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/special/tests/data/gsl.npz (/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/special/tests/data/gsl.npz)
WARNING Could not tell if file was meant for purelib or platlib, so it was mapped to platlib: /tmp/mesonpy-2qdx8tuh/install/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/special/tests/data/local.npz (/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/special/tests/data/local.npz)
WARNING Could not tell if file was meant for purelib or platlib, so it was mapped to platlib: /tmp/mesonpy-2qdx8tuh/install/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/cython_blas.pyx (/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/cython_blas.pyx)
WARNING Could not tell if file was meant for purelib or platlib, so it was mapped to platlib: /tmp/mesonpy-2qdx8tuh/install/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/cython_lapack.pyx (/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/cython_lapack.pyx)
WARNING Could not tell if file was meant for purelib or platlib, so it was mapped to platlib: /tmp/mesonpy-2qdx8tuh/install/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/cython_blas.pxd (/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/cython_blas.pxd)
WARNING Could not tell if file was meant for purelib or platlib, so it was mapped to platlib: /tmp/mesonpy-2qdx8tuh/install/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/cython_lapack.pxd (/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/cython_lapack.pxd)
WARNING Could not tell if file was meant for purelib or platlib, so it was mapped to platlib: /tmp/mesonpy-2qdx8tuh/install/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/_blas_subroutines.h (/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/_blas_subroutines.h)
WARNING Could not tell if file was meant for purelib or platlib, so it was mapped to platlib: /tmp/mesonpy-2qdx8tuh/install/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/_lapack_subroutines.h (/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/_lapack_subroutines.h)
WARNING Could not tell if file was meant for purelib or platlib, so it was mapped to platlib: /tmp/mesonpy-2qdx8tuh/install/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/_blas_subroutine_wrappers.f (/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/_blas_subroutine_wrappers.f)
WARNING Could not tell if file was meant for purelib or platlib, so it was mapped to platlib: /tmp/mesonpy-2qdx8tuh/install/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/_lapack_subroutine_wrappers.f (/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/scipy/linalg/_lapack_subroutine_wrappers.f)
Copying files to wheel...
Traceback (most recent call last):x86_64-linux-gnu.so                           
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/pep517/in_process/_in_process.py", line 280, in <module>
    main()
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/pep517/in_process/_in_process.py", line 263, in main
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/pep517/in_process/_in_process.py", line 204, in build_wheel
    return _build_backend().build_wheel(wheel_directory, config_settings,
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/mesonpy/__init__.py", line 780, in build_wheel
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/mesonpy/__init__.py", line 720, in wheel
    wheel = _WheelBuilder(self).build(self._install_plan, self._copy_files, self._build_dir)
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/mesonpy/__init__.py", line 315, in build
    self._install_file(whl, counter, origin, destination)
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/mesonpy/__init__.py", line 285, in _install_file
    assert elf.rpath in ('', '$ORIGIN/'), f'Unexpected RPATH in {origin}: {elf.rpath}'
AssertionError: Unexpected RPATH in /tmp/mesonpy-2qdx8tuh/build/mesonpy-wheel-libs/scipy/_lib/_ccallback_c.cpython-39-x86_64-linux-gnu.so: /home/rgommers/anaconda3/envs/scipy-meson/lib

```

</details>

The warnings come from specifying `install_dir`, e.g. for `scipy/__config__.py`:
```meson
generate_config = custom_target(
  'generate-config',
  install: true,
  build_always_stale: true,
  build_by_default: true,
  output: '__config__.py',
  input: '../tools/config_utils.py',
  command: [py3, '@INPUT@', '@OUTPUT@'],
  install_dir: py3.get_install_dir() + '/scipy'
)
```
Unless there's a way to add the purelib/platlib info in the `meson.build` file (which I don't think there is?), those warnings should not be shown. Side note: if heuristics are used, it seems odd that `.py` is mapped to `platlib`.

For the error:
```
  File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/mesonpy/__init__.py", line 285, in _install_file
    assert elf.rpath in ('', '$ORIGIN/'), f'Unexpected RPATH in {origin}: {elf.rpath}'
AssertionError: Unexpected RPATH in /tmp/mesonpy-2qdx8tuh/build/mesonpy-wheel-libs/scipy/_lib/_ccallback_c.cpython-39-x86_64-linux-gnu.so: /home/rgommers/anaconda3/envs/scipy-meson/lib
```
The path `/home/rgommers/anaconda3/envs/scipy-meson/lib` is indeed where `libopenblas.so` et al. are located.

Finally a question: how do I build in a directory that's not `/tmp`? In this case I wanted to debug, but the build output had already disappeared because it was living in a tmpdir. I don't see a flag in `build` that lets me do this.

rgommers:
> IMO, I think we should probably be running it by default, so that user installs via sdist are less prone to breaking, but that does have drawbacks.
> 
> What do you want to do?

I think it's hard to determine what is less fragile. So far I haven't been able to produce a wheel yet because both `auditwheel` and `mesonpy._elf` gave exceptions. Rpath rewriting is not completely trivial, so more likely to break than just skipping that step. 

It's also not clear to me what the rules are for including shared libraries. Let's take a concrete example from SciPy:
```
$ ldd build/scipy/linalg/_flinalg.cpython-39-x86_64-linux-gnu.so
        linux-vdso.so.1 (0x00007ffc21290000)
        libopenblas.so.0 => /home/rgommers/anaconda3/envs/scipy-dev/lib/libopenblas.so.0 (0x00007ff0b3f6c000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007ff0b3d79000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007ff0b3c35000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007ff0b3c14000)
        libgfortran.so.5 => /home/rgommers/anaconda3/envs/scipy-dev/lib/./libgfortran.so.5 (0x00007ff0b3a6b000)
        /usr/lib64/ld-linux-x86-64.so.2 (0x00007ff0b5fb1000)
        libquadmath.so.0 => /home/rgommers/anaconda3/envs/scipy-dev/lib/././libquadmath.so.0 (0x00007ff0b3a2f000)
        libgcc_s.so.1 => /home/rgommers/anaconda3/envs/scipy-dev/lib/././libgcc_s.so.1 (0x00007ff0b3a1a000)
```
What is going to be included here? Does the answer change if, for example, `libgfortran` or `libopenblas` come from `/usr/lib/`?

FFY00:
> I did a bit of testing. First, a trivial fix for this PR:

Ah, yes! I forgot to add the file to `meson.build`. We should add a build action in the CI. The `pep621` should not be specified there, as it is not always necessary, and we put it in `get_requires_for_*` when it is.

> Unless there's a way to add the purelib/platlib info in the `meson.build` file (which I don't think there is?), those warnings should not be shown.

We can work around this by using the new Meson tags feature, but that would be a dirty hack. Other possibility would be to add a setting in `project.toml` for users to clarify where the files should be placed.

> Side note: if heuristics are used, it seems odd that `.py` is mapped to `platlib`.

`.py` does not mean it should necessarily go to purelib, it might still be architecture dependent (eg. a generated Python source file that depends on the architecture). If you have Meson file to install the file to platlib, the heuristics should not be putting it in purelib just because it is a `.py`. Because we can't tell where it was supposed to be installed, we put it in the safest place, which is platlib.

> For the error:
> 
> ```
>   File "/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/mesonpy/__init__.py", line 285, in _install_file
>     assert elf.rpath in ('', '$ORIGIN/'), f'Unexpected RPATH in {origin}: {elf.rpath}'
> AssertionError: Unexpected RPATH in /tmp/mesonpy-2qdx8tuh/build/mesonpy-wheel-libs/scipy/_lib/_ccallback_c.cpython-39-x86_64-linux-gnu.so: /home/rgommers/anaconda3/envs/scipy-meson/lib
> ```
> 
> The path `/home/rgommers/anaconda3/envs/scipy-meson/lib` is indeed where `libopenblas.so` et al. are located.

Right, I have to look into this. The assert is there because I'd rather to blow up on unexpected values than to possibly silently introduce a bug for letting it work in a situation we did not take into account.
In this case, I think we might be taking the file from the build directory, rather from the dummy installation, which would be a bug on our part, so this would be a good catch. I have to look into it.

> Finally a question: how do I build in a directory that's not `/tmp`? In this case I wanted to debug, but the build output had already disappeared because it was living in a tmpdir. I don't see a flag in `build` that lets me do this.

This is not yet documented, as I wanted to resolve the auditwheel situation and then document all config options, but you can pass a `builddir` option (eg. `python -m build -Cbuilddir=build`) to build from a local Meson build directory. This is extremely useful in large projects like SciPy, to cache the build.

FFY00:
> So far I haven't been able to produce a wheel yet because both `auditwheel` and `mesonpy._elf` gave exceptions. Rpath rewriting is not completely trivial, so more likely to break than just skipping that step.

You can remove that assert if you want, it is not very problematic.

> What is going to be included here?

Everything that is not provided by the manylinux tag should be included.

> Does the answer change if, for example, `libgfortran` or `libopenblas` come from `/usr/lib/`?

As opposed to where?

> EDIT: another question: what about name mangling and problems with duplication? If multiple packages pick up `libopenblas`, now I have multiple copies (e.g., `import numpy; import scipy.linalg`). So now I've created a huge problem even if name mangling was done correctly and I can load both copies - because I now have two copies of openblas, which means two threadpools instead of one, etc.
> 
> It still feels like something that's impossible to get right automatically.

Yes, that is one of the drawbacks of this approach. How problematic would that be? One possible solution would be excluding those symbols from name mangling, but this requires making sure both libraries have compatible ABIs, as far as those symbols are concerned at least.

I don't think we have any way around these sort of problems, as PyPI is just not fit to deal with native dependencies. One option would be focusing on improving the Conda ecosystem as an alternative, and have upstreams like SciPy build Conda packages (to be distributed directly or via a well-known channel like conda-forge).

rgommers:
> You can remove that assert if you want, it is not very problematic.

Okay, I did that and then it built me a wheel. Unpacking it I run into:

image

Looking back in the build log, I do see:
```
WARNING Duplicate name: '_zeros.cpython-39-x86_64-linux-gnu.so'
```

A much bigger problem is that the vendoring code seems to do completely the wrong thing: it vendors the SciPy extension modules in the top-level code, and they've disappeared from their regular in-tree place:

image

External dependencies are not vendored as far as I can tell.

rgommers:
> > It still feels like something that's impossible to get right automatically.
> 
> Yes, that is one of the drawbacks of this approach. How problematic would that be? One possible solution would be excluding those symbols from name mangling, but this requires making sure both libraries have compatible ABIs, as far as those symbols are concerned at least.

I think that `mesonpy` should do the basic approach here: simply zip up the files that are installable, add correct metadata, and call it a day. This is what `wheel` does now, and yes it has gaps but in practice it mostly works. As we can see here the deeper problems aren't solvable, and this vendoring introduces obvious new problems that are significantly worse than the existing ones.

The main thing you're trying to solve here is that a wheel depends on locally installed libraries from the system, and then a system upgrade _may_ break the installed package. This isn't all that likely though for many packages (e.g. BLAS/LAPACK have a stable ABI/API and actual breakage is very unlikely), and people by now know this may happen. And if it does, typically what happens is that `import mypackage` breaks. Which is much better than things like deadlocks at runtime because I have duplicate libraries being loaded.

If you want to keep vendoring, I think it should be opt-in. The use case of building a pure C library within the project itself and that library then needing to be vendored is a real use case, but for now there are no users who actually need it.

> I don't think we have any way around these sort of problems, as PyPI is just not fit to deal with native dependencies. 

Yes, that is true. I think we have to live with that for the time being.

> One option would be focusing on improving the Conda ecosystem as an alternative, and have upstreams like SciPy build Conda packages (to be distributed directly or via a well-known channel like conda-forge).

We already do that (conda-forge is a first class citizen by now), but I don't think it's realistic to drop PyPI support for libraries like SciPy - there's too much demand for it. It _is_ possible I think for packages with much more complex dependencies, like the geospatial stack.

And separately, yes we should work on the "PyPI interop with other package managers" to get a better handle on this stuff. So we can at least express dependencies that are now missing.

FFY00:
> Looking back in the build log, I do see:

Uh, we had issues with that in the past, but I thought I had fix it, perhaps this is a regression. We definitely need a test case for this.

> A much bigger problem is that the vendoring code seems to do completely the wrong thing: it vendors the SciPy extension modules in the top-level code, and they've disappeared from their regular in-tree place:

That is not the vendoring code, we are very likely just not using the correct path when we adding the files to the wheel, which is resulting in the files being added to the root. It only seems to be affecting one code path, the one I touched in this PR, so I likely just messed things up somewhere.

> External dependencies are not vendored as far as I can tell.

Yes, they are not, only internal libraries, but that is required.

FFY00:
> I think that `mesonpy` should do the basic approach here: simply zip up the files that are installable, add correct metadata, and call it a day. This is what `wheel` does now, and yes it has gaps but in practice it mostly works.

Then I don't think we should have an option to run auditwheel, the user can do that afterwards if they want.

> As we can see here the deeper problems aren't solvable, and this vendoring introduces obvious new problems that are significantly worse than the existing ones.

The problems here are not due to vendoring, they are a bug in the wheel generation.

> The main thing you're trying to solve here is that a wheel depends on locally installed libraries from the system, and then a system upgrade _may_ break the installed package. This isn't all that likely though for many packages (e.g. BLAS/LAPACK have a stable ABI/API and actual breakage is very unlikely), and people by now know this may happen. And if it does, typically what happens is that `import mypackage` breaks. Which is much better than things like deadlocks at runtime because I have duplicate libraries being loaded.

Other thing to consider is that you will have external dependencies, while having them you can never be truly manylinux compatible, and `pip install ...` can result in broken modules if the user does not have the necessary libraries, which will of course fail with obscure errors (for a normal user at least).

rgommers:
> Yes, they are not, only internal libraries, but that is required.

Okay, if it's internal only, then this should be a do-nothing operation for SciPy. We don't have internal libraries that are not Python extensions.

FFY00:
Yep, this is what a wheel that vendors a library looks like ATM:

```
$ unzip -l /tmp/mesonpy-test-deg32gtn/link_against_local_lib-1.0.0-cp310-cp310-linux_x86_64.whl
Archive:  /tmp/mesonpy-test-deg32gtn/link_against_local_lib-1.0.0-cp310-cp310-linux_x86_64.whl
  Length      Date    Time    Name
---------  ---------- -----   ----
       65  2022-01-10 21:38   link_against_local_lib-1.0.0.dist-info/METADATA
       88  2022-01-10 21:38   link_against_local_lib-1.0.0.dist-info/WHEEL
    25216  2022-01-10 21:38   example.cpython-310-x86_64-linux-gnu.so
    17128  2022-01-10 21:38   .mesonpy.libs/libexample.so
      431  2022-01-10 21:38   link_against_local_lib-1.0.0.dist-info/RECORD
---------                     -------
    42928                     5 files
```

rgommers:
> Other thing to consider is that you will have external dependencies, while having them you can never be truly manylinux compatible

We are `manylinux` compatible, because the PyPI wheels are carefully vendoring one specific version of `libopenblas` - something that cannot be done with general builds. It is unnecessary to be `manylinux` compatible for building from source, because you will be compatible with the system you're on. Redistributing locally built wheels to machines running other Linux distros is not a valid use case imho - in that case you should be more careful and use/reproduce our manylinux wheels.

FFY00:
I understand, but for local builds you lose the benefit of the binary wheels where you carefully vendor `libopenblas`, as things will break as soon as the user updates the system and OpenBLAS is updated.

But like I said, there is no perfect solution :shrug:

FFY00:
> Uh, we had issues with that in the past, but I thought I had fix it, perhaps this is a regression. We definitely need a test case for this.

I remember now! Sorry, with all the confusion of christmas, new years, and the Conda plugin system implementation, this feel out of my mind. This is the issue that I attempted to fix in https://github.com/mesonbuild/meson/pull/9474.

Meson is returning the path of some modules without the subdirs (eg. `{moduledir_shared}/_zeros.cpython-310-x86_64-linux-gnu.so`), causing the file to be installed to the root of `platlib`. We can switch to install location matching heuristics in the meanwhile.

Can you try again?

> You can remove that assert if you want, it is not very problematic.

I am still trying to understand exactly the situation here, so you will need to keep commenting it out or running Python with `-O`.

It seems to me that Meson is setting the RPATH to `libdir`, but I don't really understand why.

rgommers:
Tested again, after commenting out the problematic assert for rpath and using:
```
$ python -m build --wheel --no-isolation --skip-dependency-check -Cbuilddir=buildwheel
$ pip install dist/scipy-1.9.0.dev0-cp39-cp39-linux_x86_64.whl
```
The wheel now gets generated and the contents look about right - files are all in the right place according to some manual spot checking.

The import fails however with:
```
    from ._qhull import *
ImportError: libopenblas.so.0: cannot open shared object file: No such file or directory
```

Checking the extension that is installed into site-packages (coming from the wheel):
```
$ ldd _qhull.cpython-39-x86_64-linux-gnu.so 
        linux-vdso.so.1 (0x00007ffee8db7000)
        libopenblas.so.0 => not found
        libc.so.6 => /usr/lib/libc.so.6 (0x00007ff9b7060000)
        /usr/lib64/ld-linux-x86-64.so.2 (0x00007ff9b735b000)
```

Compare to the same extension in the wheel build dir:
```
$ cd buildwheel/scipy/spatial/
$ ldd _qhull.cpython-39-x86_64-linux-gnu.so
        linux-vdso.so.1 (0x00007ffc90728000)
        libopenblas.so.0 => /home/rgommers/anaconda3/envs/scipy-meson/lib/libopenblas.so.0 (0x00007f48118b7000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f48116c4000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007f4811580000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f481155f000)
        libgfortran.so.5 => /home/rgommers/anaconda3/envs/scipy-meson/lib/./libgfortran.so.5 (0x00007f48113b6000)
        /usr/lib64/ld-linux-x86-64.so.2 (0x00007f4813984000)
        libquadmath.so.0 => /home/rgommers/anaconda3/envs/scipy-meson/lib/././libquadmath.so.0 (0x00007f481137a000)
        libgcc_s.so.1 => /home/rgommers/anaconda3/envs/scipy-meson/lib/././libgcc_s.so.1 (0x00007f4811365000)
```
The unmangled dependencies are fine, the mangled have somehow messed with `libopenblas.so.0` so it's no longer found. The other changes are transitive dependencies of `libopenblas`.

rgommers:
Ah it's because of the RPATH in the `.so`'s:

Correct:
```
$ readelf -d ../../../buildwheel/scipy/spatial/_qhull.cpython-39-x86_64-linux-gnu.so  | grep ibrary
 0x0000000000000001 (NEEDED)             Shared library: [libopenblas.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000f (RPATH)              Library rpath: [/home/rgommers/anaconda3/envs/scipy-meson/lib]
```

After rewriting:
```
$ readelf -d _qhull.cpython-39-x86_64-linux-gnu.so  | grep ibrary
 0x0000000000000001 (NEEDED)             Shared library: [libopenblas.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN/../../.scipy.mesonpy.libs]
```

FFY00:
Do you have the build environment enabled when trying to use the wheel? It need `libopenblas.so.0` to be present on the system, if you are using some sort of isolated environment, `LD_LIBRARY_PATH` should be set to include the libs folder.

FFY00:
How does setuptools behave in this case?

rgommers:
> Do you have the build environment enabled when trying to use the wheel? 

Yes of course. The build env == the runtime env. It's an activated conda environment.

> It need `libopenblas.so.0` to be present on the system, if you are using some sort of isolated environment, `LD_LIBRARY_PATH` should be set to include the libs folder.

We should never have a need for `LD_LIBRARY_PATH`.  `libopenblas.so.0` is present right where it was found during the build. `mesonpy` is simply rewriting the rpath here when it shouldn't be touching it.

> How does setuptools/distutils behave in this case?

Let me check. Probably the same as Meson.

FFY00:
> This is where `libopenblas.so` lives. It's an external dependency, so that seems correct.

But the linker should already search `libdir` by default, Meson shouldn't have to set the RPATH for this.

`ldconfig -p | grep  libopenblas` should be showing the library already if it is in `libdir`. Can you post the output of that command?

rgommers:
Rebuilding SciPy with `distutils` takes a while (doing that now), but checked a local NumPy build:
```
 0x000000000000000f (RPATH)              Library rpath: [/home/rgommers/anaconda3/envs/numpy-dev/lib]
```

And for completeness, `conda-build`:
```
0x000000000000000f (RPATH)              Library rpath: [$ORIGIN/../../../..]
```

rgommers:
> But the linker should already search `libdir` by default, Meson shouldn't have to set the RPATH for this.

No, because we're using the linker from the operating system (see `/usr/lib64/ld-linux-x86-64.so.2` output from `ldd`), so there is no way it is going to know about a `libdir` in a conda environment.

> `ldconfig -p | grep libopenblas` should be showing the library already if it is in `libdir`. Can you post the output of that command?

```
$ which ldconfig
/usr/bin/ldconfig
$ ldconfig -p | grep libopenblas
# empty output
```

FFY00:
I understand. Can you post the Meson discovery log then?

rgommers:
`meson-logs/meson-log.txt`:

<details>

```
Build started at 2022-01-21T23:39:57.426955
Main binary: /home/rgommers/anaconda3/envs/scipy-meson/bin/python3.9
Build Options: -Dprefix=/home/rgommers/anaconda3/envs/scipy-meson '--native-file /home/rgommers/code/bldscipy/.mesonpy-native-file.ini'
Python system: Linux
The Meson build system
Version: 0.61.0.rc1
Source dir: /home/rgommers/code/bldscipy
Build dir: /home/rgommers/code/bldscipy/buildwheel
Build type: native build
Project name: SciPy
Project version: 1.9.0.dev0
Sanity testing C compiler: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-cc
Is cross compiler: False.
Sanity check compiler command line: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-cc sanitycheckc.c -o sanitycheckc.exe -Wno-unused-function -Wno-conversion -Wno-misleading-indentation -Wno-incompatible-pointer-types -D_FILE_OFFSET_BITS=64 -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,-rpath,/home/rgommers/anaconda3/envs/scipy-meson/lib -Wl,-rpath-link,/home/rgommers/anaconda3/envs/scipy-meson/lib -L/home/rgommers/anaconda3/envs/scipy-meson/lib
Sanity check compile stdout:

-----
Sanity check compile stderr:

-----
Running test binary command: /home/rgommers/code/bldscipy/buildwheel/meson-private/sanitycheckc.exe
C compiler for the host machine: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-cc (gcc 9.4.0 "x86_64-conda-linux-gnu-cc (GCC) 9.4.0")
C linker for the host machine: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-cc ld.bfd 2.36.1
Sanity testing C++ compiler: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++
Is cross compiler: False.
Sanity check compiler command line: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ sanitycheckcpp.cc -o sanitycheckcpp.exe -fvisibility-inlines-hidden -std=c++17 -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -D_FILE_OFFSET_BITS=64 -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,-rpath,/home/rgommers/anaconda3/envs/scipy-meson/lib -Wl,-rpath-link,/home/rgommers/anaconda3/envs/scipy-meson/lib -L/home/rgommers/anaconda3/envs/scipy-meson/lib
Sanity check compile stdout:

-----
Sanity check compile stderr:

-----
Running test binary command: /home/rgommers/code/bldscipy/buildwheel/meson-private/sanitycheckcpp.exe
C++ compiler for the host machine: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ (gcc 9.4.0 "x86_64-conda-linux-gnu-c++ (GCC) 9.4.0")
C++ linker for the host machine: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ ld.bfd 2.36.1
Running compile:
Working directory:  /tmp/tmpzg1r2zdx
Command line:  cython /tmp/tmpzg1r2zdx/testfile.pyx -o /tmp/tmpzg1r2zdx/output.exe --fast-fail 

Code:
 print("hello world")
Compiler stdout:
 
Compiler stderr:
 /home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/Cython/Compiler/Main.py:369: FutureWarning: Cython directive 'language_level' not set, using 2 for now (Py2). This will change in a later release! File: /tmp/tmpzg1r2zdx/testfile.pyx
  tree = Parsing.p_module(s, pxd, full_module_name)

Cython compiler for the host machine: cython (cython 0.29.26)
Sanity testing C compiler: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-cc
Is cross compiler: False.
Sanity check compiler command line: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-cc sanitycheckc.c -o sanitycheckc.exe -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -D_FILE_OFFSET_BITS=64 -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,-rpath,/home/rgommers/anaconda3/envs/scipy-meson/lib -Wl,-rpath-link,/home/rgommers/anaconda3/envs/scipy-meson/lib -L/home/rgommers/anaconda3/envs/scipy-meson/lib
Sanity check compile stdout:

-----
Sanity check compile stderr:

-----
Running test binary command: /home/rgommers/code/bldscipy/buildwheel/meson-private/sanitycheckc.exe
C compiler for the build machine: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-cc (gcc 9.4.0 "x86_64-conda-linux-gnu-cc (GCC) 9.4.0")
C linker for the build machine: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-cc ld.bfd 2.36.1
Sanity testing C++ compiler: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++
Is cross compiler: False.
Sanity check compiler command line: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ sanitycheckcpp.cc -o sanitycheckcpp.exe -fvisibility-inlines-hidden -std=c++17 -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -D_FILE_OFFSET_BITS=64 -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,-rpath,/home/rgommers/anaconda3/envs/scipy-meson/lib -Wl,-rpath-link,/home/rgommers/anaconda3/envs/scipy-meson/lib -L/home/rgommers/anaconda3/envs/scipy-meson/lib
Sanity check compile stdout:

-----
Sanity check compile stderr:

-----
Running test binary command: /home/rgommers/code/bldscipy/buildwheel/meson-private/sanitycheckcpp.exe
C++ compiler for the build machine: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ (gcc 9.4.0 "x86_64-conda-linux-gnu-c++ (GCC) 9.4.0")
C++ linker for the build machine: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ ld.bfd 2.36.1
Using cached compile:
Cached command line:   

Code:
 print("hello world")
Cached compiler stdout:
 
Cached compiler stderr:
 /home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/Cython/Compiler/Main.py:369: FutureWarning: Cython directive 'language_level' not set, using 2 for now (Py2). This will change in a later release! File: /tmp/tmpzg1r2zdx/testfile.pyx
  tree = Parsing.p_module(s, pxd, full_module_name)

Cython compiler for the build machine: cython (cython 0.29.26)
Build machine cpu family: x86_64
Build machine cpu: x86_64
Host machine cpu family: x86_64
Host machine cpu: x86_64
Target machine cpu family: x86_64
Target machine cpu: x86_64
Running compile:
Working directory:  /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpamas4px5
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-cc /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpamas4px5/testfile.c -o /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpamas4px5/output.obj -c -Wno-unused-function -Wno-conversion -Wno-misleading-indentation -Wno-incompatible-pointer-types -D_FILE_OFFSET_BITS=64 -O0 -Wunused-but-set-variable -Wno-unused-but-set-variable 

Code:
 extern int i;
int i;

Compiler stdout:
 
Compiler stderr:
 
Compiler for C supports arguments -Wno-unused-but-set-variable: YES 
Fortran compiler for the host machine: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-gfortran (gcc 9.4.0 "GNU Fortran (GCC) 9.4.0")
Fortran linker for the host machine: /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-gfortran ld.bfd 2.36.1
Program cython found: YES (/home/rgommers/anaconda3/envs/scipy-meson/bin/cython)
Program pythran found: YES (/home/rgommers/anaconda3/envs/scipy-meson/bin/pythran)
Program cp found: YES (/usr/bin/cp)
Program python3 found: YES (/home/rgommers/anaconda3/envs/scipy-meson/bin/python3.9)
Pkg-config binary for 1 is not cached.
Pkg-config binary missing from cross or native file, or env var undefined.
Trying a default Pkg-config fallback at pkg-config
Found pkg-config: /home/rgommers/anaconda3/envs/scipy-meson/bin/pkg-config (0.29.2)
Determining dependency 'python-3.9' with pkg-config executable '/home/rgommers/anaconda3/envs/scipy-meson/bin/pkg-config'
env[PKG_CONFIG_LIBDIR]: /home/rgommers/anaconda3/envs/scipy-meson/lib/pkgconfig
env[PKG_CONFIG_PATH]: 
Called `/home/rgommers/anaconda3/envs/scipy-meson/bin/pkg-config --modversion python-3.9` -> 0
3.9
env[PKG_CONFIG_LIBDIR]: /home/rgommers/anaconda3/envs/scipy-meson/lib/pkgconfig
env[PKG_CONFIG_PATH]: 
Called `/home/rgommers/anaconda3/envs/scipy-meson/bin/pkg-config --cflags python-3.9` -> 0
-I/home/rgommers/anaconda3/envs/scipy-meson/include/python3.9
env[PKG_CONFIG_LIBDIR]: /home/rgommers/anaconda3/envs/scipy-meson/lib/pkgconfig
env[PKG_CONFIG_ALLOW_SYSTEM_LIBS]: 1
env[PKG_CONFIG_PATH]: 
Called `/home/rgommers/anaconda3/envs/scipy-meson/bin/pkg-config --libs python-3.9` -> 0

env[PKG_CONFIG_LIBDIR]: /home/rgommers/anaconda3/envs/scipy-meson/lib/pkgconfig
env[PKG_CONFIG_PATH]: 
Called `/home/rgommers/anaconda3/envs/scipy-meson/bin/pkg-config --libs python-3.9` -> 0

Running command: /home/rgommers/anaconda3/envs/scipy-meson/bin/python3.9 -c import os; os.chdir(".."); import numpy; print(numpy.get_include())
--- stdout ---
/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/numpy/core/include

--- stderr ---


Running compile:
Working directory:  /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpr08gmfds
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-cc -L/home/rgommers/anaconda3/envs/scipy-meson/lib /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpr08gmfds/testfile.c -o /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpr08gmfds/output.exe -Wno-unused-function -Wno-conversion -Wno-misleading-indentation -Wno-incompatible-pointer-types -D_FILE_OFFSET_BITS=64 -O0 -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,-rpath,/home/rgommers/anaconda3/envs/scipy-meson/lib -Wl,-rpath-link,/home/rgommers/anaconda3/envs/scipy-meson/lib 

Code:
 #include<stdio.h>
        
        int main(void) {
            printf("%ld\n", (long)(sizeof(void *)));
            return 0;
        }
Compiler stdout:
 
Compiler stderr:
 
Program stdout:

8

Program stderr:


Running compile:
Working directory:  /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpivvcaske
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-cc /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpivvcaske/testfile.c -o /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpivvcaske/output.obj -c -Wno-unused-function -Wno-conversion -Wno-misleading-indentation -Wno-incompatible-pointer-types -D_FILE_OFFSET_BITS=64 -O0 --print-search-dirs 

Code:
 
Compiler stdout:
 install: /home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/
programs: =/home/rgommers/anaconda3/envs/scipy-meson/bin/../libexec/gcc/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../libexec/gcc/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/bin/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/bin/
libraries: =/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/lib/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/lib/../lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/lib/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/lib/../lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/usr/lib/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/usr/lib/../lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/usr/lib/

Compiler stderr:
 
Library npymath found: YES
Library npyrandom found: YES
Running command: /home/rgommers/anaconda3/envs/scipy-meson/bin/python3.9 -c import pybind11; print(pybind11.get_include())
--- stdout ---
/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/pybind11/include

--- stderr ---


Running command: /home/rgommers/anaconda3/envs/scipy-meson/bin/python3.9 -c import os; print(os.environ.get("SCIPY_USE_PYTHRAN", 1))
--- stdout ---
1

--- stderr ---


Running command: /home/rgommers/anaconda3/envs/scipy-meson/bin/python3.9 -c import os; os.chdir(".."); import pythran; print(os.path.dirname(pythran.__file__));
--- stdout ---
/home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages/pythran

--- stderr ---


Pkg-config binary for 1 is cached.
Determining dependency 'openblas' with pkg-config executable '/home/rgommers/anaconda3/envs/scipy-meson/bin/pkg-config'
env[PKG_CONFIG_PATH]: 
Called `/home/rgommers/anaconda3/envs/scipy-meson/bin/pkg-config --modversion openblas` -> 0
0.3.12
env[PKG_CONFIG_PATH]: 
Called `/home/rgommers/anaconda3/envs/scipy-meson/bin/pkg-config --cflags openblas` -> 0
-I/home/rgommers/anaconda3/envs/scipy-meson/include
env[PKG_CONFIG_ALLOW_SYSTEM_LIBS]: 1
env[PKG_CONFIG_PATH]: 
Called `/home/rgommers/anaconda3/envs/scipy-meson/bin/pkg-config --libs openblas` -> 0
-L/home/rgommers/anaconda3/envs/scipy-meson/lib -lopenblas
env[PKG_CONFIG_PATH]: 
Called `/home/rgommers/anaconda3/envs/scipy-meson/bin/pkg-config --libs openblas` -> 0
-L/home/rgommers/anaconda3/envs/scipy-meson/lib -lopenblas
Running compile:
Working directory:  /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpbdycvgq0
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ -L/home/rgommers/anaconda3/envs/scipy-meson/lib /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpbdycvgq0/testfile.cpp -o /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpbdycvgq0/output.exe -fvisibility-inlines-hidden -std=c++17 -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -D_FILE_OFFSET_BITS=64 -O0 -fpermissive -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,-rpath,/home/rgommers/anaconda3/envs/scipy-meson/lib -Wl,-rpath-link,/home/rgommers/anaconda3/envs/scipy-meson/lib 

Code:
 #include<stdio.h>
        
        int main(void) {
            printf("%ld\n", (long)(sizeof(void *)));
            return 0;
        }
Compiler stdout:
 
Compiler stderr:
 
Program stdout:

8

Program stderr:


Running compile:
Working directory:  /home/rgommers/code/bldscipy/buildwheel/meson-private/tmp31bj7jvz
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ /home/rgommers/code/bldscipy/buildwheel/meson-private/tmp31bj7jvz/testfile.cpp -o /home/rgommers/code/bldscipy/buildwheel/meson-private/tmp31bj7jvz/output.obj -c -fvisibility-inlines-hidden -std=c++17 -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -D_FILE_OFFSET_BITS=64 -O0 -fpermissive --print-search-dirs 

Code:
 
Compiler stdout:
 install: /home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/
programs: =/home/rgommers/anaconda3/envs/scipy-meson/bin/../libexec/gcc/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../libexec/gcc/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/bin/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/bin/
libraries: =/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/lib/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/lib/../lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/lib/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/lib/../lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/usr/lib/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/usr/lib/../lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/usr/lib/

Compiler stderr:
 
Run-time dependency openblas found: YES 0.3.12
Dependency openblas found: YES 0.3.12 (cached)
Running compile:
Working directory:  /home/rgommers/code/bldscipy/buildwheel/meson-private/tmp1otdpurn
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-cc -L/home/rgommers/anaconda3/envs/scipy-meson/lib /home/rgommers/code/bldscipy/buildwheel/meson-private/tmp1otdpurn/testfile.c -o /home/rgommers/code/bldscipy/buildwheel/meson-private/tmp1otdpurn/output.exe -Wno-unused-function -Wno-conversion -Wno-misleading-indentation -Wno-incompatible-pointer-types -D_FILE_OFFSET_BITS=64 -O0 -std=c99 -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,-rpath,/home/rgommers/anaconda3/envs/scipy-meson/lib -Wl,-rpath-link,/home/rgommers/anaconda3/envs/scipy-meson/lib 

Code:
 #include <stdio.h>
#include <limits.h>

        #if defined __stub_open_memstream || defined __stub___open_memstream
        fail fail fail this function is not going to work
        #endif
        
int main(void) {
            void *a = (void*) &open_memstream;
            long long b = (long long) a;
            return (int) b;
        }
Compiler stdout:
 
Compiler stderr:
 /home/rgommers/code/bldscipy/buildwheel/meson-private/tmp1otdpurn/testfile.c: In function 'main':
/home/rgommers/code/bldscipy/buildwheel/meson-private/tmp1otdpurn/testfile.c:9:32: error: 'open_memstream' undeclared (first use in this function)
    9 |             void *a = (void*) &open_memstream;
      |                                ^~~~~~~~~~~~~~
/home/rgommers/code/bldscipy/buildwheel/meson-private/tmp1otdpurn/testfile.c:9:32: note: each undeclared identifier is reported only once for each function it appears in

Running compile:
Working directory:  /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpqkgh33b_
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-cc -L/home/rgommers/anaconda3/envs/scipy-meson/lib /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpqkgh33b_/testfile.c -o /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpqkgh33b_/output.exe -Wno-unused-function -Wno-conversion -Wno-misleading-indentation -Wno-incompatible-pointer-types -D_FILE_OFFSET_BITS=64 -O0 -std=c99 -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,-rpath,/home/rgommers/anaconda3/envs/scipy-meson/lib -Wl,-rpath-link,/home/rgommers/anaconda3/envs/scipy-meson/lib 

Code:
 #include <stdio.h>
        int main(void) {

        /* With some toolchains (MSYS2/mingw for example) the compiler
         * provides various builtins which are not really implemented and
         * fall back to the stdlib where they aren't provided and fail at
         * build/link time. In case the user provides a header, including
         * the header didn't lead to the function being defined, and the
         * function we are checking isn't a builtin itself we assume the
         * builtin is not functional and we just error out. */
        #if !0 && !defined(open_memstream) && !0
            #error "No definition for __builtin_open_memstream found in the prefix"
        #endif

        #ifdef __has_builtin
            #if !__has_builtin(__builtin_open_memstream)
                #error "__builtin_open_memstream not found"
            #endif
        #elif ! defined(open_memstream)
            __builtin_open_memstream;
        #endif
        return 0;
        }
Compiler stdout:
 
Compiler stderr:
 /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpqkgh33b_/testfile.c: In function 'main':
/home/rgommers/code/bldscipy/buildwheel/meson-private/tmpqkgh33b_/testfile.c:12:14: error: #error "No definition for __builtin_open_memstream found in the prefix"
   12 |             #error "No definition for __builtin_open_memstream found in the prefix"
      |              ^~~~~
/home/rgommers/code/bldscipy/buildwheel/meson-private/tmpqkgh33b_/testfile.c:20:13: error: '__builtin_open_memstream' undeclared (first use in this function)
   20 |             __builtin_open_memstream;
      |             ^~~~~~~~~~~~~~~~~~~~~~~~
/home/rgommers/code/bldscipy/buildwheel/meson-private/tmpqkgh33b_/testfile.c:20:13: note: each undeclared identifier is reported only once for each function it appears in

Checking for function "open_memstream" : NO 
Configuring messagestream_config.h using configuration
Running compile:
Working directory:  /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpao65w5ub
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpao65w5ub/testfile.cpp -o /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpao65w5ub/output.obj -c -fvisibility-inlines-hidden -std=c++17 -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -D_FILE_OFFSET_BITS=64 -O0 -fpermissive -fvisibility=hidden 

Code:
 extern int i;
int i;

Compiler stdout:
 
Compiler stderr:
 
Compiler for C++ supports arguments -fvisibility=hidden: YES 
Running compile:
Working directory:  /tmp/tmp3ekh9d9x
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ /tmp/tmp3ekh9d9x/testfile.cpp -o /tmp/tmp3ekh9d9x/output.obj -D_FILE_OFFSET_BITS=64 -c -std=c++14 

Code:
 int i = static_cast<int>(0);
Compiler stdout:
 
Compiler stderr:
 
Compiler accepts -std=c++14: YES
Running compile:
Working directory:  /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpxtjnop1j
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ -L/home/rgommers/anaconda3/envs/scipy-meson/lib /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpxtjnop1j/testfile.cpp -o /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpxtjnop1j/output.exe -fvisibility-inlines-hidden -std=c++17 -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -D_FILE_OFFSET_BITS=64 -O0 -fpermissive -std=c++14 -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,-rpath,/home/rgommers/anaconda3/envs/scipy-meson/lib -Wl,-rpath-link,/home/rgommers/anaconda3/envs/scipy-meson/lib 

Code:
 #include<stdio.h>
        
        int main(void) {
            printf("%ld\n", (long)(sizeof(void*)));
            return 0;
        }
Compiler stdout:
 
Compiler stderr:
 
Program stdout:

8

Program stderr:


Checking for size of "void*" : 8
Run-time dependency threads found: YES
Running compile:
Working directory:  /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpbbfx2o5f
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ -L/home/rgommers/anaconda3/envs/scipy-meson/lib /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpbbfx2o5f/testfile.cpp -o /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpbbfx2o5f/output.exe -fvisibility-inlines-hidden -std=c++17 -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -D_FILE_OFFSET_BITS=64 -O0 -fpermissive -std=c++14 -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,-rpath,/home/rgommers/anaconda3/envs/scipy-meson/lib -Wl,-rpath-link,/home/rgommers/anaconda3/envs/scipy-meson/lib 

Code:
 #include<stdio.h>
        
        int main(void) {
            printf("%ld\n", (long)(sizeof(void*)));
            return 0;
        }
Compiler stdout:
 
Compiler stderr:
 
Program stdout:

8

Program stderr:


Checking for size of "void*" : 8
Dependency threads found: YES unknown (cached)
Using cached compile:
Cached command line:   

Code:
 extern int i;
int i;

Cached compiler stdout:
 
Cached compiler stderr:
 
Compiler for C++ supports arguments -fvisibility=hidden: YES (cached)
Using cached compile:
Cached command line:   

Code:
 extern int i;
int i;

Cached compiler stdout:
 
Cached compiler stderr:
 
Compiler for C supports arguments -Wno-unused-but-set-variable: YES (cached)
Dependency openblas found: YES 0.3.12 (cached)
Using cached compile:
Cached command line:   

Code:
 extern int i;
int i;

Cached compiler stdout:
 
Cached compiler stderr:
 
Compiler for C supports arguments -Wno-unused-but-set-variable: YES (cached)
Running compile:
Working directory:  /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpiy5b7yld
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpiy5b7yld/testfile.cpp -o /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpiy5b7yld/output.obj -c -fvisibility-inlines-hidden -std=c++17 -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -D_FILE_OFFSET_BITS=64 -O0 -fpermissive -Wformat-truncation -Wno-format-truncation 

Code:
 extern int i;
int i;

Compiler stdout:
 
Compiler stderr:
 
Compiler for C++ supports arguments -Wno-format-truncation: YES 
Running compile:
Working directory:  /home/rgommers/code/bldscipy/buildwheel/meson-private/tmp38z0eyqh
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ /home/rgommers/code/bldscipy/buildwheel/meson-private/tmp38z0eyqh/testfile.cpp -o /home/rgommers/code/bldscipy/buildwheel/meson-private/tmp38z0eyqh/output.obj -c -fvisibility-inlines-hidden -std=c++17 -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /home/rgommers/anaconda3/envs/scipy-meson/include -D_FILE_OFFSET_BITS=64 -O0 -fpermissive -Wclass-memaccess -Wno-class-memaccess 

Code:
 extern int i;
int i;

Compiler stdout:
 
Compiler stderr:
 
Compiler for C++ supports arguments -Wno-class-memaccess: YES 
Build targets in project: 204

SciPy 1.9.0.dev0

  User defined options
    Native files: /home/rgommers/code/bldscipy/.mesonpy-native-file.ini
    prefix      : /home/rgommers/anaconda3/envs/scipy-meson

Found ninja-1.10.2.git.kitware.jobserver-1 at /home/rgommers/anaconda3/envs/scipy-meson/bin/ninja
Running compile:
Working directory:  /tmp/tmpoynv8q1p
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-c++ /tmp/tmpoynv8q1p/testfile.cpp -o /tmp/tmpoynv8q1p/output.obj -D_FILE_OFFSET_BITS=64 -c -std=c++14 

Code:
 int i = static_cast<int>(0);
Compiler stdout:
 
Compiler stderr:
 
Compiler accepts -std=c++14: YES
Running compile:
Working directory:  /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpjf17_g59
Command line:  /home/rgommers/anaconda3/envs/scipy-meson/bin/x86_64-conda-linux-gnu-gfortran /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpjf17_g59/testfile.f90 -o /home/rgommers/code/bldscipy/buildwheel/meson-private/tmpjf17_g59/output.obj -c -Wno-conversion -D_FILE_OFFSET_BITS=64 -O0 --print-search-dirs 

Code:
 
Compiler stdout:
 install: /home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/
programs: =/home/rgommers/anaconda3/envs/scipy-meson/bin/../libexec/gcc/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../libexec/gcc/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/bin/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/bin/
libraries: =/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/lib/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/lib/../lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/lib/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/lib/../lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/usr/lib/x86_64-conda-linux-gnu/9.4.0/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/usr/lib/../lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../../x86_64-conda-linux-gnu/lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../lib/gcc/x86_64-conda-linux-gnu/9.4.0/../../../:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/lib/:/home/rgommers/anaconda3/envs/scipy-meson/bin/../x86_64-conda-linux-gnu/sysroot/usr/lib/

Compiler stderr:
 ]
```

</details>

FFY00:
I fixed up commits to handle this situation, can you reset to the remote and try again?

rgommers:
> Wheels build this way will not be reproducible, or rather, will be less reproducible, so I'd recommend avoiding this approach when building wheels for distribution, but it should be suitable for development.

Sure, for distribution one should always take care to build manylinux compliant wheels. And same for other OSes, need to make things relocatable.

rgommers:
Argh, why is `pip` dumb, first it refuses and then it overwrites a `conda`-installed `numpy`:
```
$ pip install scipy-1.9.0.dev0-cp39-cp39-linux_x86_64.whl 
Processing ./scipy-1.9.0.dev0-cp39-cp39-linux_x86_64.whl
Requirement already satisfied: numpy>=1.18.5 in /home/rgommers/anaconda3/envs/scipy-meson/lib/python3.9/site-packages (from scipy==1.9.0.dev0) (1.21.5)
scipy is already installed with the same version as the provided wheel. Use --force-reinstall to force an installation of the wheel.


(scipy-meson) rgommers@machine:~/code/bldscipy/dist (test-mesonpy)$ pip install scipy-1.9.0.dev0-cp39-cp39-linux_x86_64.whl --force-reinstall
Processing ./scipy-1.9.0.dev0-cp39-cp39-linux_x86_64.whl
Collecting numpy>=1.18.5
  Downloading numpy-1.22.1-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (16.8 MB)
     |████████████████████████████████| 16.8 MB 4.4 MB/s            
Installing collected packages: numpy, scipy
  Attempting uninstall: numpy
    Found existing installation: numpy 1.21.5
    Uninstalling numpy-1.21.5:
      Successfully uninstalled numpy-1.21.5
  Attempting uninstall: scipy
    Found existing installation: scipy 1.9.0.dev0
    Uninstalling scipy-1.9.0.dev0:
      Successfully uninstalled scipy-1.9.0.dev0
Successfully installed numpy-1.22.1 scipy-1.9.0.dev0
```

rgommers:
Test collection starting without errors, so looks promising. Let's see if the test suite gets to the finish line.

rgommers:
All tests passed 🎉 

That's enough for tonight:) Thanks Filipe!

FFY00:
Finally tracked down the origin of the CI fail on Python 3.7 :sweat_smile:

bnavigator:
> Have you thought about how to make this configurable? We established that there are valid reasons to do this bundling, and also valid reasons to skip it (xref [#20 (comment)](https://github.com/FFY00/meson-python/issues/20#issuecomment-1004346950)).

I think I read the whole thread, and please forgive me if I have missed something, but now that this is merged and released, how do I build a SciPy wheel without the vendoring? Vendoring libopenblas in the openSUSE RPM package is not an option. Yet I want to use meson-python so that I get the proper `.dist-info` metadata.

Right now, I get tons of rpmlint errors like this:
```
[ 1412s] python310-scipy.x86_64: E: binary-or-shlib-defines-rpath (Badness: 10000) /usr/lib64/python3.10/site-packages/scipy/_lib/_ccallback_c.cpython-310-x86_64-linux-gnu.so (RUNPATH: :$ORIGIN/../../.scipy.mesonpy.libs)
```

rgommers:
I think we can consider that a bug. Setting RPATH like that is needed _only_ in case there's a project-internal shared library. In this case, RPATH is still set unconditionally on Linux. 

I would say this is covered by gh-125. Except I didn't know it caused actual failures in Linux packaging. I'll comment there. Thanks @bnavigator

rgommers:
> Vendoring libopenblas in the openSUSE RPM package is not an option. 

Note that that is not what is actually happening right now. The RPATH entry just points at a non-existing directory.

PR #36, opened 2022-02-10
docs: add initial documentation

FFY00:


rgommers:
Nice.

https://meson-python.readthedocs.io/en/latest/ works, https://meson-python.readthedocs.io/en/stable/ still needs a trigger to build for the latest release - it's a dummy page from the README right now.

PR #37, opened 2022-02-10
Cross-platform support?

FFY00:


tests/test_wheel.py
```diff
@@ -34,6 +35,14 @@
 
 if platform.system() == 'Linux':
     PLATFORM_TAG = f'linux_{platform.machine()}'
+    SHARED_LIB_EXT = 'so'
+elif platform.system() == 'Darwin':
+    PLATFORM_TAG = '_'.join((
```
blink1073, line 40: I think this needs to use the same logic as `Project.platform_tag`

FFY00: Yeah, that's probably for the best.

blink1073:
You might need to use `PosixPath` or replace `os.sep` with backslash when forming paths on Windows to give to a subprocess.

tests/test_wheel.py
```diff
@@ -31,9 +32,14 @@
 else:
     raise NotImplementedError(f'Unknown implementation: {platform.python_implementation()}')
 
+PLATFORM_TAG = sysconfig.get_platform().replace('-', '_').replace('.', '_')
```
blink1073, line 35: Actually, it might make sense to do  what [`cibuildwheel`](https://github.com/pypa/cibuildwheel/blob/bc3106a4d0cc1f1a604806f1e3795b0051125124/cibuildwheel/macos.py#L202) does for the target mac version for both cases.

FFY00:
> You might need to use `PosixPath` or replace `os.sep` with backslash when forming paths on Windows to give to a subprocess.

Can you expand on this? `pathlib.Path` should already give us the correct separator, no?

blink1073:
`subprocess.CalledProcessError: Command '['meson', 'setup', '--native-file=D:\\a\\mesonpy\\mesonpy\\tests\\packages\\purelib-and-platlib\\.mesonpy-native-file.ini', '--prefix=C:\\hostedtoolcache\\windows\\Python\\3.7.9\\x64', 'D:\\a\\mesonpy\\mesonpy\\tests\\packages\\purelib-and-platlib', 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\mesonpy-55qv0lvi\\build']' returned non-zero exit status 1.`

In my experience, those backslashes confuse `subprocess`.  I've had to replace `os.sep` with `/`, or use `PosixPath` to ensure `/` is used.

tests/test_wheel.py
FFY00: Perhaps, but I'd prefer to keep this simple until we need it. But that is good to keep in mind, thanks!

FFY00:
Hum, I haven't experienced that. In this case it seems like that is not the root of the problem, but Meson failing to build the test project. Thanks for pointing it out, I will keep in mind :sweat_smile:

blink1073:
FWIW, I tried replacing a few instances of `pathlib.Path` with `pathlib.PurePosixPath` and was able to get [farther along](https://github.com/blink1073/mesonpy/runs/5205735969?check_suite_focus=true).

blink1073:
Might need to use `vcpkg` based on this [issue](https://github.com/mesonbuild/meson/issues/4029).

Here is an [example](https://github.com/Microsoft/vcpkg/blob/master/ports/fribidi/portfile.cmake).

Alternatively, there is a `pkg-config` available for Windows from (conda-forge](https://anaconda.org/conda-forge/pkg-config/files)

blink1073:
As for the Python 3 lib failure, I think it might be related to venv on Windows: https://github.com/mesonbuild/meson/blob/fe2c0ff1e20403e4c0e26c705e3742d0e1ab1344/mesonbuild/modules/python.py#L198

rgommers:
> As for the Python 3 lib failure, I think it might be related to venv on Windows

That sounds right. For such a basic test, finding `libpython` should not require pkg-config/vcpkg/etc.

FFY00:
Thanks for the pointers. They are really helpful :blush:

rgommers:
Great! 🎉

What is the next step - testing the `main` branch with SciPy first, or making a new release?

Re the `xfail`s in the last commit: it may be useful to add a reference to the issue about this on the Meson repo (or file one if there's isn't one yet). That's going to save some searching later on.

FFY00:
Yes, I am planning on doing that. I did do an initial search and am not sure if this is exactly the same as the issue I found, so I need to look into it a bit more, but I did verify that the problem is indeed that Meson isn't finding the Python installation.

PR #58, opened 2022-05-20
mesonpy: include uncommited changes in sdist

### LINKED ISSUE #53: Do not ignore uncommitted changes in repo
`meson-python` should pick up uncommitted changes to files that are under version control. xref https://github.com/scipy/scipy/pull/16187 with a lot of details. The main reasons are:
1. It is very confusing to silently ignore changes, it will lead to lost time by users and contributors.
2. Even if you understand it, it's quite cumbersome to have to commit changes before running `pip install .` (or similar) each time.
3. For complex packages, it is very common that patches have to be applied before building sdists, wheels, or other packages (conda/rpm/deb/etc.). It is now extremely hard to do so.

Reproducible releases should be made through a CI job anyway, so "the current behavior is more reproducible" doesn't really apply. And I don't think there is any other reason to keep the current behavior?

FFY00:
Fixes #53

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 import collections
 import contextlib
 import functools
+import io
 import itertools
 import json
 import os
@@ -23,6 +24,7 @@ import shutil
 import subprocess
 import sys
 import sysconfig
+import tarfile
 import tempfile
 import textwrap
 import typing
@@ -538,7 +540,8 @@ class Project():
         assert isinstance(version, str)
         return version
 
-    @property
+    @property  # type: ignore[misc]
+    @functools.lru_cache(maxsize=1)
     def metadata(self) -> bytes:  # noqa: C901
         """Project metadata."""
         # the rest of the keys are only available when using PEP 621 metadata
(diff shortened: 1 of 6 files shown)


FFY00:
@rgommers can you have a look?

rgommers (commented):
This doesn't seem to work. It now builds an sdist rather than error out, and shows this warning:
```
WARNING: Repository has uncommitted changes that will not be included in the dist tarball
```
However, the uncommitted changes are not actually included in the sdist - and including desired changes was the goal here.

FFY00:
Ah, sorry, I misread the Meson commit. Then it is non-trivial to implement this, I will have a look later.

rgommers:
https://github.com/scipy/scipy/pull/16187#issuecomment-1133294137 points to the Flit code for this. Flit does seem to have the behavior we want: do not include files that are not tracked, but do include uncommitted changes to tracked files. The code in Flit to do this is quite concise. Seems like a good option to just copy that.

eli-schwartz:
> Ah, sorry, I misread the Meson commit. Then it is non-trivial to implement this, I will have a look later.

Yeah, Meson doesn't include any prepackaged support for doing this. :( Any way you slice it, `meson dist` will want to use `git archive` as a transfer mechanism and I don't think that *can* handle uncommitted changes; it fundamentally utilizes a git tree ref.

Previous versions of Meson actually warned you on uncommitted changes, because those changes would not be disted. The most recent release outright errors, and offers a flag to restore the 0.61.0 behavior of merely warning you that those changes won't be included. The logic behind the change is that it was easy to not notice the warning, so we changed the emphasis of the warning itself.

The flit approach is pretty good if your goal is to include uncommitted changes. It does mean that you'll no longer get some things that git-archive does, such as gitattributes `export-ignore` / `export-subst`, but SciPy isn't using either one so that probably does not matter.

FFY00:
@rgommers can you give this a try before merging just to confirm it indeed works as expected for SciPy? Thanks!

rgommers:
Thanks @FFY00, this appears to work. At least I tested editing a `.py` file, running `$ python -m build --sdist --skip-dependency-check --no-isolation` to pick up `meson-python` from this branch, and then getting a built sdist that indeed includes that change to the `.py` file.

One small thing to improve perhaps, this warning may be confusing:
```bash
+ meson dist --allow-dirty --no-tests --formats gztar                                               
WARNING: Repository has uncommitted changes that will not be included in the dist tarball
Created /home/rgommers/code/scipy/.mesonpy-dqpt326g/build/meson-dist/SciPy-1.9.0.dev0.tar.gz
Successfully built scipy-1.9.0.dev0.tar.gz
```
How about logging right above it something like _"`meson-python` detected changes to tracked files and will include those in the sdist (please ignore the warning 'Repository has uncommitted changes ...' below)"_ ?

Whether `pip install .` with build isolation works, I can only test after 0.5.0 is up on PyPI I think. Will reserve some time for that tomorrow morning.

FFY00:
> How about logging right above it something like _"`meson-python` detected changes to tracked files and will include those in the sdist (please ignore the warning 'Repository has uncommitted changes ...' below)"_ ?

That makes sense, but we are currently not able to detect this, and it would probably require some work to do so. I will push a 0.5.0 with this PR as-is for now and then revisit this.

PR #63, opened 2022-05-24
mesonpy: make pep621 a hard dependency

### LINKED ISSUE #56: `pep621` is not optional but the dependency isn't declared
gh-33 talks about replacing `pep621` but really seems to mean "rename" rather than replace. There seems to be a more urgent issue here: `pep621` is imported here: https://github.com/FFY00/meson-python/blob/main/mesonpy/__init__.py#L552-L553 but is not declared as a dependency. So I'm seeing:
```
ModuleNotFoundError: No module named 'pep621'
```
in the CI job that tests building via sdist for SciPy ([log](https://dev.azure.com/scipy-org/SciPy/_build/results?buildId=18207&view=logs&j=5428a243-ce9d-5a90-882b-245b865f3bbe&t=61cfec6f-65f8-54f8-b88d-e2ddf9ae8303). Why is a non-optional dependency not declared? The comment even explicitly discusses raising an exception - it should not be necessary for `meson-python` users like SciPy to list `pep621` as a direct dependency. It should be either vendored, declared as non-optional, or made actually optional.

FFY00:
Fixes #56

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -54,7 +54,6 @@ __version__ = '0.4.0'
 class _depstr:
     ninja = 'ninja >= 1.10.0'
     patchelf_wrapper = 'patchelf-wrapper'
-    pep621 = 'pep621 >= 0.3.0'
     wheel = 'wheel >= 0.36.0'  # noqa: F811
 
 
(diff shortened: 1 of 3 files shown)


rgommers:
Follow-up: the `ninja` dependency is already present in `pyproject.toml` as well, so can go too.

FFY00:
Yeah, I noticed that when touching the code, I think I moved it a while back.

PR #66, opened 2022-05-28
Clean up duplicate dependency on ninja

rgommers:
`ninja` is a dependency in `pyproject.toml` already, so it can be removed in `__init__.py`.

Follows up on https://github.com/FFY00/meson-python/pull/63#issuecomment-1136311433


FFY00 (approved):
Thanks!

PR #78, opened 2022-06-10
CI: Run tests in Cygwin for Cygwin CI

DWesl:
The current version installs the Cygwin base packages (but not any python versions), then runs the rest of the steps using the previously-installed Windows-native python.  This should run the tests in a Cygwin python.

### DIFF
--- a/.github/workflows/tests-cygwin.yml
+++ b/.github/workflows/tests-cygwin.yml
@@ -33,7 +33,7 @@ jobs:
           packages: >-
             python3${{ matrix.python-minor }}
             python3${{ matrix.python-minor }}-pip
-            cmake make ninja gcc-core
+            cmake make ninja gcc-core gcc-g++
 
       - name: Install nox
         shell: bash.exe -eo pipefail -o igncr "{0}"
--- a/.github/workflows/tests-cygwin.yml
+++ b/.github/workflows/tests-cygwin.yml
@@ -44,7 +44,7 @@ jobs:
             python3${{ matrix.python-minor }}
             python3${{ matrix.python-minor }}-pip
             cmake make ninja gcc-core gcc-g++ git
-            gnupg gnupg2 meson
+            gnupg gnupg2 meson pkg-config
 
       - name: Tell Cygwin git about this repository
         # This addresses the "fatal: detected dubious ownership in
--- a/.github/workflows/tests-cygwin.yml
+++ b/.github/workflows/tests-cygwin.yml
@@ -56,7 +56,6 @@ jobs:
           TMP: "/tmp"
           TEMP: "/tmp"
         run: |
-          cd "${GITHUB_WORKSPACE}"
           nox -s test-3.${{ matrix.python-minor }}
 
       - name: Send coverage report
--- a/.github/workflows/tests-cygwin.yml
+++ b/.github/workflows/tests-cygwin.yml
@@ -42,7 +42,7 @@ jobs:
             python3${{ matrix.python-minor }}
             python3${{ matrix.python-minor }}-pip
             cmake make ninja gcc-core gcc-g++ git
-            gnupg gnupg2
+            gnupg gnupg2 meson
 
       - name: Install nox
         shell: bash.exe -eo pipefail -o igncr "{0}"
--- a/.github/workflows/tests-cygwin.yml
+++ b/.github/workflows/tests-cygwin.yml
@@ -17,6 +17,8 @@ jobs:
     runs-on: windows-latest
     env:
       FORCE_COLOR: true
+      SHELLOPTS: igncr
+      CHERE_INVOKING: true
     strategy:
       fail-fast: false
       matrix:
@@ -43,6 +45,15 @@ jobs:
             python3${{ matrix.python-minor }}-pip
             cmake make ninja gcc-core gcc-g++ git
             gnupg gnupg2 meson
+
+      - name: Tell Cygwin git about this repository
+        # This addresses the "fatal: detected dubious ownership in
+        # repository" and "fatal: not in a git directory" errors
+        # encountered when trying to run Cygwin git in a directory not
+        # owned by the current user. This happens when the tests run
+        # Cygwin git in a directory outside the Cygwin filesystem.
+        run: git config --global --add safe.directory '*'
+        shell: C:\cygwin\bin\env.exe CYGWIN_NOWINPATH=1 CHERE_INVOKING=1 C:\cygwin\bin\bash.exe -leo pipefail -o igncr {0}
 
       - name: Install nox
         shell: bash.exe -eo pipefail -o igncr "{0}"
--- a/.github/workflows/tests-cygwin.yml
+++ b/.github/workflows/tests-cygwin.yml
@@ -30,6 +30,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - uses: actions/cache@v3
+        with:
+          path: 'C:\cygwin'
+          key: ${{ runner.os }}-cygwin-python-3${{ matrix.python-minor }}-${{ hashfiles('pyproject.toml') }}
+
       - name: Setup Cygwin
         uses: cygwin/cygwin-install-action@v2
         with:
--- a/.github/workflows/tests-cygwin.yml
+++ b/.github/workflows/tests-cygwin.yml
@@ -33,7 +33,7 @@ jobs:
           packages: >-
             python3${{ matrix.python-minor }}
             python3${{ matrix.python-minor }}-pip
-            cmake make ninja
+            cmake make ninja gcc-core
 
       - name: Install nox
         shell: bash.exe -eo pipefail -o igncr "{0}"
--- a/.github/workflows/tests-cygwin.yml
+++ b/.github/workflows/tests-cygwin.yml
@@ -33,7 +33,7 @@ jobs:
           packages: >-
             python3${{ matrix.python-minor }}
             python3${{ matrix.python-minor }}-pip
-            cmake make ninja gcc-core gcc-g++
+            cmake make ninja gcc-core gcc-g++ git
 
       - name: Install nox
         shell: bash.exe -eo pipefail -o igncr "{0}"
--- a/.github/workflows/tests-cygwin.yml
+++ b/.github/workflows/tests-cygwin.yml
@@ -24,6 +24,9 @@ jobs:
           - '9'
 
     steps:
+      - name: Tell git to use proper newlines
+        run: git config --global core.autocrlf input
+
       - name: Checkout
         uses: actions/checkout@v2
 
@@ -39,6 +42,8 @@ jobs:
         shell: bash.exe -eo pipefail -o igncr "{0}"
         env:
           PATH: "/bin:/usr/bin:/usr/local/bin:/usr/lib/lapack"
+          TMP: "/tmp"
+          TEMP: "/tmp"
         run: |
             /usr/bin/python -m pip install nox
             nox --version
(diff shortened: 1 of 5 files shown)


FFY00 (approved):
Hi @DWesl, thank you so much for looking into this! It is highly appreciated.

The PR looks great, I just want to wait for a new GitPython release before merging so that we can be sure that is why the tests are currently failing.

FFY00:
I think it's gonna be a while, so I am gonna rebase and merge this PR.

tests/test_wheel.py
```diff
@@ -54,6 +54,10 @@
     SHARED_LIB_EXT = 'dylib'
 elif platform.system() == 'Windows':
     SHARED_LIB_EXT = 'pyd'
+elif platform.system().startswith('CYGWIN'):
+    # I'm using import library (link-time) rather than dynamic library (run-time)
+    # because that's what was already here for Cygwin
+    SHARED_LIB_EXT = 'dll.a'
```
rgommers, line 60: This should be `.pyd` as well I guess? Either that or `.so` or `.dll`. But `dll.a` is the import library, not the extension module.

DWesl:
> I think it's gonna be a while, so I am gonna rebase and merge this PR.

Should we install GitPython from the repository rather than from PyPI so the fix is included?  I think the tests still fail with the released version.

FFY00:
We are now installing from the repo but there are still errors, I don't know what's going on. @DWesl can you try running the tests with nox on a Cygwin environment?

dnicolodi:
@DWesl The current way to generate wheel tags from extension filenames is doomed. I think the sane approach is the one in #202. It should already work with Cygwin but IIUC our CI setup is incomplete. Do you mind if I cherry pick the CI extension commits from this PR into that one?

DWesl:
> We are now installing from the repo but there are still errors, I don't know what's going on. @DWesl can you try running the tests with nox on a Cygwin environment?

The most recent commit passes locally, except for what I suspect are local network errors.



> @DWesl The current way to generate wheel tags from extension filenames is doomed. I think the sane approach is the one in #202. It should already work with Cygwin but IIUC our CI setup is incomplete. Do you mind if I cherry pick the CI extension commits from this PR into that one?

You are welcome to try it.

dnicolodi:
CI is not very happy. There seems to be something wrong with `git` in Cygwin.

.github/workflows/tests-cygwin.yml
```diff
@@ -56,7 +56,6 @@ jobs:
           TMP: "/tmp"
           TEMP: "/tmp"
         run: |
-          cd "${GITHUB_WORKSPACE}"
```
dnicolodi: The code that fails is this https://github.com/mesonbuild/meson-python/blob/ef67e0e03c3224a8703809ce6eafb8ac5ce352d5/tests/conftest.py#L33-L51 which is called for example here https://github.com/mesonbuild/meson-python/blob/ef67e0e03c3224a8703809ce6eafb8ac5ce352d5/tests/conftest.py#L96
It should not depend on CWD of the tests runner. Why it fails although is not very clear.

DWesl: I was going to say "GitPython hasn't done a release since they merged the changes to get it working on Cygwin", but apparently that's only true on GitHub; PyPI had two releases a month ago.

`actions/checkout` is done with Windows git, not Cygwin, so there would be (different) errors operating on the `meson-python` repository, but the errors are apparently from a brand-new repository, so that's not it either.

My next guess would be `fork()` failure since CI runs into that a decent bit, but this project doesn't have many dependencies, nor does it have many C extension modules, and also I think I've found the actual `stdout` and `stderr` in the logs, which show a different message.

There's no reasons coming to mind for why `rm -rf .git; git init; git config gpgsign false` would fail, especially with that error

dnicolodi: Indeed. By the way @DWesl , do you know if there is a n easy way to cache the cygwin installation done by the `cygwin/cygwin-install-action`? Debugging this on the CI having to wait for Cygwin to install at each run is a bit tedious. On a branch I setup caching for the Python dependencies, cutting down the `pip install` time considerably, but installing Cygwin still takes anything between one minute and a half and three minutes.

dnicolodi: The fix for the `git` issue is this: https://github.com/mesonbuild/meson-python/blob/68e1d391e5a81d588da35ebaabe908473dbcbfee/.github/workflows/tests.yml#L91-L98

.github/workflows/tests-cygwin.yml
```diff
@@ -30,6 +30,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - uses: actions/cache@v3
+        with:
+          path: 'C:\cygwin'
+          key: ${{ runner.os }}-cygwin-python-3${{ matrix.python-minor }}-${{ hashfiles('pyproject.toml') }}
+
```
dnicolodi, line 37: This is not enough. With this the installation path is populated but the installation will be run anyway. I don't think this saves any time. Anyhow, I can live with the CI job taking a couple of minutes.

.github/workflows/tests-cygwin.yml
DWesl: So it's the second issue, but git for some reason decided to have two error messages for that.  Lovely

.github/workflows/tests-cygwin.yml
DWesl: The installation will be run, but that cache includes the list of installed packages; the install step will only download and extract packages if there are new versions (which means I should expire the cache at least once a month).  I would very much prefer to run the installer to be sure the dlls get rebased properly (which means I need to run `rebase-trigger full` so it knows to do that).

You're probably going to have more runins with the length of the CI job; most of the ones I've built take half an hour.  Should I take this back out again?

dnicolodi: I don't think we will be merging this PR either way. The work in #202 supersedes all of it. I just have one funny test error on Cygwin to track down: https://github.com/mesonbuild/meson-python/actions/runs/3456021871/jobs/5768534392 We setup two virtual environments as part of the tests, for some reason the second time we do it, it fails. The annoying thing is that it fails with a backtrace that does not match my copy of the CPython 3.9.15 source code and I haven't been able to locate the Cygwin Python sources

DWesl: If you've got a Cygwin install, run `setup-x86_64.exe` again and check the `Src?` tick box; you may also want to install the `cygport` package to simplify the build process (`cygport python39.cygport download prep compile test install package` if you want to do a local install, using [the Cygwin source package mentioned above](https://cygwin.com/cgi-bin2/package-cat.cgi?file=x86_64%2Fpython39-src%2Fpython39-3.9.10-1-src&grep=python39-src) and [a local download repository](https://cygwin.com/package-server.html))

dnicolodi: I don't have any Windows machine were to install Cygwin. I was hoping that a git repository with the sources exists somewhere. With your indications above I found it, https://cygwin.com/git-cygwin-packages/?p=git/cygwin-packages/python39.git but it seems the sources are maintained as a set of patches... This is so '90ies... 😄

dnicolodi: Wait, the git repo above has the history only up to Python 3.9.9 but version 3.9.10 is the current one. I guess the git repo is not where the development happens.

DWesl: > I don't have any Windows machine were to install Cygwin. I was hoping that a git repository with the sources exists somewhere. With your indications above I found it, https://cygwin.com/git-cygwin-packages/?p=git/cygwin-packages/python39.git but it seems the sources are maintained as a set of patches... This is so '90ies... 😄

Well, the source tarball from python.org and a bunch of packages that haven't been accepted upstream.  I think the standardized script to apply the packages, build and install the software, and create a distribution might be the main update to the package format since it was introduced back then.  

> Wait, the git repo above has the history only up to Python 3.9.9 but version 3.9.10 is the current one. I guess the git repo is not where the development happens.

Not really; the maintainer usually does the build locally from the upstream sources then uploads the resulting packages; synchronizing the build script with the repository is new within the past few years and not everyone has updated their muscle memory.

Poking around the mirror `cygwin/cygwin-install-action` uses by default, I found https://mirrors.kernel.org/sourceware/cygwin/x86_64/release/python39/python39-3.9.10-1-src.tar.xz, which should have everything you need.

As for the venv failure; I used nox here, which probably uses some virtual environment setup for dependency management and seems to get through the "install dependencies" part just fine; running `python -m venv test-venv; . test-venv/bin/activate; python -Im ensurepip --upgrade --default-pip` worked fine as well, so I'm not sure what's going on there.

dnicolodi: Thanks. I'm looking into this. The venv creation that fails is the second one we do in code in the testsuite. See https://github.com/mesonbuild/meson-python/blob/ef67e0e03c3224a8703809ce6eafb8ac5ce352d5/tests/conftest.py#L62-L82

DWesl: That bit of the traceback looks like it was added in https://cygwin.com/git-cygwin-packages/?p=git/cygwin-packages/python39.git;a=blob;f=00189-use-rpm-wheels.patch;h=e5cf8f7831a34d4d8a0a6b543358d495cd182ace;hb=42494e325a050ba03638568d7318f8f0075e25fb
which seems to have been borrowed from Fedora or a related distribution.

DWesl: It looks like it's failing because it's not finding wheels in `/usr/share/python-wheels`; would running `python -m pip wheel --wheel-dir /usr/share/python-wheels wheel setuptools pip` help with that?

dnicolodi: Thank you for the pointers, this helps a lot! Putting these wheels there should be responsibility of the system vendor, not something we need to patch. I need to debug this a bit more. At least now I know what to look for.

dnicolodi: Lovely, `/usr/share/python-wheels/` does not even exist, despite I explicitly install the `python39-pip` and `python39-setuptools` packages

dnicolodi: @DWesl can you please check the content of `/usr/share/python-wheels/` on your Cygwin system? Also, what does it happen if you do something like `python -m venv $(mktemp -d)`?

dnicolodi: Accordingly to https://cygwin.com/cgi-bin2/package-grep.cgi there is no Cygwin package that installs something in` /usr/share/python-wheels/ ` so I am not sure I understand how the patch you linked is even supposed to work.

DWesl: > @DWesl can you please check the content of `/usr/share/python-wheels/` on your Cygwin system? 

I have a few hundred wheels in there, because I often create my own packages and make sure my packages install wheels there so I don't have to sit waiting for scipy to build every time I want to test something with tox.  I think it worked fine before, but I haven't checked.

> Also, what does it happen if you do something like `python -m venv $(mktemp -d)`?

Works the same as the other one, once I found the temporary environment.

It looks like the function assumes there is some file for the package, but only looks for architecture independent packages; is it perhaps finding only sdists and architecture-dependent wheels?

dnicolodi: Uh? Why do you have wheel files in there? Also, wheel files are only a distribution bundle, they need to be unpacked to be used, so if you install a wheel package you don't have any wheel file at the end, except in a cache, maybe. However, `/usr/share/python-wheels/` is not the default path of the `pip` cache. 

When `venv` creates a new virtual environment it optionally populates it with the `setuptools` and `pip` packages. Because `pip` does not exist yet in the virtual environment, the `ensurepip` module is used for doing it. Customarily, the Python distribution comes with two wheel files for `setuptools` and `pip` bundled with it and these are unpacked by `ensurepip` into the virtual environment. Cygwin decided to patch `ensurepip` to look for the wheel it needs in `/usr/share/python-wheels/` but it does not place any wheel there. How is this supposed to work? Who should place the wheels there?

Now that we understood the issue, I'm patching it in the CI setup. But Python as distribute by Cygwin is broken.

dnicolodi: The fix https://github.com/mesonbuild/meson-python/blob/f0ce0f55f98128f477246842ca141729f10ecee5/.github/workflows/tests.yml#L117-L128
Now the tests succeed on Cygwin too!

DWesl: Success!

DWesl:
Superseded by #202

PR #83, opened 2022-06-20
Support the maintenance branches of CPython

tacaswell:
The version reported by the 3.10 CPython branch is "3.10.5+" which fails the
metadata check.

This strips any trailing '+'.


FFY00 (approved):
Thanks! I missed this :sweat_smile:

PR #87, opened 2022-06-22
ENH: add support for editable wheels

FFY00:
~~This is just an initial approach, that is known to not work on all
use-cases. It simply uses .pth files to inject search locations for the
default import finder, for this reason generated files that should be
installed inside modules will not be available.~~

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -49,6 +49,56 @@ frontend, like ``builddir``, to specify the build directory to use/re-use. You
 can find more information about them in the `build options page`_.
 
 
+Editable installs
+-----------------
+
+Editable installs allow you to install the project in such a way where you can
+edit the project source and have those changes be reflected in the installed
+module, without having to explicitly rebuild and reinstall.
+
+The way it works on ``meson-python`` specifically, is that when you import a
+module of the project, it will be rebuilt on the fly. This means that project
+imports will take more time than usual.
+
+You can use pip to install the project in editable mode.
+
+
+.. code-block::
+
+   python -m pip install -e .
+
+
+It might be helpful to see the output of the Meson_ commands. This is offered
+as a **provisional** feature, meaning it is subject to change.
+
+If you want to temporarily enable the output, you can set the
+``MESONPY_EDITABLE_VERBOSE`` environment variable to a non-empty value. If this
+environment variable is present during import, the Meson_ commands and their
+output will be printed.
+
+
+.. code-block::
+
+   MESONPY_EDITABLE_VERBOSE=1 python my_script.py
+
+
+
+This behavior can also be enabled by default by passing the ``editable-verbose``
+config setting when installing the project.
+
+
+.. code-block::
+
+   python -m pip install -e . --config-settings editable-verbose=true
+
+
+This way, you won't need to always set ``MESONPY_EDITABLE_VERBOSE`` environment
+variable, the Meson_ commands and their output will always be printed.
+
+The ``MESONPY_EDITABLE_VERBOSE`` won't have any effect during the project
+install step.
+
+
 How does it work?
 =================
 
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -49,6 +49,7 @@ else:
 
 import mesonpy._compat
 import mesonpy._elf
+import mesonpy._introspection
 import mesonpy._tags
 import mesonpy._util
 import mesonpy._wheelfile
@@ -146,56 +147,6 @@ def _as_python_declaration(value: Any) -> str:
     elif isinstance(value, Iterable):
         return '[' + ', '.join(map(_as_python_declaration, value)) + ']'
     raise NotImplementedError(f'Unsupported type: {type(value)}')
-
-
-@functools.lru_cache()
-def _debian_python() -> bool:
-    """Check if we are running on Debian-patched Python."""
-    if sys.version_info >= (3, 10):
-        return 'deb_system' in sysconfig.get_scheme_names()
-    try:
-        import distutils
-        try:
-            import distutils.command.install
-        except ModuleNotFoundError:
-            raise ModuleNotFoundError('Unable to import distutils, please install python3-distutils')
-        return 'deb_system' in distutils.command.install.INSTALL_SCHEMES
-    except ModuleNotFoundError:
-        return False
-
-
-@functools.lru_cache()
-def _debian_distutils_paths() -> Mapping[str, str]:
-    # https://ffy00.github.io/blog/02-python-debian-and-the-install-locations/
-    assert sys.version_info < (3, 12) and _debian_python()
-
-    with warnings.catch_warnings():
-        warnings.filterwarnings('ignore', category=DeprecationWarning)
-        import distutils.dist
-
-    distribution = distutils.dist.Distribution()
-    install_cmd = distribution.get_command_obj('install')
-    install_cmd.install_layout = 'deb'  # type: ignore[union-attr]
-    install_cmd.finalize_options()  # type: ignore[union-attr]
-
-    return {
-        'data': install_cmd.install_data,  # type: ignore[union-attr]
-        'platlib': install_cmd.install_platlib,  # type: ignore[union-attr]
-        'purelib': install_cmd.install_purelib,  # type: ignore[union-attr]
-        'scripts': install_cmd.install_scripts,  # type: ignore[union-attr]
-    }
-
-
-@functools.lru_cache()
-def _sysconfig_paths() -> Mapping[str, str]:
-    sys_vars = sysconfig.get_config_vars().copy()
-    sys_vars['base'] = sys_vars['platbase'] = sys.base_prefix
-    if _debian_python():
-        if sys.version_info >= (3, 10):
-            return sysconfig.get_paths('deb_system', vars=sys_vars)
-        else:
-            return _debian_distutils_paths()
-    return sysconfig.get_paths(vars=sys_vars)
 
 
 class Error(RuntimeError):
@@ -461,9 +412,9 @@ class _WheelBuilder():
         origin file and the Meson destination path.
         """
         warnings.warn('Using heuristics to map files to wheel, this may result in incorrect locations')
-        sys_paths = _sysconfig_paths()
+        sys_paths = mesonpy._introspection.SYSCONFIG_PATHS
         # Try to map to Debian dist-packages
-        if _debian_python():
+        if mesonpy._introspection.DEBIAN_PYTHON:
             search_path = origin
             while search_path != search_path.parent:
                 search_path = search_path.parent
@@ -634,7 +585,7 @@ class _WheelBuilder():
         rebuild_commands = self._project.build_commands(install_path)
 
         import_paths = set()
-        for name, raw_path in _sysconfig_paths().items():
+        for name, raw_path in mesonpy._introspection.SYSCONFIG_PATHS.items():
             if name not in ('purelib', 'platlib'):
                 continue
             path = pathlib.Path(raw_path)
@@ -833,7 +784,7 @@ class Project():
         We will try to reconfigure the build directory if possible to avoid
         expensive rebuilds.
         """
-        sys_paths = _sysconfig_paths()
+        sys_paths = mesonpy._introspection.SYSCONFIG_PATHS
         setup_args = [
             f'--prefix={sys.base_prefix}',
             os.fspath(self._source_dir),
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -597,7 +597,7 @@ class _WheelBuilder():
 
         return wheel_file
 
-    def build_editable(self, directory: Path) -> pathlib.Path:
+    def build_editable(self, directory: Path, verbose: bool = False) -> pathlib.Path:
         self._project.build()  # ensure project is built
 
         wheel_file = pathlib.Path(directory, f'{self.name}.whl')
@@ -631,6 +631,7 @@ class _WheelBuilder():
                     import_paths={_as_python_declaration(import_paths)},
                     top_level_modules={_as_python_declaration(top_level_modules)},
                     rebuild_commands={_as_python_declaration(rebuild_commands)},
+                    verbose={verbose},
                 )
             ''').strip().encode()
             whl.writestr(
@@ -672,10 +673,12 @@ class Project():
         working_dir: Path,
         build_dir: Optional[Path] = None,
         meson_args: Optional[MesonArgs] = None,
+        editable_verbose: bool = False,
     ) -> None:
         self._source_dir = pathlib.Path(source_dir).absolute()
         self._working_dir = pathlib.Path(working_dir).absolute()
         self._build_dir = pathlib.Path(build_dir).absolute() if build_dir else (self._working_dir / 'build')
+        self._editable_verbose = editable_verbose
         self._install_dir = self._working_dir / 'install'
         self._meson_native_file = self._source_dir / '.mesonpy-native-file.ini'
         self._meson_cross_file = self._source_dir / '.mesonpy-cross-file.ini'
@@ -897,10 +900,11 @@ class Project():
         source_dir: Path = os.path.curdir,
         build_dir: Optional[Path] = None,
         meson_args: Optional[MesonArgs] = None,
+        editable_verbose: bool = False,
     ) -> Iterator[Project]:
         """Creates a project instance pointing to a temporary working directory."""
         with tempfile.TemporaryDirectory(prefix='.mesonpy-', dir=os.fspath(source_dir)) as tmpdir:
-            yield cls(source_dir, tmpdir, build_dir, meson_args)
+            yield cls(source_dir, tmpdir, build_dir, meson_args, editable_verbose)
 
     @functools.lru_cache()
     def _info(self, name: str) -> Dict[str, Any]:
@@ -1110,7 +1114,7 @@ class Project():
         return file
 
     def editable(self, directory: Path) -> pathlib.Path:
-        file = self._wheel_builder.build_editable(directory)
+        file = self._wheel_builder.build_editable(directory, self._editable_verbose)
         assert isinstance(file, pathlib.Path)
         return file
 
@@ -1151,7 +1155,7 @@ def _project(config_settings: Optional[Dict[Any, Any]]) -> Iterator[Project]:
     meson_args_cli_keys = tuple(f'{key}-args' for key in meson_args_keys)
 
     for key in config_settings:
-        known_keys = ('builddir', *meson_args_cli_keys)
+        known_keys = ('builddir', 'editable-verbose', *meson_args_cli_keys)
         if key not in known_keys:
             import difflib
             matches = difflib.get_close_matches(key, known_keys, n=3)
@@ -1170,6 +1174,7 @@ def _project(config_settings: Optional[Dict[Any, Any]]) -> Iterator[Project]:
             key: config_settings.get(f'{key}-args', ())
             for key in meson_args_keys
         }),
+        editable_verbose=bool(config_settings.get('editable-verbose'))
     ) as project:
         yield project
 
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -385,6 +385,32 @@ class _WheelBuilder():
             return stable[0]
         return None
 
+    @property
+    def top_level_modules(self) -> Collection[str]:
+        modules = set()
+        for type_ in self._wheel_files:
+            for path, _ in self._wheel_files[type_]:
+                top_part = path.parts[0]
+                # file module
+                if top_part.endswith('.py'):
+                    modules.add(top_part[:-3])
+                else:
+                    # native module
+                    for extension in _EXTENSION_SUFFIXES:
+                        if top_part.endswith(extension):
+                            modules.add(top_part[:-len(extension)])
+                            # XXX: We assume the order in _EXTENSION_SUFFIXES
+                            #      goes from more specific to last, so we go
+                            #      with the first match we find.
+                            break
+                    else:  # nobreak
+                        # skip Windows import libraries
+                        if top_part.endswith('.a'):
+                            continue
+                        # package module
+                        modules.add(top_part)
+        return modules
+
     def _is_native(self, file: Union[str, pathlib.Path]) -> bool:
         """Check if file is a native file."""
         self._project.build()  # the project needs to be built for this :/
@@ -605,7 +631,6 @@ class _WheelBuilder():
         wheel_file = pathlib.Path(directory, f'{self.name}.whl')
 
         install_path = self._source_dir / '.mesonpy' / 'editable' / 'install'
-        top_level_modules = self._project.top_level_modules
         rebuild_commands = self._project.build_commands(install_path)
 
         import_paths = set()
@@ -631,7 +656,7 @@ class _WheelBuilder():
                     project_path={_as_python_declaration(self._source_dir)},
                     build_path={_as_python_declaration(self._build_dir)},
                     import_paths={_as_python_declaration(import_paths)},
-                    top_level_modules={_as_python_declaration(top_level_modules)},
+                    top_level_modules={_as_python_declaration(self.top_level_modules)},
                     rebuild_commands={_as_python_declaration(rebuild_commands)},
                     verbose={verbose},
                 )
@@ -973,34 +998,6 @@ class Project():
             version = self._meson_version
         assert isinstance(version, str)
         return version
-
-    @property
-    def top_level_modules(self) -> Collection[str]:
-        modules = set()
-        for type_ in self._install_plan:
-            for _, details in self._install_plan[type_].items():
-                path = pathlib.Path(details['destination'])
-                if len(path.parts) >= 2 and path.parts[0]:
-                    top_part = path.parts[1]
-                    # file module
-                    if top_part.endswith('.py'):
-                        modules.add(top_part[:-3])
-                    else:
-                        # native module
-                        for extension in _EXTENSION_SUFFIXES:
-                            if top_part.endswith(extension):
-                                modules.add(top_part[:-len(extension)])
-                                # XXX: We assume the order in _EXTENSION_SUFFIXES
-                                #      goes from more specific to last, so we go
-                                #      with the first match we find.
-                                break
-                        else:  # nobreak
-                            # skip Windows import libraries
-                            if top_part.endswith('.a'):
-                                continue
-                            # package module
-                            modules.add(top_part)
-        return modules
 
     @cached_property
     def metadata(self) -> bytes:
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -49,18 +49,13 @@ import mesonpy._util
 import mesonpy._wheelfile
 
 from mesonpy._compat import (
-    Collection, Iterator, Literal, Mapping, ParamSpec, Path, typing_get_args
+    Collection, Iterator, Literal, Mapping, ParamSpec, Path, cached_property,
+    typing_get_args
 )
 
 
 if typing.TYPE_CHECKING:  # pragma: no cover
     import pyproject_metadata  # noqa: F401
-
-
-if sys.version_info >= (3, 8):
-    from functools import cached_property
-else:
-    cached_property = lambda x: property(functools.lru_cache(maxsize=None)(x))  # noqa: E731
 
 
 __version__ = '0.13.0.dev0'
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -301,6 +301,8 @@ class _WheelBuilder():
     @property
     def _debian_python(self) -> bool:
         """Check if we are running on Debian-patched Python."""
+        if sys.version_info >= (3, 10):
+            return 'deb_system' in sysconfig.get_scheme_names()
         try:
             import distutils
             try:
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -194,10 +194,14 @@ class _WheelBuilder():
         return bool(self._wheel_files['platlib'])
 
     @property
+    def normalized_name(self) -> str:
+        return self._project.name.replace('-', '_')
+
+    @property
     def basename(self) -> str:
         """Normalized wheel name and version (eg. meson_python-1.0.0)."""
         return '{distribution}-{version}'.format(
-            distribution=self._project.name.replace('-', '_'),
+            distribution=self.normalized_name,
             version=self._project.version,
         )
 
@@ -541,6 +545,9 @@ class _WheelBuilder():
                         self._install_path(whl, counter, origin, destination)
 
         return wheel_file
+
+    def build_editable(self, directory: Path) -> pathlib.Path:
+        self._build()
 
 
 MesonArgsKeys = Literal['dist', 'setup', 'compile', 'install']
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -1105,11 +1105,9 @@ class Project():
 
     def wheel(self, directory: Path) -> pathlib.Path:
         """Generates a wheel (binary distribution) in the specified directory."""
-        wheel = self._wheel_builder.build(self._build_dir)
-
-        final_wheel = pathlib.Path(directory, wheel.name)
-        shutil.move(os.fspath(wheel), final_wheel)
-        return final_wheel
+        file = self._wheel_builder.build(directory)
+        assert isinstance(file, pathlib.Path)
+        return file
 
     def editable(self, directory: Path) -> pathlib.Path:
         file = self._wheel_builder.build_editable(directory)
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -604,6 +604,8 @@ class _WheelBuilder():
             hook_module_name = f'_mesonpy_hook_{self.normalized_name.replace(".", "_")}'
             hook_install_code = textwrap.dedent(f'''
                 MesonpyFinder.install(
+                    project_name={_as_python_declaration(self._project.name)},
+                    hook_name={_as_python_declaration(hook_module_name)},
                     project_path={_as_python_declaration(self._source_dir)},
                     build_path={_as_python_declaration(self._build_dir)},
                     import_paths={_as_python_declaration(import_paths)},
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -519,7 +519,7 @@ class _WheelBuilder():
         Some files might need to be fixed up to set the RPATH to the internal
         library directory on Linux wheels for eg.
         """
-        location = os.fspath(destination).replace(os.path.sep, '/')
+        location = destination.as_posix()
         counter.update(location)
 
         # fix file
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -148,6 +148,56 @@ def _as_python_declaration(value: Any) -> str:
     raise NotImplementedError(f'Unsupported type: {type(value)}')
 
 
+@functools.lru_cache()
+def _debian_python() -> bool:
+    """Check if we are running on Debian-patched Python."""
+    if sys.version_info >= (3, 10):
+        return 'deb_system' in sysconfig.get_scheme_names()
+    try:
+        import distutils
+        try:
+            import distutils.command.install
+        except ModuleNotFoundError:
+            raise ModuleNotFoundError('Unable to import distutils, please install python3-distutils')
+        return 'deb_system' in distutils.command.install.INSTALL_SCHEMES
+    except ModuleNotFoundError:
+        return False
+
+
+@functools.lru_cache()
+def _debian_distutils_paths() -> Mapping[str, str]:
+    # https://ffy00.github.io/blog/02-python-debian-and-the-install-locations/
+    assert sys.version_info < (3, 12) and _debian_python()
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings('ignore', category=DeprecationWarning)
+        import distutils.dist
+
+    distribution = distutils.dist.Distribution()
+    install_cmd = distribution.get_command_obj('install')
+    install_cmd.install_layout = 'deb'  # type: ignore[union-attr]
+    install_cmd.finalize_options()  # type: ignore[union-attr]
+
+    return {
+        'data': install_cmd.install_data,  # type: ignore[union-attr]
+        'platlib': install_cmd.install_platlib,  # type: ignore[union-attr]
+        'purelib': install_cmd.install_purelib,  # type: ignore[union-attr]
+        'scripts': install_cmd.install_scripts,  # type: ignore[union-attr]
+    }
+
+
+@functools.lru_cache()
+def _sysconfig_paths() -> Mapping[str, str]:
+    sys_vars = sysconfig.get_config_vars().copy()
+    sys_vars['base'] = sys_vars['platbase'] = sys.base_prefix
+    if _debian_python():
+        if sys.version_info >= (3, 10):
+            return sysconfig.get_paths('deb_system', vars=sys_vars)
+        else:
+            return _debian_distutils_paths()
+    return sysconfig.get_paths(vars=sys_vars)
+
+
 class Error(RuntimeError):
     def __str__(self) -> str:
         return str(self.args[0])
@@ -298,56 +348,6 @@ class _WheelBuilder():
 
         return text.encode()
 
-    @property
-    def _debian_python(self) -> bool:
-        """Check if we are running on Debian-patched Python."""
-        if sys.version_info >= (3, 10):
-            return 'deb_system' in sysconfig.get_scheme_names()
-        try:
-            import distutils
-            try:
-                import distutils.command.install
-            except ModuleNotFoundError:
-                raise ModuleNotFoundError('Unable to import distutils, please install python3-distutils')
-            return 'deb_system' in distutils.command.install.INSTALL_SCHEMES
-        except ModuleNotFoundError:
-            return False
-
-    @property
-    def _debian_distutils_paths(self) -> Mapping[str, str]:
-        # https://ffy00.github.io/blog/02-python-debian-and-the-install-locations/
-        assert sys.version_info < (3, 12) and self._debian_python
-
-        with warnings.catch_warnings():
-            warnings.filterwarnings('ignore', category=DeprecationWarning)
-            import distutils.dist
-
-        distribution = distutils.dist.Distribution({
-            'name': self._project.name,
-        })
-        install_cmd = distribution.get_command_obj('install')
-        install_cmd.install_layout = 'deb'  # type: ignore[union-attr]
-        install_cmd.finalize_options()  # type: ignore[union-attr]
-
-        return {
-            'data': install_cmd.install_data,  # type: ignore[union-attr]
-            'headers': install_cmd.install_headers,  # type: ignore[union-attr]
-            'platlib': install_cmd.install_platlib,  # type: ignore[union-attr]
-            'purelib': install_cmd.install_purelib,  # type: ignore[union-attr]
-            'scripts': install_cmd.install_scripts,  # type: ignore[union-attr]
-        }
-
-    @property
-    def _sysconfig_paths(self) -> Mapping[str, str]:
-        sys_vars = sysconfig.get_config_vars().copy()
-        sys_vars['base'] = sys_vars['platbase'] = sys.base_prefix
-        if self._debian_python:
-            if sys.version_info >= (3, 10):
-                return sysconfig.get_paths('deb_system', vars=sys_vars)
-            else:
-                return self._debian_distutils_paths
-        return sysconfig.get_paths(vars=sys_vars)
-
     @cached_property
     def _stable_abi(self) -> Optional[str]:
         """Determine stabe ABI compatibility.
@@ -461,9 +461,9 @@ class _WheelBuilder():
         origin file and the Meson destination path.
         """
         warnings.warn('Using heuristics to map files to wheel, this may result in incorrect locations')
-        sys_paths = self._sysconfig_paths
+        sys_paths = _sysconfig_paths()
         # Try to map to Debian dist-packages
-        if self._debian_python:
+        if _debian_python():
             search_path = origin
             while search_path != search_path.parent:
                 search_path = search_path.parent
@@ -634,7 +634,7 @@ class _WheelBuilder():
         rebuild_commands = self._project.build_commands(install_path)
 
         import_paths = set()
-        for name, raw_path in self._sysconfig_paths.items():
+        for name, raw_path in _sysconfig_paths().items():
             if name not in ('purelib', 'platlib'):
                 continue
             path = pathlib.Path(raw_path)
@@ -833,6 +833,7 @@ class Project():
         We will try to reconfigure the build directory if possible to avoid
         expensive rebuilds.
         """
+        sys_paths = _sysconfig_paths()
         setup_args = [
             f'--prefix={sys.base_prefix}',
             os.fspath(self._source_dir),
@@ -841,6 +842,15 @@ class Project():
             # TODO: Allow configuring these arguments
             '-Ddebug=false',
             '-Doptimization=2',
+
+            # XXX: This should not be needed, but Meson is using the wrong paths
+            #      in some scenarios, like on macOS.
+            #      https://github.com/mesonbuild/meson-python/pull/87#discussion_r1047041306
+            '--python.purelibdir',
+            sys_paths['purelib'],
+            '--python.platlibdir',
+            sys_paths['platlib'],
+
             # user args
             *self._meson_args['setup'],
         ]
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -42,6 +42,11 @@ if sys.version_info < (3, 11):
 else:
     import tomllib
 
+if sys.version_info >= (3, 10):
+    import importlib.resources as importlib_resources
+else:
+    import importlib_resources
+
 import mesonpy._compat
 import mesonpy._elf
 import mesonpy._tags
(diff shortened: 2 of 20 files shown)


FFY00:
This is just an initial implementation, it will likely not be merged.

FFY00:
This new implementation uses import hooks, which is what I wanted to do initially. Since the complexity required to make the old approach work turned out to be a bit big, it makes more sense to do it like this from the start.

The idea is that we built on top of the [Traversable](https://docs.python.org/3/library/importlib.resources.abc.html#importlib.resources.abc.Traversable) protocol (`EditablePath`) and only trigger a rebuild when the data is accessed, this will delay rebuilds until the very last moment and prevent unnecessary rebuilds by just importing the module.

rgommers:
> During development with compiled languages (which I think is the main target of meson-python)

I'd say it is "Python projects containing compiled languages" - typically the majority of contributors (at least for large projects like NumPy, SciPy, Pandas) touch docs, tests, and pure Python code only, or at least way more often than native code. 

I agree that when you're touching C/C++/Cython/Fortran, editable builds are less relevant. But I've had regular requests from collaborators for editable installs, in particular from those who work with IDEs.

rgommers:
The way we've set it up for SciPy now to work with Meson is that running the test suite always first rebuilds (necessary to ensure the out-of-place install is not out of date). This takes a fraction of a second, so avoiding rebuilds on imports isn't essential I'd say. (not sure it matters regarding code complexity here).

rgommers:
> The idea is that we built on top of the [Traversable](https://docs.python.org/3/library/importlib.resources.abc.html#importlib.resources.abc.Traversable) protocol (`EditablePath`) and only trigger a rebuild when the data is accessed, this will delay rebuilds until the very last moment and prevent unnecessary rebuilds by just importing the module.

The Python stdlib docs are extremely sparse as usual. I have to admit I don't immediately grasp how this all works, a sketch of what actually happens with this implementation when one runs `pip install -e .` and then `import pkg` would be useful.

dnicolodi:
> I agree that when you're touching C/C++/Cython/Fortran, editable builds are less relevant.

I don't think that's the case. I do a lot of cross-project development where I need to test the changes in one project within another one. In this case editable installations are useful.

lithomas1:
Wow, this is pretty cool. Thanks for putting up the PR @FFY00.
I just have a couple quick clarifying questions.

For this to work, would every project would need to inject this hook(MesonpyFinder) into sys.meta_path in their __init__.py?

mesonpy/_editable.py
```diff
@@ -0,0 +1,155 @@
...
+    ) -> Optional[importlib.machinery.ModuleSpec]:
+        if not fullname.startswith('aaaa'):
+            return None
+        spec_path = self._base_path.joinpath(*fullname.split('.'))
+        return importlib.util.spec_from_loader(fullname, EditableLoader(fullname, spec_path))
+
+
+def install(path: str) -> None:
+    editable = Editable(os.path.abspath(path))
+    sys.meta_path.append(MesonpyFinder(editable))
```
lithomas1, line 155: Would a prepend be more appropriate here?

FFY00: Doesn't matter much, but my reasoning for simply appending was to not slow down any of the other imports. But this is now outdated given our new approach.

FFY00:
After discussing with @rgommers, we thought it would be better to simplify the initial approach. Instead of creating our own loader and reader, we will just intercept the import call with a custom finder, which won't change any of the import machinery and will simply trigger a rebuild + install to a dummy directory with we will make sure we add to `sys.path`. This means that the module would still be imported via the normal finder and loader. I think this will be easier to understand after you see the code.

FFY00:
> For this to work, would every project would need to inject this hook(MesonpyFinder) into sys.meta_path in their **init**.py?

Nop! Our editable wheel will install a `.pth` file that will do that. This code is still missing from the PR.

Also, the custom finder will need to be in `sys.meta_path` before the module is imported, as its goal is to hook into the import process, and the module `__init__.py` gets run on import, so that wouldn't work :stuck_out_tongue:

mesonpy/_editable.py
```diff
@@ -0,0 +1,70 @@
...
+import subprocess
+import sys
+
+from types import ModuleType
+from typing import Optional, Union
+
+from mesonpy._compat import Collection, Sequence
+
+
+class MesonpyFinder(importlib.abc.MetaPathFinder):
```
lithomas1, line 13: Sorry, I'm still a little confused on how this is going to be injected into sys.meta_path. 

Is the plan to write a pth file, that does arbritrary code execution? Or go down the sitecustomize route.

Also, where would this file live? IIUC, for it to be importable by a pth file or sitecustomize, it needs to live in site-packages, or meson-python needs to be installed.

I guess requiring meson-python to be installed is OK???, but if pip doesn't cleanup our pth files/wherever we're injecting this, it could be really bad.

FFY00: I pushed more code, but it still isn't complete. Hopefully it would help clarify things for you.

> Sorry, I'm still a little confused on how this is going to be injected into sys.meta_path.
> 
> Is the plan to write a pth file, that does arbritrary code execution? Or go down the sitecustomize route.
 
`.pth` file that executes code.

> I guess requiring meson-python to be installed is OK???

I think so, given this is for editable installs.

> but if pip doesn't cleanup our pth files/wherever we're injecting this, it could be really bad.

It should, we do not install the `.pth` file ourselves. We create a wheel for the editable install, which pip then installs on the system, `pip uninstall` should clean everything up. Maybe taking a look at https://peps.python.org/pep-0660/ could help clarify how this is supposed to work.

mesonpy/__init__.py
```diff
@@ -192,11 +188,15 @@ def _has_extension_modules(self) -> bool:
         # Assume that all code installed in {platlib} is Python ABI dependent.
         return bool(self._wheel_files['platlib'])
 
+    @property
+    def normalized_name(self) -> str:
+        return self._project.name.replace('-', '_')
```
henryiii, line 193: Not related to editable installs: shouldn't this also map `"."` to `"_"`?

FFY00: Good catch! But not really here, though I should definitely be doing that in the module name.

FFY00:
Alright, I haven't tested it yet, but it's getting pretty late, so I should go. The code *should* now be complete, but I expect stuff to be broken -- as I said, I didn't do any testing.

mesonpy/__init__.py
```diff
@@ -537,6 +541,9 @@ def build(self, directory: Path) -> pathlib.Path:
 
         return wheel_file
 
+    def build_editable(self, directory: Path) -> pathlib.Path:
+        self._build()
```
dnicolodi: This hunk seems unrelated.

mesonpy/__init__.py
```diff
@@ -542,7 +546,49 @@ def build(self, directory: Path) -> pathlib.Path:
...
 
     def build_editable(self, directory: Path) -> pathlib.Path:
-        self._build()
+        self._project.build()  # ensure project is built
+
+        wheel_file = pathlib.Path(directory, f'{self.name}.whl')
+
+        project_path = self._source_dir / '.mesonpy-editable'
+        top_level_modules = self._project.top_level_modules
+        rebuild_commands = [list(cmd) for cmd in self._project.build_commands(project_path)]
```
dnicolodi, line 555: Why the list comprehension?

mesonpy/__init__.py
```diff
@@ -733,11 +779,29 @@ def _wheel_builder(self) -> _WheelBuilder:
...
+    def source_dir(self) -> pathlib.Path:
+        return self._source_dir
+
+    @property
+    def working_dir(self) -> pathlib.Path:
+        return self._working_dir
+
+    @property
+    def build_dir(self) -> pathlib.Path:
+        return self._build_dir
```
dnicolodi, line 792: Why expose these as properties? I don't see them used outside private code, and there accessing the property directly is fine. I'm not a big fan of indirection if not needed.

dnicolodi (commented):
Just a few superficial comments.

mesonpy/__init__.py
```diff
@@ -815,21 +881,46 @@ def version(self) -> str:
...
         if not self.pep621:
-            return textwrap.dedent(f'''
+            data = textwrap.dedent(f'''
                 Metadata-Version: 2.1
                 Name: {self.name}
                 Version: {self.version}
-            ''').strip().encode()
+            ''').strip()
+            if editable:
+                data += f'\nRequires-Dist: mesonpy=={__version__}'
```
lithomas1, line 904:
```suggestion
                data += f'\nRequires-Dist: meson-python=={__version__}'
```

mesonpy/__init__.py
```diff
@@ -815,21 +881,46 @@ def version(self) -> str:
...
+
         # re-import pyproject_metadata to raise ModuleNotFoundError if it is really missing
         import pyproject_metadata  # noqa: F401, F811
         assert self._metadata
+
+        metadata = dataclasses.replace(self._metadata)
+        # add extra dependency required for the editable hook
+        if editable:
+            from packaging.requirements import Requirement
+            metadata.dependencies.append(Requirement(f'mesonpy=={__version__}'))
```
lithomas1, line 915:
```suggestion
            metadata.dependencies.append(Requirement(f'meson-python=={__version__}'))
```

mesonpy/__init__.py
```diff
@@ -815,21 +881,46 @@ def version(self) -> str:
...
 
-    @cached_property
-    def metadata(self) -> bytes:
+    @property
+    def top_level_modules(self) -> Collection[str]:
+        return {
+            os.path.dirname(self._copy_files[file])
+            for files in self._install_plan.values()
+            for file, details in files.items()
+            if details['destination'].startswith('{purelib}') or details['destination'].startswith('{platlib}')
```
lithomas1, line 890:
```suggestion
            if details['destination'].startswith('{py_purelib}') or details['destination'].startswith('{py_platlib}')
```

mesonpy/__init__.py
```diff
@@ -542,6 +545,51 @@ def build(self, directory: Path) -> pathlib.Path:
...
+            )
+
+            # install hook module
+            hook_module_name = f'_mesonpy_hook_{self.normalized_name.replace(".", "_")}'
+            whl.writestr(
+                f'{hook_module_name}.py',
+                textwrap.dedent(f'''
+                    import mesonpy._editable
+                    mesonpy._editable.MesonpyFinder.install(
+                        project_path={project_path},
```
lithomas1, line 571:
```suggestion
                        project_path="{project_path}",
```

mesonpy/__init__.py
```diff
@@ -815,21 +881,46 @@ def version(self) -> str:
         assert isinstance(version, str)
         return version
 
-    @cached_property
-    def metadata(self) -> bytes:
+    @property
+    def top_level_modules(self) -> Collection[str]:
+        return {
+            os.path.dirname(self._copy_files[file])
```
lithomas1, line 887: This line doesn't look correct. 

IIUC, we need to normpath, and then replace "/" with ".". Also, the paths have to be relative to the destination anchor, I think.

lithomas1 (commented):
I left some comments, addressing typos I saw when I pulled this. Hopefully, this helps you out a little. 

Other then the suggestions, you'll need to add ``mesonpy/_editable`` to the ``meson.build`` file.

mesonpy/__init__.py
FFY00: :upside_down_face: I was tired yesterday :rofl:

mesonpy/__init__.py
FFY00: This was the last thing I did yesterday, it was indeed broken, I re-wrote it with slightly better logic.

FFY00:
Alright, this should now be finally working :tada: 

The only thing is that currently does not detect top-level native modules.

@lithomas1 want to give it a try?

lithomas1:
> Alright, this should now be finally working 🎉
> 
> The only thing is that currently does not detect top-level native modules.
> 
> @lithomas1 want to give it a try?

Nice this is pretty awesome 🚀 (It is especially cool that you can just edit .pyx files and have the changes sync up too).

The only issue I have is slow rebuilds.
I attached time.time() calls to measure the speed of rebuilds and currently its' taking ~3-4 seconds for a regular rebuild, and ~34 seconds(!!) when I modified a .pyx file.

FFY00:
How does that compare to just running `ninja` directly? It shouldn't be taking much longer than that, and this is something we can't really do much about.
Using `Traversable`s we should be able to avoid the extra install step, but that shouldn't take that long anyway. We could also try to optimize the import hook code, but I feel that will be insignificant compared to the build step.

Perhaps we should have a way to disable swallowing the Meson output, so that developers can see the log when Python starts up?

eli-schwartz:
Where is that time being spent?

lithomas1:
> How does that compare to just running `ninja` directly? It shouldn't be taking much longer than that, and this is something we can't really do much about. Using `Traversable`s we should be able to avoid the extra install step, but that shouldn't take that long anyway. We could also try to optimize the import hook code, but I feel that will be insignificant compared to the build step.
> 
> Perhaps we should have a way to disable swallowing the Meson output, so that developers can see the log when Python starts up?

Installs seem to be the bottleneck. It seems to take ~ 3 secs. We are installing a lot of files, though (.mesonpy/editable/install is around 68 MB). 

I think disabling meson output is the right step, otherwise, someone might control-c thinking python is hanging, when we're actually rebuilding, which could potentially mess up future builds.

mesonpy/_editable.py
```diff
@@ -0,0 +1,91 @@
...
+        self._top_level_modules = top_level_modules
+        self._rebuild_commands = rebuild_commands
+
+    def __repr__(self) -> str:
+        return f'{self.__class__.__name__}({self._project_path})'
+
+    @functools.lru_cache(maxsize=1)
+    def rebuild(self) -> None:
+        # delete old install directory
+        shutil.rmtree(self._project_path)
```
lithomas1, line 48: Is there a reason we are deleting the install directory? 
It seem easier to re-use it and pass ``--only-changed`` to the install command?

FFY00: That makes sense.

mesonpy/__init__.py
FFY00: To convert to the correct type.

mesonpy/__init__.py
```diff
@@ -542,6 +553,60 @@ def build(self, directory: Path) -> pathlib.Path:
...
+                MesonpyFinder.install(
+                    project_path='{project_path}',
+                    import_paths={list(import_paths)},
+                    top_level_modules={list(top_level_modules)},
+                    rebuild_commands={rebuild_commands},
+                )
+            ''').strip().encode()
+            whl.writestr(
+                f'{hook_module_name}.py',
+                (importlib_resources.files('mesonpy') / '_editable.py').read_bytes() + hook_install_code,
```
eli-schwartz, line 592: This seems like a perfectly reasonable use of `importlib.resources.read_binary()`, which works on all versions of python without adding more dependencies. I would simply add a `warnings.filterwarnings()` to disable all deprecation notices from `importlib.resources`.

eli-schwartz (changes requested):

mesonpy/__init__.py
dnicolodi: Correct type for what? This makes tuples into lists, but it does not seem like lists are required. Are they? If they are, can't we just return the correct type?

mesonpy/__init__.py
FFY00: Yes, it's reasonable, but I'd rather make the code future-compatible (using a backport for older versions), than the other way round.

eli-schwartz: But it's not making the code future-compatible.

It's adding a dependency in order to avoid a function that is deprecated with no plans to remove. The function that it's avoiding simply makes use of `(files(f), p).read_bytes()` on newer versions of python, so there's absolutely no reason to deprecate it in the first place, despite which, `importlib.resources` prefers to tell people that it's "literally impossible" to use the ***stdlib*** without either version checks or deprecation warnings.

I think version checks are ugly, and so is using a backports package when the lowest supported python has an identically functioning and also more readable API that does everything the project needs. The `files()` API isn't even a good pathlib idiom anyway -- you cannot pass multiple path components to the initializer function. :rofl:

henryiii: Having a simple definition doesn't mean `read_binary()` won't be removed. Using deprecated API is setting yourself up for pain later. My suggestion is always to use the latest API and consider version checks and backports zero cost, they will not even need to be installed on newer versions of Python. Tools like pyupgrade can remove the version checked code for you when you drop a version.

henryiii: Also, I would never filter deprecation warnings to use an old API in new code. If it's removed, you would have been warned, it would be completely self inflicted.

mesonpy/__init__.py
```diff
@@ -542,6 +553,60 @@ def build(self, directory: Path) -> pathlib.Path:
...
+
+            # install non-code schemes
+            for scheme in self._SCHEME_MAP:
+                if scheme in ('purelib', 'platlib', 'mesonpy-libs'):
+                    continue
+                for destination, origin in self._wheel_files[scheme]:
+                    destination = pathlib.Path(self.data_dir, scheme, destination)
+                    whl.write(origin, destination.as_posix())
+
+        return wheel_file
```
henryiii, line 608: Didn't this used to add `meson-python==...` somewhere? I'm not seeing it now, was going to comment on that.

mesonpy/__init__.py
eli-schwartz: No, I mean that the author said it wasn't going to be removed, and also in Jaraco's discuss.python.org thread where the fact that "people dislike this deprecation and think I was unreasonable to add it" was discussed, no one really seemed convinced that it needed to be removed, multiple PyPA or CPython developers chipped in to say they thought it was unfortunate or in one case "So may I know when this slowing down will happen? Because I’m not going to use that module and will actively oppose people using it until then."

I would actually consider it outright reckless to fully and entirely remove the function from the stdlib for at least until python 3.10 is the oldest version of python still supported. Which is only barely not the most recent version of python still supported!

Because that (`pyupgrade --py310-plus`) is exactly when pyupgrade can start migrating that kind of code automatically, even without version checks.

Also -- if you're going to version gate the code anyway, why not simply use the stdlib only, and at the call site, do a version check and use `(files(p), f).read_binary()` on python 3.10, and `read_binary(f, p)` on lower versions. Fewer dependencies == win, and the version check can still be trivially removed by pyupgrade.

eli-schwartz: > Also, I would never filter deprecation warnings to use an old API in new code. If it's removed, you would have been warned, it would be completely self inflicted.
> 
> (The one special case is if you can talk the library into keeping the API and removing the warning in future versions, then you can hide it on older versions)

I would *usually* not filter deprecation warnings to use an old API in new code. This is an exceptional case because the deprecation itself seems unreasonable and a *lot* of people are complaining that the module as a whole is moving "recklessly" fast and making it unreasonably difficult to write acceptable code without simply pretending that it was not added to the stdlib at all until 3.10, which is a really lousy feeling to have to feel about the stdlib.

eli-schwartz: I like `files()` too, especially since it allows me to lookup arbitrary files without adding `__init__.py` to every data directory just so the older API works. :D However, I often cannot use it... (I cannot use it in Meson, so instead I have a pending branch that will go add use of `path()`).

mesonpy/__init__.py
```diff
@@ -817,21 +892,49 @@ def version(self) -> str:
...
+                path = pathlib.Path(details['destination'])
+                if len(path.parts) >= 2 and path.parts[0]:
+                    top_part = path.parts[1]
+                    # file module
+                    if top_part.endswith('.py'):
+                        modules.add(top_part[:-3])
+                    else:
+                        # native module
+                        for extension in _EXTENSION_SUFFIXES:
+                            if top_part.endswith(extension):
```
dnicolodi, line 909: This cab be written
```
if top_part.endswith(extension, _EXTENSION_SUFFIXES):
    ...
```

mesonpy/__init__.py
rgommers: version checks are okay, they're a very minor technical debt. taking on another runtime dependency should be avoided, that's a big cost.

mesonpy/__init__.py
```diff
@@ -731,11 +796,23 @@ def _wheel_builder(self) -> _WheelBuilder:
             self._copy_files,
         )
 
+    def build_commands(self, install_dir: Optional[pathlib.Path] = None) -> Collection[Sequence[str]]:
+        return (
+            ('meson', 'compile', *self._meson_args['compile'],),
```
lithomas1, line 801: This is now broken for me again.

On import, this is invoking just plain ``meson compile`` without a builddir, which is wrong.

mesonpy/__init__.py
FFY00: Yep, since we don't have any dependencies on the `_editables` module (internal or external), I changed it hook to just copy its contents and append the custom code.

dnicolodi:
The tests failure on macOS are due to this https://github.com/mesonbuild/meson-python/blob/6845f905747030813314e78a752b62c35134d44c/mesonpy/_tags.py#L96-L111 Probably triggered by the recent release of a Python 3.7 minor version. It was meant to happen. I'll look into adjusting the tests.

FFY00:
:astonished: 

I was able to reproduce, my CPU and RAM get pinned to 100%. My best guess is that during the rebuild, a python interpreter gets spawned, which somehow triggers a new rebuild? SciPy is a complex build, something there might be importing a `scipy` module.

rgommers:
> something there might be importing a `scipy` module.

Yes, I think that does happen. In NumPy as well. Probably in other projects too - it's fairly common, and tends to work fine (it's a bit hacky of course, but it's hard not to do this when you have for example codegen components in your package that you also need at build time).

FFY00:
Still missing the tests.

FFY00:
I have the missing test now. While adding it, I discovered an issue with the `top_level_modules` implementation, which should be fixed now.

FFY00:
Ah, how nice... tests are failing on meson because the site-packages path is mangled. I need to test that to see what bit of the code exactly is throwing it off, I will do that monday.

mesonpy/__init__.py
```diff
@@ -623,6 +623,18 @@ def build_editable(self, directory: Path, verbose: bool = False) -> pathlib.Path
         }
         rebuild_commands = [list(cmd) for cmd in self._project.build_commands(install_path)]
 
+        # XXX: Meson is messing up the install paths on macOS, let's just try
+        #      searching for site-packages directories in the install paths and
+        #      add matches to the import paths
```
dnicolodi, line 628: Ugh? The install paths used by Meson are the ones specified in `sysconfig`. Why do you think Meson is doing something wrong? Please note that the installation directory schema used on macOS is different when installing in an virtual env.

FFY00: > The install paths used by Meson are the ones specified in `sysconfig`

They aren't. They are based on `sysconfig` but they're processed in some cases, which is what seems I am running into here.

https://github.com/mesonbuild/meson/blob/b249470a4c1de39a0a56d9237f767fe04d6efa05/mesonbuild/modules/python.py#L470

I have already recommended that Meson start trying to find Python interpreter in the target prefix, instead of finding *some* interpreter in `PATH` and trying to derive the paths by introspecting that interpreter and trying to override the prefix, Python paths are just too fragile for that to work reliably. Even the `install_env` is utterly conceptually broken, unfortunately layout paths are just not portable between interpreters. After my previous interactions with the maintainers, I am not really motivated to try to fix Python path handling on Meson.

That said, I am not sure if the current issue I am running into is the result of the prefix overriding logic or something else, what I know is that the Meson is trying to install to are definitely not the ones indicated by `sysconfig`.

Debugging shows a path mismatch on both Homebrew Python 3.11, and Apple Python.

Here are the paths on Homebrew Python 3.11:

Meson install path
```
/opt/homebrew/opt/python@3.11/Frameworks/Python.framework/Versions/3.11/opt/homebrew/lib/python3.11/site-packages
```

sysconfig path
```
/opt/homebrew/lib/python3.11/site-packages
```

rgommers: > I have already recommended that Meson start trying to find Python interpreter in the target prefix, instead of finding some interpreter in PATH and trying to derive the paths by introspecting that interpreter and trying to override the prefix, Python paths are just too fragile for that to work reliably.

That does sound sensible to me. It'd be great if there were an upstream issue for this, and the XXX comment referring to it.

eli-schwartz: > Even the install_env option is utterly conceptually broken, unfortunately layout paths are just not portable between interpreters.

... Then it's a good thing thats not really what Meson is trying to do? But I'm pretty sure we discussed this in the past already, nothing new to be said.

> After my previous interactions with the maintainers, I am not really motivated to try to fix Python path handling on Meson.

There's a non-zero chance you're not the only one burned out from that PR. :shrug:

I have some ideas floating around in the back of my head that may be relevant to this whole topic, but for over a year, I've been discovering so many more interesting and enjoyable things to do for Meson than to discuss python layout paths. Most of those things involve not thinking about the python module at all. Maybe next year. Maybe not.

Considering we have to work with the `sysconfig` API of all things :P it may make sense to defer all changes until sysconfig can actually handle the situation. Hopefully sysconfig can at least help resolve the remaining holdouts for porting off of distutils before 3.12 final lands.

mesonpy/__init__.py
```diff
@@ -291,6 +291,8 @@ def entrypoints_txt(self) -> bytes:
     @property
     def _debian_python(self) -> bool:
         """Check if we are running on Debian-patched Python."""
+        if sys.version_info >= (3, 10):
```
dnicolodi: I think the right version threshold is 3.10.3, see also https://github.com/mesonbuild/meson/pull/11133

FFY00:
Okay, I have now fixed all the last conflicts from the new release, this should be ready to go in.

mesonpy/__init__.py
dnicolodi: I'm with @eli-schwartz and @rgommers here. We should strive for minimal dependencies. If we can make this work with an API available on older Python versions, I would go that way.

mesonpy/__init__.py
```diff
@@ -102,7 +107,7 @@ def _init_colors() -> Dict[str, str]:
 _STYLES = _init_colors()  # holds the color values, should be _COLORS or _NO_COLORS
 
 
-_EXTENSION_SUFFIXES = frozenset(importlib.machinery.EXTENSION_SUFFIXES)
+_EXTENSION_SUFFIXES = list(importlib.machinery.EXTENSION_SUFFIXES)
```
dnicolodi, line 111: `importlib.machinery.EXTENSION_SUFFIXES` is already a `list`. No coercion is necessary here.

mesonpy/__init__.py
```diff
@@ -1116,3 +1271,27 @@ def build_wheel(
     out = pathlib.Path(wheel_directory)
     with _project(config_settings) as project:
         return project.wheel(out).name
+
+
+def build_editable(
```
dnicolodi, line 1239: I think this should be decorated with `@_pyproject_hook()`

mesonpy/__init__.py
```diff
@@ -1116,3 +1271,27 @@ def build_wheel(
...
+        config_settings = {}
+    if 'builddir' not in config_settings:
+        config_settings['builddir'] = os.path.join('.mesonpy', 'editable', 'build')
+
+    out = pathlib.Path(wheel_directory)
+    with _project(config_settings) as project:
+        return project.editable(out).name
+
+
+def get_requires_for_build_editable(
```
dnicolodi, line 1257: This too.

mesonpy/_editable.py
```diff
@@ -0,0 +1,90 @@
...
+import sys
+
+from types import ModuleType
+from typing import List, Optional, Union
+
+
+if sys.version_info >= (3, 9):
+    from collections.abc import Sequence
+else:
+    from typing import Sequence
```
dnicolodi, line 15: This can be imported from `_compat`.

mesonpy/__init__.py
dnicolodi: Spurious hunk.

mesonpy/_editable.py
```diff
@@ -18,6 +19,37 @@
...
+            warnings.warn('Both NO_COLOR and FORCE_COLOR environment variables are set, disabling color')
+        return _NO_COLORS
+    elif 'FORCE_COLOR' in os.environ or sys.stdout.isatty():
+        return _COLORS
+    return _NO_COLORS
+
+
+_STYLES = _init_colors()  # holds the color values, should be _COLORS or _NO_COLORS
+
+
```
dnicolodi, line 52: Do we really need to duplicate all this here and in `__init__`?

dnicolodi (changes requested):
Just a few superficial comments.

mesonpy/__init__.py
```diff
@@ -547,6 +576,62 @@ def build(self, directory: Path) -> pathlib.Path:
...
+            path = pathlib.Path(raw_path)
+            import_paths.add(install_path / path.relative_to(path.anchor))
+
+        install_path.mkdir(parents=True, exist_ok=True)
+
+        with mesonpy._wheelfile.WheelFile(wheel_file, 'w') as whl:
+            self._wheel_write_metadata(whl)
+            whl.writestr(
+                f'{self.distinfo_dir}/direct_url.json.py',
+                self._source_dir.as_uri().encode(),
```
rgommers, line 600: This does not seem all that happy for me, when testing this PR with SciPy on Linux:
```
Successfully built scipy
Installing collected packages: scipy
  *** Error compiling '/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/scipy-1.11.0.dev0.dist-info/direct_url.json.py'...
    File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/scipy-1.11.0.dev0.dist-info/direct_url.json.py", line 1
      file:///home/rgommers/code/scipy
           ^^
  SyntaxError: invalid syntax
```

mesonpy/_editable.py
FFY00: No, this module is standalone.

mesonpy/_editable.py
FFY00: This module is meant to be standalone, it gets copied to the editable hook.

FFY00:
> I'm with @eli-schwartz and @rgommers here. We should strive for minimal dependencies. If we can make this work with an API available on older Python versions, I would go that way.

I am planning to submit a followup PR for that, but right now I just wanted to get this merged, so that it is easier to make other changes.

mesonpy/_editable.py
dnicolodi: Why does it need to be?

mesonpy/__init__.py
dnicolodi: Why do we need a copy at all?

mesonpy/_editable.py
```diff
@@ -0,0 +1,153 @@
...
+_NO_COLORS = {color: '' for color in _COLORS}
+
+
+def _init_colors() -> Mapping[str, str]:
+    """Detect if we should be using colors in the output. We will enable colors
+    if running in a TTY, and no environment variable overrides it. Setting the
+    NO_COLOR (https://no-color.org/) environment variable force-disables colors,
+    and FORCE_COLOR forces color to be used, which is useful for thing like
+    Github actions.
+    """
```
dnicolodi, line 40: Whoever decided to standardize on this interface needs to reconsider its life choices. What was wrong with `COLORS=0` and `COLORS=1` ? I guess the `colors.org` domain was taken...

henryiii: Rich uses this, by the way - If you set both, you get styles (like bold and underline), but not color. Probably not the original intent, but an interesting idea.

I also like setting FORCE_COLOR to 3, just in case levels are supported - 8 being 16 colors, 2 being 256, and 3 true color. Not sure anyone has actually done that, but it would be nice. Most libraries just expect it set or truthy.

The original NO_COLOR was written by someone who really, really hated colors and didn't care one bit about colors being forced on, and didn't specify any standard for turning it on. According to the standard, setting `NO_COLOR` to an empty string or false value also disables color (distinguishing between an empty and a missing environment variable is terrible, IMO, especially because TOML doesn't have a null value)

lithomas1 (changes requested):
Looks good to me. 
This needs docs before landing, though, so I am going to block this.

Thanks again for adding this @FFY00.

mesonpy/_editable.py
FFY00: I mean, `NO_COLOR` has been around for a bit, we started using `FORCE_COLOR` as its counterpart in https://github.com/pypa/build, we could have a used `COLORS` to but :shrug:

henryiii: FORCE_COLOR is starting to be a standard too, mostly as a counterpart to NO_COLOR, but a lot of tools use it now, like pytest and such. It's best for everyone to use the same var, so you can set it once in a CI file and get it everywhere, rather than having to look up each tool's custom settings.

FFY00:
> This needs docs before landing, though, so I am going to block this.

What exactly are you looking for in the docs? The next big item to work on is reworking the docs, as discussed in #197, so I avoided adding things just to then replace them right away, also I really wanted to get this out as soon as possible so that we can get people testing it.

lithomas1:
> > This needs docs before landing, though, so I am going to block this.
> 
> What exactly are you looking for in the docs? The next big item to work on is reworking the docs, as discussed in #197, so I avoided adding things just to then replace them right away, also I really wanted to get this out as soon as possible so that we can get people testing it.

Its hard to test something without having documentation. Also, I think this is basically going to get zero visibility without any docs, so we shouldn't rush on releasing this, IMO.

As to what I'm looking for, I am mainly looking for how to configure ``MesonpyFinder``( e.g. how to use the verbose option, what does "_MESONPY_EDITABLE_SKIP" do). I already know the answers to these just by reading the code, but for anyone not already familiar with meson-python this is going to be challenging.

lithomas1:
I don't have time to test this this week, so I'm just going to assume that if it works for scipy, its probably going to work for pandas.

mesonpy/_editable.py
eli-schwartz: https://imgs.xkcd.com/comics/standards.png

There's a standard for FORCE_COLOR, another standard for NO_COLOR, and a third standard for "check for both CLICOLOR as well as CLICOLOR_FORCE".

There's also exporting `TERM=dumb`, because dumb terminals don't support color codes, and APPNAME_NO_COLOR for when you want 60 environment variables all doing the same thing within a *single* standard.

CMake has CMAKE_COLOR_MAKEFILE if you want colors in cmake, but not in the generated Makefiles?

GCC has GCC_COLORS=, grep has GREP_COLORS=, for when you want to not only specify that colors shall be used but also a format string specifying a mapping of levels to colors.

FFY00:
> Its hard to test something without having documentation. Also, I think this is basically going to get zero visibility without any docs, so we shouldn't rush on releasing this, IMO.

I added some simple instructions, which I think should be enough for now.

> As to what I'm looking for, I am mainly looking for how to configure `MesonpyFinder`( e.g. how to use the verbose option, what does "_MESONPY_EDITABLE_SKIP" do). I already know the answers to these just by reading the code, but for anyone not already familiar with meson-python this is going to be challenging.

I mentioned the verbose option in the docs.

`_MESONPY_EDITABLE_SKIP` is private, it should not be used by users. We need it to use it internally to prevent rebuild loops.

mesonpy/_editable.py
dnicolodi: Yeah, I know. It is simply that `NO_COLOR` is backward in all the ways you want to look at it. And I'm surprised that so many folks blindly followed the trend just because someone spent the money to buy a domain name... Getting the general setting from the `COLORS=` environmental variable, with the possibility of having application specific settings with `APPNAME_COLORS=` would have made so much more sense, and is what most of the GNU tools have been doing for ages.

eli-schwartz: Note: despite my slight sarcasm above, I actually do think that application-specific variables makes some sense. If you use one of these global standards, then on some operating systems, tools like `ls` will force on colored output, which is *guaranteed* to break existing scripts that parse the output.

henryiii: Yes, I like having both a global setting and a custom one, just in case you do really want to only turn on mypy, pytest, tox, etc. Most of those do have that. 95% of the time it's fine to set `FORCE_COLOR` in a CI environment and forget about it.

No argument that COLORS/APP_COLORS would have been better. And for the custom app one, you can do whatever you'd like, since the point of it is it's not a standard. :)

docs/index.rst
```diff
@@ -49,6 +49,31 @@ frontend, like ``builddir``, to specify the build directory to use/re-use. You
...
+edit the project source and have those changes be reflected in the installed
+module.
+
+The way it works on ``meson-python`` specifically, is that when you import a
+module of the project, it will be rebuilt. This means that project imports will
+take more time than usual.
+
+It might be helpful to see the output of the Meson_ commands. That can be
+achieved by using the ``editable-verbose`` config setting when installing the
+project, or afterwards, by setting the ``MESONPY_EDITABLE_VERBOSE`` environment
```
rgommers, line 65: A few thoughts:
- for `editable-verbose`, can you please add the exact thing to add to the command line? this isn't clear
- should we mark `MESONPY_EDITABLE_VERBOSE` as provisional (subject to change)?
- it needs to be specified what happens when both the cli flag and the env var are given for the initial build

rgommers:
One issue I just ran into is that I did an editable build of SciPy (which failed), then a `git clean -xdf`, then a non-editable build, and the regular build is now broken because `scipy-editable-hook.pth` got left behind. Questions:
- Should the `.pth` file only be installed right at the end, once we're sure that the build has succeeded?
- Does `pip uninstall scipy` clean up the `.pth` file? (I understand `git clean -xdf` won't)
- For `git clean -xdf` type things, is it possible to give a very clear error message when the `.pth` file gets left behind? So the user at least gets told to manually delete that file to get things working again.

docs/index.rst
rgommers: Also, I am already seeing that we need verbosity levels. Rebuilding 2 files followed by 1000+ lines of _"Installing xxx"_ is not an optimal user experience.

mesonpy/_editable.py
```diff
@@ -0,0 +1,153 @@
...
+                for path in self._import_paths:
+                    if path in sys.path:
+                        sys.path.remove(path)
+                sys.path = self._import_paths + sys.path
+        # return none (meaning we "didn't find" the module) and let the normal
+        # finders find/import it
+        return None
+
+    @classmethod
+    def install(
```
rgommers, line 129: This function needs two changes:
- add a `self._debug(f'{{cyan}}{{bold}}+` print command like for `rebuild` above
- call `meson install --only-changed` rather than reinstalling all files

rgommers: actually, inside the `.rebuild` method I think. It happens before the `+ successfully rebuilt` message.

rgommers:
Everything does seem to work as advertised on SciPy now as far as I can tell - nice! My comments above are mostly usability paper cuts.

FFY00:
> Should the `.pth` file only be installed right at the end, once we're sure that the build has succeeded?

We build a wheel, and pip installs it, so it will only ever get installed if the build succeeds.

> Does `pip uninstall scipy` clean up the `.pth` file? (I understand `git clean -xdf` won't)

Yes, it should.

> For `git clean -xdf` type things, is it possible to give a very clear error message when the `.pth` file gets left behind? So the user at least gets told to manually delete that file to get things working again.

Yes.

mesonpy/_editable.py
FFY00: This method only installs the hook, the meson install is part of the rebuild process. It seems we lost `--only-changed` in the refactoring, will correct.

docs/index.rst
FFY00: > for `editable-verbose`, can you please add the exact thing to add to the command line? this isn't clear

> it needs to be specified what happens when both the cli flag and the env var are given for the initial build

Done.

> it needs to be specified what happens when both the cli flag and the env var are given for the initial build

Sure.

> Also, I am already seeing that we need verbosity levels. Rebuilding 2 files followed by 1000+ lines of _"Installing xxx"_ is not an optimal user experience.

This should be fixed by adding the missing `--only-changed` argument.

FFY00:
> I don't have time to test this this week, so I'm just going to assume that if it works for scipy, its probably going to work for pandas.

I did some basic testing and it seems working.

```
Python 3.10.8 (main, Nov  1 2022, 14:18:21) [GCC 12.2.0] on linux
>>> import pandas
+ rebuilding /home/anubis/git/pandas
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: /usr/bin/ninja
[2/2] Generating write_version_file with a custom command
ninja: Entering directory `/home/anubis/git/pandas/.mesonpy/editable/build'
[1/1] Generating write_version_file with a custom command
Installing subdir /home/anubis/git/pandas/pandas/io/clipboard to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/io/clipboard
Installing subdir /home/anubis/git/pandas/pandas/io/excel to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/io/excel
Installing subdir /home/anubis/git/pandas/pandas/io/formats to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/io/formats
Installing subdir /home/anubis/git/pandas/pandas/io/json to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/io/json
Installing subdir /home/anubis/git/pandas/pandas/io/parsers to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/io/parsers
Installing subdir /home/anubis/git/pandas/pandas/_config to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/_config
Installing subdir /home/anubis/git/pandas/pandas/_libs to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/_libs
Installing subdir /home/anubis/git/pandas/pandas/_testing to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/_testing
Installing subdir /home/anubis/git/pandas/pandas/api to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/api
Installing subdir /home/anubis/git/pandas/pandas/arrays to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/arrays
Installing subdir /home/anubis/git/pandas/pandas/compat to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/compat
Installing subdir /home/anubis/git/pandas/pandas/core to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/core
Installing subdir /home/anubis/git/pandas/pandas/errors to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/errors
Installing subdir /home/anubis/git/pandas/pandas/plotting to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/plotting
Installing subdir /home/anubis/git/pandas/pandas/tests to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/tests
Installing subdir /home/anubis/git/pandas/pandas/tseries to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/tseries
Installing subdir /home/anubis/git/pandas/pandas/util to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas/util
Installing _version_meson.py to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas
Installing /home/anubis/git/pandas/pandas/__init__.py to /home/anubis/git/pandas/.mesonpy/editable/install/usr/lib/python3.10/site-packages/pandas
Preserved 1961 unchanged files, see meson-logs/install-log.txt for the full list
+ successfully rebuilt
>>> pandas.__version__
'2.0.0.dev0+919.g4fe3ec05a3.dirty-test-editable'
```

mesonpy/_editable.py
lithomas1: Are you sure ``--only-changed`` is working?

What if I delete a file from the source directory(using ``install_subdir`` to install files)? This doesn't seem to be reflected(the file is not deleted from the install directory).

docs/index.rst
```diff
@@ -49,6 +49,57 @@ frontend, like ``builddir``, to specify the build directory to use/re-use. You
...
+
+
+
+This behavior can also be enabled by default by passing the ``editable-verbose``
+config setting when installing the project.
+
+
+.. code-block::
+
+   python -m pip install -e . --config-settings editable-verbose=
```
lithomas1, line 93: Do we error when this is passed but we are not doing an editable install?

mesonpy/_editable.py
eli-schwartz: `meson install` doesn't delete files that were installed by older versions of the project but then deleted from source.

`--only-changed` has no effect on that -- it just ***avoids doing I/O on files that would otherwise be installed***, if the destination already exists and the heuristic of "file timestamps" indicates that the source isn't newer than the destination.

docs/index.rst
FFY00: Yep.

https://github.com/mesonbuild/meson-python/blob/89af86c2769ca6d900c3d60b6fcbd48479fd1f86/mesonpy/__init__.py#L1119-L1128

mesonpy/_editable.py
FFY00: I don't think this is a huge issue, but definitely something we'll want to fix.

@eli-schwartz would it be reasonable to request this as a feature from Meson? Or should we just implement it ourselves.
`--only-changed` not removing deleted files seems like a footgun to me.

eli-schwartz: If the file was deleted, then "probably" any actual references to it inside another file were deleted as well, so having an extra pointless file may not matter.

Although sometimes it does matter -- for example, if you rename `mesonbuild/mesonlib/*.py` to `mesonbuild/utils/*.py` and then add `mesonbuild/mesonlib.py` with the old contents of `mesonbuild/mesonlib/__init__.py` (`from .foo import *` remapped to `.utils.foo`). And then you re-run `python -m build` to produce wheels, and upload them to PyPI, but setuptools preserves the old code in `build/`, does not clean it up, and the wheel is broken because `mesonbuild.mesonlib` is still used, it just changed from a package to a module.

(Guess how I know that one.)

Meson doesn't have that problem if you install fresh as part of making a brand new wheel. Or if you install fresh to a DESTDIR as part of a unix packaging workflow. Or if you run `ninja uninstall && ninja install` to delete the old version and install the new one.

But it does have that problem -- it *has* to have that problem -- if you simply install a newer version in-place on top of an older version.

eli-schwartz: @FFY00,

I think having `meson install --only-changed` produce results that are intentionally different from `meson install` is probably wrong. The point of this option is to say "I'm pretty sure I don't have timestamp bugs such as changed files that don't have updated modification times, so go ahead and optimize the work when you're positive that it doesn't make a difference".

It sounds like what you want is better named `--uninstall-old`.

FFY00: Yes, that is why it matters, and also why I don't think it's a *huge* issue, but definitely something we want to fix :blush: 

> But it does have that problem -- it _has_ to have that problem -- if you simply install a newer version in-place on top of an older version.

It could keep track of the installed files in a metadata file, and when installing with `--only-changed`, it would look at it, delete any files not present in the new installation, and update the metadata file.

Does this sound reasonable? I can open a Meson issue if you think it's worth considering. Otherwise, we can implement that here ourselves, but it IMO being a footgun for `--only-changed` in Meson, I'd rather see it fixed in the source directly.

FFY00: (sorry, when I was writing the message above the UI hadn't updated with your last reply)

> I think having `meson install --only-changed` produce results that are intentionally different from `meson install` is probably wrong. The point of this option is to say "I'm pretty sure I don't have timestamp bugs such as changed files that don't have updated modification times, so go ahead and optimize the work when you're positive that it doesn't make a difference".

I personally disagree, but understand your position.

> It sounds like what you want is better named `--uninstall-old`.

That seems reasonable.

lithomas1: I guess it does seem reasonable for meson not to touch other files.

We also probably can figure out which files are changed through introspection, as long as intro-install-plan and intro-installed match what happens without ``--only-changed``. (Then, we find a list of the installed files, and delete everything else in the folder not in installed files).

docs/index.rst
```diff
@@ -49,6 +49,57 @@ frontend, like ``builddir``, to specify the build directory to use/re-use. You
 can find more information about them in the `build options page`_.
 
 
+Editable installs
+-----------------
+
+Editable installs allow you to install the project in such a way where you can
+edit the project source and have those changes be reflected in the installed
+module.
```
rgommers, line 57: I suggest adding ", without having to explicitly rebuild and reinstall".

docs/index.rst
```diff
@@ -49,6 +49,57 @@ frontend, like ``builddir``, to specify the build directory to use/re-use. You
...
+
+The way it works on ``meson-python`` specifically, is that when you import a
+module of the project, it will be rebuilt. This means that project imports will
```
rgommers, line 60: I suggest "it will be rebuilt on the fly"

docs/index.rst
```diff
@@ -49,6 +49,57 @@ frontend, like ``builddir``, to specify the build directory to use/re-use. You
...
+take more time than usual.
+
+You can use pip, or other package manager, to install the project in editable
```
rgommers, line 63: Is there another installer / package manager that does editable installs? If not, better to leave it out.

docs/index.rst
rgommers, line 93: `editable-verbose=` seems to be missing something (from a usability perspective), better to make it `editable-verbose=true`.

rgommers (approved):
The docs look quite good to me now, just a few minor suggestions to clarify things.

lithomas1 (approved):
LGTM

FFY00:
I somehow missed the notification for Ralf's review, so I only noticed yesterday :rofl:

PR #89, opened 2022-06-29
DOC: extend documentation on what's included/excluded in sdist and wheels

rgommers:
For the `.gitattributes` content, I tested that this works as advertised in SciPy. The info came from gh-58.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -58,9 +58,24 @@ Python tool (pip_, `pypa/build`_, etc.) to build and install the project.
 ``mesonpy`` will build a Python sdist (source distribution) or wheel (binary
 distribution) from Meson_ project.
 
+Source distribution (sdist)
+---------------------------
+
 The source distribution is based on ``meson dist``, so make sure all your files
-are inluded there. In git projects, Meson_ will not include files that are not
-checked into git, keep that in mind when developing.
+are included there. In git projects, Meson_ will not include files that are not
+checked into git, keep that in mind when developing. By default, all files
+under version control will be included in the sdist. In order to exclude files,
+use ``export-ignore`` or ``export-subst`` attributes in ``.gitattributes`` (see
+the ``git-archive`` documentation for details; ``meson dist`` uses
+``git-archive`` under the hood).
+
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -71,11 +86,17 @@ similarly, the user will be warned. These warnings are not necessarily a reason
...
+
+If the project itself includes shared libraries that are needed by other files,
+those libraries will be included in the wheel, and native files will have their
 library search path extended to include the directory where the libraries are
 placed.
 
+If the project depends on external shared libraries, those libraries will not
+be included in the wheel. Vendoring those libraries or finding a suitable runtime
+dependency that provides those libraries is the responsibility of the user
+of ``meson-python``.
```
FFY00, line 98: The user of `meson-python` would be the package developer, not necessarily the people trying to compile or install the project. As such, this sentence seems a bit confusing, because the people that need to find the runtime dependency are the ones compiling/installing the project.

Here's my suggestion.
```suggestion
be included in the wheel. Vendoring those libraries or finding a suitable runtime
dependency that provides those libraries is the responsibility of the user
of binary distributions built by ``meson-python``.
```

Maybe we can also point people that want to distribute their artifacts to `auditwheel` and `delocate`?

FFY00 (changes requested):
Thank you! This looks great, I just have a comment.

docs/index.rst
rgommers: Ah, good point. With "user" I actually meant the package developer. But it could be either, so this needs a bigger rephrase.
- The package developer should vendor those libraries or find a suitable runtime dependency, OR
- The user must install that shared library in some other way (e.g., through a distro package manager).

rgommers: For SciPy, we want to do the vendoring, and how to tweak the vendoring mechanism is a little TBD right now.

rgommers: Okay, I went for an extended version of the above and also included auditwheel/delocate.

FFY00 (approved):
Sorry for the delay, thanks!

Issue #86, opened 2022-06-22
tests/test_sdist.py::test_contents_unstaged fails outside git clone


mgorny:
The following test fails when run from GitHub archive rather than full `git clone`. IIUC `in_git_repo_context` is supposed to make it use a temporary git repo, so I suppose something goes wrong:

```pytb
_______________________________________________________ test_contents_unstaged ________________________________________________________

package_pure = PosixPath('/tmp/meson-python/tests/packages/pure')
tmpdir = local('/tmp/pytest-of-mgorny/pytest-1/test_contents_unstaged0')

    def test_contents_unstaged(package_pure, tmpdir):
        new_data = textwrap.dedent('''
        def bar():
            return 'foo'
        ''').strip()
    
        with open('pure.py', 'r') as f:
            old_data = f.read()
    
        try:
            with in_git_repo_context(), open('pure.py', 'w') as f, open('crap', 'x'):
                f.write(new_data)
    
>           sdist_path = mesonpy.build_sdist(os.fspath(tmpdir))

/tmp/meson-python/tests/test_sdist.py:52: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/tmp/meson-python/mesonpy/__init__.py:869: in build_sdist
    return project.sdist(out).name
/tmp/meson-python/mesonpy/__init__.py:794: in sdist
    self._meson('dist', '--allow-dirty', '--no-tests', '--formats', 'gztar')
/tmp/meson-python/mesonpy/__init__.py:468: in _meson
    return self._proc('meson', *args)
/tmp/meson-python/mesonpy/__init__.py:463: in _proc
    subprocess.check_call(list(args))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

popenargs = (['meson', 'dist', '--allow-dirty', '--no-tests', '--formats', 'gztar'],), kwargs = {}, retcode = 1
[...]

Issue #65, opened 2022-05-26
Tests fail when building from source distribution


awvwgk:
Testing the installed package against the tests available in the `tests` directory of the sdist results in a failure due to

```
=================================== FAILURES ===================================
____________________________ test_contents_unstaged ____________________________

package_pure = PosixPath('$SRC_DIR/tests/packages/pure')
tmpdir = local('/tmp/pytest-of-conda/pytest-0/test_contents_unstaged0')

    def test_contents_unstaged(package_pure, tmpdir):
        new_data = textwrap.dedent('''
        def bar():
            return 'foo'
        ''').strip()
    
        with open('pure.py', 'r') as f:
            old_data = f.read()
    
        try:
            with in_git_repo_context(), open('pure.py', 'w') as f, open('crap', 'x'):
                f.write(new_data)
    
>           sdist_path = mesonpy.build_sdist(os.fspath(tmpdir))

$SRC_DIR/tests/test_sdist.py:52: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
$PREFIX/lib/python3.10/site-packages/mesonpy/__init__.py:798: in build_sdist
    return project.sdist(out).name
$PREFIX/lib/python3.10/site-packages/mesonpy/__init__.py:724: in sdist
    self._meson('dist', '--allow-dirty', '--no-tests', '--formats', 'gztar')
$PREFIX/lib/python3.10/site-packages/mesonpy/__init__.py:438: in _meson
    return self._proc('meson', *args)
$PREFIX/lib/python3.10/site-packages/mesonpy/__init__.py:434: in _proc
    subprocess.check_call(list(args))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

popenargs = (['meson', 'dist', '--allow-dirty', '--no-tests', '--formats', 'gztar'],)
kwargs = {}, retcode = 1
cmd = ['meson', 'dist', '--allow-dirty', '--no-tests', '--formats', 'gztar']

    def check_call(*popenargs, **kwargs):
        """Run command with arguments.  Wait for command to complete.  If
        the exit code was zero then return, otherwise raise
        CalledProcessError.  The CalledProcessError object will have the
        return code in the returncode attribute.
    
[...]

PR #92, opened 2022-07-05
TST: fix a failing test which assumes it's running in a git repo

rgommers:
Running tests from an sdist is important for packagers, and should work without failures. So simply skip the one test that assumes it is running in a git repo.

Closes gh-65
Closes gh-86

### DIFF
--- a/tests/test_sdist.py
+++ b/tests/test_sdist.py
@@ -4,6 +4,10 @@ import os
 import tarfile
 import textwrap
 
+from pathlib import Path
+
+import pytest
+
 import mesonpy
 
 from .conftest import in_git_repo_context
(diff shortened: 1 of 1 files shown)


rgommers:
Failing tests are unrelated, but I know what the problem is - I'll send a separate PR to fix those.

FFY00:
I don't think this is the correct fix. [`in_git_repo_context`](https://github.com/rgommers/meson-python/blob/b7b005974b012ce9a29fa8294b2f911d4edd3013/tests/conftest.py#L32) should create a brand-new repo and use it for the tests, but something does not seem to be working correctly there.

FFY00:
This needs to be better documented, but the `package_*` fixtures will change the current directory to the target package.

Issue #93, opened 2022-07-05
`test_wheel.py::test_local_lib` failure in CI


rgommers:
Test failure just observed in https://github.com/FFY00/meson-python/runs/7202624920?check_suite_focus=true. This should be due to the Meson 0.63.0 release on July 3rd (2 days ago):

```
________________________________ test_local_lib ________________________________
virtual_env = PosixPath('/tmp/mesonpy-test-venv-n30nt69r/bin/python')
wheel_link_against_local_lib = PosixPath('/tmp/pytest-of-runner/pytest-0/test0/mesonpy-test-1jnqmmle/link_against_local_lib-1.0.0-py3-none-linux_x86_64.whl')
    @pytest.mark.skipif(platform.system() != 'Linux', reason='Unsupported on this platform for now')
    def test_local_lib(virtual_env, wheel_link_against_local_lib):
        subprocess.check_call([virtual_env, '-m', 'pip', 'install', wheel_link_against_local_lib])
>       subprocess.check_output([
            virtual_env, '-c', 'import example; print(example.example_sum(1, 2))'
        ]).decode() == '3'
virtual_env = PosixPath('/tmp/mesonpy-test-venv-n30nt69r/bin/python')
wheel_link_against_local_lib = PosixPath('/tmp/pytest-of-runner/pytest-0/test0/mesonpy-test-1jnqmmle/link_against_local_lib-1.0.0-py3-none-linux_x86_64.whl')
tests/test_wheel.py:119: 
...
---------------------------- Captured stdout setup -----------------------------
! Using Meson to generate the project metadata (no `project` section in pyproject.toml)
+ meson setup --native-file=/home/runner/work/meson-python/meson-python/tests/packages/link-against-local-lib/.mesonpy-native-file.ini -Ddebug=false -Dstrip=true -Doptimization=2 --prefix=/opt/hostedtoolcache/Python/3.11.0-beta.3/x64 /home/runner/work/meson-python/meson-python/tests/packages/link-against-local-lib /home/runner/work/meson-python/meson-python/tests/packages/link-against-local-lib/.mesonpy-7s3pci72/build
The Meson build system
Version: 0.63.0
Source dir: /home/runner/work/meson-python/meson-python/tests/packages/link-against-local-lib
Build dir: /home/runner/work/meson-python/meson-python/tests/packages/link-against-local-lib/.mesonpy-7s3pci72/build
Build type: native build
Project name: link-against-local-lib
Project version: 1.0.0
[...]

PR #94, opened 2022-07-05
TST: fix module name in a the purelib-and-platlib C extension

rgommers:
The second argument to `PyModuleDef` is the module name, and that should match the `PyInit_modulename` and the name used in `py.extension_module('modulename')`.

Note that this is somewhat related to gh-93, and I found this when looking at that issue. The test that exercises this code is marked as `xfail` though, so this PR only prevents a possible future issue with this test.

### DIFF
--- a/tests/packages/purelib-and-platlib/plat.c
+++ b/tests/packages/purelib-and-platlib/plat.c
@@ -18,7 +18,7 @@ static struct PyModuleDef module = {
     methods,
 };
 
-PyMODINIT_FUNC PyInit_foo(void)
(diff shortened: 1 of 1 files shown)


FFY00 (approved):

FFY00:
It seems it is looking for `example`, not `plat`?

rgommers:
> It seems it is looking for `example`, not `plat`?

That's another test that is failing, I don't quite understand yet why. This fix is for `packages/purelib-and-platlib/`, the failing one is in `packages/link-against-local-lib`.

PR #96, opened 2022-07-06
BUG: include license file in the metadata directory

### LINKED ISSUE #88: `license` field in `pyproject.toml` is ignored
I'm just noticing this now in the SciPy wheels: the license file specified in `pyproject.toml` is ignored by `meson-python`.

The relevant metadata is:
```toml
[project]
name = "SciPy"
license = {file = "LICENSE.txt"}
```
which should be correct I think, see https://peps.python.org/pep-0621/#license. From that text: _The `file` key has a string value that is a relative file path to the file which contains the license for the project_

Now what is not specified is what should be done with that. What PEP 621 further states is _"The `text` key has a string value which is the license of the project whose meaning is that of the License field from the [core metadata](https://packaging.python.org/specifications/core-metadata/). These keys are mutually exclusive, so a tool MUST raise an error if the metadata specifies both keys."_ Which I think is insufficient, because we need both the license of a package, and the license of individual wheels, which are definitely not the same (wheels can vendor platform-specific components which are under different licenses for example).

In addition to the `license` field, we have the PyPI classifier:
```toml
classifiers = [
    "License :: OSI Approved :: BSD License",
```

Now how should all this work in practice? Here's what I think:

1. The PyPI classifier needs the license of the project itself (fine as is for SciPy).
2. The `Meta > License` field in the left sidebar on the PyPI page (see, e.g., https://test.pypi.org/project/scipy/) needs that same license. But how is it populated, from the classifier or from a `text` field in the PEP 621 license key?
3. The file specified with `license = {file = "LICENSE.txt"}` needs to be included in both sdist and wheels. It does not need to be installed onto the end user's system when the wheel is installed (which makes sense, because there is nowhere for that file to go, it's at the top level in the tree, next to `scipy/` rather than under it).

The above means PEP 621 is wrong. There is an alternative that meets 1-3 above:
- Use `license = {text = "BSD 3-Clause"}` in the metadata, don't specify `file` there.
- Explicitly include the license file by something like `py3.install_sources(['LICENSE.txt'])` in the top-level `meson.build` file
  - The problem with this is that we _don't_ want that file to be installed, but only included in the sdist and wheels. Currently I don't think there is a way to do this?

FFY00:
Fixes #88


FFY00:
@rgommers can you have a look?

FFY00:
I merged, but please let me know what you think and if you ran into any issues.

rgommers:
Thanks @FFY00, that did the job:)

PR #101, opened 2022-07-08
REL: update version to 0.7.0.dev0

rgommers:
This is good practice I think, but more importantly: it makes it easy to distinguish between versions installed from source during development and versions installed from PyPI or conda-forge. It has happened to me multiple times already that I see a possible issue in SciPy that may be related, or that I know is fixed in `main` but not the latest release, and want to check the version to distinguish between those two.

I suggest to make a commit like this right after a version is tagged. One thing to discuss: should it be `0.7.0.dev0` or `0.6.1.dev0`? I'm used to the former because of maintaining separate release branches to make bugfix-only releases from.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,4 +1,4 @@
-project('mesonpy', version: '0.6.0')
+project('mesonpy', version: '0.7.0.dev0')
 
 py_mod = import('python')
 py = py_mod.find_installation()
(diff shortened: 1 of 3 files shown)


FFY00 (approved):
That is a good idea, though it can make tagging bugfix releases slightly more annoying :sweat_smile:

Issue #82, opened 2022-06-14
Wrong script permission for sdist


windreamer:
Assume we have a package with a python script included. And in source tree the script has mode 755.

The corresponding `meson.build` is
```python
project('library')

scripts = files('scripts/example')
install_data(scripts, install_dir: get_option('bindir'), install_mode: 'rwxr-xr-x')
```

The package tree
```
.
├── [-rw-rw-r--]  meson.build
├── [-rw-rw-r--]  pyproject.toml
└── [drwxrwxr-x]  scripts
    └── [-rwxrwxr-x]  example
```

The `pyproject.toml`
```toml
[build-system]
build-backend = 'mesonpy'
requires = ['meson-python']

[project]
name = 'library'
version = '1.0.0
```

And after we build the package:
```
pyproject-build . -Cbuilddir=build
```

We will find the sdist file in `meson-dist` is good 
```
✗ tar tfvz build/meson-dist/library-undefined.tar.gz
drwxrwxr-x windreamer/windreamer 0 2022-06-14 09:38 library-undefined/
-rw-rw-r-- windreamer/windreamer 139 2022-06-14 09:33 library-undefined/meson.build
-rw-rw-r-- windreamer/windreamer 115 2022-06-14 09:33 library-undefined/pyproject.toml
drwxrwxr-x windreamer/windreamer   0 2022-06-14 09:33 library-undefined/scripts/
-rwxrwxr-x windreamer/windreamer  75 2022-06-14 09:33 library-undefined/scripts/example
```

But the final sdist file is not good (`library-1.0.0/scripts/example` should have a 755 mode)
```
✗ tar tfvz dist/library-1.0.0.tar.gz
-rw-r--r-- 0/0             139 1970-01-01 00:00 library-1.0.0/meson.build
-rw-r--r-- 0/0             115 1970-01-01 00:00 library-1.0.0/pyproject.toml
-rw-r--r-- 0/0              75 1970-01-01 00:00 library-1.0.0/scripts/example
-rw-r--r-- 0/0              51 1970-01-01 00:00 library-1.0.0/PKG-INFO
```

We need the script in 755 mode, as `build` by default create wheel package via sdist instead of via source code. and the mode of the script in wheel will be determined by the file mode of sdist.

FFY00:
> Why do we only copy file sizes?

No reason, we should be copying the permissions too.

PR #102, opened 2022-07-09
BUG: preserve file permissions for sdist

rgommers:
Closes gh-82

Note that the bug report is only about the sdist. The wheel generation uses `shutil.copy2`, which should do the right thing already.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -863,7 +863,9 @@ class Project():
                     continue
 
                 info = tarfile.TarInfo(member.name)
-                info.size = os.path.getsize(path)
+                file_stat = os.stat(path)
+                info.size = file_stat.st_size
+                info.mode = int(oct(file_stat.st_mode)[-3:])
 
                 # rewrite the path if necessary, to match the sdist distribution name
                 if dist_name != meson_dist_name:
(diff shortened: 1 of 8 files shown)


rgommers:
Ah fun to test - git does not preserve permissions aside from the executable bit. The checkout depends on `umask` (see https://stackoverflow.com/questions/22239508/git-pull-is-setting-664-instead-of-644-permissions), so we see `644` on Linux/macOS and `666` on Windows in CI.

For installing a wheel, only the executable bit matters, `pip` does not preserve the rest: https://github.com/pypa/pip/blob/d0051d3266ba11e7410a43a040dfd409af88a1db/src/pip/_internal/utils/unpacking.py#L105-L110

So the bug fix is fine here, but for a robust test it looks like we should only test the executable bit.

rgommers:
@FFY00 would you then prefer this to be a separate test, rather than included in `test_contents_subdirs`?

FFY00:
Yep, I would, but this is acceptable too.

h-vetinari:
> so we see `644` on Linux/macOS and `666` on Windows in CI.

Permissions on windows don't follow the unix permission model anyway. You probably know this [article](https://fasterthanli.me/articles/i-want-off-mr-golangs-wild-ride), and it talks about much, _much_ more than that, but I would just reflect this in the windows tests - or perhaps even better - just test for the main executable bit.

FFY00 (approved):
Thanks!

I have added a test for the built wheel and skipped the executable bit tests on Windows.

rgommers:
thanks!

PR #103, opened 2022-07-12
BUG: fix ABI tag calculation when Meson returns a full path

FFY00:
Signed-off-by: Filipe Laíns <lains@riseup.net>

PR #104, opened 2022-07-13
ENH: use Meson paths based on {moduledir_shared}

FFY00:
This seems to have been fixed, so we can now start to use them.

See https://github.com/FFY00/meson-python/issues/95#issuecomment-1179215646

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 build-backend = 'mesonpy'
 backend-path = ['.']
 requires = [
-  'meson',
+  'meson>=0.63.0',
   'ninja',
   'pyproject-metadata>=0.5.0',
   'tomli>=1.0.0',
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -134,11 +134,7 @@ class _WheelBuilder():
     _SCHEME_MAP: ClassVar[Dict[str, Tuple[str, ...]]] = {
         'scripts': ('{bindir}',),
         'purelib': ('{py_purelib}',),
-        # {moduledir_shared} should be listed in platlib here too, but currently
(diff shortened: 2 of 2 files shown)


FFY00:
@eli-schwartz from what you've told me, this has been fixed, but we ran into https://github.com/mesonbuild/meson/issues/10601. It could be that only some code paths are fixed, and others may be now a bit buggy. With #103, we should be able to handle full paths.
This PR changes the code to trust paths that use the `{moduledir_shared}` placeholder, but I am worried that the paths are not as fixed as we thought they were, given that we ran into https://github.com/mesonbuild/meson/issues/10601. Could you have a look? What do you think?

pyproject.toml
```diff
@@ -2,7 +2,7 @@
 build-backend = 'mesonpy'
 backend-path = ['.']
 requires = [
-  'meson>=0.62.0',
+  'meson>=0.63.0',
```
rgommers, line 5: Just a comment that yes this should be fine, and a reminder-to-self that I should bump the version numbers for SciPy

eli-schwartz:
@FFY00 can you verify that this is fixed in meson git master?

It should use the py_platlibdir tags, not moduledir_shared.

rgommers:
I just re-tried with different `meson` versions:
- `0.62.2` produces `{moduledir_shared}` in `intro-install_plan.json` for SciPy
- in `0.63.0`, `0.63.1`, `0.63.2` and `master` there is no more `{moduledir_shared}` at all
- the absolute path issue is fixed in `master` but not in `0.63.2`

For this PR, I think it's unblocked and can be merged when CI is green. The changes seem correct. The only reason not to merge it would be if we want to do another release based off of `main` where we still need `0.62.x` in order to avoid the "absolute path" issue (which is `0.63.x`-only).

PR #112, opened 2022-07-26
TST: mark test_local_lib as xfail on github actions

FFY00:


FFY00:
I still haven't been able to track down the reason it fails on github actions, but it's unrelated.

PR #113, opened 2022-07-26
TST: force colors on CI

FFY00:


FFY00:
And we've uncovered a bug in the test suite :rofl:

PR #116, opened 2022-07-26
MAINT: ensure minimum Meson version is consistent in pyproject.toml

rgommers:
This dependency was bumped from 0.60.0 to 0.62.0 in gh-58, but only in one of the two places where it was needed. These two places should probably always be kept consistent with each other, even though technically it's not _required_.


FFY00 (approved):
Yep, good catch!

Issue #54, opened 2022-05-16
Passing through project-specific config options


rgommers:
The docs currently mention one option: https://meson-python.readthedocs.io/en/latest/usage/build-options.html. This is a `meson-python` build option. I am trying to figure out if/how it is possible to pass project-specific options through. For example, in SciPy we have a `meson_options.txt` file containing:
```meson
option('blas', type: 'string', value: 'openblas',
        description: 'option for BLAS library switching')
option('lapack', type: 'string', value: 'openblas',
        description: 'option for LAPACK library switching')
```

I don't think it is currently possible to use those with either `build` or `pip`? Something like:
```bash
$ python -m build "-C--Dblas=mkl --Dlapack=lapack"
```
Note that the `build` docs are a bit fuzzy on `-C--` for config settings, or how to pass on multiple settings.

Is this possible at all, or is it planned for the future?

FFY00:
Yes, I am working on it, but that will play with the default Meson options in the config file and I am struggling to decide on the design.

Passing multiple backend config options in pypa/build works just like in Meson with `-D`, it ends up looking a bit weird but there isn´t much we can do about that.

```
$ python -m build -C-Dblas=mkl -C-Dlapack=lapack
```

rgommers:
Cool, good to know. This one is not urgent, it's more a nice-to-have for SciPy 1.9.0 I'd say.

andyfaff:
I'm currently trying to setup [cibuildwheel infrastructure for scipy](https://github.com/andyfaff/scipy/pull/28).
The scipy [pyproject.toml](https://github.com/andyfaff/scipy/blob/cibw2/pyproject.toml) specifies the use of `mesonpy` as the build backend (ultimately using meson/ninja underneath). cibuildwheel just asks `pip` or `build` to make the wheel.

However, in order to target `cp3*-macosx_arm64` cross compilation has to be used, because there is an `arm64` host on an `x86_64` build machine (Github Actions). This requires the use of a [cross-compile file](https://mesonbuild.com/Cross-compilation.html#changing-cross-file-settings). I can find documentation on how to supply this file directly to a meson build (I think it gets placed in the build directory). However, there doesn't seem to be a way of supplying the cross-compile file as options to a `pip/build`.

e.g. I can't do:
```
python -m build . <my_cross_compile_build_flags_to_be_supplied_to_meson>
```

Will this PR enable this?

rgommers:
[...]

Issue #108, opened 2022-07-22
Specify options for meson in pyproject.toml


llohse:
Is it possible to specify options for meson in the pyproject.toml? If yes, could you document how?

Consider a project that has optional python bindings. I would need to enable them, when building via pip / pypa build.
mesonpep517 has `meson-options` that get passed to the meson call, but I could not find anything comparable in meson-python.

rgommers:
That does sound like a reasonable feature request to me. It's not possible right now to do this in `pyproject.toml`; there is no `meson-python` specific section there.

xref `mesonpep517` docs: https://thiblahute.gitlab.io/mesonpep517/pyproject.html#mesonoptions

rgommers:
Based on https://github.com/ufo-kit/ufo-core/pull/189#issuecomment-1192463588, I think this doesn't necessarily need to live in `pyproject.toml`. If it's a _default_ option, then it can live in `meson.build` files. If it's not default but a _user-specific customization_, then it should not be done via editing `pyproject.toml` but via the CLI interface similar to the `mesonpep517` link above. The latter is covered by gh-122.

@llohse does that work for you?

llohse:
Mhm, not really.  What I have in mind is rather a convenience feature as follows.
An option that is *not* default when building the project (via meson) but *is* default when building it via pep517.

For example, consider a C/C++ library with *optional* python bindings. When installing the library via pep517, it would not make sense to compile it without the python bindings.
Always having to call `pip install package --build-option="--with-python"` is be rather inconvenient and forcing the package to build the python bindings by default also might not always be what you want.

It might be possible to implement this in meson.build by somehow detecting when being called by meson-python (although I would not know how) but I am not sure if that would be any cleaner.

rgommers:
Ah okay, thanks for the clarification. It's clear to me now. So gh-122 is a prerequisite, but this makes sense as a follow-up. The interface should follow whatever comes out of gh-122. Something like

```toml
[tool.meson-python]
setup-args = "-Doption=value"
install-args = "-Dother-option=other-value"
```

> It might be possible to implement this in meson.build by somehow detecting when being called by meson-python (although I would not know how) but I am not sure if that would be any cleaner.

No, that seems worse. `meson` should not know at all who is calling it, only what arguments are passed to it.

eli-schwartz:
> It might be possible to implement this in meson.build by somehow detecting when being called by meson-python (although I would not know how) but I am not sure if that would be any cleaner.

[...]

Issue #35, opened 2022-02-02
Skip subprojects when building the wheel


tristan957:
`WARNING File could not be mapped to an equivalent wheel directory: /tmp/mesonpy-udbdqxaw/install/usr/lib64/libxxhash.a ({libdir_static}/libxxhash.a)`

Example project: https://github.com/hse-project/hse-python. Use the `tristan957/mesonpy` branch.

dogtopus:
I also have a upcoming project that can benefit from this. Currently I have pure-static external subprojects which there's currently no way to skip the installation on the meson side other than specifying `--skip-subprojects`. It's fine (although hacky and error prone) when manually invoking meson but with a builder script it becomes a problem.

FFY00:
Hi @tristan957 and @dogtopus, thank you for opening this issue, and sorry for the late reply. There is no location in wheels for static libraries, and they aren't needed at runtime, so we should probably just ignore them. I will try to look into a PR tomorrow.

tristan957:
Very cool. Happy to see some movement on this issue

rgommers:
> There is no location in wheels for static libraries, and they aren't needed at runtime, so we should probably just ignore them. I will try to look into a PR tomorrow.

I should have commented on this earlier, but: this statement is unfortunately not 100% accurate. NumPy ships static libraries (`libnpymath.a` and `libnpyrandom.a`) that live inside `site-packages/numpy/`. Rebuilding those from source for consuming libraries is highly nontrivial, so I don't think that NumPy will be able to get rid of them very easily. 

Shipping static libraries is obviously not the best design, and it makes assumptions about the compiler toolchain that consumers are using (must be MSVC-compatible on Windows, GCC-compatible on Linux and Clang-compatible on macOS).

I suspect that when NumPy moves to Meson, we can probably work around this problem by not using `static_library(..., install: true)` (which I think is what produces `{libdir_static}`) but `custom_target`.

PR #122, opened 2022-07-31
ENH: pass through `config_settings` coming from build frontends

rgommers:
Closes gh-54

This came up when Spack tried to package SciPy 1.9.0 (see https://github.com/spack/spack/pull/31810). It turned out that most of the infra is already there, `config_settings` simply needed plumbing through. I have tested that this works for SciPy, when building a wheel against a non-default BLAS/LAPACK.

@FFY00 in gh-54 you said _"I am struggling to decide on the design"_, but as far as I can tell there is no UI design aspect here? This is up to build front-ends, all a backend has to do is pass on the `config_settings` dict that is passed into `build_sdist` and `build_wheel`?

I haven't added tests yet, I thought I'd first put this up for feedback. 

EDIT: `config_settings` is part of PEP 517: https://peps.python.org/pep-0517/#config-settings

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -406,6 +406,7 @@ class Project():
 
     def __init__(
         self,
+        config_settings: Dict[Any, Any],
         source_dir: Path,
         working_dir: Path,
         build_dir: Optional[Path] = None,
@@ -460,7 +461,7 @@ class Project():
                 self._meson_native_file.write_text(native_file_data)
 
         # configure the meson project; reconfigure if the user provided a build directory
-        self._configure(reconfigure=bool(build_dir) and not native_file_mismatch)
+        self._configure(config_settings, reconfigure=bool(build_dir) and not native_file_mismatch)
 
         # set version if dynamic (this fetches it from Meson)
         if self._metadata and 'version' in self._metadata.dynamic:
@@ -476,17 +477,22 @@ class Project():
         with mesonpy._util.cd(self._build_dir):
             return self._proc('meson', *args)
 
-    def _configure(self, reconfigure: bool = False) -> None:
+    def _configure(self, config_settings: Dict[Any, Any], reconfigure: bool = False) -> None:
         """Configure Meson project.
 
         We will try to reconfigure the build directory if possible to avoid
         expensive rebuilds.
         """
-        setup_args = [
+        setup_args = []
+        for key, value in config_settings.items():
+            setup_args.append(f'{key}={value}')
+
+        setup_args += [
             f'--prefix={sys.base_prefix}',
             os.fspath(self._source_dir),
             os.fspath(self._build_dir),
         ]
+
         if reconfigure:
             setup_args.insert(0, '--reconfigure')
 
@@ -494,14 +500,13 @@ class Project():
             self._meson(
                 'setup',
                 f'--native-file={os.fspath(self._meson_native_file)}',
-                # TODO: Allow configuring these arguments
                 '-Ddebug=false',
                 '-Doptimization=2',
                 *setup_args,
             )
         except subprocess.CalledProcessError:
             if reconfigure:  # if failed reconfiguring, try a normal configure
-                self._configure()
+                self._configure(config_settings)
             else:
                 raise
 
@@ -539,12 +544,16 @@ class Project():
     @contextlib.contextmanager
     def with_temp_working_dir(
         cls,
+        config_settings: Optional[Dict[Any, Any]] = None,
         source_dir: Path = os.path.curdir,
         build_dir: Optional[Path] = None,
     ) -> Iterator[Project]:
         """Creates a project instance pointing to a temporary working directory."""
+        if config_settings is None:
+            config_settings = {}
+
         with tempfile.TemporaryDirectory(prefix='.mesonpy-', dir=os.fspath(source_dir)) as tmpdir:
-            yield cls(source_dir, tmpdir, build_dir)
+            yield cls(config_settings, source_dir, tmpdir, build_dir)
 
     @functools.lru_cache()
     def _info(self, name: str) -> Dict[str, Any]:
@@ -901,8 +910,11 @@ def _project(config_settings: Optional[Dict[Any, Any]]) -> Iterator[Project]:
     if config_settings is None:
         config_settings = {}
 
+    build_dir = config_settings.pop('builddir', None)
+
     with Project.with_temp_working_dir(
-        build_dir=config_settings.get('builddir'),
+        config_settings=config_settings,
+        build_dir=build_dir,
     ) as project:
         yield project
 


FFY00:
Thanks for opening this. Unfortunately, it is not that simple :sweat_smile:. Blindly passing the options will cause breakage if you mess with any of the options we rely on. That said, the `-D` option would not be an issue, so we can allow that option for now and then work on the other ones.

> @FFY00 in [gh-54](https://github.com/FFY00/meson-python/issues/54) you said _"I am struggling to decide on the design"_, but as far as I can tell there is no UI design aspect here? This is up to build front-ends, all a backend has to do is pass on the `config_settings` dict that is passed into `build_sdist` and `build_wheel`?

Well, we do multiple meson calls! Currently, we do `setup`, `compile`, and `install`. So, there are indeed some design decisions there. There is also specifying the extra Meson arguments in the config and then adding/overriding them on the call.

eli-schwartz:
> Blindly passing the options will cause breakage if you mess with any of the options we rely on. That said, the `-D` option would not be an issue, so we can allow that option for now and then work on the other ones.

I think the only option you rely on is `-Dprefix={sys.base_prefix}` (yes, that can be a -D option)?

> Well, we do multiple meson calls! Currently, we do `setup`, `compile`, and `install`.

Only the `meson setup` program is relevant here, I guess. :) The other two programs merely happen to share a multi-call program launcher.

FFY00:
> I think the only option you rely on is `-Dprefix={sys.base_prefix}` (yes, that can be a -D option)?

IIRC changing the prefix should only be an issue in the heuristics code paths, but yeah, we do set that. We also use `--native-file`, so we cannot let users override that, at least blindly. Similarly, cross builds will need some extra handling.

> Only the `meson setup` program is relevant here, I guess. :) The other two programs merely happen to share a multi-call program launcher.

Yes, but I would want to allow users to pass `--{ninja,vs,xcode}-args` to `meson compile` if they need, or tweak `--jobs` and `--load-average`.

In `meson install`, some users might want `--skip-subprojects`, but I am still not entirely sure how to handle that. See #35 for eg.

eli-schwartz:
> We also use `--native-file`, so we cannot let users override that, at least blindly.

These will simply stack, so passing your native file as the very last option would work and override any settings the user specified.

TBH I'd want to configure the python installation being compiled for 🤷‍♂️ but that might be challenging for build requirements.

FFY00:
> These will simply stack, so passing your native file as the very last option would work and override any settings the user specified.

Yep, though the UX of quietly overriding it is not great, so we'd probably want to try detecting those situations and either throw an error or a warning. Is there anything fancy in the syntax that would make parsing not trivial?

> TBH I'd want to configure the python installation being compiled for man_shrugging but that might be challenging for build requirements.

Well, that goes against PEP 517 :sweat_smile:. For cross builds, I have thrown around the idea of defining a new hook specifically for that, so that build backends may support that use case correctly. That needs a new PEP though, so I am not sure when I'll get around to do it.

eli-schwartz:
> Yep, though the UX of quietly overriding it is not great, so we'd probably want to try detecting those situations and either throw an error or a warning. Is there anything fancy in the syntax that would make parsing not trivial?

The file must be a valid configparser file, but the RHS of a binaries entry can have a limited amount of meson.build syntax. `mesonbuild.coredata -> class MachineFileParser(configparser.ConfigParser)` is the reference implementation. Off the top of my head, the things which might actually be expected to see in a binaries entry are:
- `[]` arrays (such as `cpp = ['ccache', 'g++']`)
- `+` concatenation
- `/` path joining
- variable references to a constants entry, e.g. a target triplet

Since you don't actually care about the contents, only whether the key exists, you can simply load it as a configparser file, and check the names of binaries being set in `cfg = ConfigParser(); cfg.read(...); cfg['binaries'].keys()`.

rgommers:
It's indeed the case that the vast majority of options need to go to `setup`. This PR as it stands does not break anything that is possible today, and just adds that capability. Filtering `config_settings` to pass on known supported options to `compile` / `install` steps should not be hard to add, your suggestions in https://github.com/FFY00/meson-python/pull/122#issuecomment-1211366152 seem like the right ones to me.

The optimization and debug levels we use can be overridden, that requires custom handling.

> Yep, though the UX of quietly overriding it is not great, so we'd probably want to try detecting those situations and either throw an error or a warning.

I'd throw an error for `--native-file` for now, and leave any parsing till the time someone needs it (let's open a new enhancement issue to track).

> That said, the `-D` option would not be an issue, so we can allow that option for now and then work on the other ones.

This isn't a great approach. Each time someone needs to pass something else to `setup`, they're going to run into a problem, come here to open an issue, and then wait for a new release. Unless there's a good reason to not pass everything on to `setup`, we should do that instead.

rgommers:
@FFY00 and I had a chat about this PR on Friday. One issue @FFY00 pointed out was that if any new options are added to the `build` or `install` stages in Meson, those are not supported through the UX proposed in this PR until they are explicitly handled. That shouldn't happen often, but it is a small downside.

I now understand better the alternative he had in mind: explicitly letting the user define which options go to `setup`, which to `build` and to `install`. That is more comprehensive and requires less special handling of options to redirect them. 

The flip side of that coin is that the UX is going to be more verbose. It's not clear exactly what that UX should be yet. The `pip` and `build` `--config-settings` docs say it must be `key=value` pairs as the argument:
- https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-config-settings
- https://pypa-build.readthedocs.io/en/stable/#python--m-build---config-setting

So likely something like:
```
-config-settings --setup-opt-key=value
-config-settings --build-opt-key=value
-config-settings --install-opt-key=value
# or shorter (can be with or without a space after -C):
-C--setup-opt-key=value
-C--build-opt-key=value
-C--install-opt-key=value

# With concrete options:
-C --setup-opt--Dblas=blis
-C --build-opt--n=8
-C --install-opt---strip
```

There's a couple of different ways of spelling this (e.g. `--setup-key` instead of `--setup-opt-key`, and one could automatically prepend one dash to `key` so it'd be `--setup-Dblas=blis` rather than `--setup--Dblas-blis`). The principle is clear though.

Opinions pro or con this idea? If separate setup/build/install stages, preferred way of spelling it?

eli-schwartz:
> if any new options are added to the `build` or `install` stages in Meson

The install stage does have some fine-tuning you can do to it, e.g. you can define a subset of tags to be installed, or skip subprojects. I'm not sure how useful either one is for wheels though. There's also --strip, but that duplicates the setup option of the same name.

In general, it seems plausible that further options might be added.

In comparison, the build stage is just "run ninja". It is both more and less interesting -- there are no options to fine-tune the resulting wheel, but several options to control the resources used (parallelism, etc.)

But, the build stage is also probably best controlled by e.g. `$SAMUFLAGS` which is already a supported interface for the c89 Unix-only version of ninja, assuming that leaving out build stage support would allow reducing the complexity of config_settings (it might not).

rgommers:
It turns out `mesonpep517` already has this: https://thiblahute.gitlab.io/mesonpep517/pyproject.html#mesonoptions. It has
```
--setup-args
--dist-args
--install-args
```
There's one missing, `--compile-args`. The naming scheme is logical though: `--<meson-subcommand-name>-args`.

Would make sense to adopt that I think.

tristan957:
`--compile-args` doesn't exist because `mesonpep517` calls out to Meson directly. I am the one that implemented the feature there.

rgommers:
Ah, good to know, thanks. Are you still happy with how that interface works in `mesonpep517` now @tristan957?

tristan957:
Having a way to specify default options in the pyproject.toml would also be nice. Not sure if that has been taken into account yet. Passing options on the command line would override the defaults in a perfect world I think

rgommers:
> renaming them to `--meson-xxx-args` might not be a bad idea.

That may be helpful for clarity. It makes very long lines even longer, but overall that is still a good tradeoff perhaps.

> Having a way to specify default options in the pyproject.toml would also be nice.

That is gh-108. It seems reasonable, and the issue submitter there also explains why it must be in `pyproject.toml` rather than in `meson.build` or `meson-options.txt`.

tristan957:
I wasn't referring to `-Dxxx` options. I was referring to `--skip-subprojects` for instance.

FFY00:
Closing in favor of #167.

PR #126, opened 2022-08-02
Project.platform_tag: Handle the case of 32bit Python on x86_64

### LINKED ISSUE #123: platform_tag is wrong in 32-bit builds on 64-bit Linux kernels
meson-python 0.8.0 creates a wheel named `scipy-1.9.0rc3-cp310-cp310-linux_x86_64.whl` when run on an i386 Docker image, for example on `ubuntu-bionic-i386` (see https://github.com/mkoeppe/sage/runs/7599750782).

Ref: https://trac.sagemath.org/ticket/34081#comment:47

mkoeppe:
Fixes #123

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -684,6 +684,11 @@ class Project():
             # https://github.com/pypa/packaging/issues/578
             parts[1] += '.0'
             platform_ = '-'.join(parts)
+        elif parts[0] == 'linux' and parts[1] == 'x86_64' and sys.maxsize == 0x7fffffff:
+            # 32-bit Python running on an x86_64 host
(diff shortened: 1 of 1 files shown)


mkoeppe:
I've tested this successfully within Sage - https://trac.sagemath.org/ticket/34081#comment:53

rgommers (approved):
Thanks @mkoeppe. The fix looks good to me. The one question I have is whether we can expect a similar issue with 32-bit Python on any other platforms - Windows or non-x86 Linux for example. This is worth checking the code for at the same time.

mkoeppe:
I only know about this case of x86 on x86_64 on Linux, sorry

FFY00 (approved):
Thanks! I changed the maxsize check to a hex literal.

PR #132, opened 2022-08-23
DOC: correct RTD links for PyPI

takluyver:
I just tried to click these links and noticed that they didn't work.


rgommers (approved):
Thanks @takluyver

Issue #127, opened 2022-08-04
Builds incomplete wheel for SciPy because "File could not be mapped to an equivalent wheel directory"


mkoeppe:
This is on gentoo with gentoo's python3.9. @rgommers 

https://github.com/mkoeppe/sage/runs/7659423834?check_suite_focus=true

```
e75b6879f2d8 /sage # cat logs/pkgs/scipy-1.9.0rc3.log 
[...]
  + meson install --destdir /sage/local/var/lib/sage/venv-python3.9/var/tmp/sage/build/scipy-1.9.0rc3/src/.mesonpy-cjaz4nzj/install
  WARNING Using heuristics to map files to wheel, this may result in incorrect locations
  WARNING File could not be mapped to an equivalent wheel directory: /sage/local/var/lib/sage/venv-python3.9/var/tmp/sage/build/scipy-1.9.0rc3/src/.mesonpy-cjaz4nzj/install/usr/lib/python-exec/python3.9/../../../lib/python3.9/site-packages/scipy/__config__.py (/usr/lib/python-exec/python3.9/../../../lib/python3.9/site-packages//scipy/__config__.py)
  WARNING File could not be mapped to an equivalent wheel directory: /sage/local/var/lib/sage/venv-python3.9/var/tmp/sage/build/scipy-1.9.0rc3/src/.mesonpy-cjaz4nzj/install/usr/lib/python-exec/python3.9/../../../lib/python3.9/site-packages/scipy/version.py (/usr/lib/python-exec/python3.9/../../../lib/python3.9/site-packages//scipy/version.py)
  WARNING File could not be mapped to an equivalent wheel directory: /sage/local/var/lib/sage/venv-python3.9/var/tmp/sage/build/scipy-1.9.0rc3/src/.mesonpy-cjaz4nzj/install/usr/lib/python-exec/python3.9/../../../lib/python3.9/site-packages/scipy/_lib/_ccallback_c.cpython-39-x86_64-linux-gnu.so ({moduledir_shared}/_ccallback_c.cpython-39-x86_64-linux-gnu.so)
[...]

PR #133, opened 2022-09-03
CI: Run tests against Python from Debian, Fedora, Centos, Gentoo, Archlinux, OpenSUSE, conda-forge

### LINKED ISSUE #76: Run tests against Debian Python in CI

mkoeppe:
Using the reusable workflows from SageMath

Fixes #76

### DIFF
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -92,20 +92,6 @@ jobs:
           path: upstream
           name: upstream
 
-  cygwin:
-    # https://github.com/sagemath/sage/blob/develop/.github/workflows/cygwin.yml
-    uses: sagemath/sage/.github/workflows/cygwin.yml@develop
-    with:
-      # FIXME: duplicated from env.TARGETS
-      targets: SAGE_CHECK=no SAGE_CHECK_PACKAGES="meson_python,scipy" scipy
-      prefix: /opt/sage-meson-python
-      sage_repo: sagemath/sage
-      sage_ref: develop
-      sage_trac_git: https://github.com/sagemath/sagetrac-mirror.git
-      sage_trac_ticket: 34081
-      upstream_artifact: upstream
-    needs: [dist]
-
   linux:
     # https://github.com/sagemath/sage/blob/develop/.github/workflows/docker.yml
     uses: sagemath/sage/.github/workflows/docker.yml@develop
@@ -114,27 +100,12 @@ jobs:
       # Full list at https://github.com/sagemath/sage/blob/develop/.github/workflows/docker.yml
       # See also https://github.com/sagemath/sage/blob/develop/tox.ini
       tox_system_factors: >-
-          ["ubuntu-focal",
-           "ubuntu-jammy",
-           "ubuntu-kinetic",
-           "debian-bullseye",
-           "debian-bookworm",
-           "debian-sid",
-           "fedora-31",
-           "fedora-32",
-           "fedora-33",
-           "fedora-34",
-           "fedora-35",
+          ["debian-bullseye",
            "fedora-36",
-           "fedora-37",
            "centos-stream-8",
-           "centos-stream-9",
            "gentoo-python3.9",
-           "gentoo-python3.10",
            "archlinux-latest",
-           "opensuse-15.3",
            "opensuse-15.4",
-           "opensuse-tumbleweed",
            "conda-forge",
            ]
       tox_packages_factors: >-
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -83,7 +83,8 @@ jobs:
 
   linux:
     # https://github.com/sagemath/sage/blob/develop/.github/workflows/docker.yml
-    uses: sagemath/sage/.github/workflows/docker.yml@develop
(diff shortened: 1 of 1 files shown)


rgommers (commented):
Wow, that is comprehensive. Thanks for working on this @mkoeppe! 

This is very heavy-duty, probably a bit too much. Could we pare it down to a single job per distro? The other ones could be left in for now as a comment, in the (much less likely) case that something works for say Fedora 37 but is broken for 36.

It looks like we'd want on job each of:
- Debian (leaving out Ubuntu, since it's basically identical)
- Fedora
- CentOS
- Gentoo
- Arch
- OpenSUSE
- conda-forge
- Cygwin

There seems to be some naming issue that caused jobs to not run.

I haven't looked at the Sage actions yet to see what they're actually doing.

rgommers:
There's already a separate Cygwin job (https://github.com/FFY00/meson-python/blob/main/.github/workflows/tests-cygwin.yml), so that can be left out of this PR probably?

mkoeppe:
Here's a better version. 
Currently scipy's `runtest.sh` is failing for me:
```
  [scipy-1.9.1]     File "/sage/local/var/lib/sage/venv-python3.9/var/tmp/sage/build/scipy-1.9.1/src/runtests.py", line 325, in main
  [scipy-1.9.1]       shutil.copyfile(os.path.join(ROOT_DIR, '.coveragerc'),
  [scipy-1.9.1]     File "/usr/lib/python3.9/shutil.py", line 264, in copyfile
  [scipy-1.9.1]       with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
  [scipy-1.9.1]   FileNotFoundError: [Errno 2] No such file or directory: '/sage/local/var/lib/sage/venv-python3.9/var/tmp/sage/build/scipy-1.9.1/src/.coveragerc'
```
@rgommers What is the recommended way to test scipy when using an sdist, not a git clone?

rgommers:
> @rgommers What is the recommended way to test scipy when using an sdist, not a git clone?

The CI log shows that a SciPy 1.9.1 wheel is successfully built and installed, so it's not sdist vs git clone. The issue here is trying to run the tests with `runtests.py`, which is a distutils-specific development wrapper. The tests can be run with one of:
- `python -c "import scipy; scipy.test()", or
- `pytest --pyargs scipy`

In case the test script is in the root of the scipy repo, make sure to do `cd ..` or some such thing, to avoid Python seeing the local source tree instead of the installed package.

Note that this way of testing is recommended also for upstream Sage - it will work independent from with which build system SciPy was built.

mkoeppe:
> * `python -c "import scipy; scipy.test()" [...]

Thanks, I've made this change now in https://trac.sagemath.org/ticket/34081

> Note that this way of testing is recommended also for upstream Sage - it will work independent from with which build system SciPy was built.

Yes, it's the same script that will run for our CI and the one here.

mkoeppe:
Force-push to re-run the tests with the updated https://trac.sagemath.org/ticket/34081

mkoeppe:
This is now ready: It reveals the failure on gentoo reported in #127 and passes on the other platforms.

FFY00 (approved):
Thank you! I had a look and this seems great.

@rgommers do you have any further concerns, or can we go ahead and merge?

rgommers:
It looks very good indeed. It's heavy CI, but for the time being this is very useful. Longer-term we may want to use a lighter package than SciPy inside the `meson-python` CI, but for now I think it'll guard against more issues due to Linux distro Python install layout.

Note that there is one more issue on Gentoo shown in this PR's CI results. That will be fixed in SciPy, see https://github.com/FFY00/meson-python/issues/127#issuecomment-1237952409. I think we can merge this now though, and CI will run more often which is useful. As soon as the SciPy patch lands, this will be 100% green.

FFY00:
Yep, and we can also revisit this later.

rgommers:
@mkoeppe there is something wrong with this after merging. The Gentoo log on this PR shows:
```
 make --no-print-directory scipy-SAGE_VENV-no-deps
[scipy-1.9.1] installing. Log file: /sage/logs/pkgs/scipy-1.9.1.log
  [scipy-1.9.1] error installing, exit status 1. End of log file:
```

While [the most recent log](https://github.com/FFY00/meson-python/runs/8255403529?check_suite_focus=true) on `main` shows:
```
 [scipy-1.8.1] installing. Log file: /sage/logs/pkgs/scipy-1.8.1.log
  [scipy-1.8.1] successfully installed.
```

I don't see anything in this PR that would cause that. Was there an upstream change?

mkoeppe:
In the log, I see
```
remote branch: u/mkoeppe/numpy_1_23_x__scipy_1_9_x

From https://github.com/sagemath/sagetrac-mirror
 * branch                  u/mkoeppe/numpy_1_23_x__scipy_1_9_x -> FETCH_HEAD
 * [new branch]            u/mkoeppe/numpy_1_23_x__scipy_1_9_x -> trac/u/mkoeppe/numpy_1_23_x__scipy_1_9_x

fatal: refusing to merge unrelated histories
```
I've now merged the most recent release tag into this branch, which should fix this issue.

rgommers:
great, thanks!

Issue #121, opened 2022-07-29
Unknown Python implementation: pypy38


oscarbenjamin:
This comes from a SciPy issue: https://github.com/scipy/scipy/issues/16737

As of SciPy 1.9.0 `pip install scipy` fails under PyPy with the error message:
```
+ meson install --destdir /tmp/pip-install-pkqrdcy5/scipy_8a1b991ddd494020a114650533cb4ba6/.mesonpy-ve4kjypi/install
      Traceback (most recent call last):
        File "/opt/hostedtoolcache/PyPy/3.8.13/x64/lib/pypy3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 156, in prepare_metadata_for_build_wheel
          hook = backend.prepare_metadata_for_build_wheel
      AttributeError: module 'mesonpy' has no attribute 'prepare_metadata_for_build_wheel'
      
      During handling of the above exception, another exception occurred:
      
      Traceback (most recent call last):
        File "/tmp/pip-build-env-mupbdlja/overlay/lib/pypy3.8/site-packages/mesonpy/__init__.py", line 714, in _calculate_file_abi_tag_heuristic_posix
          return mesonpy._tags.StableABITag(tag)
        File "/tmp/pip-build-env-mupbdlja/overlay/lib/pypy3.8/site-packages/mesonpy/_tags.py", line 39, in __init__
          raise ValueError(f'Invalid PEP 3149 stable ABI tag, expecting pattern `{self._REGEX.pattern}`')
      ValueError: Invalid PEP 3149 stable ABI tag, expecting pattern `^abi(?P<abi_number>[0-9]+)$`
      
      During handling of the above exception, another exception occurred:
      
      Traceback (most recent call last):
        File "/opt/hostedtoolcache/PyPy/3.8.13/x64/lib/pypy3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 363, in <module>
          main()
        File "/opt/hostedtoolcache/PyPy/3.8.13/x64/lib/pypy3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 345, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
        File "/opt/hostedtoolcache/PyPy/3.8.13/x64/lib/pypy3.8/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 160, in prepare_metadata_for_build_wheel
[...]

PR #135, opened 2022-09-09
BUG: handle PyPy tags correctly

FFY00:
@mattip can you have a look?

EDIT: addresses gh-121

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,6 +21,8 @@ jobs:
           - macos
           - windows
         python:
+          - 'pypy-3.8'
+          - 'pypy-3.9'
           - '3.7'
           - '3.8'
           - '3.9'
--- a/mesonpy/_tags.py
+++ b/mesonpy/_tags.py
@@ -4,8 +4,6 @@
 
 import abc
 import re
-import sys
-import warnings
 
 from typing import Any, Optional
 
@@ -61,7 +59,7 @@ class StableABITag(Tag):
         return hash(str(self))
 
 
-class LinuxInterpreterTag(Tag):
+class InterpreterTag(Tag):
     def __init__(self, value: str) -> None:
         parts = value.split('-')
         if len(parts) < 2:
@@ -74,7 +72,7 @@ class LinuxInterpreterTag(Tag):
         self._interpreter_version = parts[1]
         self._additional_information = parts[2:]
 
-        if self.implementation not in ('cpython', 'pypy', 'pypy3'):
+        if self.implementation != 'cpython' and not self.implementation.startswith('pypy'):
             raise NotImplementedError(
                 f'Unknown Python implementation: {self.implementation}. '
                 'Please report this to https://github.com/FFY00/mesonpy/issues '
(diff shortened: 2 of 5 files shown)


FFY00:
Also, if you have a some PyPy wheels of different versions I can have a look, please let me know.

mattip:
Are these the wheel tags? You can see some of them in the [numpy wheels](https://pypi.org/project/numpy/#files), here are some for [cryptography](https://pypi.org/project/cryptography/#files). I don't think anyone is producing python2.7 wheels, much less pypy2.7 ones.

FFY00:
Yep, these are the wheel tags. We generate the tags from the extension names. See `test_interpreter_tag` (prev. `test_linux_interpreter_tag`), it receives the extension name (eg. `pypy38-pp73-x86_64-linux-gnu` from `_multiarray_tests.pypy38-pp73-x86_64-linux-gnu.so`) and generated the Python and ABI tags.

rgommers:
pre-commit's `flake8` check is unhappy. `codecov` is is unhappy too, but that's a "don't care" I'd say.

rgommers (approved):
Thanks @FFY00. This LGTM modulo the minor CI hiccup. I checked what this actually looks like for a NumPy PyPy Windows wheel:
- The wheel filename is `numpy-1.23.3-pp38-pypy38_pp73-win_amd64.whl`
- The Python extension modules within that wheel have extensions `.pypy38-pp73-win_amd64.pyd`

Using `implementation.startswith('pypy')` looks like a good idea, that should be robust. PEP 3149 is vague in its description for `pypy` and can be interpreted as wanting `pypy-38` type encoding (with a dash), while PyPy actually does `pypy38`. So it's not impossible that this may change in the future.

The simplification into a single `InterpreterTag` makes sense to me.

rgommers:
Adding a CI job can wait. It would be nice if anyone could confirm that this fixes things though.

mesonpy/_tags.py
```diff
@@ -105,21 +104,18 @@ def python(self) -> str:
         if self.implementation == 'cpython':
             # The Python tag for CPython does not seem to include the flags suffixes.
             return f'cp{self.interpreter_version}'.rstrip('dmu')
-        elif self.implementation in ('pypy', 'pypy3'):
-            interpreter_version = f'{sys.version_info[0]}{sys.version_info[1]}'
-            return f'pp{interpreter_version}'
+        elif self.implementation.startswith('pypy'):
+            return f'pp{self.implementation[4:]}'
```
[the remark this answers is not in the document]
FFY00: PEP 3149 is not quite clear enough, but I'd make the argument that PyPy is following it. PyPy's release model is fairly different from CPython, as certain PyPy version provides many Python versions.

Here, `pypy38` would be the interpreter name, and `pp73` would be the interpreter version, which is different from the Python version.

.github/workflows/tests.yml
```diff
@@ -21,6 +21,9 @@ jobs:
           - macos
           - windows
         python:
+          - 'pypy-3.7'
+          - 'pypy-3.8'
+          - 'pypy-3.9'
```
mattip, line 26: I would drop pypy-3.7 since it is not really supported and the scientific python community doesn't really support python3.7 anymore. If CI is resource constrained, it is actually probably sufficient to only run one PyPy version, perhaps 3.9.

FFY00: Ah, gotcha. I added it because the current PyPy version still supports it.

rgommers:
Does this CI addition actually do what you wanted it to do? Last line of the log says `CPython` (and test session further down says `Python 3.9.13`):
 
image

FFY00:
Actually it doesn't! But not because of that :sweat_smile: 

We setup CPython 3.9 for nox, as that is more stable, and then use nox to run the test against the target interpreter.

PR #136, opened 2022-09-11
DOC: add a simple C extension example

peytondmurray:
## Summary

I've been using meson-python for building C extensions to python a lot recently. Starting out, I would have benefited from a really simple example with some documentation. This PR adds a simple example adapted from https://docs.python.org/3/extending/extending.html.

## Changes

- Added a `spam` example C extension. This is mostly taken from https://docs.python.org/3/extending/extending.html, but I added some more comments and modified the project structure a bit - hopefully users find the changes instructive.
- Added a test that checks that the example builds, installs, and can be imported. 
  - If I remember correctly if a problem is encountered with setting up fixtures, the whole pytest session fails. So for that reason the test for the `spam` example builds and imports the example in a virtual environment, all inside the test (I didn't make the wheel built from the example into a pytest fixture).
- Added a missing `assert` in `test_wheel.py`.

Feedback is welcome, please let me know if there's anything here that I can improve!

### DIFF
--- a/docs/examples/spam/pyproject.toml
+++ b/docs/examples/spam/pyproject.toml
@@ -1,0 +1,11 @@
+[build-system]
+requires = [
+    "meson-python",
+]
+build-backend = "mesonpy"
+
+[project]
+name = "spam"
+description = "Example C extension for Python"
+version = "0.0.1"
+requires-python = ">=3.8.0"
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 .mesonpy-native-file.ini
 docs/_build
 *.pyc
+.cache/
+dist/
--- a/docs/examples/spam/meson.build
+++ b/docs/examples/spam/meson.build
@@ -1,0 +1,37 @@
+project(
+  'spam',
+  'c',
+)
+
+py_mod = import('python')
+py = py_mod.find_installation()
+
+# Specify directories to be included - usually this is where your header files live.
+# Note that specifying the path to Python.h is not necessary with meson-python.
+includes = include_directories(
+  [
+    'src'
+  ]
+)
+
+srcs = [
+  'src/spammodule.c',
+]
+
+# If some files (such as .py files) need to be copied to site-packages,
+# this is where they get specified. Files get copied to
+# <python directory>/site-packages/<subdir>
+py.install_sources(
+  [
+    'src/__init__.py'
+  ],
+  subdir: 'spam'
+)
+
+py.extension_module(
+  '_spam',
+  srcs,
+  install: true,
+  subdir: 'spam',
+  include_directories: includes
+)
--- a/docs/examples/spam/src/__init__.py
+++ b/docs/examples/spam/src/__init__.py
@@ -1,0 +1,6 @@
+from ._spam import system
+
+
+__all__ = [
+    'system'
+]
--- a/docs/examples/spam/src/spammodule.c
+++ b/docs/examples/spam/src/spammodule.c
@@ -1,0 +1,83 @@
+// This example is adapted from the spam example taken from the python documentation
+// and is available here: https://docs.python.org/3/extending/extending.html
+#define PY_SSIZE_T_CLEAN
+#include <Python.h>
+
+static PyObject *SpamError;
+
+// Define a method that will be part of the module. This method
+// passes the arguments to `system()` and returns the output.
+static PyObject * spam_system(PyObject *self, PyObject *args) {
+    const char *command;
+    int sts;
+
+    if (!PyArg_ParseTuple(args, "s", &command)) return NULL;
+    sts = system(command);
+    return PyLong_FromLong(sts);
+}
+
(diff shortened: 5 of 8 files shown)


eli-schwartz:
> This is a pretty common pattern to expose methods from subpackages as part of a larger package (e.g. `np.linspace`, etc), so I added a rule to ignore it. I'm open to other ideas if there's a better option that people prefer.

One option is using `__all__` here.

docs/examples/spam/pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[build-system]
+requires = [
+    "meson>=0.63.0",
+    "meson-python",
+    "patchelf",
+    "wheel",
```
FFY00, line 6: Only `meson-python` should be needed here, as it will request the necessary dependencies itself. Unless the Meson version needs extra bounds.

setup.cfg
```diff
@@ -2,6 +2,7 @@
 max-line-length = 127
 max-complexity = 10
 extend-ignore = E203
+per-file-ignores = __init__.py:F401
```
FFY00, line 5: nitpick: I would prefer a `# noqa: F401` comment instead.

tests/conftest.py
```diff
@@ -18,9 +18,8 @@
 
 
 @contextlib.contextmanager
-def cd_package(package):
+def cd_package(package_path):
```
FFY00, line 21: I would prefer to keep this fixture as-is. It seems you just need a chdir helper, like https://docs.python.org/3.11/library/contextlib.html#contextlib.chdir, so you can just extract that code and use the helper in this function.

FFY00 (changes requested):
Thanks! This is great :blush: 

I have some comments.

Additionally, can you link the example package in the docs?

peytondmurray:
Okay, I think I broke something in the tests while addressing comments just now. Let me get this figured out...

peytondmurray:
Okay, I'm re-pushing the old version - it looks like I've found something a bit weird. Previously, the `docs/examples/spam/pyproject.toml` required `python>=3.9`, but in order to satisfy the flake8 pre-commit hook, I changed this requirement to `python>=3.8`. Now the spam example won't build. I'll keep investigating...

tests/conftest.py
peytondmurray: :+1: Done!

FFY00 (approved):
Thanks! This looks great :blush:

rgommers:
Very nice to have a more complete usage example, thanks @peytondmurray!

A test is still failing:
```
 FAILED tests/docs/examples/test_spam.py::test_build_and_import - FileNotFound
```

peytondmurray:
@rgommers That test should definitely have

```
@pytest.mark.skipif(platform.system() != 'Linux', reason='Unsupported on this platform for now')
```

because it calls out to `ls` to check whether `spam.system` works, so that should solve the failing windows tests. But that's not even the reason those tests are failing - they fail when the virtual environment tries to install the wheel - they say that the virtual environment python executable can't be found. I assume this is why the only other test that makes use of the `virtual_env` fixture skips windows too?

For the `(macos, pypy-3.8)` and `(ubuntu, pypy-3.8)` tests, it looks like the `encodings` module is not found during the step where the virtual environment is being created. I'm not very familiar with pypy, is this something that you've encountered here already? It looks like the only other test that uses the `virtual_env` fixture is skipped when run on the `PyPy` platform.

rgommers:
I haven't seen `ModuleNotFoundError: No module named 'encodings'` before. Could be a PyPy, `venv` or Nox bug. I don't see what is going on in your `conftest.py` with
```
subprocess.check_call([virtual_env, '-m', 'pip', '-qqq', 'install', wheel])
```
though. This is stacking up more layers on something that already looks fragile (there's a `FIXME: windows` comment in the `virtual_env` fixture).

The `pypy-3.9` runs are passing while the `pypy-3.8` ones are failing, so I guess it's a PyPy 3.8 related issue. You could skip it for your new test, or we could just drop the PyPy 3.8 tests. Maybe @FFY00 has a more insightful suggestion.

peytondmurray:
Okay, I've marked the test to be skipped if the test platform isn't linux, or if pypy-3.8 is being used.

FFY00:
> I haven't seen `ModuleNotFoundError: No module named 'encodings'` before.

This is what happens when `sys.path` is bad. `encodings` is the first module the interpreter loads, so if `sys.path` is mangled making the module unimportable, this is what we see. Unfortunately, PyPy on Windows has a nasty bug that causes this to happen on some occasions, I believe it's https://foss.heptapod.net/pypy/pypy/-/issues/3321, whose fix landed in 3.9, so that's why it is passing there. Skipping on PyPy 3.8 is the right call.

rgommers:
The `pypy-39` CI job is also still failing on the new `spam` package test.

FFY00:
Rebased on `main` now that we merged the fix for the Sage tests. Let's see if everything passes now!

PR #137, opened 2022-09-12
BUG: fix native file override

eli-schwartz:
This attempted to override "python3", which Meson never used for find_installation. So the native file did nothing.

Override the correct name.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -614,7 +614,7 @@ class Project():
         # write the native file
         native_file_data = textwrap.dedent(f'''
             [binaries]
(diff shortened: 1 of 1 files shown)


bnavigator:
Works for me. See https://github.com/FFY00/meson-python/issues/51#issuecomment-1243143059. You should probably add a test for this.

rgommers (approved):
LGTM, thanks @eli-schwartz. And thanks @bnavigator for testing.

I think a test is useful, but not required to merge this. Let's open a follow-up issue for that. We want this fix in the new release that is going out this week.

FFY00 (approved):

PR #139, opened 2022-09-13
MAINT: skip setting rpath if it's not needed

### LINKED ISSUE #125: Only use and depend on `patchelf` if it is actually needed
Sage ran into an issue where `patchelf` is built from source and break with GCC 6.3 on Debian when trying to use `meson-python` with SciPy (see https://trac.sagemath.org/ticket/34081, Cc @mkoeppe). This is a little unfortunate, because `patchelf` is not actually needed. What it's used for here is to add a non-existing folder to the RPATH - which is not useful:

```python
        # fix file
            # add .mesonpy.libs to the RPATH of ELF files
            if self._is_elf(os.fspath(origin)):
                # copy ELF to our working directory to avoid Meson having to regenerate the file
                new_origin = self._libs_build_dir / pathlib.Path(origin).relative_to(self._project._build_dir)
                os.makedirs(new_origin.parent, exist_ok=True)
                shutil.copy2(origin, new_origin)
                origin = new_origin
                # add our in-wheel libs folder to the RPATH
                elf = mesonpy._elf.ELF(origin)
                libdir_path = f'$ORIGIN/{os.path.relpath(f".{self._project.name}.mesonpy.libs", destination.parent)}'
                if libdir_path not in elf.rpath:
                    elf.rpath = [*elf.rpath, libdir_path]
```

Of course an issue is that `meson-python` cannot currently know when setting up the build environment whether the project has standalone compiled libraries that are going to be included in `.mesonpy.libs`. This would be easy to make controllable via a boolean flag in `pyproject.toml` though - and then the vast majority of Python packages that do not have such libraries don't have to deal with the unnecessary `patchelf` dependency.

It seems to me like it's time for a `[meson-python]` table in `pyproject.toml`, with at least two entries: one for "has standalone libraries" and one for "pure" to avoid the fragile heuristics around that.

FFY00:
I also fixed `ELF.rpath`, which is needed for the tests, and added a test for setting the rpath in the first place in this PR.

Fixes #125

### DIFF
--- a/mesonpy/_elf.py
+++ b/mesonpy/_elf.py
@@ -17,7 +17,7 @@ class ELF:
         self._needed: Optional[Collection[str]] = None
 
     def _patchelf(self, *args: str) -> str:
-        return subprocess.check_output(['patchelf', *args, self._path]).decode()
+        return subprocess.check_output(['patchelf', *args, self._path], stderr=subprocess.STDOUT).decode()
 
     @property
     def rpath(self) -> Collection[str]:
--- a/mesonpy/_elf.py
+++ b/mesonpy/_elf.py
@@ -23,7 +23,7 @@ class ELF:
     def rpath(self) -> Collection[str]:
         if self._rpath is None:
             rpath = self._patchelf('--print-rpath').strip()
-            self._rpath = rpath.split(';') if rpath else []
+            self._rpath = rpath.split(':') if rpath else []
         return frozenset(self._rpath)
 
     @rpath.setter
--- a/mesonpy/_elf.py
+++ b/mesonpy/_elf.py
@@ -21,9 +21,10 @@ class ELF:
 
     @property
     def rpath(self) -> Collection[str]:
-        if not self._rpath:
-            self._rpath = self._patchelf('--print-rpath').strip().split(';')
-        return self._rpath
+        if self._rpath is None:
+            rpath = self._patchelf('--print-rpath').strip()
+            self._rpath = rpath.split(';') if rpath else []
+        return frozenset(self._rpath)
 
     @rpath.setter
     def rpath(self, value: Collection[str]) -> None:
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -247,7 +247,16 @@ def test_detect_wheel_tag_script(wheel_executable):
 
 
 @pytest.mark.skipif(platform.system() != 'Linux', reason='Unsupported on this platform for now')
-def test_unneeded_rpath(wheel_purelib_and_platlib, tmpdir):
+def test_rpath(wheel_link_against_local_lib, tmpdir):
+    artifact = wheel.wheelfile.WheelFile(wheel_link_against_local_lib)
+    artifact.extractall(tmpdir)
+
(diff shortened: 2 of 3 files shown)


tests/test_wheel.py
```diff
@@ -150,3 +154,21 @@ def test_executable_bit(wheel_executable_bit):
             assert executable_bit, f'{info.filename} should have the executable bit set!'
         else:
             assert not executable_bit, f'{info.filename} should not have the executable bit set!'
+
+
+def test_rpath(wheel_link_against_local_lib, tmpdir):
```
rgommers, line 159: These added test should should be skipped unless the platform is `linux`.

mesonpy/_elf.py
```diff
@@ -21,9 +21,10 @@ def _patchelf(self, *args: str) -> str:
 
     @property
     def rpath(self) -> Collection[str]:
-        if not self._rpath:
-            self._rpath = self._patchelf('--print-rpath').strip().split(';')
-        return self._rpath
+        if self._rpath is None:
+            rpath = self._patchelf('--print-rpath').strip()
```
rgommers, line 25: This doesn't work portably it looks like. For example [in this CI log](https://github.com/FFY00/meson-python/actions/runs/3049095667/jobs/4914826119#step:7:340):
```
 E               subprocess.CalledProcessError: Command '['patchelf', '--print-rpath', '/tmp/pytest-of-runner/pytest-0/test_uneeded_rpath0/purelib_and_platlib-1.0.0.data/platlib/plat.{EXT_SUFFIX}']' returned non-zero exit status 1.
```

rgommers (changes requested):
Thanks @FFY00. There are some issues shown by the CI results. The code change looks like an improvement, and will solve the most important part of gh-125. It looks like it won't yet solve the unnecessary dependency issue. That can be left for later if you prefer.

rgommers:
Ugh, this still depends on other fixes:
```
patchelf: getting info about '/tmp/pytest-of-runner/pytest-0/test_unneeded_r"
 b"path0/plat.cpython-39-x86_64-linux-gnu.so': No such file or directory
```
That's because the test package it's built on is still xfailed:
```python
@pytest.mark.xfail(reason='Meson bug')
def test_purelib_and_platlib(wheel_purelib_and_platlib):
    artifact = wheel.wheelfile.WheelFile(wheel_purelib_and_platlib)

    expecting = {
        f'plat{EXT_SUFFIX}',
```
This passes locally with Meson master. After the other open PR for tags is merged, it should pass with released Meson as well.

rgommers:
You just lost the commits that made this PR pass CI. I'll re-add them.

FFY00:
Sorry :sweat_smile:

rgommers:
I think that's it, but hard to be sure since I had to manually re-apply fixes. Let's first ensure CI is green, and then if desired at the end rewrite history to the desired commit granularity.

FFY00:
I have squashed the fix with the affected commit, the tests seem to be passing, are you okay with me pushing?

rgommers:
Sure, go ahead. CI is green (codecov isn't relevant, and Sage isn't currently testing the right thing). So rewrite history as you wish and then merge it I'd say.

Issue #141, opened 2022-09-14
Wheel ABI tag heuristics producing `py3-none` again on Arch Linux


rgommers:
To reproduce this, install `meson` master (currently at commit 83d18d137) and `meson-python` main (currently at commit 672f9ff) in a development environment, and then this can be reproduced with both the `purelib-and-platlib` test case in this repo and with SciPy main:
```
$ cd tests/packages/purelib-and-platlib
$ python -m build --wheel . --no-isolation --skip-dependency-check
....
[1/1] purelib_and_platlib-1.0.0.data/purelib/pure.pyy                                
Successfully built purelib_and_platlib-1.0.0-py3-none-linux_x86_64.whl
```

Meson produced this `intro-install_plan.json`:
```json
{
   "targets":{
      "/home/rgommers/code/tmp/meson-python/tests/packages/purelib-and-platlib/build/plat.cpython-310-x86_64-linux-gnu.so":{
         "destination":"{py_platlib}/plat.cpython-310-x86_64-linux-gnu.so",
         "tag":"runtime"
      }
   },
   "python":{
      "/home/rgommers/code/tmp/meson-python/tests/packages/purelib-and-platlib/pure.py":{
         "destination":"{py_purelib}/pure.py",
         "tag":"runtime"
      }
   }
}
```
That looks correct. The platlib part of the path gets lost though:
```
$ cd dist
$ unzip purelib_and_platlib-1.0.0-py3-none-linux_x86_64.whl
$ tree .
.
├── plat.cpython-310-x86_64-linux-gnu.so
├── purelib_and_platlib-1.0.0.data
│   └── purelib
│       └── pure.py
├── purelib_and_platlib-1.0.0.dist-info
│   ├── METADATA
│   ├── RECORD
│   └── WHEEL
└── purelib_and_platlib-1.0.0-py3-none-linux_x86_64.whl
```

The offending change is gh-130.

Running the test suite yields a failure too, which looks potentially relevant but is failing also with commit 58b0034 (the last one before gh-130):

<details>

```
___________________________________________ test_configure_data ____________________________________________

wheel_configure_data = PosixPath('/tmp/pytest-of-rgommers/pytest-11/test0/mesonpy-test-i_23__v1/configure_data-1.0.0-py3-none-any.whl')

    def test_configure_data(wheel_configure_data):
        artifact = wheel.wheelfile.WheelFile(wheel_configure_data)
    
>       assert wheel_contents(artifact) == {
[...]

PR #143, opened 2022-09-14
BUG: fix issues with wheel tags, and add more tests to cover tags

FFY00:
Closes gh-141

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -386,12 +386,17 @@ class _WheelBuilder():
                 return f.read(4) in (
                     b'\xfe\xed\xfa\xce',  # 32-bit
                     b'\xfe\xed\xfa\xcf',  # 64-bit
+                    b'\xcf\xfa\xed\xfe',  # arm64
                     b'\xca\xfe\xba\xbe',  # universal / fat (same as java class so beware!)
                 )
             elif platform.system() == 'Windows':
                 return f.read(2) == b'MZ'
 
-        raise NotImplementedError(f'Unknown system: {platform.system()}')
+        # For unknown platforms, check for file extensions.
+        _, ext = os.path.splitext(file)
+        if ext in ('.so', '.a', '.out', '.exe', '.dll', '.dylib', '.pyd'):
+            return True
+        return False
 
     def _warn_unsure_platlib(self, origin: pathlib.Path, destination: pathlib.Path) -> None:
         """Warn if we are unsure if the file should be mapped to purelib or platlib.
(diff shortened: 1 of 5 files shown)


mesonpy/__init__.py
```diff
@@ -364,10 +375,23 @@ def _select_abi_tag(self) -> Optional[mesonpy._tags.Tag]:  # noqa: C901
...
+            elif platform.system() == 'Darwin':
+                return f.read(4) in (
+                    b'\xfe\xed\xfa\xce',  # 32-bit
+                    b'\xfe\xed\xfa\xcf',  # 64-bit
+                    b'\xca\xfe\xba\xbe',  # universal / fat (same as java class so beware!)
+                )
+            elif platform.system() == 'Windows':
+                return f.read(2) == b'MZ'
+
+        raise NotImplementedError(f'Unknown system: {platform.system()}')
```
rgommers, line 394: You can't do this, that is guaranteed to break things on other platforms. I suggest emitting a warning and returning `True`, because that's the safest choice.

rgommers (changes requested):
Thanks @FFY00. This looks like an improvement, it seems to fix things in the test case I just submitted in combination with Meson master.

One change is needed, to not error out on unknown platforms.

rgommers:
Also, CI seems to be unhappy.

rgommers:
More Mypy buggy-ness:
```
mesonpy/__init__.py:839: error: Returning Any from function declared to return "bool"  [no-any-return]
``
This works fine locally on both Linux and macOS, and the code is correct - just a little complicated with property decorators:(

rgommers:
> More Mypy buggy-ness:

It depends on the exact Mypy version _and_ the minor version of Python ... sigh. On Python 3.9 this error disappears, and another one appears. Guess we have to live with it, Mypy isn't going to mature any time soon.

rgommers:
This was all green modulo the mypy issue, the last commit should fix that.

rgommers (approved):

rgommers:
This is happy, in it goes. Thanks @FFY00!

FFY00:
Thanks for all the work on getting this PR ready for merging. For next time, I'd prefer PRs that do multiple things to be rebase-merged, instead of squashed, the rule I try to abide by is one commit per logical change. In case you are not used to this kind of workflow, I recommend using fixup commits if you need to modify commits, as explained in https://openinput.readthedocs.io/en/latest/contributing/index.html#workflow. Thanks :blush:

rgommers:
Sure, that is a good idea in most cases. I am familiar with it, will aim to rebase-merge where possible.

In cases where it would break `git bisect` it may be a bit questionable though, then it's more of a judgement call and squash-merge or a merge commit may be more helpful.

FFY00:
Thanks! All commits being self-contained is also something I strive to achieve, so in this workflow you would apply the fixes needed to the PR to each respective commit.

PR #145, opened 2022-09-15
TST: add a regular Python package structure to test

rgommers:
This follows up on gh-141.

@FFY00 what still needs adding is parametrizing the CI tests over Meson versions. We need at least the latest stable version and `master` (the check for `none` in the filename will fail with master, and anyway testing against master is needed). I'm not familiar with `nox`, so it's not clear to me if this should be done inside `noxfile.py` or in GitHub Actions.


rgommers:
CI failure I intended to demonstrate:
```
 E       AssertionError: assert 'none' not in 'mypkg-2.3.4-py3-none-linux_x86_64.whl'
E         'none' is contained here:
E           mypkg-2.3.4-py3-none-linux_x86_64.whl
E         ?                 ++++
```
Looks like gh-143 in its current state fixes that.

tests/test_wheel.py
```diff
@@ -6,15 +6,15 @@
 import subprocess
 import sys
 import sysconfig
-import importlib.metadata
 
 import pytest
 import wheel.wheelfile
 
 try:
+    import importlib.metadata  # not available in Python 3.7
```
FFY00, line 14: Why not just use `importlib_metadata` in that case?

https://github.com/python/importlib_metadata

tests/test_wheel.py
```diff
@@ -83,6 +88,8 @@ def test_scipy_like(wheel_scipy_like):
         #'mypkg/submod/unknown_filetype.npq',
     }
     assert wheel_contents(artifact) == expecting
+    assert 'none' not in wheel_filename(artifact)
```
FFY00, line 91: Wouldn't in make more sense to use `artifact.parsed_filename.group('abi')` instead?

FFY00 (approved):
Thanks! This looks great, only some minor nitpicks.

tests/test_wheel.py
rgommers: (a) because I didn't know it existed, and (b) I used `importlib.metadata` to avoid a dependency (could have used https://packaging.pypa.io/en/latest/version.html for version comparison too, but decided not to). I think it's fine as is - and I don't really care about extra Python 3.7 testing given that it's already dropped in SciPy and other similar projects), but happy to change if you prefer to add a new test dependency.

tests/test_wheel.py
rgommers: I prefer this for both readability and robustness I think. There must not be a `none` anywhere, so as a test that is independent from the implementation this seem better to me.

FFY00: Yes, but you can check the exact value of the different groups. This is just a nitpick though.

rgommers: Let's do both then.

rgommers:
The Windows tests for some reason are run with Mingw, even though that is never explicitly installed or asked for:
```
C compiler for the host machine: ccache gcc (gcc 11.2.0 "gcc.exe (MinGW-W64 x86_64-posix-seh, built by Brecht Sanders) 11.2.0")
```
That is fairly annoying, and the tests should probably use MSVC instead.

rgommers:
More fun with extension file names, on Windows they look like `extmod.cp311-win_amd64.pyd` and not `.cpython-`.

rgommers:
Okay, the tags issue on Windows is sorted out now. Next bit of fun, we're packaging `.dll.a` files next to each `.pyd` extension module:
```
 E         +  'mypkg/extmod.cp39-win_amd64.dll.a',
E            'mypkg/extmod.cp39-win_amd64.pyd',
```
This is happening in the SciPy nightlies as well, so it didn't come from anything merged this week. I'll look into it.

rgommers:
Getting there. Python 3.7 on Windows is kinda broken:
```
 found pkg-config 'C:\\Strawberry\\perl\\bin\\pkg-config.BAT' but it is Strawberry Perl and thus broken. Ignoring...

Found Pkg-config: NO
Could not find Python3 library 'D:\\a\\meson-python\\meson-python\\.nox\\test-3-7\\python37.dll'
Run-time dependency python3 found: NO (tried sysconfig)

..\..\meson.build:6:0: ERROR: Pkg-config binary for machine 1 not found. Giving up.

 =========================== short test summary info ===========================
ERROR tests/test_wheel.py::test_scipy_like - NotImplementedError: Unknown Pyt...
ERROR tests/test_wheel.py::test_purelib_and_platlib - subprocess.CalledProces...
======== 42 passed, 5 skipped, 1 xfailed, 2 errors in 73.61s (0:01:13) ========
```

Probably a Nox bug, because I don't see us explicitly installing `pkg-config`. Let's skip the test there.

rgommers:
Okay, this is finally happy. It contains:
- one bug fix for wheel tags on Windows,
- one important new test case that matches the main things SciPy and Pandas need, including use of Cython (new test dependency), `install_subdir`, installing generated files, and handling unknown file types,
- enabling the previously `platlib_and_purelib` test, which was xfailed so far, when possible (it works with Meson master),
- making `nox` output in CI more verbose, so that we can see what packages are being installed.

Two follow-ups are needed, opening upstream issues in Meson for:
1. the file extensions for Python extension modules on Windows (is missing `cpython-`),
2. not installing the `.dll.a` import libraries on Windows for Python extension modules

rgommers:
This is all green. Only the opensuse test just hangs, it's been going for 1hr 30min. Let's get this in.

PR #146, opened 2022-09-15
CI: add concurrency control to GitHub Actions workflows

rgommers:
This ensures that multiple jobs don't run for the same action. If a new commit is pushed, this cancels the jobs for the older commit. This is typically what is desired, and especially for longer-running jobs.


FFY00 (approved):
Yep, that makes sense, thanks!

PR #147, opened 2022-09-25
BUG: Fix meson-python failing on installing install_subdirs

lithomas1:
You can only write a single file, or a directory that you can't specify the destination for with wheel.

I copy pasted the code and adjusted from wheel to do it.

cc @rgommers @FFY00


rgommers (approved):
Thanks @lithomas1! I have tested that this works with the test case for `install_subdir` in gh-145.

rgommers:
Mypy and the pre-commit check seem to be a bit unhappy still.

rgommers:
This is happy now, and also works in combination with the scipy-like test in gh-148. So in it goes. Thanks @lithomas1!

Issue #115, opened 2022-07-26
Scripts go to the incorrect location


FFY00:
The following snippet installs to `.data/data/bin/example-script`, instead of `.data/scripts/example-script`.

```meson
install_data(
    'example-script.py',
    rename: 'example-script',
    install_dir: get_option('bindir'),
)
```

This happens because Meson returns the full path in this situation, but still, we should not be putting the file in `.data/data` like this.

rgommers:
gh-148 improves a test for this. Meson master uses `{bindir}`, so the file lands in `scripts/`. Meson 0.63.2 still contains absolute paths, so the file lands in `data/bin/`. So once we make Meson 0.64 the minimum supported version, this will fix itself and we can clean up the test for it.

PR #148, opened 2022-09-27
TST: fix executable bit test

rgommers:
Note that Meson does have the correct install path in `intro-install_plan.json`, containing `{bindir}`.
This gets translated into `script/` somewhere, at least for a conda-installed Python on Arch Linux. This code was added in gh-102, and it's not quite clear why this line was commented out.

For now, let's just make sure the test is not failing.


rgommers:
This all looks good except for the Mypy failure, which is fixed in gh-149. Somehow even the Gentoo job turned green.

rgommers:
This is all happy, the one failure is a weird job runner issue where the log isn't visible and it can't even be re-run. Fixes the local failures I'm seeing.

rgommers:
Ah, the issue with `bin` vs script` is discussed in gh-115. We should be good here. Let's merge this, and leave gh-115 open for the follow-up to ensure everything is 100% correct (it will be with Meson master it looks like, not yet with `0.63.2`). I will comment there, and just update the code comment here to point at that issue.

PR #152, opened 2022-09-28
.github/workflows/ci-sage.yml: Make sure git is available

mkoeppe:
... in the build containers

### DIFF
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -101,7 +101,7 @@ jobs:
            "conda-forge",
            ]
       tox_packages_factors: >-
-          ["standard",
+          ["standard-develop",
            ]
       # Sage distribution packages to build
       targets: SAGE_CHECK=no SAGE_CHECK_PACKAGES="meson_python,scipy" scipy
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ test = [
   'GitPython',
   'auditwheel',
   'Cython',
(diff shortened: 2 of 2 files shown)


FFY00 (approved):

FFY00:
Looks like our tests are failing.

mkoeppe:
Yes, see https://github.com/FFY00/meson-python/issues/150#issuecomment-1261410641

mkoeppe:
Let me know if I should change anything in the `pytest` options so that you can see enough detail.

FFY00:
I suspect the fails are due to an old `pyproject-metadata` dependency.

mkoeppe:
Updated in https://git.sagemath.org/sage.git/commit/?id=004197d2eb3fec2104e25053b2b90045c61e95d3 to 0.6.1

FFY00:
Just pushed a fix for the `pyproject-metadata` version mismatch, hopefully everything will pass now.

FFY00:
The archlinux test passed, so I think we are good :tada:

mkoeppe:
Failures on `centos-stream-8`, `gentoo-python3.9`

FFY00:
Alright, those fails are different. The gentoo one seems like it might be a genuine issue, and the centos one is due to some weirdness in the tests. I am gonna mark them as not required and try to tackle them in followup PRs.

rgommers:
Gentoo will be the same issue as we had before for building SciPy, that is gh-127.

PR #156, opened 2022-09-29
Use `mac_ver` instead of `sysconfig.get_platform` when determining the platform tag and document the MacOS platform tag handling

FFY00:

### DIFF
--- a/docs/usage/specific-behaviors.rst
+++ b/docs/usage/specific-behaviors.rst
@@ -1,0 +1,17 @@
+******************
+Specific Behaviors
+******************
+
+This page documents how our backend handles certain situations.
+
+
+``MACOSX_DEPLOYMENT_TARGET``
+============================
+
+
+The target macOS version can by changed by setting the
+``MACOSX_DEPLOYMENT_TARGET`` environment variable.
+
+If ``MACOSX_DEPLOYMENT_TARGET`` is set, we will use the selected version for the
+wheel tag, otherwise ``platform.mac_ver()`` (i.e., the current macOS version on
+the build machine) will be used instead.
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -269,6 +269,12 @@ class _WheelBuilder():
                     'platform tag to {target}{reset}'.format(target=target, **_STYLES)
                 )
                 parts[1] = target
+            else:
+                # If no target macOS version is specified fallback to
+                # platform.mac_ver() instead of sysconfig.get_platform() as the
+                # latter specifies the target macOS version Python was built
+                # against.
+                parts[1] = platform.mac_ver()[0]
 
             if parts[1] in ('11', '12'):
                 # Workaround for bug where pypa/packaging does not consider macOS
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -148,6 +148,7 @@ If you really need the data to be installed where it was previously (eg.
 
    usage/start
    usage/build-options
+   usage/specific-behaviors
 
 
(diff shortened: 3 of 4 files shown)


rgommers (commented):
Now that I see this, I am wondering if we should first fix it further before documenting it. The correct thing to use seems to be:
```python
>>> import platform
>>> platform.mac_ver()
('12.4', ('', '', ''), 'arm64')
```

Or maybe we can note in this PR that the current behavior is subject to change, and if users want to be sure of the correct behavior, they should explicitly set `MACOSX_DEPLOYMENT_TARGET`? That is helpful either way I'd say.

FFY00:
I pushed a commit changing to `mac_ver` and adjusted the documentation.

rgommers:
The change looks correct, however the tests are unhappy. I'll test locally.

docs/usage/specific-behaviors.rst
```diff
@@ -0,0 +1,16 @@
...
+
+``MACOSX_DEPLOYMENT_TARGET``
+============================
+
+
+The target MacOS version can by changed by setting the
+``MACOSX_DEPLOYMENT_TARGET`` environment variable.
+
+If ``MACOSX_DEPLOYMENT_TARGET`` is set, we will use the selected version for the
+wheel tag, otherwise ``platform.mac_ver()`` will be used instead.
```
rgommers, line 16: How about this clarification:
```
wheel tag, otherwise ``platform.mac_ver()`` (i.e., the current macOS version on
the build machine) will be used instead.
```

docs/usage/specific-behaviors.rst
```diff
@@ -0,0 +1,16 @@
...
+******************
+
+This page documents how our backend handles certain situations.
+
+
+``MACOSX_DEPLOYMENT_TARGET``
+============================
+
+
+The target MacOS version can by changed by setting the
```
rgommers, line 12: minor: spelling should be `macOS` (no capital M)

rgommers: and same in the code comment in `__init__.py`

mesonpy/__init__.py
```diff
@@ -269,6 +269,12 @@ def platform_tag(self) -> str:
                     'platform tag to {target}{reset}'.format(target=target, **_STYLES)
                 )
                 parts[1] = target
+            else:
+                # If no target MacOS version is specified fallback to
+                # platform.mac_ver() instead of sysconfig.get_platform() as this
+                # one specifies the target MacOS version Python was built
```
rgommers, line 275: I suggest changing "this one" to "the latter" to be a little clearer

rgommers (commented):
Tested by installing locally and running `python -m build --wheel --no-isolation --skip-dependency-check`:

- Old wheel filename (`main`): `scipy-1.10.0.dev0-cp39-cp39-macosx_11_0_arm64.whl`
- New wheel filename (this PR): `scipy-1.10.0.dev0-cp39-cp39-macosx_12_4_arm64.whl`

So the bug fix LGTM. Just a few comments on wording.

rgommers:
The test failures are all from
```python
    assert name.group('plat') == sysconfig.get_platform().replace('-', '_').replace('.', '_')
```
so that's kind of a self-consistency check. Given that that check is supposed to still be true for other platforms, it looks like it'd be useful to create that as a single test function.

Testing related: adding `x86_64` and `arm64` CPython tags to `test_interpreter_tag` may be good too.

rgommers (approved):
LGTM now!

PR #161, opened 2022-09-30
BUG: ignore macOS minor version

### LINKED ISSUE #160: 0.9.0 release produces macOS wheel platform tag that cannot be installed on Homebrew Python
@mkoeppe noticed a problem with macOS wheel tags with Homebrew Python on Sage, see https://github.com/FFY00/meson-python/issues/91#issuecomment-1263004679. The linked Sage issue (https://trac.sagemath.org/ticket/34081#comment:90) shows:
```
[scipy-1.9.1] Using pip 22.2.2 from /Users/mkoeppe/s/sage/sage-rebasing/worktree-gcc11/local/var/lib/sage/venv-python3.10/lib/python3.10/site-packages/pip (python 3.10)
[scipy-1.9.1] Looking in links: /Users/mkoeppe/s/sage/sage-rebasing/worktree-gcc11/local/var/lib/sage/venv-python3.10/var/lib/sage/wheels
[scipy-1.9.1] ERROR: scipy-1.9.1-cp310-cp310-macosx_12_6_x86_64.whl is not a supported wheel on this platform.
```

@mkoeppe can you confirm that you built this wheel on macOS 12.6? If so, the wheel tag should be correct. However, there is a problem of course if `pip` refuses to install it, which we may have to work around. 

Also, can you determine when it is installable? If you rename the wheel from `12_6` to say `12_3`, `12_0`, `11_0`, it should become installable.

FFY00:
Fixes #160

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -275,6 +275,9 @@ class _WheelBuilder():
                 # latter specifies the target macOS version Python was built
                 # against.
                 parts[1] = platform.mac_ver()[0]
+                # Only pick up the major version
+                # https://github.com/FFY00/meson-python/issues/160
+                parts[1] = parts[1].split('.')[0]
 
             if parts[1] in ('11', '12'):
                 # Workaround for bug where pypa/packaging does not consider macOS
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -51,7 +51,9 @@ else:
 platform_ = sysconfig.get_platform()
 if platform.system() == 'Darwin':
     parts = platform_.split('-')
-    parts[1] = platform.mac_ver()[0]
+    parts[1] = platform.mac_ver()[0].split('.')[0]
+    if parts[1] in ('11', '12'):
+        parts[1] += '.0'
     platform_ = '-'.join(parts)
 PLATFORM_TAG = platform_.replace('-', '_').replace('.', '_')
 


mesonpy/__init__.py
```diff
@@ -275,6 +275,9 @@ def platform_tag(self) -> str:
                 # latter specifies the target macOS version Python was built
                 # against.
                 parts[1] = platform.mac_ver()[0]
+                # Only pick up the major version
+                # https://github.com/FFY00/meson-python/issues/160
+                parts[1] = parts[1].split('.')[0]
```
rgommers, line 280: And append `'.0'` then, right? We want `11.0`, `12.0`, not `11`, `12`

FFY00: We are already doing that below.

tests/test_wheel.py
```diff
@@ -51,7 +51,9 @@
 platform_ = sysconfig.get_platform()
 if platform.system() == 'Darwin':
     parts = platform_.split('-')
-    parts[1] = platform.mac_ver()[0]
+    parts[1] = platform.mac_ver()[0].split('.')[0]
+    if parts[1] in ('11', '12'):
```
rgommers, line 55: Make it `> '11'`, to avoid this breaking as soon as macOS 13 comes out?

rgommers (changes requested):
I now tested this on macOS arm64, and it works correctly there. However, the logic is wrong for macOS 10.x - there we _do_ want to preserve the major version, it should result in (e.g.) `10_13` and not `10_0`.

So it should be something like:
```python
parts[1] = platform.mac_ver()[0]
if parts[1] > '11':
    parts[1] = parts[1].split('.')[0]
```

I can't test this, as I don't have a 10.x machine available. But from reading the code, this case isn't handled.

tests/test_wheel.py
FFY00: The idea is that `packaging` would have this fixed by then.

rgommers: Even with that fix, it still needs to be `13.0` for any 13.x OS version in the absence of `MACOSX_DEPLOYMENT_PLATFORM`. The `packaging` fix is only for when `MACOSX_DEPLOYMENT_PLATFORM` is explicitly set.

PR #163, opened 2022-10-03
BUG: also handle macOS 10.X versions correctly

rgommers:
This is a follow-up to gh-161. From my review comment there:

However, the logic is wrong for macOS 10.x - there we _do_ want to preserve the major version, it should result in (e.g.) `10_13` and not `10_0`.


FFY00 (approved):
Thanks!

PR #165, opened 2022-10-06
TST: update the test for `install_subdir` for fix in meson 0.63.3

rgommers:
We weren't sure if the fix we needed was getting backported, but it has been.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,7 @@ test = [
   'auditwheel',
   'Cython',
   'pyproject-metadata>=0.6.1',
+  'importlib_metadata; python_version<"3.8"'
 ]
 docs = [
   'furo>=2021.08.31',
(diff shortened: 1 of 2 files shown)


FFY00 (approved):

rgommers:
Python 3.7 `importlib` thing messing with the results again - will push a fix later.

FFY00:
I think it would be easier to just add the `importlib.metadata` backport as a test dependency on Python 3.7.

rgommers:
@FFY00 `codecov` failure is spurious, as it pretty much always is. This is pretty annoying. The default codecov targets are known to be problematic. Do you mind if I open a PR to set the targets to 1%/0% so we stop it from failing? We can keep the integration, because the links in CI to the full coverage html report is useful.

eli-schwartz:
After being repeatedly bothered by this in Meson, I eventually set it to `informational: true` which is supposed to never fail for any reason.

rgommers:
> After being repeatedly bothered by this in Meson, I eventually set it to `informational: true` which is supposed to never fail for any reason.

Ah nice, that is a newer setting, didn't know about that yet - and better than setting target to 0%.

PR #167, opened 2022-10-07
ENH: allow users to pass options directly to Meson

FFY00:

EDIT: closes gh-54

### DIFF
--- a/tests/packages/user-args/pyproject.toml
+++ b/tests/packages/user-args/pyproject.toml
@@ -1,0 +1,9 @@
+[build-system]
+build-backend = 'mesonpy'
+requires = ['meson-python']
+
+[tool.mesonpy.args]
+dist = ['config-dist']
+setup = ['config-setup']
+compile = ['config-compile']
+install = ['config-install']
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -32,8 +32,8 @@ import typing
 import warnings
 
 from typing import (
-    Any, ClassVar, DefaultDict, Dict, List, Optional, Set, TextIO, Tuple, Type,
-    Union
+    Any, ClassVar, DefaultDict, Dict, List, Optional, Sequence, Set, TextIO,
+    Tuple, Type, Union
 )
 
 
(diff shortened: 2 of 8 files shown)


tests/packages/user-args/pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[build-system]
+build-backend = 'mesonpy'
+requires = ['meson-python']
+
+[tool.mesonpy.args]
```
dnicolodi, line 5: Does this need to be `mesonpy`? I think we are settling to `meson-python` as project name and it is nice to try to be consistent.

doronbehar:
@FFY00 would it be possible to give an example of how to use `--config-settings KEY=VALUE` with this PR? I wish to test whether it's possible to compile scipy with `--cross-file` with this patch..

FFY00:
It should be `python -m build -C--cross-file=yourfile`.

doronbehar:
> It should be `python -m build -C--cross-file=yourfile`.

Is `-C` equivalent to `--config-settings`? I don't understand what is this `build` module you refer to with `python -m`..

eli-schwartz:
> Is `-C` equivalent to `--config-settings`? I don't understand what is this `build` module you refer to with `python -m`..

https://github.com/pypa/build

rgommers:
Great to see this land!

One comment left unaddressed, and let's do a few doc updates in a follow-up PR. @FFY00 are you planning to do that? If not I can do it - already have a need to point people to it:)

tests/packages/user-args/pyproject.toml
FFY00: Whops, I missed this comment :persevere:

rgommers:
> It should be `python -m build -C--cross-file=yourfile`.

That seems wrong, I think it should be `python -m build -Csetup_args="--cross-file=yourfile"`. If we want `-C--cross-file` to work, it looks like it should be special-cased like `builddir`.

This landed with underscores `*_args`, which is a little unusual for CLI arguments, so wanted to check that that was intended?

It's nontrivial to figure out the right invocations here, so we need quite a few examples in the docs. Here is another one:
```
pip install . --config-settings compile_args="-j1"
```

Another follow-up, unrecognized inputs should raise or give a warning, rather than be silently ignored. This will happily continue a build:
```
pip install . --no-build-isolation --config-settings nonsense=true
```

rgommers:
> This landed with underscores `*_args`, which is a little unusual for CLI arguments, so wanted to check that that was intended?

This was done in gh-218.

> Another follow-up, unrecognized inputs should raise or give a warning, rather than be silently ignored.

This was done in gh-219.

I'll open a new issue for the rest.

Issue #158, opened 2022-09-29
Only use and depend on patchelf if it is actually needed


rgommers:
This is a follow-up to gh-125. The fix for that issue ensured that runpath entries that were not needed are no longer inserted. The original issue description of gh-125 was also about the dependency on `patchelf` on Linux. This is still injected on Linux when `patchelf` is not already installed, in `get_requires_for_build_wheel`. Ideally this would not be the case, because `patchelf` is not needed unless an rpath/runpath entry needs to be made. That is typically not the case. And it can lead to problems - in gh-125 a `patchelf` build from source was triggered and it failed.

At the moment we don't have any real-world users (AFAIK) that do require `patchelf`. So a useful follow-up would be to not require `patchelf` unless it is actually needed. How to achieve this is TBD, there are various options - from a dedicated flag for it in `pyproject.toml`, to simply documenting how it works and that end users need to take care of this themselves, to perhaps smarter solutions.

This isn't urgent, but it would be a nice improvement.

FFY00:
Fixed by #64

PR #175, opened 2022-10-15
ENH: use system ninja if adequate

henryiii:
This uses the same solution as patchelf - if the system has it (and it's a new enough version, ~~1.7+ according to the release notes~~, 1.8.2 according to the meson error if missing), then use that. This fixes the build on systems that have a new enough ninja present but are not supportable by binary wheels.

See https://github.com/scikit-build/ninja-python-distributions/issues/127 for discussion.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -1019,7 +1019,15 @@ def get_requires_for_build_wheel(
         # we may need patchelf
         if not shutil.which('patchelf'):
             # patchelf not already accessible on the system
-            dependencies.append(_depstr.patchelf)
+            if _env_ninja_command() is not None:
+                # we have ninja available, so we can run Meson and check if the project needs patchelf
+                with _project(config_settings) as project:
+                    if not project.is_pure:
+                        dependencies.append(_depstr.patchelf)
+            else:
+                # we can't check if the project needs patchelf, so always add it
+                # XXX: wait for https://github.com/mesonbuild/meson/pull/10779
+                dependencies.append(_depstr.patchelf)
 
     return dependencies
 
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -64,14 +64,6 @@ else:
 __version__ = '0.11.0.dev0'
 
 
-class _depstr:
-    """Namespace that holds the requirement strings for dependencies we *might*
-    need at runtime. Having them in one place makes it easier to update.
-    """
-    patchelf = 'patchelf >= 0.11.0'
-    wheel = 'wheel >= 0.36.0'  # noqa: F811
-
-
 _COLORS = {
     'cyan': '\33[36m',
     'yellow': '\33[93m',
@@ -82,6 +74,16 @@ _COLORS = {
     'reset': '\33[0m',
 }
 _NO_COLORS = {color: '' for color in _COLORS}
+_NINJA_REQUIRED_VERSION = '1.8.2'
+
+
+class _depstr:
+    """Namespace that holds the requirement strings for dependencies we *might*
+    need at runtime. Having them in one place makes it easier to update.
+    """
+    patchelf = 'patchelf >= 0.11.0'
+    wheel = 'wheel >= 0.36.0'  # noqa: F811
+    ninja = f'ninja >= {_NINJA_REQUIRED_VERSION}'
 
 
 def _init_colors() -> Dict[str, str]:
@@ -565,6 +567,12 @@ class Project():
         self._build_dir = pathlib.Path(build_dir).absolute() if build_dir else (self._working_dir / 'build')
         self._install_dir = self._working_dir / 'install'
         self._meson_native_file = self._source_dir / '.mesonpy-native-file.ini'
+        self._env = os.environ.copy()
+
+        # prepare environment
+        ninja_path = _env_ninja_command()
+        if ninja_path is not None:
+            self._env.setdefault('NINJA', str(ninja_path))
 
         # load config -- PEP 621 support is optional
         self._config = tomllib.loads(self._source_dir.joinpath('pyproject.toml').read_text())
@@ -637,7 +645,7 @@ class Project():
     def _proc(self, *args: str) -> None:
         """Invoke a subprocess."""
         print('{cyan}{bold}+ {}{reset}'.format(' '.join(args), **_STYLES))
-        subprocess.check_call(list(args))
+        subprocess.check_call(list(args), env=self._env)
 
     def _meson(self, *args: str) -> None:
         """Invoke Meson."""
@@ -957,6 +965,37 @@ def _project(config_settings: Optional[Dict[Any, Any]]) -> Iterator[Project]:
         yield project
 
 
+def _env_ninja_command(*, version: str = _NINJA_REQUIRED_VERSION) -> Optional[pathlib.Path]:
+    """
+    Returns the path to ninja, or None if no ninja found.
+    """
+    required_version = tuple(int(v) for v in version.split('.'))
+    env_ninja = os.environ.get('NINJA', None)
+    ninja_candidates = [env_ninja] if env_ninja else ['ninja', 'ninja-build', 'samu']
+    for ninja in ninja_candidates:
+        ninja_path = shutil.which(ninja)
+        if ninja_path is None:
+            continue
+
+        result = subprocess.run([ninja_path, '--version'], check=False, text=True, capture_output=True)
+
+        try:
+            candidate_version = tuple(int(x) for x in result.stdout.split('.')[:3])
+        except ValueError:
+            continue
+        if candidate_version < required_version:
+            continue
+        return pathlib.Path(ninja_path)
+
+    return None
+
+
+def get_requires_for_build_sdist(
+    config_settings: Optional[Dict[str, str]] = None,
+) -> List[str]:
+    return [_depstr.ninja] if _env_ninja_command() is None else []
+
+
 def build_sdist(
     sdist_directory: str,
     config_settings: Optional[Dict[Any, Any]] = None,
@@ -972,12 +1011,16 @@ def get_requires_for_build_wheel(
     config_settings: Optional[Dict[str, str]] = None,
 ) -> List[str]:
     dependencies = [_depstr.wheel]
-    with _project(config_settings) as project:
-        if not project.is_pure and platform.system() == 'Linux':
-            # we may need patchelf
-            if not shutil.which('patchelf'):  # XXX: This is slightly dangerous.
-                # patchelf not already acessible on the system
-                dependencies.append(_depstr.patchelf)
+
+    if _env_ninja_command() is None:
+        dependencies.append(_depstr.ninja)
+
+    if sys.platform.startswith('linux'):
+        # we may need patchelf
+        if not shutil.which('patchelf'):
+            # patchelf not already accessible on the system
+            dependencies.append(_depstr.patchelf)
+
     return dependencies
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,6 @@ build-backend = 'mesonpy'
 backend-path = ['.']
 requires = [
   'meson>=0.63.3',
-  'ninja',
   'pyproject-metadata>=0.5.0',
   'tomli>=1.0.0; python_version<"3.11"',
   'typing-extensions>=3.7.4; python_version<"3.8"',
@@ -27,7 +26,6 @@ classifiers = [
 dependencies = [
   'colorama; os_name == "nt"',
   'meson>=0.63.3',
-  'ninja',
   'pyproject-metadata>=0.5.0', # not a hard dependency, only needed for projects that use PEP 621 metadata
   'tomli>=1.0.0; python_version<"3.11"',
   'typing-extensions>=3.7.4; python_version<"3.8"',
@@ -43,10 +41,13 @@ test = [
   'pytest>=6.0',
   'pytest-cov[toml]',
   'pytest-mock',
+  'GitPython',
   'auditwheel',
   'Cython',
   'pyproject-metadata>=0.6.1',
   'wheel',
+  'importlib_metadata; python_version<"3.8"',
+  'ninja',
 ]
 docs = [
   'furo>=2021.08.31',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,12 +41,10 @@ test = [
   'pytest>=6.0',
   'pytest-cov[toml]',
   'pytest-mock',
-  'GitPython',
   'auditwheel',
   'Cython',
   'pyproject-metadata>=0.6.1',
   'wheel',
-  'importlib_metadata; python_version<"3.8"',
   'ninja',
 ]
 docs = [
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length = 127
-max-complexity = 10
+max-complexity = 12
 extend-ignore = E203
--- a/tests/test_pep517.py
+++ b/tests/test_pep517.py
@@ -37,10 +37,17 @@ def test_get_requires_for_build_wheel(monkeypatch, package, system_patchelf, nin
     monkeypatch.setattr(subprocess, 'run', run)
 
     expected = {mesonpy._depstr.wheel}
-    if system_patchelf is None and sys.platform.startswith('linux'):
-        expected |= {mesonpy._depstr.patchelf}
-    if ninja is None or [int(x) for x in ninja.split('.')] < [1, 8, 2]:
+
+    ninja_available = ninja is not None and [int(x) for x in ninja.split('.')] >= [1, 8, 2]
+
+    if not ninja_available:
         expected |= {mesonpy._depstr.ninja}
+
+    if (
+        system_patchelf is None and sys.platform.startswith('linux')
+        and (not ninja_available or (ninja_available and package != 'pure'))
+    ):
+        expected |= {mesonpy._depstr.patchelf}
 
     with cd_package(package):
         assert set(mesonpy.get_requires_for_build_wheel()) == expected
--- a/tests/test_pep517.py
+++ b/tests/test_pep517.py
@@ -1,6 +1,10 @@
 # SPDX-License-Identifier: MIT
 
-import platform
+import shutil
+import subprocess
+import sys
+
+from typing import List
 
 import pytest
 
@@ -9,28 +13,34 @@ import mesonpy
 from .conftest import cd_package
 
 
-if platform.system() == 'Linux':
-    VENDORING_DEPS = {mesonpy._depstr.patchelf}
-else:
-    VENDORING_DEPS = set()
+@pytest.mark.parametrize('package', ['pure', 'library'])
+@pytest.mark.parametrize('system_patchelf', ['patchelf', None], ids=['patchelf', 'nopatchelf'])
+@pytest.mark.parametrize('ninja', [None, '1.8.1', '1.8.3'], ids=['noninja', 'oldninja', 'newninja'])
+def test_get_requires_for_build_wheel(monkeypatch, package, system_patchelf, ninja):
+    def which(prog: str) -> bool:
+        if prog == 'patchelf':
+            return system_patchelf
+        if prog == 'ninja':
+            return ninja and 'ninja'
+        if prog in ('ninja-build', 'samu'):
+            return None
+        # smoke check for the future if we add another usage
+        raise AssertionError(f'Called with {prog}, tests not expecting that usage')
 
+    def run(cmd: List[str], *args: object, **kwargs: object) -> subprocess.CompletedProcess:
+        if cmd != ['ninja', '--version']:
+            # smoke check for the future if we add another usage
+            raise AssertionError(f'Called with {cmd}, tests not expecting that usage')
+        return subprocess.CompletedProcess(cmd, 0, f'{ninja}\n', '')
 
-@pytest.mark.parametrize(
-    ('package', 'system_patchelf', 'expected'),
-    [
-        ('pure', True, set()),  # pure and system patchelf
-        ('library', True, set()),  # not pure and system patchelf
-        ('pure', False, set()),  # pure and no system patchelf
-        ('library', False, VENDORING_DEPS),  # not pure and no system patchelf
-    ]
-)
-def test_get_requires_for_build_wheel(mocker, package, expected, system_patchelf):
-    mock = mocker.patch('shutil.which', return_value=system_patchelf)
+    monkeypatch.setattr(shutil, 'which', which)
+    monkeypatch.setattr(subprocess, 'run', run)
 
-    if mock.called:  # sanity check for the future if we add another usage
-        mock.assert_called_once_with('patchelf')
+    expected = {mesonpy._depstr.wheel}
+    if system_patchelf is None and sys.platform.startswith('linux'):
+        expected |= {mesonpy._depstr.patchelf}
+    if ninja is None or [int(x) for x in ninja.split('.')] < [1, 8, 2]:
+        expected |= {mesonpy._depstr.ninja}
 
     with cd_package(package):
-        assert set(mesonpy.get_requires_for_build_wheel()) == expected | {
-            mesonpy._depstr.wheel,
-        }
+        assert set(mesonpy.get_requires_for_build_wheel()) == expected


rgommers:
Thanks @henryiii! I agree that checking if `ninja` is already available is the right thing to do. 

The tests seem to be unhappy still:
```
ERROR: Could not detect Ninja v1.8.2 or newer
```

henryiii:
Ahh, I should uninstall Ninja before running the tests locally. I also might’ve forgotten to run the test after my last change. I’m hoping it’s just that some unit tests are not activating the isolated environment. but I’ll check.

I’ll be traveling for the next couple of days so I might not be able to finish this until Monday.

mesonpy/__init__.py
```diff
@@ -951,6 +952,27 @@ def _project(config_settings: Optional[Dict[Any, Any]]) -> Iterator[Project]:
         yield project
 
 
+def _get_ninja_if_needed() -> List[str]:
+    # only add ninja if it is not present or is too low a version
+    ninja = shutil.which('ninja')
+    if ninja is None:
+        return []
+
+    result = subprocess.run(['ninja', '--version'], check=False, text=True, capture_output=True)
```
eli-schwartz, line 961: This should use the returned value of shutil.which above.

mesonpy/__init__.py
```diff
@@ -951,6 +952,27 @@ def _project(config_settings: Optional[Dict[Any, Any]]) -> Iterator[Project]:
         yield project
 
 
+def _get_ninja_if_needed() -> List[str]:
+    # only add ninja if it is not present or is too low a version
+    ninja = shutil.which('ninja')
```
eli-schwartz, line 957: Note that finding ninja can be slightly complicated. Meson's internal implementation looks like this:

```python
def detect_ninja_command_and_version(version: str = '1.8.2', log: bool = False) -> T.Tuple[T.List[str], str]:
    env_ninja = os.environ.get('NINJA', None)
    for n in [env_ninja] if env_ninja else ['ninja', 'ninja-build', 'samu']:
        prog = ExternalProgram(n, silent=True)
        if not prog.found():
            continue
        try:
            p, found = Popen_safe(prog.command + ['--version'])[0:2]
        except (FileNotFoundError, PermissionError):
            # Doesn't exist in PATH or isn't executable
            continue
[...]
```

Note that on some distros, the command is arbitrarily renamed to "ninja-build" because "ninja" was once owned by an IRC client software last developed in 2002.

And samurai can work as well, under the name `samu`.

Meson will detect either of these cases as fallbacks if "ninja" itself does not exist, or returns a version from an IRC client that is less than 1.8.2.

eli-schwartz (changes requested):

mesonpy/__init__.py
henryiii: It doesn't matter if we are only looking for one command. But when we look for three, yes, this would be better.

mesonpy/__init__.py
henryiii: > returns a version from an IRC client that is less than 1.8.2.

Ahh, okay.

henryiii:
It looks like it hits the non-standard:
```
Using Meson to generate the project metadata (no `project` section in pyproject.toml)
```

Then it breaks because it's trying to run Meson outside building sdists/wheels. Is it possible to generate the metadata without ninja being present? Or could this step be delayed until inside the isolated environment via the hooks (even `prepare_metadata_for_build_wheel` runs inside the isolated environment).

mesonpy/__init__.py
```diff
@@ -953,18 +953,25 @@ def _project(config_settings: Optional[Dict[Any, Any]]) -> Iterator[Project]:
 
 
 def _get_ninja_if_needed() -> List[str]:
-    # only add ninja if it is not present or is too low a version
-    ninja = shutil.which('ninja')
-    if ninja is None:
-        return []
+    env_ninja = os.environ.get('NINJA', None)
+    ninja_canidates = [env_ninja] if env_ninja else ['ninja', 'ninja-build', 'samu']
+    for ninja in ninja_canidates:
```
eli-schwartz, line 958: Spelling nit: missing the first "d" in "candidates" (the "candidate_version" variable uses correct spelling below).

mesonpy/__init__.py
eli-schwartz: Probably because "ninja" is the most likely to be found.

henryiii: Yeah, I realized that due to the version check, this order is safe.

mesonpy/__init__.py
henryiii: Thanks! I'll fix this but need to know the best direction to go for the generating project metadata step which seems to happen outside the build hooks. Or does it? It seems rather unlikely it would be doing project generation in the get requires hooks, which should be the only unprotected hooks. Maybe it's fine to add these to the testing requirements, as long as it's also testing inside PEP 517 isolated builds and passing. Is there a dependent project I could quickly try with this branch that's smaller than scipy?

eli-schwartz: I recently offered a bit of review to a project that might be more convenient to test: https://github.com/dfm/python-fsps

eli-schwartz: > for the generating project metadata step which seems to happen outside the build hooks. Or does it? It seems rather unlikely it would be doing project generation in the get requires hooks, which should be the only unprotected hooks.

I'm not entirely sure what all this is doing or *why*, but immediately under the changes you make, it's running a context manager that sets up a build directory, and then checks if "is_pure", and if not pure it goes and adds patchelf.

henryiii: Prefect, I also think I see what I need to do in the tests. It'll take a bit and I'm teaching today, so tonight or tomorrow I'll fix it, add pep 517 tests, and verify that the linked project can build with the change.

tests/conftest.py
```diff
@@ -103,3 +105,59 @@ def fixture(tmp_dir_session):
...
+            'tomli',
+            'wheel',
+            'colorama',
+            'meson',
+            'ninja',
+            'ninja',
+            'patchelf',
+            'pyproject-metadata',
+            'tomli',
+            'typing-extensions',
```
eli-schwartz, line 152: tomli and ninja are repeated twice, dunno if that is intentional.

mesonpy/__init__.py
henryiii: If someone could retriever the tests, I've added ninja as a testing dependency and added a "PEP 518 test" (which is really a bit more of an integration test for PEP 517 isolation, but pulls in the PEP 518 requirement table too and PEP 518 is short and memorable - I can change the name though if you'd prefer.

tests/conftest.py
henryiii: Probably not.

henryiii: Fixed & sorted.

henryiii:
Ahh, this is "it's running a context manager that sets up a build directory". Okay, so it's running Meson to see if it it is pure or not. Is there a reason to use Meson if it's pure? I think it probably should add patchelf regardless - it won't use it in the pure case, and it's would only be a downside if not present on the system _and_ for a pure Python build. This is faster too.

henryiii:
This works now for me locally after uninstalling ninja:

```
Successfully built fsps-0.4.3.dev1+gd689181.d20221020.tar.gz and fsps-0.4.3.dev1+gd689181.d20221020-cp310-cp310-macosx_12_0_x86_64.whl
```

henryiii:
I think I fixed the Windows and Linux tests, macOS was passing - you can tell what I'm developing on. :)

henryiii:
This is working except on pypy & Windows, and those are failing due to the `virtual_env` fixture not working on those platforms, so it can't do the isolated test. I'm using pytest-virtualenv elsewhere, which works on both platforms, so I've pushed a commit that uses that instead. (I wonder if this is related to the skips on the other tests using `virtual_env`)?

henryiii:
Could you try triggering the CI again? I think I fixed the platform incompatibilities by changing virtualenv implementations.

eli-schwartz:
> NotImplementedError: Unknown Python implementation: mesonpy. Please report this to https://github.com/FFY00/mesonpy/issues and include information about the Python distribution you are using.

What a fascinating error. Why is this even able to happen?

pyproject.toml
```diff
@@ -3,7 +3,7 @@ build-backend = 'mesonpy'
 backend-path = ['.']
 requires = [
   'meson>=0.63.3',
-  'ninja',
+  'ninja>=1.8.2',
```
FFY00, line 6: I think we should remove this line, no?

henryiii: Ahh, yes.

henryiii:
Looks like Win Python 3.7 might be broken by this, though I'm not sure how. Will try to look into it in the next day or two. Something is going really wrong with the ABI tags, seems to be related to the stable ABI. Seems mesonpy is reporting itself as the Python implementation on Windows 3.7 only. 🙄

FFY00:
Looks like some of the tag parsing code is broken for windows tags. Windows does not have standardized tags so this all is just looking at what happens in practice and writing code around it 😑. Interesting how this only triggers on Python 3.7.

FFY00:
> What a fascinating error. Why is this even able to happen?

It parsing code for a non-standardized format, so it's a bit fragile.

dnicolodi:
The parsing code that generates the funny error is removed in #191. I'll be happy to provide another review for this PR. Can you please squash the commits into meaningful patches?

henryiii:
>  squash the commits into meaningful patches

Should I wait till after #191?

henryiii:
Ahh, so this can happen without my changes: https://github.com/mesonbuild/meson-python/issues/189 - okay, can clean up.

henryiii:
Everything that was not related was PR'd separately already, so it's just one commit, the ninja change + related tests.

tests/test_pep518.py
```diff
@@ -0,0 +1,21 @@
...
+
+from .conftest import cd_package
+
+
+@pytest.mark.parametrize(
+    ('package'),
+    [
+        'scipy-like',
+    ]
+)
```
dnicolodi, line 11: I'm not so familiar with how pytest parametrized tests work, but what is the point of having a test parameter that is always the same? I'm also not sure I grok what this test is doing. It installs the `build` package in a virtualenv, but `build` has been added to the test dependencies. Why can't we just use `subprocess.run()`? Also, I don't see the need to test the `build` command line in addition to the PEP517 hook functions. Can you elaborate?

mesonpy/__init__.py
```diff
@@ -995,6 +996,34 @@ def _project(config_settings: Optional[Dict[Any, Any]]) -> Iterator[Project]:
...
+        ninja_path = shutil.which(ninja)
+        if ninja_path is None:
+            continue
+
+        result = subprocess.run([ninja_path, '--version'], check=False, text=True, capture_output=True)
+
+        try:
+            candidate_version = tuple(int(x) for x in result.stdout.split('.')[:3])
+        except ValueError:
+            # The meson search function skips forward if the version is not readable or too low
```
dnicolodi, line 1012: I don't get the reference to meson in this comment. I would simply drop the comment.

mesonpy/__init__.py
```diff
@@ -995,6 +996,34 @@ def _project(config_settings: Optional[Dict[Any, Any]]) -> Iterator[Project]:
...
+            continue
+        if candidate_version < (1, 8, 2):
```
dnicolodi, line 1014: I would have defines the minimum version as a module variable and used it where you need it. This is the second place where it used and in this cases it is inevitable that some day someone will update one but not the other. I would do something like:
```python
# up where other constants are defined
_NINJA_REQUIRED_VERSION = (1, 8, 2)
# in the _depstr class definition
    ninja = 'ninja >= {}'.format('.'.join(_NINJA_REQUIRED_VERSION))
# here
if candidate_version < _NINJA_REQUIRED_VERSION
```
or something along these linesm (also see above about the function argument).

mesonpy/__init__.py
```diff
@@ -995,6 +996,34 @@ def _project(config_settings: Optional[Dict[Any, Any]]) -> Iterator[Project]:
...
+            continue
+        return []
+
+    return [_depstr.ninja]
+
+
+def get_requires_for_build_sdist(
+    config_settings: Optional[Dict[str, str]] = None,
+) -> List[str]:
+    return _get_ninja_if_needed()
```
dnicolodi, line 1024:
```python
if _env_has_ninja_command():
   return []
return [_depsr.ninja]
```

mesonpy/__init__.py
```diff
@@ -1009,13 +1038,13 @@ def build_sdist(
...
     config_settings: Optional[Dict[str, str]] = None,
 ) -> List[str]:
-    dependencies = [_depstr.wheel]
-    with _project(config_settings) as project:
-        if not project.is_pure and platform.system() == 'Linux':
-            # we may need patchelf
-            if not shutil.which('patchelf'):  # XXX: This is slightly dangerous.
-                # patchelf not already acessible on the system
-                dependencies.append(_depstr.patchelf)
+    dependencies = [_depstr.wheel, *_get_ninja_if_needed()]
```
dnicolodi, line 1041:
```python
dependencies = [_depstr.wheel]
if not _env_has_ninja_command():
    dependencies.append(_depstr.ninja)
```

mesonpy/__init__.py
```diff
@@ -995,6 +996,34 @@ def _project(config_settings: Optional[Dict[Any, Any]]) -> Iterator[Project]:
         yield project
 
 
+def _get_ninja_if_needed() -> List[str]:
```
dnicolodi, line 999: This does not "get" ninja, it only checks if it is available. I would name this `_env_has_ninja_command()` or something, and then return a boolean instead of the dependency specification. I would probably also make sense to specify the required version as an argument: 
```python
def _env_has_ninja_command(version: str = _NINJA_REQUIRED_VERSION) -> bool:
    ...
```

dnicolodi (commented):
Some style comments and some questions regarding the new tests added.

tests/test_pep518.py
dnicolodi: Also, the Sage jobs in CI fail on this test, but there isn't enough in the logs to understand what goes wrong.

henryiii: The idea here is several projects could be tested. I wasn't very familiar with what would be good from the existing test suite, so I just picked one that looked important; the idea is more could be added. I can add a more if you want. If we never want more, the parametrization can be dropped.

The idea behind the virtual environment is to get an environment without the test requirements of the package (like ninja!) installed. However, due to the isolated build, I am wondering if it's really necessary, as long as you trust the isolated build to work. I might try it, maybe it wasn't needed?

mesonpy/__init__.py
henryiii: This algorithm was copied from meson. Maybe linking to the lines would be better?

mesonpy/__init__.py
henryiii: The current design was based on scikit-build-core, which may need to add multiple dependencies (cmake too, not just ninja), which is why it was designed this way. You are right, a flag is simpler with only one. (And technically might be simpler with two, might revisit there!)

tests/test_pep518.py
henryiii: > build command line in addition to the PEP517 hook functions

Because build creates an isolated environment and downloads ninja, while the hook functions will always be using ninja from the test requirements. It's an integration test to make sure it builds without ninja present.

mesonpy/__init__.py
dnicolodi: I think the algorithm is straightforward and does not need explanation. A link to the implementation in meson is doomed to become stale quickly.

tests/test_pep518.py
dnicolodi: Don't the PEP517 parametrized tests test just this?

henryiii: That's a (heavily mocked) unit test. This is an integration test to make sure you can actually build a package. The naming is rather terrible, I think this should be called "isolated_build" or something like that.

henryiii: Before this got added, test_wheel.py was much closer to an integration test. Now it's more noticeably not, since it will never add ninja, that's already a test dependency. This is a full integration test.

And to answer the question above, it isn't necessary to use virtualenv, it's still adding ninja to the isolated environment with subprocess.run.

I'd like to check the output and verify it's adding ninja. Not sure how to handle that test if there's already a ninja on the tester's system, though.

henryiii: And if you don't care about an integration test, I could drop it.

mesonpy/__init__.py
henryiii: It can be a permalink. The idea though is this ideally needs to match, since it needs to evaluate the version of the same ninja that meson would. It doesn't have to be perfect, but it's best if it's as close to in-sync as possible.

tests/test_pep518.py
dnicolodi: Now I understand better. Thank you. However, note that running the test in a virtualenv in not enough to remove `ninja` from the environment if it is available in `$PATH` and on Linux I expect `ninja` to be available on most CI setups and on macOS too, probably. The check implemented here would thus find `ninja` and do not instruct `build` to install it. To remove `ninja` from the environment you need an higher degree of trickery, which I'm not sure is worth implementing.

mesonpy/__init__.py
```diff
@@ -1009,19 +1010,21 @@ def _get_ninja_if_needed() -> List[str]:
...
 
 
 def get_requires_for_build_sdist(
     config_settings: Optional[Dict[str, str]] = None,
 ) -> List[str]:
-    return _get_ninja_if_needed()
+    dependencies = []
+
+    if not _env_has_ninja_command():
+        dependencies.append(_depstr.ninja)
```
dnicolodi, line 1027: This looks wrong:
```python
    if not _env_has_ninja_command():
        return _depstr.ninja
```

henryiii: You can't return a string here, it has to be a list. I did rewrite it though to look nicer.

mesonpy/__init__.py
henryiii: Haven't added the argument yet.

tests/test_pep518.py
dnicolodi: > And if you don't care about an integration test, I could drop it.

I think the integration test is valuable, better one test more than one test less, but I'm not sure the test tests does what you say.

mesonpy/__init__.py
```diff
@@ -82,6 +73,16 @@ class _depstr:
...
+_NINJA_REQUIRED_VERSION = (1, 8, 2)
+
+
+class _depstr:
+    """Namespace that holds the requirement strings for dependencies we *might*
+    need at runtime. Having them in one place makes it easier to update.
+    """
+    patchelf = 'patchelf >= 0.11.0'
+    wheel = 'wheel >= 0.36.0'  # noqa: F811
+    ninja = f'ninja >= {".".join(_NINJA_REQUIRED_VERSION)}'
```
dnicolodi, line 85: You took my code snipped too seriusly: this is wrong 😄 You need something like:
```python
    ninja = f'ninja >= {".".join(str(i) for i in _NINJA_REQUIRED_VERSION)}'
```

mesonpy/__init__.py
dnicolodi: Which string?

mesonpy/__init__.py
henryiii: Ahhh!

mesonpy/__init__.py
henryiii: You fixed it. :)

dnicolodi: 😄

mesonpy/__init__.py
dnicolodi: The permalink has the drawback that it will document only the algorithm as implemented in the current version of Meson and will not take into account possible future changes to it. If Meson ever changes, the link would actually be detrimental. I would just document the requirement at the beginning of the function. Something like:

```python
# Verify that a suitable ``ninja`` command is available. This function needs and 
# the corresponding one in Meson need to resolve to the same executable.
# Mimic as close as possible the algorithm used by Meson.
```

However, this brings another question: why do we need to delegate to Meson again running `ninja`? If we have this function that finds a suitable `ninja` command, we can simply run `ninja -C ...` ourselves instead than going through meson to do the same for us. Or does `meson setup` fail if `ninja` is not available?

tests/test_pep518.py
henryiii: No, it wasn't to get ninja off the system environment (and still isn't), it was to get ninja out of the local environment. But since we are using an isolated build anyway (with build), that's just an extra layer that wasn't needed, and has been dropped - this is a subprocess.run now[^1]. Currently, it does pull in ninja if I uninstall it locally, but I'm assuming it's present in the CI so using that instead of ninja. (which is nice as well, but it would be better if we could control it).

Maybe a simple 1/2 solution would be to check to see if ninja is on the system, then require or disallow "ninja" in the build output based on that. Not ideal to adapt to the test system instead of of adapting the system to the test, but better than not caring at all if it adds ninja.

[^1]: In scikit-build-core, I also install it after building it, which does require the virtualenv.

mesonpy/__init__.py
henryiii: I believe meson setup looks for ninja, so if it was off enough, it could be problematic. I wonder if meson could expose the ninja search for meson-python to just use?

eli-schwartz: Meson does not expose a public documented python API, this is an intentional decision.

But it's honestly not all that important. The Meson algorithm is just about covering some common bases to ensure that some sort of ninja is found. meson-python wants to support samurai because it's right for meson-python to allow it -- not because it "has to since Meson does".

tests/test_pep518.py
eli-schwartz: If it's important to remove it from the global system environment, maybe you can just create "poisoned" fake executables and prepend them to the system PATH, which report version 0.1 and/or exit with a failing return status?

dnicolodi: I see. If this is the goal, choosing `scipy-like` as a test project may not be the best choice, as it has extra build requirements that need to be installed in the isolated environment. Maybe a test project without extra dependencies could be used.

Anyhow, the test is failing in the CI. Can you adjust it so that the output of the executed command is spitted out in case of test failure so the situation can be debugged? Thank you!

mesonpy/__init__.py
dnicolodi: @eli-schwartz 's and mine comments crossed each other...

dnicolodi: @eli-schwartz This function only check that a `ninja` like command exist. It install `ninja` form PyPI if it does not. It leaves it to meson to invoke it. Thus if this function covers something that Meson does not, we have a problem.

eli-schwartz: - Meson might add support for things that this then misses out on, but the worst outcome is installing an extra ninja via pip, so nothing will break
- meson-python can guarantee that if it found a good ninja, Meson always find it too -- simply export $NINJA pointing to the executable you want. Even though meson will of course find it anyway because it engages in the same general approach to sensible checking... You can guarantee it beyond all shadow of a doubt.


Meson's python API may be internal and subject to change if someone decides that for refactoring purposes it would look better somewhere else...

... but the $NINJA environment variable is a public user-facing feature and can absolutely 100% be relied on.

henryiii: So instead of returning true/false, this can return the location of the discovered ninja, then $NINJA can be set with it. Still is truthy based on if it was found.

tests/test_pep518.py
henryiii: It does print out the result if it fails:

```
---------------------------- Captured stderr setup -----------------------------
/sage/local/var/lib/sage/venv-python3.10/bin/python3: No module named build
```

(Only visible in the logs artifact, since it's truncating the output for the GHA log)

henryiii: I don't know how to tell sage CI to add build to the venv it's creating, though. Doesn't seem to be using `.[test]` or nox.

dnicolodi: I looked at the GitHub Actions workflows involved and I didn't see anything that allows to specify python packages as test requisites. The last to touch the sage CI jobs definitions is @mkoeppe Maybe he knows more...

mkoeppe: The relevant file in Sage is build/pkgs/meson_python/dependencies_check [build/pkgs/meson_python/dependencies_check](https://github.com/sagemath/sagetrac-mirror/compare/develop...u/mkoeppe/numpy_1_23_x__scipy_1_9_x#diff-5a691b20a9521d9e7de0a407f41c8a473a55cf4549d3a2c66c4dbc0aacc59c81)

mkoeppe: but the easiest way to make `build` available is to add `python_build` before `meson_python` in https://github.com/mesonbuild/meson-python/blob/main/.github/workflows/ci-sage.yml#L110

henryiii:
Hmm, for the new tests I get:
```
ERROR: No matching distribution found for pyproject-metadata>=0.5.0
```

And an old test, `tests/test_wheel.py::test_local_lib`, starts failing:

```
ImportError: dynamic module does not define module export function (PyInit_example)
```

tests/test_pep518.py
henryiii: So I believe at least `pyproject-metadata` needs to be added to https://github.com/sagemath/sagetrac-mirror/blob/853a6995c01958d6ec26c8c68cb266944eb8dc5b/build/pkgs/meson_python/dependencies_check? Or I could just have these tests get skipped if the dependencies are not present, that would work too, though not as reassuring.

FFY00: @mkoeppe can you let us know what is the workflow to submit the required changes?

henryiii: I can also add a importorskip and "fix" the test by not testing on sage CI because it's missing dependencies. Then this could be added later. @mkoeppe please advise. If I don't hear anything, I'll probably go that route when I add `$NINJA`.

FFY00:
There have been some fixes, so I am rebasing this on `main`.

mkoeppe:
These failures in the testsuite as run by the Sage CI:
```
WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ConnectTimeoutError(<pip._vendor.urllib3.connection.HTTPSConnection object at 0x7f833feab640>, 'Connection to 192.0.2.0 timed out. (connect timeout=15)')': /simple/pyproject-metadata/
ERROR: Could not find a version that satisfies the requirement pyproject-metadata>=0.5.0 (from versions: none)
ERROR: No matching distribution found for pyproject-metadata>=0.5.0
```
... will be fixed in #313.

henryiii:
Nice, after those two go in, I'll rebase & work on adding `$NINJA`.

FFY00:
@henryiii, can you update the branch?

henryiii:
Yes, I should be able to in about 30 mins.

henryiii:
I'm getting `meson dist --allow-dirty --no-tests --formats gztar` breaking in the new test in Sage, but I can't see any info as to why it breaks.

mkoeppe:
I see `Dist currently only works with Git or Mercurial repos` in https://github.com/mesonbuild/meson-python/actions/runs/3481218227/jobs/5822146022#step:14:326

henryiii:
Hmm, interesting! Any idea what that means, though? This is triggering build isolation, so it should be in a virtual environment, but it should still be pointing at the original source. Unless it's in the sdist -> wheel step. Then it's getting and SDist and making a wheel.

mkoeppe:
All tests in the Sage CI operate from an sdist, not from the meson-python clone.

FFY00:
It seems that Meson isn't being able to produce a dist when it doesn't have VCS, which is reasonable, but we'll need to deal with that.

dnicolodi (commented):
Nice. Maybe a couple of wrong merge resolutions during the rebase?

tests/test_pep518.py
```diff
@@ -15,7 +15,6 @@
 @pytest.mark.parametrize(
     'build_arg', ['', '--wheel'], ids=['sdist_to_wheel', 'wheel_directly']
 )
-@pytest.mark.xfail(sys.platform.startswith('cygwin'), reason='ninja pip package not available on cygwin', strict=True)
 def test_pep518(pep518, package, build_arg, tmp_path):
```
dnicolodi, line 18: It would be cool to extend this test to be an integration test for the two cases: ninja from a package and ninja from the system. However I'm not so sure about what strategy we could use to realize that. If we want to keep the test not dependent on a specific test environment setup, the only thing I can think about is to provide a wheelhouse that does not have ninja, mask a possibly existing ninja by adding a script that returns an error or a wrong version number to a directory we prepend to `$PATH`, and verify that the build fails. Add ninja to the wheelhouse and verify that the build succeeds.

dnicolodi:
I just merged #212 Can you please rebase again? We should see the pep518 test fail on platforms that do not have a system ninja.

dnicolodi:
If you rebase and add `ninja` to the wheelhouse setup, I'll work on the pep518 tests.

dnicolodi:
Seems that all the GitHub Actions runners have ninja installed...

dnicolodi:
Never mind, `ninja` is there because we install it as part of the test dependencies. I think we can merge this and work on extended tests later.

mesonpy/__init__.py
```diff
@@ -972,12 +1011,16 @@ def get_requires_for_build_wheel(
...
+
+    if _env_ninja_command() is None:
+        dependencies.append(_depstr.ninja)
+
+    if sys.platform.startswith('linux'):
+        # we may need patchelf
+        if not shutil.which('patchelf'):
+            # patchelf not already accessible on the system
+            dependencies.append(_depstr.patchelf)
+
```
FFY00, line 1023: Please do not change the `patchelf` logic here, this is a regression of #158.

FFY00 (approved):
I just have one thing that needs to be fixed before merging. Otherwise, it looks good to me! Thank you @henryiii :blush:

mesonpy/__init__.py
henryiii: I can check at home, but I'm pretty sure you can't use `_project` which uses `meson` without `ninja`. I think the only loss is for pure python projects, which probably shouldn't be using meson?

FFY00: You should be able to, Meson generates the introspection metadata during the build directory setup. ninja should only be needed for the build step, which only happens when `build_wheel` is called.

dnicolodi: @henryiii the problem is about `patchelf`, not `ninja`. You lost a `if not project.is_pure` in the condition.

dnicolodi: > You should be able to, Meson generates the introspection metadata during the build directory setup. ninja should only be needed for the build step, which only happens when `build_wheel` is called.

AFAIK, `meson setup` fails if it does not find `ninja`.

eli-schwartz: > I think the only loss is for pure python projects, which probably shouldn't be using meson?

Not actually true, tbh. There's a couple reasons why you might want to use Meson, for example, Meson wants to use Meson. :D Also, if you have files that need to be installed outside of a python install scheme, for example configuration files to /etc or, when installing as a user, config files to `~/.config/foobar.ini`, scripts to `~/.local/bin/foobar`, desktop files to `~/.local/share/applications/foobar.desktop`, and python packages to `~/.venvs/foobar/lib/python3.X/site-packages/foobar-1.0.dist-info/`.

Admittedly, that use case is "do not use wheels", so it would be an enhanced mode when using Meson directly instead of via this backend, but the principle is there. :D 

I also have a Meson PR to add `meson setup builddir/ --backend=none` that doesn't generate a ninja backend, only installdata, which could be useful for pure python projects (again, such as Meson itself).

henryiii: We can't run meson inside get_requires to find out if a project is pure unless meson can be run without ninja, I believe, without a "backend=none" feature. The only loss here is a pure Python project will require patchelf (possibly pre-installed) on linux systems until such a feature is added to meson. Which seems a lot better the current ninja requirement, which is hurting both pure and non pure projects.

FFY00: Okay, then we should do the test regardless if ninja is available, odds are if you are using a local patchelf executable, you are also using a local ninja, and always add patchelf if not. Hopefully, we will be able to change that soon, if/when Eli's PR gets in.

eli-schwartz: Then please express your support for https://github.com/mesonbuild/meson/pull/10779 :)

henryiii: @FFY00 Is there a change necessary? Also, this could be solved by adding a TOML option turning off patchelf checking. Something like `tool.meson-python.needs-patchelf=False` could disable the check entirely, and if you have a pure Python package, you could disable the check (also useful if you don't need patchelf for other reasons).  This avoids paying the cost of running meson just to get need_requires, too.

henryiii: One alternate suggestion: For packages that have a known set of platforms, you can check to see if the platform is supported by a binary wheel, and only add it if it is. See, for example, https://github.com/maxbachmann/RapidFuzz/blob/e6bc56c1adf18b19008c99be6e3318b8d895c10f/_custom_build/backend.py#LL33. We could do that for patchelf, which would only add it if it's safe to add (supported by a wheel), which would allow pure Python packages to avoid breaking if on a system with no patchelf and no wheel.

eli-schwartz:
> Erroring out seems like the right default at least. If we do need a way to create an sdist outside of a VCS, can that be a new issue with its own rationale? Seems low-prio to me, and should not be blocking for dealing with system `ninja`.

Maybe a better way to describe this would be "if we do need a way to create an sdist from an sdist"? I'm not sure I understand the general desirability of this, although apparently some people do want for academic reasons.

For test mocking specifically, there should probably be a setup fixture that runs `git init && git add -A . && git commit -m "create test project repo"` or thereabouts

henryiii:
There is a context manager that does that and I found it and now use it (which is why the tests are passing now) when I realized the other tests had to be doing this too. It can't be a fixture since it needs the package directory as input or it needs to have the working directory already changed to the package directory.

(It could be a fixture that takes arguments via a mark, actually, but probably not worth it?)

eli-schwartz:
Ah okay, never mind then. :D

mesonpy/__init__.py
dnicolodi: I think there is nothing to change in this patch. I'm working on adding some tests, but testing this from the perspective of running `python -m build ...` makes the tests very ugly, especially trying to avoid repeating expensive test setup...

FFY00: I pushed a commit with the changes I was asking. Please review and let me know if you think there will be any issues.

henryiii: Okay, that's fine. I'm still not sure forcing all projects to run `meson setup` just so you can keep pure Python projects from failing (in an environment where binary projects will still fail, mind you!) is a necessary optimization to make. Regardless, hopefully https://github.com/mesonbuild/meson/pull/10779 fixes the no-ninja case or there's eventually a patchelf configuration option, and it won't matter, and it's not any slower or worse than what is done before this PR, it's clearly a net improvement.

henryiii: (I underwent what you meant now :) )

PR #176, opened 2022-10-15
ENH: use tomllib on Python 3.11+

henryiii:
Use tomllib if available.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -35,7 +35,11 @@ from typing import (
     Union
 )
 
-import tomli
+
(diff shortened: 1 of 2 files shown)


rgommers:
@henryiii no those have been happening on other PRs and we need to sort them out. This PR LGTM, thanks for making it!

rgommers:
Eh wait, let me double check. There are more failures now, not just the 2-3 distros from before.

FFY00 (approved):
Thanks!

FFY00:
The change looks good, we just need to figure out what is triggering these CI fails.

PR #177, opened 2022-10-15
BUG: add generated files to sdist

lithomas1:
Generated files were not being added to the sdist.
This is useful for pandas, since we run a versioneer script to generate the sdist.

I think scipy avoids this, since they don't add generate a version file for their sdist?

I'll add a test, sometime in the future, but I'm a little busy now, so feel free to push to this branch in the meantime.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -943,7 +943,7 @@ class Project():
                     # MESON_DIST_ROOT could have a different base name
                     # than the actual sdist basename, so we need to rename here
                     file = meson_dist.extractfile(member.name)
-                    member.name = str(pathlib.Path(dist_name, *member_parts[1:]))
+                    member.name = str(pathlib.Path(dist_name, *member_parts[1:]).as_posix())
                     tar.addfile(member, file)
                     continue
 
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -935,6 +935,9 @@ class Project():
                 path = self._source_dir.joinpath(*member_parts[1:])
 
                 if not path.is_file():
+                    # Could be a generated file, we still need to copy in that case
+                    if member.isfile():
+                        tar.addfile(member, meson_dist.extractfile(member.name))
                     continue
 
                 info = tarfile.TarInfo(member.name)
@@ -943,6 +946,7 @@ class Project():
                 info.mode = int(oct(file_stat.st_mode)[-3:], 8)
 
                 # rewrite the path if necessary, to match the sdist distribution name
+                #print("File is:", path)
                 if dist_name != meson_dist_name:
                     info.name = pathlib.Path(
                         dist_name,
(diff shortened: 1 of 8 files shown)


lithomas1:
cc @rgommers @FFY00

rgommers:
Thanks @lithomas1. I don't think this specific change is a good idea - it's going to pull in all sorts of files that were not intended to be included. I think what you want instead to include specific files that are not under version control (like `version.py` generated by `versioneer`) is either `meson.add_dist_script()` or the `if fs.exists...` code from https://github.com/FFY00/meson-python/issues/159#issuecomment-1265962676. 

> I think scipy avoids this, since they don't add generate a version file for their sdist?

We had a similar version-related file with the git hash, but got rid of it when moving to Meson. We may want to bring it back, but it hasn't been a priority so far.

lithomas1:
I'm a little confused. I'm pretty sure the only way to get generated files into the sdist is by adding them with ``meson.add_dist_script()``(which I'm using with pandas), so there wouldn't be any unwanted generated files. 

One thing that meson-python might consider doing, is parsing a MANIFEST file like setuptools. The .targ.z meson generates includes a bunch of extra things, and its probably much easier to do this filtering in meson.

FFY00:
> it's going to pull in all sorts of files that were not intended to be included

Ralf, can you elaborate on that? AFAICT it will only pull files that are present in the Meson dist but not on the git checkout.

rgommers:
> > it's going to pull in all sorts of files that were not intended to be included
> 
> Ralf, can you elaborate on that? AFAICT it will only pull files that are present in the Meson dist but not on the git checkout.

I think I misunderstood what's going on here. A quick try with a breakpoint under `path.is_file()` showed files that we didn't want to include on SciPy, and the PR title/description seemed to say "all generated files". I did some more testing with the actual change in this PR, and it seems fine indeed.

lithomas1:
Cool, I'll add tests in the future then.

FFY00:
I pushed a small refactoring to make the code a bit easier to maintain, I hope you don't mind. Let me know if you want me to avoid that in the future.

lithomas1:
> I pushed a small refactoring to make the code a bit easier to maintain, I hope you don't mind. Let me know if you want me to avoid that in the future.

No worries. Thanks for fixing the pre-commit as well. 
I'm going to be busy since midterms are rolling around again, so feel free to finish up anything I have open in case I don't circle back to it in time.

lithomas1:
pre-commit.ci autofix

lithomas1:
@rgommers @FFY00 This is as green as it gets.

mesonpy/__init__.py
```diff
@@ -903,6 +903,19 @@ def sdist(self, directory: Path) -> pathlib.Path:
                     continue
                 path = self._source_dir.joinpath(*member_parts[1:])
 
+                if not path.exists() and member.isfile():
+                    # File doesn't exists on the source diractory but exists on
```
FFY00, line 907:
```suggestion
                    # File doesn't exists on the source directory but exists on
```

FFY00 (approved):
Thank you @lithomas1!

PR #180, opened 2022-10-17
TST: allow custom test selection

henryiii:
Currently, you cannot select a specific testfile/test to run when running with nox. This allows specific tests to be selected and is backward compatible.

```bash
nox -s test-3.10 -- tests/test_pep517.py::test_get_requires_for_build_wheel
```

Didn't work before this PR.


FFY00 (approved):
Thanks! This makes sense.

PR #185, opened 2022-10-20
BUG: generate entry_points.txt file

### LINKED ISSUE #184: entry-point and other metadata management
Hi,
I am trying to migrate some projects from setuptools+numpy-distutils `to meson-python`. Apparently the wheels generated are valid and both the modules and extensions are at the proper place in it. But I did not figure out how to have entry-point properly declared.

https://github.com/kif/fabio/blob/meson/pyproject.toml

With `setuptools`,  wheels had a `dist-info/entry-points.txt` containing the list of scripts to generate at installation.
Apparently, this is now done with [project.scripts] in the `pyproject.toml` file but no entry-point is generated from then. Do you have an idea of what could go wrong ?

Thanks for your help.

Jerome

FFY00:
Fixes #184

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -150,6 +150,7 @@ class _WheelBuilder():
     def __init__(
         self,
         project: Project,
+        metadata: Optional[pyproject_metadata.StandardMetadata],
         source_dir: pathlib.Path,
         install_dir: pathlib.Path,
         build_dir: pathlib.Path,
@@ -157,6 +158,7 @@ class _WheelBuilder():
         copy_files: Dict[str, str],
     ) -> None:
         self._project = project
+        self._metadata = metadata
         self._source_dir = source_dir
         self._install_dir = install_dir
         self._build_dir = build_dir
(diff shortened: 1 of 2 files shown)


kif:
Hi,
Thanks for the fast fix, but I don't manage to get the `entry_points.txt` generated at all within the wheel ...

kif:
The property `entrypoints_txt` is apparently never used.

FFY00:
Can you tell me how are you testing?

FFY00:
By default, pypa/build builds with isolation from the current environment, meaning it sets up a new environment with all the requirements to build in, so that stuff from the current environment doesn't end up leaking to the build. So, you need to pass the `-n` option to disable this behavior. If you are running meson-python from source, you'll need to pass `-x` to disable the dependency check, as meson-python will not actually be installed.

kif:
My bad ... at least with those options, the property is call and I should be able to debug...

kif:
Apparently the console scripts, registered under `[project.scripts]` are properly added to the entry-points file but not those from the `[project.gui_scripts]`. I had a look at the metadata object  and the `gui_scripts` entry in the `StandardMetadata` object is empty.

FFY00:
Ah great! The field on `pyproject.toml` is actually called `gui-scripts`, not `gui_scripts`, do that should explain why it is missing 😅

https://peps.python.org/pep-0621/#entry-points

We should definitely detect these typos and warn the user though!

kif:
Apparently the bug was on my side where `[project.gui-scripts]` was written with an underscore, not a dash...

PR #191, opened 2022-11-01
ENH: delegate computing wheel tags to the packaging module

### LINKED ISSUE #142: Unknown Python implementation: pyston (cannot install SciPy >= 1.9.0 on Pyston)
This is similar to #121, only it's for the lesser-known (but awesome) [Pyston](https://github.com/pyston/pyston).

```
NotImplementedError: Unknown Python implementation: pyston. Please report this to https://github.com/FFY00/mesonpy/issues and include information about the Python distribution you are using.
```

Could you add support for it?

### LINKED ISSUE #189: Curious build failure with Python 3.7 on Windows
Running `python -m build .` in GitHub CI ends with this error:
```
+ meson install --destdir C:\Users\RUNNER~1\AppData\Local\Temp\build-via-sdist-x9mu6s_a\siphash24-1.1.dev0\.mesonpy-wgvn__mt\install
ninja: Entering directory `C:\Users\RUNNER~1\AppData\Local\Temp\build-via-sdist-x9mu6s_a\siphash24-1.1.dev0\.mesonpy-wgvn__mt\build'
ninja: no work to do.
Installing siphash24.pyd to C:\Users\RUNNER~1\AppData\Local\Temp\build-via-sdist-x9mu6s_a\siphash24-1.1.dev0\.mesonpy-wgvn__mt\install\hostedtoolcache\windows\Python\3.7.9\x64\Lib\site-packages
Installing siphash24.lib to C:\Users\RUNNER~1\AppData\Local\Temp\build-via-sdist-x9mu6s_a\siphash24-1.1.dev0\.mesonpy-wgvn__mt\install\hostedtoolcache\windows\Python\3.7.9\x64\Lib\site-packages
    return mesonpy._tags.StableABITag(tag)

  File "C:\Users\RUNNER~1\AppData\Local\Temp\build-env-0_b2_gmq\lib\site-packages\mesonpy\__init__.py", line 310, in _calculate_file_abi_tag_heuristic_windows
    return mesonpy._tags.StableABITag(tag)
  File "C:\Users\RUNNER~1\AppData\Local\Temp\build-env-0_b2_gmq\lib\site-packages\mesonpy\_tags.py", line 37, in __init__
    raise ValueError(f'Invalid PEP 3149 stable ABI tag, expecting pattern `{self._REGEX.pattern}`')
ValueError: Invalid PEP 3149 stable ABI tag, expecting pattern `^abi(?P<abi_number>[0-9]+)$`

  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\pep517\in_process\_in_process.py", line 351, in <module>
    main()
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\pep517\in_process\_in_process.py", line 333, in main
  File "C:\hostedtoolcache\windows\Python\3.7.9\x64\lib\site-packages\pep517\in_process\_in_process.py", line 250, in build_wheel
    metadata_directory)
  File "C:\Users\RUNNER~1\AppData\Local\Temp\build-env-0_b2_gmq\lib\site-packages\mesonpy\__init__.py", line 987, in build_wheel
  File "C:\Users\RUNNER~1\AppData\Local\Temp\build-env-0_b2_gmq\lib\site-packages\mesonpy\__init__.py", line 935, in wheel
    wheel = self._wheel_builder.build(self._build_dir)
  File "C:\Users\RUNNER~1\AppData\Local\Temp\build-env-0_b2_gmq\lib\site-packages\mesonpy\__init__.py", line 578, in build
    wheel_file = pathlib.Path(directory, f'{self.name}.whl')
  File "C:\Users\RUNNER~1\AppData\Local\Temp\build-env-0_b2_gmq\lib\site-packages\mesonpy\__init__.py", line 189, in name
    python_tag=self.python_tag,
  File "C:\Users\RUNNER~1\AppData\Local\Temp\build-env-0_b2_gmq\lib\site-packages\mesonpy\__init__.py", line 244, in python_tag
    selected_tag = self._select_abi_tag()
  File "C:\Users\RUNNER~1\AppData\Local\Temp\build-env-0_b2_gmq\lib\site-packages\mesonpy\__init__.py", line 375, in _select_abi_tag
    tags = self._files_by_tag()
  File "C:\Users\RUNNER~1\AppData\Local\Temp\build-env-0_b2_gmq\lib\site-packages\mesonpy\__init__.py", line 361, in _files_by_tag
    tag = self._calculate_file_abi_tag_heuristic(file)
  File "C:\Users\RUNNER~1\AppData\Local\Temp\build-env-0_b2_gmq\lib\site-packages\mesonpy\__init__.py", line 346, in _calculate_file_abi_tag_heuristic
    return self._calculate_file_abi_tag_heuristic_windows(filename)
  File "C:\Users\RUNNER~1\AppData\Local\Temp\build-env-0_b2_gmq\lib\site-packages\mesonpy\__init__.py", line 312, in _calculate_file_abi_tag_heuristic_windows
    return mesonpy._tags.InterpreterTag(tag)
  File "C:\Users\RUNNER~1\AppData\Local\Temp\build-env-0_b2_gmq\lib\site-packages\mesonpy\_tags.py", line 83, in __init__
    f'Unknown Python implementation: {self.implementation}. '
NotImplementedError: Unknown Python implementation: mesonpy. Please report this to https://github.com/FFY00/mesonpy/issues and include information about the Python distribution you are using.

```
Funny enough, the same builds on macOS and Linux and on Windows with Python 3.8 and later. I have absolutely no clue about what may be wrong.

### LINKED ISSUE #190: Use introspection to generate wheel tags
In #189 it become clear that it is not really possible to map extension filenames to wheel tags in a robust way: the file names do not contain ABI tags. However, the python implementation which is running meson-python is the same as the one for which the extensions are being built. Therefore, we can introspect the interpreter to get the wheel tags. Using the `packaging` module seems the most straightforward way to do this.

The only open question is how to decide to go for an implementation ABI tag or a stable ABI tag. I don't think that Meson python module has a way to generate extension modules for the stable ABI, so maybe this is not important right now. But the current meson-python has some support to auto-detect this, so maybe it should be supported somehow.

Should we keep validating the extension filename suffixes against the suffixes accepted by the Python interpreter? The list of accepted suffixes is provided in `importlib.machinery.EXTENSION_SUFFIXES` thus the check is easily implemented. Should deeper checks be done?

dnicolodi:
Use the wheel contents only to determine whether the wheel contains python ABI dependent modules or other platform dependent code.

Fixes #142, fixes #189, fixes #190.


dimpase:
For testing purposes, it would be great to have this fix as a pip-installable package. Is it too much to ask?

dnicolodi:
`pip install git+https://github.com/dnicolodi/meson-python.git@packaging` should work.

You can use something like this:
```toml
[build-system]
requires = [
    "wheel",
    "meson >= 0.63.3",
    "meson-python @ git+https://github.com/dnicolodi/meson-python.git@packaging",
]
build-backend = "mesonpy"
```
in a `pyproject.toml` file.

dimpase:
Thanks. I forgot that pip can install from git repos...

dimpase:
I can confirm that with this fix I am able to install scipy 1.9.3 on an M1 macOS 13. Thanks!

dimpase:
something funny with `pypy` tags, as far as I can see from failing tests here:
```
AssertionError: assert 'pp38' == 'pypy38_pp73'
```

dnicolodi:
Indeed, although it looks like a bug in `wheel`. The wheel filename parsing is confused by filenames in the form `purelib_and_platlib-1.0.0-pp38-pypy38_pp73-manylinux_2_31_x86_64.whl` while `packaging` is fine with it:
```python
>>> import packaging.utils
>>> name, version, build, tags = packaging.utils.parse_wheel_filename('purelib_and_platlib-1.0.0-pp38-pypy38_pp73-manylinux_2_31_x86_64.whl')
>>> t = next(iter(tags))
>>> t.interpreter
'pp38'
>>> t.abi
'pypy38_pp73'
>>> t.platform
'manylinux_2_31_x86_64'
```
Although, I am not sure what is the canonical form for pypy tags.

dnicolodi:
Ah, no. Much simpler. I inverted interpreter and ABI tags in the test. The naming is confusing... Can someone please restart the tests?

dnicolodi:
Some other tests fail because bugs in the tests, namely they look for the string `any` anywhere in the wheel filename and that triggers on `manylinux`. However, this highlights the fact that `packaging` uses something like `manylinux_2_31_x86_64` as platform tags while `wheel` usually generates something that looks like `linux_x86_64`. Is there a standard for these platform tags? I don't see why one version or the other should be preferred, except that we are not checking against the `manylinux` specification, thus putting that in the platform tag may be misleading.  Should we simply use `linux_$architecture`?

dnicolodi:
Windows test were failing for an actual bug. Fixed. Can someone please kick the tests again?

tristan957:
CC @rgommers

belm0:
please mark this as fixing #142 also

dnicolodi:
Thanks @rgommers The two test failures seem unrelated, but I don't have access to the CI artifacts to check what exactly went wrong. Can someone please copy the relevant log portion out?

rgommers:
I think these 3 Sage jobs are known failures that we need to get around to fixing. Failing tests:
```
  [meson_python-git]   FAILED tests/test_wheel.py::test_local_lib - subprocess.CalledProcessError: C...
  [meson_python-git]   FAILED tests/docs/examples/test_spam.py::test_build_and_import - subprocess.C...
```

Can you really not access GitHub Actions logs? They should be public.

dnicolodi:
On `centos-stream-8-python3.9`:
```
ERROR: spam-0.0.1-cp39-cp39-manylinux_2_28_x86_64.whl is not a supported wheel on this platform.
```
This seems a case of `packaging` not agreeing with himself about what's the correct platform tag.

dnicolodi:
The test failure on `gentoo-python3.9` and `opensuse-15.4` seem to be unrelated, indeed.

dnicolodi:
> Can you really not access GitHub Actions logs? They should be public.

I do, but I was under the impression that the complete logs are uploaded as artifacts. What is available in the GitHub Actions logs is not very helpful to see what exactly went wrong in the case of the Gentoo job, for example.

rgommers (commented):
Thanks @dnicolodi! This is the right approach overall I think. We've had too many issues with wheel tags, and it's mostly reverse-engineering what `packaging` does, which is not much fun. The test coverage of produced wheel filenames in `test_wheel` still looks fine.

The main question I have here is whether or not this is a regression for cross-compilation. It seems like it should be, since with this PR we get packaging from the native interpreter and there is no attempt to detect whether we are cross-compiling. When deriving tags from file extensions, that naturally worked out correctly.

dnicolodi:
I don't think PEP517 is intended to support cross-compilation: it is assumed that the interpreter controlling the build is the one for which the packages are being built. `meson-python` enforces that passing the current interpreter as the python program to build for to `meson` via the patform declaration file. `meson` in turn derives the extension modules file names by running the python interpreter. I don't think this allows for cross-compilation.

dimpase:
> > Can you really not access GitHub Actions logs? They should be public.
> 
> I do, but I was under the impression that the complete logs are uploaded as artifacts. What is available in the GitHub Actions logs is not very helpful to see what exactly went wrong in the case of the Gentoo job, for example.

Gentoo has its own issues with the new scipy: https://bugs.gentoo.org/878883 - not sure if that's the cause of the problem here, though.

dimpase:
> I think these 3 Sage jobs are known failures that we need to get around to fixing. Failing tests:
> 
> ```
>   [meson_python-git]   FAILED tests/test_wheel.py::test_local_lib - subprocess.CalledProcessError: C...
>   [meson_python-git]   FAILED tests/docs/examples/test_spam.py::test_build_and_import - subprocess.C...

How can one get more details on Sage failures? E.g. one would like to see the main Sage's `config.log`, etc.

> ```
> 
> Can you really not access GitHub Actions logs? They should be public.

dnicolodi:
> Gentoo has its own issues with the new scipy: https://bugs.gentoo.org/878883 - not sure if that's the cause of the problem here, though.

This is a different problem. I think it is due to the fact that the version of `packaging` vendored by the `pip` available on Gentoo and the one installed in the test environment do not agree on what is the right platform tag. 

Definitely, using `packaging` for generating the platform tag results in very specific tags. I can find a way to relax them, but I don't know how how relevant is the issue of using an old `pip` to install a package build with the latest release of the build tools.

rgommers:
> I don't think PEP517 is intended to support cross-compilation: it is assumed that the interpreter controlling the build is the one for which the packages are being built. `meson-python` enforces that passing the current interpreter as the python program to build for to `meson` via `--native-file`. `meson` in turn derives the extension modules file names by running the python interpreter. I don't think this allows for cross-compilation.

I don't think there's such an assumption, rather it's not considered at all. The only thing PEP 517 does is provide a way to invoke a build system; what you do after that is completely up to you, as long as you produce an sdist or wheel.

> Or, from another prospective, how would you run a cross compilation with `build` or `pip` or other interface to PEP517 backends?

Here's an example: https://github.com/scipy/scipy/blob/6912c638fc13c4eb67b09412c33557f53a821055/.github/workflows/wheels.yml#L205-L220

I'll note that the separate line
```bash
meson setup --cross-file $(pwd)/tools/wheels/cross_arm64.txt build
```
is a workaround for `meson-python` not support `config_settings` yet (see gh-54, gh-167). Once that lands, I'd expect to be able to use `--cross-file` directly. Either way, I think this PR breaks that SciPy cross-compilation setup.

rgommers:
> I can find a way to relax them, but I don't know how how relevant is the issue of using an old `pip` to install a package build with the latest release of the build tools.

Unfortunately, quite relevant. If the produced wheels don't work with `pip` versions packaged in Linux distros, there's going to be a lot of complaints.

dnicolodi:
> > I can find a way to relax them, but I don't know how how relevant is the issue of using an old `pip` to install a package build with the latest release of the build tools.
> 
> Unfortunately, quite relevant. If the produced wheels don't work with `pip` versions packaged in Linux distros, there's going to be a lot of complaints.

If `packaging` does no agree with himself on some platforms there is little that can be done, expect dumb down the platform tags. I can work on doing this, but having to wait days to get the tests run is not an efficient way to do it...

dnicolodi:
@rgommers I amended the path to rewrite the tags to be more generic. This should restore compatibility with `pip` back to version 8.1.0 and `auditwheel` 1.0.0. Can someone please spin the tests once more?

rgommers:
> It definitely does but there is no way to obtain correct wheel tags without being able to execute the Python interpreter for which the wheel are being built. 

Well it works in `main` now - so I guess that's an advantage of the current approach. Maybe we need a hybrid of the two methods? Get both, and when they mismatch warn, when one method has support but the other does not, then continue. E.g. Pyston is an unknown interpreter for `meson-python` now, but is supporting in `packaging`. Cross-compilation is the other way around.

> Also, how does the Meson `python` module even work without being able to execute the Python interpreter it is building for?

It doesn't need to execute it, only to introspect it somehow. Sane cross-compilation support is one of the major advantages of using Meson, so we should keep it in mind (and add a few tests for it in this repo at some point).

dnicolodi:
Ops, forgot to adjust the tests. At least I know that the platform tags are indeed modified as expected. @rgommers can you spin the tests once more, please?

dnicolodi:
> It doesn't need to execute it, only to introspect it somehow.

It does introspect it by executing it, see https://github.com/mesonbuild/meson/blob/97ec20e90142c229f62a6d20371f44df0b8dd41e/mesonbuild/modules/python.py#L327

dnicolodi:
> Pyston is an unknown interpreter for `meson-python` now, but is supporting in `packaging`. Cross-compilation is the other way around.

It is not possible to generate correct wheel tags only looking at the extension module filenames. Some required information about the interpreter is not encoded in the file names. Also the extension file names are not required to contain any information about the interpreter for which they are compiled. For example, the Meson `python` module does not add the ABI tag to the extension filenames for Python 3.7. With current Meson, `meson-python` does not work at all for Python 3.7.

dnicolodi:
Argh! Forgot the type annotations... Fixed, but I'll wait the current test run to finish before pushing the fix.

dnicolodi:
Remaining test failures are most likely unrelated to this PR: two `ImportError: dynamic module does not define module export function (PyInit_example)` and one with unspecified multiple test errors.

dnicolodi:
The only way I see to support cross compilation is to have a way to specify the wheel tags explicitly instead than computing them. I am not sure whether there is a way that works across build frontends to pass configuration options. As a last resort, environment variables can be used. Looking at the scipy example, this is anyway more or less what it is done already.

rgommers:
> The only way I see to support cross compilation is to have a way to specify the wheel tags explicitly instead than computing them. I am not sure whether there is a way that works across build frontends to pass configuration options. As a last resort, environment variables can be used. Looking at the scipy example, this is anyway more or less what it is done already.

Yes, I think you are right. It's not done manually there, but it shouldn't be fully robust - we're probably getting it right a little by accident. Here's a [CI log](https://github.com/scipy/scipy/actions/runs/3350870278/jobs/5551959221) that shows:
```
Building wheels for collected packages: scipy
  Created temporary directory: /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-wheel-5kel266f
  Destination directory: /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-wheel-5kel266f
  Building wheel for scipy (pyproject.toml): started
  Running command Building wheel for scipy (pyproject.toml)
  Building wheel for scipy (pyproject.toml): finished with status 'done'
  Created wheel for scipy: filename=scipy-1.10.0.dev0-cp38-cp38-macosx_11_0_arm64.whl size=21059121 sha256=878586e93ccc537df2e101857924b860d5386df745882fd263cd5240b371266c
  Stored in directory: /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/pip-ephem-wheel-cache-r8_tyzmx/wheels/e4/9c/9e/cfec93e95f4b4c7f6445dfaa96cf45ba0d6be751a6c69cd040
Successfully built scipy
```

dnicolodi:
The build log has
```
...
Program python3 found: YES (/Users/runner/hostedtoolcache/Python/3.8.14/x64/bin/python)
...
Linking target scipy/_lib/_test_ccallback.cpython-38-darwin.so
...
Copying files to wheel...
...
  [3/1234] scipy/_lib/_test_ccallback.cpython-38-darwin.so
```
So there is no indication that Meson knows that the build is for an arm64 Python. I think things work out just by accident. I also don't see how the wheel ends up being named `scipy-1.10.0.dev0-cp38-cp38-macosx_11_0_arm64.whl` as there is nothing in the extension modules filenames to indicate that they are built for arm64.

dnicolodi:
I found the trick: https://github.com/scipy/scipy/blob/6912c638fc13c4eb67b09412c33557f53a821055/.github/workflows/wheels.yml#L168
```python
>>> import sysconfig
>>> sysconfig.get_platform()
'macosx-10.15-x86_64'
>>> import os
>>> os.environ['_PYTHON_HOST_PLATFORM']="macosx-12.0-arm64"
>>> sysconfig.get_platform()
'macosx-12.0-arm64'
```
So even the current solution uses the python used to run `meson-python` and thus `meson` to determine the wheel tags, simply Python is tricked into reporting the "right" information.

eli-schwartz:
Substantially, the introspection used to find out information is running a python script that imports a data file referenced by the environment variable `_PYTHON_SYSCONFIGDATA_NAME`, which is basically a very fancy python dict.

It happens to be that the python script which gets run is shipped with the python stdlib (and defaults to finding the data file bundled alongside it. The script is called "sysconfig" but I suspect at this late point in my post, you've figured that one out. ;)

The state of the art in python cross compilation is, AFAIK, mostly to set that variable.

...

The other approach would be to specify to meson that the python interpreter you're building for isn't called `/path/to/cross/python` but rather called `qemu-user /path/to/python`. That will run fine, albeit a bit of a heavyweight approach. It also doesn't fully generalize, because it assumes that qemu supports your OS. There may be a similar tool for whatever other OS one does use, however.

dnicolodi:
Thanks @eli-schwartz 

It seems that setting `_PYTHON_SYSCONFIGDATA_NAME` would work with the approach for generating the wheel tags used here... as long as all the things the `packaging` module introspects to generate the wheel tags are correctly covered. I'm not sure if this is the case. At a cursory look, the introspection of the running interpreter and of the surrounding system is quite deep in some cases.

dnicolodi:
The failed tests are the known broken ones.

From what I can understand this patches do not make cross compilation any less supported than what it is now: it requires setting some undocumented environmental variables to coerce the host machine Python to lay about its identity. From what I can tell, the scipy cross compilation example @rgommers brought up would continue to work as it is with this PR merged.

This PR fixes at least three know issues (#142, #189, and #160 for good) and removes quite a bit of complexity.

I think there is nothing holding this PR back and I would like to ask for it to be merged.

rgommers:
I'm okay with following up on cross-compilation separately (thanks for investigating - the info in this thread will be useful!), and having this move forward. @FFY00 wrote most of the current `_tags.py` code, so let's wait for their review.

mesonpy/__init__.py
```diff
@@ -267,166 +301,34 @@ def _debian_python(self) -> bool:
...
     @cached_property
-    def platform_tag(self) -> str:
-        if self.is_pure:
-            return 'any'
-        # XXX: Choose the sysconfig platform here and let something like auditwheel
-        #      fix it later if there are system dependencies (eg. replace it with a manylinux tag)
-        platform_ = sysconfig.get_platform()
-        parts = platform_.split('-')
-        if parts[0] == 'macosx':
-            target = os.environ.get('MACOSX_DEPLOYMENT_TARGET')
```
FFY00, line 293: We need this code. This breaks `MACOSX_DEPLOYMENT_TARGET` support.

mesonpy/__init__.py
```diff
@@ -267,166 +301,34 @@ def _debian_python(self) -> bool:
...
+            if os.name == 'nt' and path.suffix == '.pyd' or path.suffix == '.so':
+                name, suffix = path.name.split('.', 1)
+                if suffix not in _EXTENSION_SUFFIXES:
+                    raise ValueError('Extension module "{}" not compatible with Python interpreter.'.format(path))
+                match = _EXTENSION_SUFFIX_REGEX.match(suffix)
+                if match is None:
+                    raise ValueError('Suffix "{}" for extension module "{}" not understood.'.format(suffix, path))
+                abis.append(match.group('abi'))
+
+        return all(abi == 'abi3' for abi in abis)
```
FFY00, line 331: We should change this function to return the ABI version.

FFY00 (changes requested):

FFY00:
Thank you for the work @dnicolodi, this is great! I have some comments I'd like to see address before merging. Let me know if you agree and/or have any feedback.

mesonpy/__init__.py
```diff
@@ -187,14 +203,32 @@ def basename(self) -> str:
...
 
+    @property
+    def tags(self) -> packaging.tags.Tag:
+        """Wheel tags."""
+        if self.is_pure:
+            return packaging.tags.Tag('py3', 'none', 'any')
+        # Get the most specific tag for the Python interpreter.
+        tag = next(packaging.tags.sys_tags())
+        if tag.platform.startswith('manylinux'):
+            tag = packaging.tags.Tag(tag.interpreter, tag.abi, _adjust_manylinux_tag(tag.platform))
```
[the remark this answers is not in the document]
dnicolodi: I don't want to claim that what my patch does is beautiful, but what you propose unfortunately does not work. `InterpreterTag.python` is unfortunately not complete as reported in #121. If we depend on `packaging` anyway, I would delegate to that for computing the interpreter tag correctly. Also, as you can see from the code of `packaging` or `wheel` or the code in `meson-python` now, `sysconfig.get_platform()` does not always return the right thing as you need to check for a 32-bit Python running on a 64-bit kernel.

All in all, I think the if we want to delegate some work to `packaging` it makes sense to delegate as much of it as possible.

mesonpy/__init__.py
```diff
@@ -187,14 +203,32 @@ def basename(self) -> str:
...
+        if not self._has_extension_modules:
+            # The wheel has platform dependent code (is not pure) but
+            # does not contain any extension module (does not
+            # distribute any file in {platlib}) thus use generic
+            # implementation and ABI tags.
+            return packaging.tags.Tag('py3', 'none', tag.platform)
+        elif self._use_stable_abi:
+            # All distributed extension modules use the stable ABI.
+            return packaging.tags.Tag(tag.interpreter, 'abi3', tag.platform)
```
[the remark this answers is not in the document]
dnicolodi: There is only one defined and it is `abi3`. See PEP 384, PEP 3149, and https://docs.python.org/3/c-api/stable.html. Also note that currently the Meson's `python` module does not offer a way to build an `abi3` tagged extension module (the extension suffix is fixed) and that the `.abi3.so` suffix is only advisory and not mandatory for `Py_LIMITED_API` modules, thus this is an heuristic at best.

mesonpy/__init__.py
dnicolodi: There is only one versioned ABI tag. See the other comment.

mesonpy/__init__.py
dnicolodi: AFAIK, none of the other PEP517 backends implement this. I tested this with `cibuildwheel` and it seemed to do the right thing. I guess `cibuildwheel` sets the `_PYTHON_HOST_PLATFORM` or an equivalent environment variable to achieve the same. As highlighted by the bug reports concerning macOS 13, getting the platform tag to agree with what `packaging` thinks it needs to be is going to be a game of whack-a-mole I would really like to avoid if possible.

dnicolodi: Yes, that's what `cibuildwheel` does: https://github.com/pypa/cibuildwheel/blob/2e6cbe2435ef410156319c98625002ed76e08785/cibuildwheel/macos.py#L219

mesonpy/__init__.py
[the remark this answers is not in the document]
dnicolodi: I don't think there is value in splitting this into separate functions and it would require either internal caching or calling multiple times possibly expensive `packaging.tags` functions. We need the tag only in a couple of places and always in its complete form. Anyhow, the reason why I decided to return a `packaging.tags.Tag` instead of a string is indeed that this way accessing the individual parts of the tag is straightforward.

mesonpy/__init__.py
dnicolodi: I think that advanced users that require the magic of `MACOS_DEPLOYMENT_TARGET` should also know about `_PYTHON_HOST_PLATFORM`. I would prefer documenting this somewhere than adding special handling here. In general, I think that `meson-python` should behave as close as possible to other Python build tools, to adhere to the principle of less surprise.

dnicolodi:
Test failures seem due to changes in Meson 0.64

rgommers:
Hmm, that seems bad:
```
found pkg-config 'C:\\Strawberry\\perl\\bin\\pkg-config.BAT' but it is Strawberry Perl and thus broken. Ignoring...
Found Pkg-config: NO
Could not find Python3 library 'D:\\a\\meson-python\\meson-python\\.nox\\test-3-10\\python310.dll'
Run-time dependency python3 found: NO (tried sysconfig)

..\..\meson.build:6:0: ERROR: Dependency lookup for python3 with method 'pkgconfig' failed: Pkg-config binary for machine 1 not found. Giving up.
```
The Perl thing is a bit of a red herring it seems: https://github.com/mesonbuild/meson/issues/9480. So what we're left with is `python3` not being found for some reason that needs investigating.

dnicolodi:
Reported as mesonbuild/meson#11017

dnicolodi:
@rgommers Fixed in #198

mesonpy/__init__.py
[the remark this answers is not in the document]
FFY00: > I don't want to claim that what my patch does is beautiful, but what you propose unfortunately does not work. `InterpreterTag.python` is unfortunately not complete as reported in #121. If we depend on `packaging` anyway, I would delegate to that for computing the interpreter tag correctly.

You are right, I opened https://github.com/pypa/packaging/pull/612. We can copy the `interpreter_tag` implementation from https://github.com/pypa/packaging/pull/613 for now.

> Also, as you can see from the code of `packaging` or `wheel` or the code in `meson-python` now, `sysconfig.get_platform()` does not always return the right thing as you need to check for a 32-bit Python running on a 64-bit kernel.

For now we can use the first value from `linux_platforms`, but I think this functionality should either be provided by `packaging` directly or `sysconfig`.

mesonpy/__init__.py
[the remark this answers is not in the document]
FFY00: Yes, but there have been talks about defining a new one, and it doesn't hurt to do some future proofing. The PEPs were written with an understanding that future versions would just update the number, as you can see from the example provided. We can do that, and have further discussion, in a future PR though.

mesonpy/__init__.py
FFY00: Yes, but that's private API and meant to be used for cross-builds specifically. We should be migrating away from relying on it, as it could have other unintended behavior changes, or even go away entirely if/when we add better cross-build support in Python.

mesonpy/__init__.py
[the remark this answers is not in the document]
eli-schwartz: > but I think this functionality should either be provided by packaging directly or sysconfig.

I've actually been thinking about this quite a bit recently, because Meson cannot depend on `packaging` or `setuptools` (and still uses distutils for more than just Debian installation schemes). I think the appropriate location is sysconfig, because it's information about the current python interpreter as it is running on the current system, and is basically set in stone as a result, whereas `packaging.tags` (in theme with the general approach of `packaging`) is about the set of all python interpreters intersecting all OSes, including the ones released after the pair that you're currently running.

Speaking of which, I heard you can do something about that now. :)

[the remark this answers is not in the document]
eli-schwartz: I guess what I'd like to see is something like `sysconfig.platform_tag(stable=False, pure=True)` and then it could tell you what abi standard the current interpreter uses, and infer e.g. the most recent manylinux/musllinux tag based on the loaded glibc or musl.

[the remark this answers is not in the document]
dnicolodi: > You are right, I opened [pypa/packaging#612](https://github.com/pypa/packaging/issues/612). We can copy the `interpreter_tag` implementation from [pypa/packaging#613](https://github.com/pypa/packaging/pull/613) for now.

I can do that but I don't see the advantage. Why calling that function is better than calling `sys_tags()` and extracting the component we are interested in? I understand that the API provided by `packaging` is not the most effective butI think the suboptimal API is better than cooking our own half baked version.

> For now we can use the last value from `linux_platforms`, but I think this functionality should either be provided by `packaging` directly or `sysconfig`.

There is the same problem on macOS. But again, I don't see why getting the thing piecewise is better than calling one function be done with it.

mesonpy/__init__.py
dnicolodi: If this is the case, support for `MACOS_DEPLOYMENT_TARGET` should go into `sysconfig` or `packaging` or whatever. `meson-python` tried to have his own code to generate the macOS platform tag, and it is wrong. Letting `packaging` deal with it makes it certain that at least `pip` and `meson-python` will agree on the tags. And makes it one less place that needs to be fixed at the next macOS major version bump, or for the next re-intepretation of the platform tag PEPs.

dnicolodi:
I'm coming around on some of the objections. I had a better look at the API provided by `packaging` and, assuming that all functions not explicitly marked as private are to be considered public, not only the ones documented in the manual, we can get most of the things we need without re-implementing too much. I'll give it a try later today.

mesonpy/__init__.py
dnicolodi: My reason for resisting to implement this is that I don't want to replicate this https://github.com/pypa/packaging/blob/606c71acce93d04e778cdbdea16231b36d6b870f/packaging/tags.py#L346-L431 into `meson-python`. However, at a closer look, that function allows to pass in a version specifier, so problem solved.

henryiii: FWIW, I've liked the way I implemented this in scikit-build-core: 

https://github.com/scikit-build/scikit-build-core/blob/3ab96051f3626f4323f681e5f29afaf029735bc0/src/scikit_build_core/builder/wheel_tag.py#L29-L54
&
https://github.com/scikit-build/scikit-build-core/blob/main/src/scikit_build_core/builder/macos.py

FFY00: Well, we only need the following snippet, to work around the `packaging` bug.

https://github.com/mesonbuild/meson-python/blob/ef67e0e03c3224a8703809ce6eafb8ac5ce352d5/mesonpy/__init__.py#L312-L320

But, arguably, we could even remove it, and require users to se `MACOSX_DEPLOYMENT_TARGET` to a vlaue `packaging` currently accepts. I would like to keep this workaround though, and just remove it if it gives us toruble.

dnicolodi: @FFY00 I don't understand. If we delegate computing the tag to `packaging` via `packaging.tags.mac_platforms()` we necessarily agree with what `packaging` considers valid. Don't we?

FFY00: See my point above.

> `MACOS_DEPLOYMENT_TARGET` is set to overwrite the target platform, making `sysconfig.get_platform` return it is wrong IMO, and making `packging` doesn't feel right either with the current provided API, as it is primarily used to see if a certain platform is compatible with the running interpreter.

FFY00: We can still use `mac_platforms` if you want, but we should support overwriting that via `MACOS_DEPLOYMENT_TARGET`.

Sorry, I wasn't clear enough above.

dnicolodi: I think I'm not making my point come across. I was thinking to the equivalent of
```python
try:
    version = tuple(map(int, os.environ['MACOS_DEPLOYMENT_TARGET'].split('.')))
except ValueError:
    version = None
platform = next(packaging.tags.mac_platforms(version))
```
where all and every normalization is delegated to `packaging`.

FFY00: Yes, that makes sense :+1:

henryiii: I would also recommend producing at least a warning if `MACOS_DEPLOYMENT_TARGET` is set lower than `sysconfig.get_platform()`, which is the version Python was compiled with. Maybe an error, as you won't really be producing something that can run on what you think it can run on.

dnicolodi:
I've been looking into this some more and the main piece missing without re-implementing parts of packaging into `meson-python` is `packaging.tags.interpreter_abi()` see pypa/packaging#611 Unfortunately the bit and pieces necessary to implement this simple function are private and it may be better to do not rely on them. It really seem that `packaging` is not meant to provide the functionality we require, but I don't know of a better match. `wheel` would be the best candidate, but it does not have (yet) a stable API.

I see two possibilities: either continue to abuse the `packaging` API like in the approach in the current patch, or bite the bullet and start re-implementing the required pieces here, with the risk of the implementations diverging because of bugs or reinterpretations of the PEPs or new platforms appearing. The first approach is not that bad, but the second would be a step into providing `meson-python` functionality directly from Meson (because of the Meson's "no dependencies outside the standard library" policy).

I'll go with the first approach, for now.

mesonpy/__init__.py
dnicolodi: > I would also recommend producing at least a warning if `MACOS_DEPLOYMENT_TARGET` is set lower than `sysconfig.get_platform()`, which is the version of macOS that Python was compiled with. Maybe an error, as you won't really be producing something that can run on what you think it can run on. (which is what I was trying to show above)

I don't understand why this is the case. The extensions are not statically linked with the Python libraries.

henryiii: It was due to https://github.com/pypa/setuptools/blob/cd7d6d56a1cf84cafea44f7cf7e357a926821f03/distutils/spawn.py#L62-L70 - which is now removed in favor of a single check - apparently it mattered once, for... 10.3. The configuration values in sysconfig were incorrect if it was compiled for an older macOS than 10.3. The modern version of that is here: https://github.com/pypa/setuptools/blob/ac727af6acd38537cdddc6c98bc9457aa1eb991a/setuptools/_distutils/util.py#L98-L117 (FYI, Python itself hasn't supported < 10.6 in forever and < 10.9 in a while, AFAIK.)

Sysconfig values are a lot less important for us than for distutils, too, so I think I was wrong.

dnicolodi:
I think the comments have all been addressed. This is ready for another round of review.

Despite there isn't in Meson yet an easy way to generate extension modules targeting the stable Python ABI, I added some synthetic tests checking that the handling of the stable ABI tag works as intended. Unfortunately, it turns out that, as I was fearing, the stable ABI tag in the extension modules filename is not only just advisory, it is not even supported on Windows: a CPython interpreter on windows does not load extension modules with filename suffix `.abi3.pyd`. Therefore, the `meson-python` handling of the stable ABI tag is "broken by design". I decided to keep the current logic in place, but a better solution has to be found if we want to support this use case. By the way, does anyone know of any project that builds against the stable ABI?

henryiii:
That's correct, on Windows you are not supposed to add the `.abi3` part of the extension, that's by design (why, I don't know). You also need to link against `python3.dll` instead of `python3.X.dll` when building. You can see some stable ABI users here: https://discuss.python.org/t/lets-get-rid-of-the-stable-abi-but-keep-the-limited-api/18458/11?u=henryiii

dnicolodi:
Thanks @henryiii I also found https://github.com/pypa/cibuildwheel/issues/78 with some information. In general there isn't much documentation about how this is to be used. I think the only build tools that supports this is `setuptools` and its documentation on this is very poor.

Clearly detecting the stable ABI from the extension modules filenames is not a solution, if it cannot be done on all platforms. Also, it seems that one of the goals is to have extension modules compatible with multiple Python versions, thus using the stable ABI wheel tag and the current Python interpreter version for the interpreter tag is not going to be enough. The only solution I see is a configuration parameter in `pyproject.toml`. Something like 
```toml
[tool.meson]
python-api = "cp32"
```

As this is an existing separate issue and Meson does not have yet an easy way to produce stable ABI extensions, I would leave solving this issue to another PR.

henryiii:
Given the mention of `cryptography` on that issue, I'm guessing Maturin supports it. I've got initial support in scikit-build-core, too, though FindPython from CMake doesn't support it so I'm only testing it on Windows. Yes, you can't auto-detect it, and you need to tell the system what your minimum version is for the tagging. In scikit-build-core:

```toml
[tool.scikit-build]
wheel.py-abi = "cp32-abi3"
```

Would target ABI3. It can also be used to build a "Pythonless" wheel (one that doesn't use Python, like ninja, cmake, or clang-format, or just has things you load via ctypes):

```toml
[tool.scikit-build]
wheel.py-abi = "py3-none"
```

SKBUILD_SOABI is set to `"abi3"` on non-Windows if this is set to `*-abi3`.

dnicolodi:
I'm a bit fuzzy on the connection between the limited API and the stable ABI. I thought that using the limited API always produces a stable ABI extension module. Doesn't it?

henryiii:
Yes. The reason I settled on requesting both API and ABI tags was to have a way to specify binary tags that do not depend on Python at all; currently I think those are just as if not more common than limited API / stable ABI extensions. You _could_ do just the api tag and `cp32` would be limited ABI and `py3` or `py2.py3` or `py37` would be a wheels that doesn't depend on the C api. I was a little worried about that being unclear; setting `py37` or `cp37` here for example means you'd support anything 3.7+, but it looks like you are locking a specific version of Python instead. Maybe it's fine? (scikit-build-core is still experimental and can be changed, if we have similar settings that would be nice)

dnicolodi:
I think #202 is a better approach in the long run.

PR #194, opened 2022-11-04
BUG: Fix breakage when used with Tox

ssssam:
Remove an unnecessary `os.path.join()` call that broke when the output directory was a relative path.

This mistake was hidden in most cases because joining two absolute paths
together with os.path.join() results in the first path being dropped.

Fixes meson-python used with Tox.

### DIFF
--- a/mesonpy/_util.py
+++ b/mesonpy/_util.py
@@ -47,7 +47,7 @@ def create_targz(path: Path) -> Iterator[Tuple[tarfile.TarFile, Optional[int]]]:
     mtime = int(source_date_epoch) if source_date_epoch else None
 
     file = typing.cast(IO[bytes], gzip.GzipFile(
-        os.path.join(path, path),
+        path,
         mode='wb',
         mtime=mtime,
     ))


ssssam:
Hmm, the tests seem to be failing with exactly the error that the patch is supposed to fix - i wonder if I should reorder the two commits.

FFY00 (approved):
Thanks @ssssam! I think this was something that got overlooked when we merged this code.

FFY00:
Hum, something is actually broken here with this change.

ssssam:
I wonder if the unit test I introduced is broken somehow, as it seems to fail in CI repeatedly. In fact, I can reproduce that locally with commit f64caec48540aaa76cd913ac50ac5c6b67fd93ca using `nox` - let me see what's wrong.

ssssam:
I'm struggling to get a unit test that works. Since the fix is fairly obvious, I've pushed a new version of this change with just the bug fix.

rgommers:
If this is all green, I guess it can be merged. @ssssam it would be great if you added a traceback in the description, that will make this easier to find if the problem re-occurs.

dnicolodi:
I'm surprised that the tests that were added failed. We may be overlooking something. I'm having a look.

dnicolodi:
The fix is obviously correct, but I've no idea how `tox` manages to trigger the issue, the path is normalized to an absolute path deep into the `build` code before calling out to the backend. Is `nox` implementing a pep517 frontend? Anyhow, merging.

PR #196, opened 2022-11-07
CI: skip tests on documentation changes

FFY00:
I think we are gonna run into #8926 because of the required checks, and AFAIK there's no sane way to handle it, we need to copy the test matrix over to a new workflow, but lets see how it goes.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,6 +6,13 @@ on:
     branches:
       - main
       - release-*
+    paths-ignore:
+      # This is spelt like this to do not ignore the docs/examples/ folder.
+      - docs/*
+      - docs/usage/*
+      - CHANGELOG.rst
+      - LICENSE
+      - README.md
   workflow_dispatch:
     # Allow to run manually
 
--- a/.github/workflows/sage.yml
+++ b/.github/workflows/sage.yml
@@ -1,4 +1,4 @@
-name: Run Sage CI for Linux
+name: sage
 
 ## This GitHub Actions workflow provides:
 ##
@@ -35,14 +35,18 @@ name: Run Sage CI for Linux
 ## systems/configurations.
 
 on:
+  push:
   pull_request:
     branches:
       - main
       - release-*
-  push:
-    branches:
-      - main
-      - release-*
+    paths-ignore:
+      # This is spelt like this to do not ignore the docs/examples/ folder.
+      - docs/*
+      - docs/usage/*
+      - CHANGELOG.rst
+      - LICENSE
+      - README.md
   workflow_dispatch:
     # Allow to run manually
 
@@ -52,10 +56,9 @@ concurrency:
 
 env:
   # Ubuntu packages to install so that the project's can build an sdist
-  DIST_PREREQ:      python3-pip
+  DIST_PREREQ: python3-pip
   # Name of this project in the Sage distribution
-  SPKG:             meson_python
-  ## REMOVE_PATCHES:   "*"
+  SPKG: meson_python
 
 jobs:
 


.github/workflows/tests.yml
```diff
@@ -6,6 +6,11 @@ on:
     branches:
       - main
       - release-*
+    paths:
```
dnicolodi, line 9: Maybe using `paths-ignore: docs/**` is a bit more robust?

FFY00: Eh, it's the same I think, as we would still need to list the README, changelog, etc...

dnicolodi: I was working on the assumption that it is better to run the tests once more that to do not run them if the patterns here are not updated.

FFY00: That's true, and given I don't think we are gonna be able to get Github to display a *skiped* icon on the check, I think that may be better. Thanks.

.github/workflows/ci-sage.yml
```diff
@@ -36,13 +36,15 @@ name: Run Sage CI for Linux
...
-      - release-*
   push:
     branches:
       - main
       - release-*
+    paths:
+      - mesonpy/**
+      - tests/**
+      - meson.build
+      - pyproject.toml
```
dnicolodi, line 47: @FFY00 is it intentional that Sage tests are not run with the same criteria as the other tests?

FFY00: Nop, this should be the same as the other workflow.

dnicolodi: The sage tests are not run on pushes to branches other than main and release branches. I thought this was intentional to do not run on branches used on forks to open PRs. Anyhow, I'll sync the triggers.

FFY00: Though this may be something we want to keep. Let's see if other people have any thoughts about this.

dnicolodi: If this is the intent, maybe it is better to use an explicit condition on the repository

dnicolodi: I checked an officially there are no limits on the numbers of workflows or jobs we run, thus maybe there is no need to be too careful. https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits On the other hand, I see that sometimes the jobs scheduling gets stuck. I don't know if that is a reaction to the many jobs scheduled or something else.

One limitation we definitely hit is 20 concurrent jobs. For debugging purposes, it is often nice to see the results on the different platforms/interpreters, but not that much on the different Linux distributions. What about making the Sage tests run only conditional to the other tests? This would cut the number of concurrent jobs and the total number of jobs in the cases when there are test failures.

I also think that running on 5 versions of CPython on 4 platforms is a bit redundant. I would at a minimum reduce the number of CPython versions tested on Homebrew. Covering 3.7 and 3.11 should be enough.

It would be more interesting to test with different versions of Meson. Right now we support 0.63.3 and 0.64 and it would be nice to run the tests for both at least in a couple of configurations. I'm thinking CPython 3.7 and 3.11 on Windows and Linux. This would add another 4 jobs.

eli-schwartz: Note also that organizations share slots, so if PRs to the Meson repository are currently running lots of CI, then meson-python has to wait -- and also vice versa, which is how I noticed.

henryiii: IMO, first and last on all OS's + at least one for the ones in-between (usually on linux, or alternate linux and Windows), and 1-2 for each special platform (like homebrew) is nearly as good as having a filled out matrix, at about half the cost. It does slow down debugging a little bit, since if 3.7 and 3.11 differ, you don't know exactly where the change occurred. But if you are getting the tests done faster, it's worth it.

I thought there was a limit of 256 or 512 jobs scheduled for a workflow? But I've not hit it (140 or so is the highest I've gone), so not sure, and it could have been lifted.

rgommers: +1 for less Python versions from me too, for the reasons others have already given.

dnicolodi: > Note also that organizations share slots

This is unfortunate. I cut down the CI jobs, but we are still running:

- CPython 3.7 and 3.11, and PyPy 3.8 on Windows, macOS, and Linux: 9 jobs
- CPython 3.8, 3.9, 3.10, and PyPy 3.7 on Linux: 4 jobs
- CPython 3.9 in Cygwin: 1 job
- Pyston 3.8 on Linux: 1 job
- CPython 3.7 and 3.11 on Homebrew: 2 jobs
- a mypy job,
- the precommit tests

that's 19 jobs plus the Sage jobs. I'm temped to axe PyPy on Windows and macOS completely and run just one version of PyPy on Linux. I think there is much more value in testing different versions of Meson.

dnicolodi: Ok. Down to 16 jobs plut Sage now. I'll add the Meson versions tests in another PR.

dnicolodi: The drawback of triggering the Sage tests like this is that the jobs status is not shown on the PR. This is not very good. The only other way I know to have the Sage jobs run conditionally to the other tests is to have them defined in the same workflow. Ideas?

dnicolodi: Moved test matrix reorganization to #221

PR #198, opened 2022-11-09
TST: Fix test packages

dnicolodi:
In some of the test packages, the Python library is looked up incorrectly using dependency('python3') while the correct way is to use import('python').find_installation().dependency().

The correct lookup method could be used, but since Meson 0.63.0 python extension modules implicitly depend on the Python library and meson-python requires Meson 0.63.3. Just drop the explicit dependency.

### DIFF
--- a/tests/packages/link-against-local-lib/meson.build
+++ b/tests/packages/link-against-local-lib/meson.build
@@ -8,13 +8,11 @@ example_lib = shared_library(
     install: true,
 )
 
-py_dep = dependency('python3')
 py_mod = import('python')
 py = py_mod.find_installation()
 
 py.extension_module(
     'example', 'examplemod.c',
-    dependencies: py_dep,
     link_with: example_lib,
     install: true,
 )
--- a/tests/packages/purelib-and-platlib/meson.build
+++ b/tests/packages/purelib-and-platlib/meson.build
@@ -3,13 +3,11 @@ project(
     version: '1.0.0',
 )
 
-py_dep = dependency('python3')
 py_mod = import('python')
 py = py_mod.find_installation()
 
 py.install_sources('pure.py')
 py.extension_module(
     'plat', 'plat.c',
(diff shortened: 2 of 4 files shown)


dnicolodi:
Ubuntu tests failing because `pip` returns with process return code != 0 if it is not the latest version, apparently...

dnicolodi:
> Ubuntu tests failing because `pip` returns with process return code != 0 if it is not the latest version, apparently...

I take it back. This is not the case at all. Fixing the test project fixed a test that was marked as know failure. The output of `pytest` can be misleading sometimes...

dnicolodi:
Notably, this also fixes the Sage tests. There is only one failing test, and it fails because:
```
Checking whether SageMath should install SPKG python3...
checking whether any of bzip2 liblzma libffi is installed as or will be installed as SPKG... no
checking for python3 >= 3.8.0, < 3.11.0 with modules sqlite3, ctypes, math, hashlib, crypt, socket, zlib, distutils.core, ssl, ensurepip... 
checking for python3... /opt/conda/bin/python3
checking ... whether /opt/conda/bin/python3 is good... no, Python 3.11.0 is too recent
configure: error: the python3 selected using --with-python=python3 is not suitable
```
which seems something that needs to be fixes within the Sage tests setup. @mkoeppe Any ideas?

FFY00 (approved):
Ah, good catch! Thank you so much :blush:

dimpase:
Sagemath does not support Python 3.11 yet: https://trac.sagemath.org/ticket/33842

dnicolodi:
@dimpase Something must be off with the Sage test suite then: it tries to run with Python 3.11. I don't see anything in out test setup that selects Python 3.11

FFY00:
AFAIK the Python is provided by the environment, and the Python in conda-forge is on 3.11.

mkoeppe:
#200 fixes this

PR #199, opened 2022-11-09
TST: make venv test fixture work on Windows and with PyPy

dnicolodi:


FFY00:
Eh, I think it may be just be easier to use the `vitualenv` API.

dnicolodi:
> I think it may be just be easier to use the `virtualenv` API.

This now works fine on all platforms with CPython. It still fails on PyPy but I'm not sure `virtualenv` helps there. Everything is failing now because I'm trying to extend the test for the example project to other Python versions and platforms.

dnicolodi:
By the way, do we link to the example from anywhere in the documentation?

dnicolodi:
This turned into a more general cleanup of the tests, but it is now ready for review.

I reviewed all XFAIL and SKIP tests. Now, other than test skipped for genuinely unimplemented features on a given platform, some tests are skipped only:

- on Python 3.7 because the extension filenames do not allow to derive the ABI tag (fixed in #191)
- on Linux and macOS when running with PyPy because creating a venv fails.

I'm still investigating the venv issue.

dnicolodi:
Rebased. Let's see if this fixes the conda tests.

dnicolodi:
The venv issues are caused by nox. I removed the use of nox from the CI setup. I think not using it actually makes things easier. Please have a look.

.github/workflows/tests.yml
```diff
@@ -42,33 +42,15 @@ jobs:
...
-            nox --version
+      - name: Install
+        run: python -m pip --disable-pip-version-check install .[test]
 
       - name: Run tests
-        run: nox -v -s test-${{ env.BASE }}
+        run: >-
+          python -m pytest --showlocals -vv --cov
+          --cov-config setup.cfg
+          --cov-report=html:htmlcov
```
FFY00, line 52: This isn't needed in CI.

FFY00:
> The venv issues are caused by nox. I removed the use of nox from the CI setup. I think not using it actually makes things easier. Please have a look.

This is not optimal, but good enough for now.

FFY00 (approved):

.github/workflows/tests.yml
FFY00:
```suggestion
```

dnicolodi:
@FFY00 Can I merge (well, rebase and merge) my own PRs once they are approved? The readthedocs check is a requirement for merging but something is still off with it, thus I cannot merge this.

FFY00:
Yep. I will merge this one since the required checks are broken, I need to sort out the readthedocs situation.

PR #202, opened 2022-11-11
ENH: determine wheel tags by Python interpreter introspection & tests and CI improvements

dnicolodi:
Looking at the wheel tags generation code in `packaging` and in `wheel` it looks like it is full of cruft to support old interpreter versions. I tried to see how much effort it takes to have the most minimal compliant implementation. This for now only tests that my implementation returns the same values as the `packaging`-based one. Pushing it here just to have the tests run in CI.


dnicolodi:
The code turns out nice and simple. The most complex part is a workaround for some macOS SDK bug. It would be nice if more of this string fiddling would be implemented in the standard library, but it does not seem as fragile as I was fearing. @FFY00 It would be nice if `sysconfig` on Windows would expose `SOABI` too.

I went ahead and replaced the use of `packaging` code with this implementation for the generation of the tags. Please have a look. @eli-schwartz 

The tests failures on macOS are due to the fact that a version of `packaging` with the macOS SDK workaround has not been released yet. I'm not sure what the best solution is here.

tests/conftest.py
```diff
@@ -15,6 +16,24 @@
 import mesonpy
 
 
+def adjust_packaging_platform_tag(platform: str) -> str:
```
rgommers, line 19: Since `packaging` is no longer used, I assume the `manylinux` part of this is also no longer needed?

The change seems right to me; neither `meson-python` nor `packaging` can know whether or not the wheel being produced is manylinux compliant, and there should be no attempt to deal with that at all (that's for `auditwheel`).

tests/test_tags.py
```diff
@@ -1,75 +1,99 @@
...
-        ('cpython-37-x86_64-linux-gnu', 'cpython', '37', ('x86_64', 'linux', 'gnu'), 'cp37', 'cp37'),
-        ('cpython-310-x86_64-linux-gnu', 'cpython', '310', ('x86_64', 'linux', 'gnu'), 'cp310', 'cp310'),
-        ('cpython-310', 'cpython', '310', (), 'cp310', 'cp310'),
-        ('cp311-win_amd64', 'cpython', '311', ('win_amd64', ), 'cp311', 'cp311'),
-        ('cpython-311-win_amd64', 'cpython', '311', ('win_amd64', ), 'cp311', 'cp311'),
-        ('cpython-310-special', 'cpython', '310', ('special',), 'cp310', 'cp310'),
-        ('cpython-310-x86_64-linux-gnu', 'cpython', '310', ('x86_64', 'linux', 'gnu'), 'cp310', 'cp310'),
-        ('pypy39-pp73-x86_64-linux-gnu', 'pypy39', 'pp73', ('x86_64', 'linux', 'gnu'), 'pypy39_pp73', 'pp39'),
-        ('pypy39-pp73-win_amd64', 'pypy39', 'pp73', ('win_amd64', ), 'pypy39_pp73', 'pp39'),
-        ('pypy38-pp73-darwin', 'pypy38', 'pp73', ('darwin', ), 'pypy38_pp73', 'pp38'),
```
rgommers, line 49: These should be kept, right? We had a lot of trouble with wheel tags, so I'd like to make sure there are no regressions - that would make packages immediately uninstallable from source.

dnicolodi: This was testing the conversion from a extension module filename suffix to a wheel tag. This approach cannot work as the filename suffix does not carry enough information (and strictly speaking it is not mandatory for it to carry any). This patch set moves away from that approach. Thus this test is not necessary anymore and the code that is being tested does not exist anymore.

tests/conftest.py
dnicolodi: This is only used in the tests to compare what `meson-python` generates with what `packaging` generates. `packaging` generates the very specific `manylinux` platform tag while `meson-python` does not. Thus we need to adapt the former version of the tag to the latter. However, while I was writing this I realized that this does not handle `muslinux` tags, thus the tests are broken on platforms using that libc. Fixing it.

rgommers: Ah okay, got it. So we use `packaging` in the test suite to ensure that there's no silent mismatches that result in wheels we produce being uninstallable - that seems quite useful.

.github/workflows/tests.yml
```diff
@@ -60,3 +60,163 @@ jobs:
...
+        # some dependencies takes considerable time. Caching the built
+        # wheels speeds up the CI job quite a bit.
+        uses: actions/cache@v3
+        with:
+          path: ${{ steps.pip-cache-path.outputs.path }}
+          key: cygwin-pip-${{ github.sha }}
+          restore-keys: cygwin-pip-
+
+      - name: Install
+        run: python -m pip --disable-pip-version-check install .[test]
```
DWesl: The nox configuration also installs `pytest-github-actions-annotate-failures`; do you wish to install that as well?

.github/workflows/tests.yml
```diff
@@ -60,3 +60,167 @@ jobs:
...
+          path: ${{ steps.pip-cache-path.outputs.path }}
+          key: cygwin-pip-${{ github.sha }}
+          restore-keys: cygwin-pip-
+
+      - name: Install
+        run: |
+          ls -la /usr/share/python-wheels/ || true
+          python -m venv /tmp/venv1
+          python -m venv /tmp/venv2
+          python -m pip --disable-pip-version-check install .[test]
```
DWesl, line 122: It looks like the nox configuration would also install `pytest-github-actions-annotate-failures`; do you want to install that as well?

DWesl (commented):
The Cygwin CI configuration looks solid to me; I especially like using `/usr/bin/env` to make pytest's description of the test environment shorter.

.github/workflows/tests.yml
```diff
@@ -60,3 +60,162 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup Cygwin
+        uses: cygwin/cygwin-install-action@v2
+        with:
+          packages: >-
+            python39
+            python39-devel
+            python39-pip
```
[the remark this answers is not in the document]
dnicolodi: I put the Python version in the matrix mostly as a reminder of which Python version this job runs. I don't expect Cygwin to update the Python version they distribute anytime soon.

.github/workflows/tests.yml
```diff
@@ -60,3 +60,163 @@ jobs:
...
+
+      - name: Get pip cache path
+        id: pip-cache-path
+        run: echo "path=$(cygpath -w $(python -m pip cache dir))" >> $GITHUB_OUTPUT
+        shell: C:\cygwin\bin\env.exe CYGWIN_NOWINPATH=1 CHERE_INVOKING=1 C:\cygwin\bin\bash.exe -leo pipefail -o igncr {0}
+
+      - name: Restore cache
+        # Cygwin Python cannot use binary wheels from PyPI. Building
+        # some dependencies takes considerable time. Caching the built
+        # wheels speeds up the CI job quite a bit.
```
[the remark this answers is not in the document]
dnicolodi: In that case the wheels will be rebuilt. That's why weels are tagged with a platform specifier. I don't foresee any problem with this. The effort to put in place and maintain this is negligible and it cuts the job time in half, which is quite nice when debugging.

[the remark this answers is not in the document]
eli-schwartz: As long as it dramatically speeds up the CI at least one in five CI runs, it sounds like a huge win to me?

I'm quite sure that there will be cache hits more than 20% of the time though. :D

.github/workflows/tests.yml
dnicolodi: I dislike stuff that annotates the diffs on GitHub it makes reviewing them a pain. But I can reinstate this one if people like it.

dnicolodi:
All green except mysterious failures on Homebrew with Python 3.10 and 3.11. I'm convinced these are test setup issues, but debugging the issues with the Cygwin job drained my divination power for the day.

.github/workflows/tests.yml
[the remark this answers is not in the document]
DWesl: The other option would be to try testing on python 3.7 or 3.8, also currently available from Cygwin, and the reason the package is no longer named python3 and python3-pip.  I don't think that would turn up too many more bug, though.

eli-schwartz:
```
  brew reinstall python@3.11 || true
```

This command fails, but is told to report the job as successful regardless of whether it actually succeeded. I'm not sure what the point of this is supposed to be, other than to make errors less obvious. :)

...

It's a github actions bug, github actions has inherently broken homebrew: https://github.com/Homebrew/discussions/discussions/3895

.github/workflows/tests.yml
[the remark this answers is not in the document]
DWesl: My concern would be the new wheel don't get added to the cache; the next run continues to pull the old cache rather than getting a new version.  It seems the consensus is that the cache will be expired at least every couple of months, so this isn't really an issue.

.github/workflows/tests.yml
DWesl: If it just lets you jump to the start of the detailed error reports in the CI logs, I can see the use for that; if it mostly clutters the diff that's less useful.

I just found it as a side effect of checking that the tests didn't use any of the `GITHUB_*` environment variables, since I remembered seeing one recently.

.github/workflows/tests.yml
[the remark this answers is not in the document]
dnicolodi: The cache is keyed with the hash of the commit that triggered the CI job but retrieved with a less specific key in LRU fashion, thus this is not a concern. You can read the documentation for `actions/cache` to get the details.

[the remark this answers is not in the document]
DWesl: That works.

dnicolodi:
> ```
>   brew reinstall python@3.11 || true
> ```
> 
> This command fails, but is told to report the job as successful regardless of whether it actually succeeded. I'm not sure what the point of this is supposed to be, other than to make errors less obvious. :)

The `brew install` fails to link the package because there are already binaries from Python 2.7 in place of the destinations of the homebrew links for the unversioned binaries. The `brew link --overwrite` solves overwrites the existing binaries with the homebrew links. The other solution I played with was to `rm -f` the I'll make sure to stick a comment explaining what that does once I solve the remaining issues, if it is possible.

> It's a github actions bug, github actions has inherently broken homebrew:

Oh. Good to know. I'll read up about it.

eli-schwartz:
> The `brew install` fails to link the package because there are already binaries from Python 2.7 in place of the destinations of the homebrew links for the unversioned binaries.

Actually the GitHub Actions specific 3.11, not 2.7, but otherwise correct.

.github/workflows/tests.yml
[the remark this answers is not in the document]
dnicolodi: The Cygwin tests are interesting because of the peculiarities of the platform. I don't expect differences between Python versions. We already have many CI jobs and I'm more thinking or remove some elements of the test matrix rather than adding some more.

dnicolodi:
The issues with the Homebrew Python are due to this https://github.com/pypa/pip/issues/11539 I still don't know if it is due to a bug in pip or something weird in how Homebrew Python is setup. It is strange that I don't seem to encounter this error in my local setup, though.

mesonpy/_tags.py
```diff
@@ -1,130 +1,124 @@
...
+    debug = pymalloc = ''
+    if _get_config_var('Py_DEBUG', hasattr(sys, 'gettotalrefcount')):
+        debug = 'd'
+    if sys.version_info < (3, 8) and _get_config_var('WITH_PYMALLOC', True):
+        pymalloc = 'm'
+    return f'cp{version[0]}{version[1]}{debug}{pymalloc}'
+
+
+def get_abi_tag() -> str:
+    abi = sysconfig.get_config_var('SOABI')
```
FFY00, line 46: The `SOABI` isn't documented anywhere, so could be removed in future Python versions. While I am okay merging this as an initial implementation, granted that all of our current tests pass, we should work on figuring out a better way to do this.

Please add an `# XXX: ...` comment documenting this.

mesonpy/_tags.py
```diff
@@ -1,130 +1,124 @@
...
+    if not abi:
+        # CPython on Windows does not have a SOABI sysconfig variable.
+        assert sys.implementation.name == 'cpython'
```
FFY00, line 49: This should be a proper check, not an assertion. It may go haywire if the asserts are disabled and the implementation doesn't actually provide `SOABI`.

mesonpy/_tags.py
```diff
@@ -1,130 +1,124 @@
...
+
+    if major >= 11:
+        # For macOS reelases up to 10.15, the major version number is
+        # actually part of the OS name and the minor version is the
+        # actual OS release.  Starting with macOS 11, the major
+        # version number is the OS release and the minor version is
+        # the patch level.  Reset the patch level to zero.
+        minor = 0
+
+    if sys.maxsize <= 2**32:
```
FFY00, line 95: nitpick: Make this a constant.

FFY00 (approved):
I have some concerns regarding the reliance on `SOABI`, as presented below, but overall this PR looks good.

mesonpy/_tags.py
```diff
@@ -1,130 +1,124 @@
...
+def get_abi_tag() -> str:
+    abi = sysconfig.get_config_var('SOABI')
+    if not abi:
+        # CPython on Windows does not have a SOABI sysconfig variable.
+        assert sys.implementation.name == 'cpython'
+        return _get_cpython_abi()
+    if abi.startswith('cpython'):
+        return 'cp' + abi.split('-')[1]
+    if abi.startswith('pypy'):
+        return '_'.join(abi.split('-')[:2])
```
FFY00, line 54: No special handling is needed for PyPy AFACT. Am I missing something?

mesonpy/_tags.py
```diff
@@ -1,130 +1,124 @@
...
+    value = sysconfig.get_config_var(name)
+    if value is None:
+        return default
+    return value
+
+
+def _get_cpython_abi() -> str:
+    version = sys.version_info
+    debug = pymalloc = ''
+    if _get_config_var('Py_DEBUG', hasattr(sys, 'gettotalrefcount')):
```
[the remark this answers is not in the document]
dnicolodi: The idea is to rely on the value of `Py_DEBUG` if set (`sysconfig.get_config_var()` does not return Null) and fallback to the `hasattr` check otherwise. Thus open-coding this would be something like:
```
debug = _get_config_var('Py_DEBUG')
if debug is None:
    debug = hasattr(sys, 'gettotalrefcount')
```
Because we have two of these checks I decided to have the tiny helper function. Happy to open code it if you prefer.

mesonpy/_tags.py
dnicolodi: `packaging` relies on the same configuration variable. As we need to be bug-compatible with `packaging` I figure this is okay. AFAIK, none of the variables available with `sysconfig.get_config_var()` is documented. Should I add the same comment everywhere we use one?

mesonpy/_tags.py
dnicolodi: The only name for the constant I can think about is `_2_POWER_32`, which is maybe more cryptic than the literal value. Do you have a specific name that would be clearer than the value itself?

mesonpy/_tags.py
dnicolodi: I agree. However, the special handling is in `packaging` and I decided to keep it. I'm happy to remove it if we are positive that it is not needed.

mesonpy/_tags.py
DWesl: `sys.maxsize < 2 ** 32` would be equivalent to `sys.maxsize <= UINT32_MAX`; since this isn't Windows, that's likely the same as `sys.maxsize <= ULONG_MAX`.  I suspect this is being used as a delineation between 32-bit Python and 64-bit Python, so you could also choose a name based on that (`PYTHON_32BIT_SYS_MAXSIZE_CUTOFF`, maybe)

mesonpy/_tags.py
```diff
@@ -1,130 +1,124 @@
...
+        # actual OS release.  Starting with macOS 11, the major
+        # version number is the OS release and the minor version is
+        # the patch level.  Reset the patch level to zero.
+        minor = 0
+
+    if sys.maxsize <= 2**32:
+        # 32-bit Python running on a 64-bit kernel.
+        if arch.startswith('ppc'):
+            arch = 'ppc'
+        arch = 'i386'
```
DWesl, line 99: It looks like `arch = 'i386'` will overwrite `arch = 'ppc'`; should that be before the conditional, or guarded by `else:`?

mesonpy/_tags.py
eli-schwartz: Maybe the suggestion was to use `if sys.maxsize < 4294967296` but I'm not sure whether that's all that much of an advantage...

FFY00: I mean, make `sys.maxsize <= 2**32` a constant, not `2**3`. Something like `_32_BIT_INTERPRETER`, just like `packaging` does.

mesonpy/_tags.py
FFY00: It does, but only for the non CPython tags. The normalized version of `SOABI` is used as the full ABI tag as a fallback, we can do the same.

IIRC we only use `SOABI` in one other place, and it has a fallback value.

mesonpy/_tags.py
FFY00: Where? The correspondent code AFAIK is

https://github.com/pypa/packaging/blob/606c71acce93d04e778cdbdea16231b36d6b870f/packaging/tags.py#L496-L499

Which only has special handling for CPython, as it's the only needed. For other implementations, the value falls back to the normalized version of `SOABI`.

FFY00: In fact, we won't generate the same tags as `packaging` if PyPy adds another section to `SOABI`, as this code is only taking the first two.

mesonpy/_tags.py
dnicolodi: We use other variables from `sysconfig.get_config_var()` all of them are undocumented. Aren't they?

mesonpy/_tags.py
dnicolodi: I'll do that but I really don't see the advantage.

mesonpy/_tags.py
henryiii: We should not be using SOABI. It is unreliable. It's better to use EXT_SUFFIX, and parse from there. Packaging is doing this in https://github.com/pypa/packaging/pull/607. Because packaging's usage of this has been wrong for so long, other implementiaonts might have broken `SOABI`'s too, just to make packing work. While `EXT_SUFFIX` is much more likely to be correct, as it's used in other places. Both variables are documented in [PEP 3149](https://peps.python.org/pep-3149/), and are supposed to be identical, minus the leading dot and extension. Also, Python itself recommends `EXT_SUFFIX` if you try to use `SO`.

dnicolodi: Sorry, I meant `wheel` https://github.com/pypa/wheel/blob/814c2efe8e40051039c5a6de6945e04ecdd162ee/src/wheel/bdist_wheel.py#L96-L99 I've been staring at all these implementations for too long. I agree that the special handling is weird. I'll remove it.

mesonpy/_tags.py
dnicolodi: Good catch! I guess I refactored this code around and I had this in a function with a `return` before.

mesonpy/_tags.py
henryiii: It is documented in [PEP 3149](https://peps.python.org/pep-3149/) and should be available on all CPythons, along with `EXT_SUFFIX`. As I mentioned on the other thread, `EXT_SUFFIX` is more reliable due to a long standing bug in packaging expecting a broken `SOABI` because PyPy has a broken SOABI. You can get a little backstory in https://foss.heptapod.net/pypy/pypy/-/issues/3816 or https://gitlab.kitware.com/cmake/cmake/-/issues/24121.

mesonpy/_tags.py
dnicolodi: @henryiii I think I don't understand. I cannot parse what you are trying to say with this sentence

> Because packaging's usage of this has been wrong for so long, other implementations might have broken `SOABI`'s too, just to make packing work (and therefor pip, etc).

Can you please try to rephrase it?

If PEP3149 documents both variables to carry the same information, if they would get out of sync, that would be a bug of the Python implementation. Wouldn't it? Can we find a Python implementation that has the wrong `SOABI` but the right `EXT_SUFFIX`?

henryiii: > Can we find a Python implementation that has the wrong `SOABI` but the right `EXT_SUFFIX`?

Absolutely: PyPy. PyPy reports the wrong SOABI, it's missing the platform information. If you use it, you make an extension that can't be loaded (since EXT_SUFFIX is used for loading). When Packaging was adapted to use SOABI, it special cases CPython, so the "general" code was based on PyPy's mistake, so it assumes SOABI is wrong. So other implementations might report wrong SOABI's because it makes packing work. It's being fixed in packaging right now (to use EXT_SUFFIX and assume it is right), but until the current Pip is old enough to not matter, PyPy is stuck reporting the wrong SOABI.

If you report the wrong `EXT_SUFFIX`, setuptools and probably other things break. If you report the _right_ `SOABI` but you are not CPython, you might not be loadable by Pip currently (is being fixed in that linked PR).

henryiii: According to the PEP, `assert sysconfig.get_config_var('EXT_SUFFIX').split(".", 1)[0].lstrip(".") == sysconfig.get_config_var('SOABI')`, but it doesn't for PyPy, so using the former is best. For just the ABI, yes, that would suffice.

What's wrong with CPython 3.7?

```console
$ docker run --rm -it quay.io/pypa/manylinux2014_x86_64
[root@f00ca105d347 /]# python3.7
Python 3.7.14 (default, Sep 12 2022, 15:16:12)
[GCC 10.2.1 20210130 (Red Hat 10.2.1-11)] on linux
>>> import sysconfg
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'sysconfg'
>>> import sysconfig
>>> sysconfig.get_config_var("EXT_SUFFIX")
'.cpython-37m-x86_64-linux-gnu.so'
>>> sysconfig.get_config_var("SOABI")
'cpython-37m-x86_64-linux-gnu'
>>>
[root@f00ca105d347 /]# pypy3.7
Python 3.7.13 (7e0ae751533460d5f89f3ac48ce366d8642d1db5, Mar 29 2022, 06:03:31)
[PyPy 7.3.9 with GCC 10.2.1 20210130 (Red Hat 10.2.1-11)] on linux
>>>> import sysconfig
>>>> sysconfig.get_config_var("EXT_SUFFIX")
'.pypy37-pp73-x86_64-linux-gnu.so'
>>>> sysconfig.get_config_var("SOABI")
'pypy37-pp73'
```

eli-schwartz: My recollection that CPython started adding more complex EXT_SUFFIX instead of just ".so", as part of the migration from 2.7 to the entire 3.x series, and that change, like most 3.x changes, happened "sometime during 3.0, 3.1, 3.2, or 3.3, although most people aren't sure which of them because it took several releases before the general populace considered it to be a viable replacement for 2.7".

dnicolodi: > Was it an early patch release of 3.7? Or something else?

CPython 3.7 on Windows just reports the extension. See #190.

henryiii: PEP 3149 was accepted in 2010, so that's what I would have figured too, which is why I was asking about why a fallback was needed for 3.7.

dnicolodi: See https://docs.python.org/3.8/whatsnew/changelog.html#python-3-8-7-final and https://github.com/python/cpython/issues/84006

henryiii: Ah, you said `.so`. Okay, sounds like a fallback is needed for Windows and Python 3.7, then.

dnicolodi: Yes, my bad. Sorry.

dnicolodi: Nope. Mayne I misunderstood the proposal, but looking at `EXT_SUFFIX` and assuming is correct does not fly.

CPython
```
>>> sysconfig.get_config_var('EXT_SUFFIX')
'.cpython-310-darwin.so'
>>> next(packaging.tags.sys_tags()).abi
'cp310'
```

PyPy:
```
>>>> next(packaging.tags.sys_tags()).abi
'pypy37_pp73'
>>>> sysconfig.get_config_var('EXT_SUFFIX')
'.pypy37-pp73-darwin.so'
```

Pyston:
```
>>> print(sysconfig.get_config_var("EXT_SUFFIX"))
'.pyston-23-x86_64-linux-gnu.so'
>>> next(packaging.tags.sys_tags()).abi
'pyston_23_x86_64_linux_gnu'
```

Also, this is the beautiful code currently proposed for `packaging` in 
```python
def _generic_abi() -> List[str]:
    """Return the ABI tag based on EXT_SUFFIX"""
    # ext_suffix is something like this. We want to keep the parts which are
    # related to the ABI and remove the parts which are related to the
    # platform:
    # - linux:   '.cpython-310-x86_64-linux-gnu.so' => cp310
    # - mac:     '.cpython-310-darwin.so'           => cp310
    # - win:     '.cp310-win_amd64.pyd'             => cp310
    # - pypy:    '.pypy38-pp73-x86_64-linux-gnu.so' => pypy38-pp73
    # - graalpy: '.graalpy-38-native-x86_64-darwin.dylib'
    #                                               => graalpy-38-native

    ext_suffix = _get_config_var("EXT_SUFFIX", warn=True)
    assert ext_suffix[0] == "."
    _, soabi, ext = ext_suffix.split(".")
    if soabi.startswith("cpython"):
        abi = "cp" + soabi.split("-")[1]
    elif soabi.startswith("pypy"):
        abi = "-".join(soabi.split("-")[:2])
    elif soabi.startswith("graalpy"):
        abi = "-".join(soabi.split("-")[:3])
    elif soabi:
        abi = soabi
    else:
        return []
    return [_normalize_string(abi)]
```
which is many things but not generic. Lovely.

dnicolodi: I'm going to stick to `SOABI` just because it takes lees code. Is really a `sysconfig.get_interpreter_abi()` function too much to ask?

henryiii: What is the result of SOABI in those cases? It should be `cpython-310-darwin`, `pypy37-pp73-darwin` (but actually is `pypy37-pp73`), and `pyston-23-x86_64-linux-gnu`, correct?

henryiii: By "assuming it is correct", I mean that it's the correct extension suffix. `pypy37-pp73.so` for example is not a valid suffix, pypy can't load it. I didn't mean the wheel abi name is "better" from `EXT_SUFFIX`, I just meant `SOABI` is sometimes wrong, while `EXT_SUFFIX` is correct. You still need to transform it for wheel tag names in either case - and it's exactly the same transform unless SOABI is buggy. And if SOABI is buggy, the implementation will hopefully try to change it (as PyPy will in a year or two).

mattip: > I don't understand how going from special casing PyPy to special casing PyPy and GrallVM Python while still getting it wrong for Pyston is an improvement

The pyston developers are welcome to add to that code or provide a comment as to what needs to be added. In a perfect world, sysconfig would provide `get_python_tag()`, `get_abi_tag()` and `get_platform_tag()`, but it doesn't, so here we are. The goal of the PR you mentioned is to fix a possible issue when PyPy fixes its SOABI value at some time in the future. It was not intended to make the code generic (which is impossible as long as packaging needs to be backward-compatible) or to cover all implementations (that is up to other projects to contribute where needed).

Both pip and wheel now use `packaging/tags.py`. It is a shame this functionality is replicated in other places.

dnicolodi: > The pyston developers are welcome to add to that code or provide a comment as to what needs to be added.

How `packaging` generates tags for Pyston cannot be changed for compatibility: compatibility of existing wheel releases with systems upgraded to an eventual future `packaging` and compatibility of systems with an old packaging with wheels generated on a system with an eventual future `packaging`.

> Both pip and wheel now use `packaging/tags.py`. It is a shame this functionality is replicated in other places.

Should `packaging` start to provide `get_python_tag()`, `get_abi_tag()` and `get_platform_tag()` functions, we may consider it. Till then the API is inconvenient to use. The algorithm is set in stone for all existing Python implementations (see above) and new Python implementations do not yet pup up like mushrooms, so keeping it in sync is not much work at all.

dnicolodi: > What is the result of `SOABI` in those cases? It should be `cpython-310-darwin`, `pypy37-pp73-darwin` (but actually is `pypy37-pp73`), and `pyston-23-x86_64-linux-gnu`, correct?

Correct.

dnicolodi:
@FFY00 I think this is now ready. After some serious debugging effort the CI should be happy now. Please take a look.

I would appreciate if we could cut a release after merging this, unless there are other important things pending. I would like to go back to release wheels for Python 3.7, which is the thing that originally prompted this work.

FFY00 (approved):
I think this looks good. Huge thanks for working on this @dnicolodi! :tada:

henryiii:
> I would appreciate if we could cut a release after merging this

I'm happy to finish off #175 as soon as this goes if if you'd like that too - that's quite important for building on non-wheel-supported platforms.

henryiii:
Great work, by the way! :)

FFY00:
We have a couple things ready that are blocked by fixes on the PR, after we get this in we can merger all the blocked stuff and make a release.

dnicolodi:
The system ninja work would be very nice to have indeed.

henryiii:
This "passes" but the pytest -> test name change messes up the require checks. Personally, I prefer having a "pass" check that depends on all the other checks, and then only require that (once per workflow, unfortunately, but still much simpler than maintaining a bunch of checks, and if you change the checks all the old merged PRs don't change to orange dots).

dnicolodi:
Yeah. Sorry, I didn't realize that the rename would have caused this issue. However, also the readthedocs job is stuck, so this is not the only thing preventing this from being merged automatically. Now that all the CI jobs are fixed, can't we simply require that all jobs succeed? I'm not sure why the status of old PRs should change changing the definition of the checks for future PRs. It seems a GitHub bug to me.

PR #206, opened 2022-11-14
Fix ABI tag detection logic to only check file basename

jcrist:
Previously this was running the regexes on the full file path, which would mistakenly match files like `/full/path/to/.build-dir/subdir/lib.so` with `build-dir/subdir/lib` as the tag rather than correctly identifying `lib.so` as not a python native extension. AFAICT the intention here was to run them on the file basename alone.

I've tested this works locally, but haven't added tests in mesonpy yet. Happy to do so with some direction as to where they should go.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -327,9 +327,9 @@ class _WheelBuilder():
             platform_ = '-'.join(parts)
         return platform_.replace('-', '_').replace('.', '_')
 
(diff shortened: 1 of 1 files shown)


dnicolodi:
Should already be fixes in #202

jcrist:
Thanks, I can confirm that #202 does indeed fix my problem. Closing.

PR #209, opened 2022-11-16
TST: adding isolated build test

henryiii:
Pulled out of 175.

### DIFF
--- a/tests/test_pep518.py
+++ b/tests/test_pep518.py
@@ -1,0 +1,24 @@
+import subprocess
+import sys
+
+import pytest
+
+from .conftest import cd_package, in_git_repo_context
+
+
+@pytest.mark.parametrize(
+    ('package'),
+    [
+        'purelib-and-platlib',
+    ]
+)
+@pytest.mark.parametrize(
+    'build_arg', ['', '--wheel'], ids=['sdist_to_wheel', 'wheel_directly']
+)
+@pytest.mark.xfail(sys.platform.startswith('cygwin'), reason='ninja pip package not available on cygwin', strict=True)
+def test_pep518(pep518, package, build_arg, tmp_path):
+    dist = tmp_path / 'dist'
+
+    with cd_package(package) as package_dir, in_git_repo_context():
+        build_args = [build_arg] if build_arg else []
+        subprocess.run([sys.executable, '-m', 'build', f'--outdir={dist}', *build_args], cwd=package_dir, check=True)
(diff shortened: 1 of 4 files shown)


henryiii:
Is there a better test than `scipy-like`? I think part of the problem is that this is getting an SDist of something that needs to build (with setuptools, you can see it trying to get setuptools on Cygwin), like maybe NumPy or something.

dnicolodi:
`scipy-like` requires Cython to build. As you cannot use the published wheel for Cython and Cython uses setuptools as build system, you need to get setuptools too. IIRC I already expressed dubs regarding choosing this particular test package. `purelib-and-platlib` may be a better choice.

tests/test_pep518.py
```diff
@@ -0,0 +1,29 @@
...
+    'build_arg', ['', '--wheel'], ids=['sdist_to_wheel', 'wheel_directly']
+)
+def test_pep518(pep518, package, build_arg, tmp_path):
+    dist = tmp_path / 'dist'
+
+    try:
+        import build  # noqa: F401
+    except ModuleNotFoundError:
+        print('Installing build for testing')
+        subprocess.run([sys.executable, '-m', 'pip', 'install', 'build'], check=True)
```
dnicolodi, line 25: Please don't do this. I don't want the tests to touch my Python environment. You added `build` to the extra-requires and it is sufficient.

henryiii: Oops, I thought someone else added this, and I see I did - it was just originally in a venv (where this made sense), and now it's not. Will remove.

tests/test_pep518.py
```diff
@@ -0,0 +1,24 @@
...
+@pytest.mark.parametrize(
+    ('package'),
+    [
+        'purelib-and-platlib',
+    ]
+)
+@pytest.mark.parametrize(
+    'build_arg', ['', '--wheel'], ids=['sdist_to_wheel', 'wheel_directly']
+)
+@pytest.mark.xfail(sys.platform.startswith('cygwin'), reason='ninja pip package not available on cygwin', strict=True)
```
henryiii, line 18: FYI, this fails on Cygwin in this branch (no ninja package), and passes in #175 - it's great to see this test does exactly what is supposed to do! :)

henryiii:
> You added build to the extra-requires and it is sufficient.

Not for Sage CI, actually. Hopefully I fixed it by adding python-build.

henryiii:
`meson dist` is still broken. It's building meson-python successfully from SDist, it's just trying to make this package (I'm guessing outside of a git repo) and breaking. Is there a way to use `meson dist` outside a git repo? Actually, I bet `in_git_repo_context()` has something to do with it. I'll look into it tomorrow.

eli-schwartz:
> Is there a way to use `meson dist` outside a git repo?

No, because `meson dist` internally runs `git archive`, plus additional handling for subprojects (which either use git and are treated the same, or else don't use git and are copied entirely on the assumption that `subprojects/*/` is effectively read-only).

PR #210, opened 2022-11-16
CI: add pass jobs for both workflows

henryiii:
This should make it easier to work with CI required jobs. There are now just three required jobs: `checks-pass`, `sage-pass` and `tests-pass`.


FFY00 (approved):
Thanks!

PR #211, opened 2022-11-16
MAINT: move config to pyproject.toml

henryiii:
`setup.cfg` is a setuptools specific file - we can move this over to `pyproject.toml`, except for flake8. Still the same total number of files.

Added a tiny bit to the pytest config. If all goes well, I'd also like to turn warnings into errors, too.


henryiii:
Rebased.

FFY00 (approved):
Looks good. It's a shame flake8 still doesn't support having the configuration in `pyproject.toml`.

PR #212, opened 2022-11-17
TST: simplify pep518 test setup

dnicolodi:


dnicolodi:
@henryiii I took the freedom to tweak the pep518 test setup in the way I had in mind. Please have a look. I think this is easier. I still don't understand why `meson-python` does not list `wheel` as a dependency, given that it requires it. For #175, once you remove `ninja` from the `meson-python` dependencies, we need to verify if it will still be placed in the`wheelhouse` because it is a build dependency for `meson-python` or not. For doing that, I simply place a breakpoint at the end of `pep518_wheelhouse()` and list the content of the `wheelhouse` with `os.listdir()`.

rgommers:
> @henryiii I took the freedom to tweak the pep518 test setup in the way I had in mind. Please have a look. I think this is easier. I still don't understand why `meson-python` does not list `wheel` as a dependency, given that it requires it

Agreed, this is not great. The intent is micro-optimization of not installing things in the build env when they are not needed for isolated builds. However, that breaks non-isolated builds - which the kind of projects that are moving to Meson need all the time. So it is more harmful than helpful. 

Maybe let's move that to a separate PR or issue?

dnicolodi:
That makes sense. We may be moving away from using `wheel` because of the issue with the promise of a stable API or the lack of it  (and because we may want to merge `meson-python` into Meson at some point) thus the problem may go away I didn't mean to resolve this issue in this PR or in #175.

tests/test_pep518.py
```diff
@@ -15,10 +15,9 @@
 @pytest.mark.parametrize(
     'build_arg', ['', '--wheel'], ids=['sdist_to_wheel', 'wheel_directly']
 )
-@pytest.mark.xfail(sys.platform.startswith('cygwin'), reason='ninja pip package not available on cygwin', strict=True)
```
henryiii, line 18: Why is this removed? The point of the test is to test that this breaks - that is, ninja's not available from PyPI on cygwin. So if this passes on Cygwin before #175 goes in, the test doesn't work. That's why it was a strict xfail.

tests/test_pep518.py
```diff
@@ -15,10 +15,9 @@
...
 def test_pep518(pep518, package, build_arg, tmp_path):
     dist = tmp_path / 'dist'
 
-    with cd_package(package) as package_dir, in_git_repo_context():
+    with cd_package(package), in_git_repo_context():
         build_args = [build_arg] if build_arg else []
-        subprocess.run([sys.executable, '-m', 'build', f'--outdir={dist}', *build_args], cwd=package_dir, check=True)
+        subprocess.check_call([sys.executable, '-m', 'build', '--outdir', str(dist), *build_args])
```
henryiii, line 23: Why are you replying `run` with `check_call`? `check_call` and `check_output` are the old, pre Python 3.5 API, and the Python docs recommend `run` instead. The rest of the calls should be changed to runs long-term, I think?

tests/conftest.py
```diff
@@ -126,47 +126,17 @@ def fixture(tmp_dir_session):
...
-            'tomli',
-            'typing-extensions',
-            'wheel',
-        ],
-        check=True,
-    )
+    meson_python = str(package_dir.parent.parent)
+    # Populate wheelhouse with wheel for the following packages and
+    # their dependencies.  Wheels are downloaded from PyPI or built
+    # from the source distribution as needed.  Sources or wheels in
```
henryiii, line 132: Ahh, it's building ninja from source on Cygwin here, that's why it's passing below.

tests/test_pep518.py
henryiii: Okay, now I know why it's passing. That's okay, I think - then we can delete `ninja` from the wheelhouse manually for #175. The only think that lightly bothers me is the `check_call` change, then.

henryiii:
Listing wheel as a dependency until recently pulled in setuptools.

tests/test_pep518.py
dnicolodi: Oh, sorry, I saw that the rest of the code uses the "old" API and I assumed it is the recommended one. Switching it back and probably adding a commit to clean up the other call sites too.

tests/test_pep518.py
dnicolodi: I think that if we remove `ninja` from `meson-python`'s `install_require` list it will not be put in the wheelhouse anymore. But the documentation is not crystal clear on this. I'll have to check.

dnicolodi: I confirm that once `ninja` is removed from meson-python's `dependencies` it is not added to the wheelhouse anymore.

PR #213, opened 2022-11-17
MAINT: rename the pyproject.toml tool config table to meson-python

FFY00:

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -625,11 +625,11 @@ class Project():
 
     def _get_config_key(self, key: str) -> Any:
         value: Any = self._config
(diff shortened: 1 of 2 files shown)


lithomas1:
Is the build backend name still going to be ``mesonpy``? It's probably best to rename everything to ``meson-python`` if possible to avoid confusion.

rgommers:
We can't really change the package name or import name without breaking backwards compatibility. As long as `mesonpy` is only used in exactly that one line of code (`build-backend = `), then it shouldn't be much of a problem. Import names can't have dashes in them, so it's defensible. I guess this is like scikit-learn vs. sklearn

rgommers (approved):
Thanks for following up on this @FFY00

PR #214, opened 2022-11-18
TST: adding marker and stricter warning checks

henryiii:
Adding a marker for the isolated build testing. You can turn off this with `-m "not isolated"` - might be useful for conda-forge, where isolated builds don't matter.

Making the tests a bit stricter, checking for warnings.


henryiii:
Nice, only warning made error is `DeprecationWarning: check_home argument is deprecated and ignored.` - which I fixed a couple of days after CPython 3.11.0 shipped. Since it's in tests and just CPython triggering a warning on itself, should be fine to just to hide.

tests/conftest.py
```diff
@@ -70,7 +71,10 @@ def tmp_dir_session(tmpdir_factory):
 
 class VEnv(EnvBuilder):
     def __init__(self, env_dir):
-        super().__init__(with_pip=True)
+        with warnings.catch_warnings():
+            if sys.version_info[:3] == (3, 11, 0):
+                warnings.filterwarnings('ignore', 'check_home argument is deprecated and ignored.', DeprecationWarning)
```
dnicolodi, line 77: Please add a comment pointing to a bug report or something detailing why this is necessary and when it can be removed.

pyproject.toml
```diff
@@ -71,16 +71,23 @@ strict = true
 lines_between_types = 1
 lines_after_imports = 2
 multi_line_output = 5
-known_first_party = "mesonpy"
+known_first_party = 'mesonpy'
```
dnicolodi, line 74: Can you please stick to the existing style for string quotes in an existing file? Having the quoting switch forth and back between the allowed styles makes the diff harder to parse.

pyproject.toml
```diff
@@ -71,16 +71,23 @@ strict = true
...
+log_cli_level = 'info'
+norecursedirs = 'tests/packages/*'
+testpaths = ['tests']
+xfail_strict = true
+filterwarnings = [
+  'error',
+]
+markers = [
+  'isolated: Isolated builds',
+]
```
dnicolodi, line 93: I don't see the point of adding more undiscoverable pytest magic. The tests run just fine on conda and shaving off a couple of seconds from the test run does not seem worth the trouble.

dnicolodi (changes requested):
The stricter warnings check may be worth adding, although  having to silence warnings triggered by the Python standard library is very ugly. The test marking much less so, IMO.

tests/conftest.py
FFY00: It was added in https://github.com/python/cpython/pull/92899. No set ETA on when it will be removed, but IIRC the policy is to have it deprecated for at least 2 versions, I am tempted to remove it then.

dnicolodi: Thanks. But what I would like to have is a note that says when the warning suppression can be removed. Most likely this will be when 3.11.1 will be released, but having a pointer to the commit where the code triggering the warning in the standard lib has been fixed would be best. I think that would be https://github.com/python/cpython/pull/98743

henryiii: Yes, that's it. Only warns on 3.11.0 exactly, 3.11.1 should be fine.

Okay to squelch this some other way if you prefer, I just copied the solution we used in build, but in build, it was a user-facing warning. Here it's in the tests, so I could use a pytest method of ignoring warnings if you prefer.

pyproject.toml
henryiii: The existing style was single quotes. I didn't realize that, so in #211 I used double quotes (probably because of this line - the only one in the file that was double quotes, but it was at the bottom where I was working). I was just trying to get the file back into a consistent format (even if it's the opposite of what I usually do, I usually follow black).

Eventually there should be a good autoformatter and this will never come up.

pyproject.toml
henryiii: To the best of my knowledge, this is a new test and has never been run on conda? When I've added similar tests to conda in the past, they generally don't work, and that's okay, since you don't do isolated builds on conda in the first place. If it does work, there's no harm done, it just adds a bit of organization.

This is not "undiscoverable", you can run `pytest --markers` and see all the markers, with descriptions - including the ones like `parametrize` and `skipif` which are already being used. Normal users shouldn't have to use this, as well, it just makes it easy for conda and maybe other packagers to skip this if they don't work.

In fact, we could trade `pep518` as a fixture for `pytest.mark.isolated`, if you want - marking it with isolated could auto-add the fixture (the fixture isn't used inside the test body, so it's fairly natural to do so).

I just changed the marking style a little for this iteration, and can change more later if you still don't like it.

pyproject.toml
henryiii: I removed this one line's change, but kept the pytest normalization to single quotes - let me know if you'd like me to leave that double, too.

dnicolodi: Sorry, I didn't realize. Please segregate coding style adjustments to their own commits. Among other things, this avoid this kind misunderstandings to occur.

pyproject.toml
dnicolodi: Tests are run on conda. This PR turns on warnings that break another test I recently added, however, the `test_pep518.py` tests run just fine there.

henryiii: Where are they run on conda? I'm fine to drop this then.

dnicolodi: See the "sage / linux / linux (conda-forge, minimal-develop)" job.

pyproject.toml
henryiii: Split this up.

henryiii:
Made changes, and updated history to be one commit per change. Also updated tmpdir to tmppath, since I noticed that elsewhere being `py.path` instead of `pathlib` - the first time I wrote this fixture, I wasn't able to use tmppath, but here it's fine.

tests/conftest.py
```diff
@@ -132,8 +132,8 @@ def disable_pip_version_check():
 
 
 @pytest.fixture(scope='session')
-def pep518_wheelhouse(tmpdir_factory):
-    wheelhouse = tmpdir_factory.mktemp('wheelhouse')
+def pep518_wheelhouse(tmp_path_factory):
+    wheelhouse = tmp_path_factory.mktemp('wheelhouse')
```
dnicolodi, line 142: Can you please check if we make use of `tmpdir_factory` anywhere else? I have the feeling that I cargo-culted this fixture elsewhere...

henryiii: Done, got rid of all usage of (pytest) tmpdir.

dnicolodi: Thank you!

Issue #192, opened 2022-11-03
Usage of private APIs in wheel


henryiii:
Currently, meson-python uses the `wheel` package, which does not have a public API (the author has been warning everyone of this for years). The current API is going to be removed in https://github.com/pypa/wheel/pull/472. Solutions:

The removal will come with a new public API, so once that gets merged, this could be set up to use that. Maybe adding a limit would be in order, too, until that happens (not fond of limits, but might be in order here, especially as it would only be in the build wheel requires hook).

The pypa/distlib library has a public Wheel interface (this is what I've been using in scikit-build-core to be future-proof for the incoming wheel change).

And of course, it can be hand rolled too.

dnicolodi:
I love the constant churn around the python packaging specification and tools. I had a look at `distlib` and its API is awful. The public API for the `wheel` package has been in the works since 2018. I think the wise thing is to wait and see how things pan out.

rgommers:
Ah yes, fun. I agree we need an upper bound here, to avoid building from source breaking at some arbitrary time whenever `wheel` removes this private API. 

> And of course, it can be hand rolled too.

Not a fan of that - the alternative to an upper bound is to vendor an existing implementation, not write yet-another-one.

FFY00:
We just need something that subclasses `zipfile.ZipFile` and logs the hashes of the written files to the metadata, I think. So, we could just extract that from `wheel` and provide it in a new package while they design a public API.

PR #215, opened 2022-11-18
ENH: drop dependency on wheel

dnicolodi:
The wheel package does not offer a public API.  Implement a minimal wheel file format writer to replace the use of the wheel package. This also reduces the number of external dependencies and will help a future integration of meson-python into Meson.  The produced wheel files are validated in the test suite inspecting them with the wheel package reader.

Fixes https://github.com/mesonbuild/meson-python/issues/192.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -46,13 +46,13 @@ import mesonpy._compat
 import mesonpy._elf
 import mesonpy._tags
 import mesonpy._util
+import mesonpy._wheelfile
 
 from mesonpy._compat import Collection, Iterator, Literal, Mapping, Path
 
 
 if typing.TYPE_CHECKING:  # pragma: no cover
     import pyproject_metadata  # noqa: F401
-    import wheel.wheelfile  # noqa: F401
 
 
 if sys.version_info >= (3, 8):
@@ -82,7 +82,6 @@ class _depstr:
     need at runtime. Having them in one place makes it easier to update.
     """
     patchelf = 'patchelf >= 0.11.0'
-    wheel = 'wheel >= 0.36.0'  # noqa: F811
     ninja = f'ninja >= {_NINJA_REQUIRED_VERSION}'
 
 
@@ -455,7 +454,7 @@ class _WheelBuilder():
 
     def _install_path(
         self,
-        wheel_file: wheel.wheelfile.WheelFile,  # type: ignore[name-defined]
+        wheel_file: mesonpy._wheelfile.WheelFile,
         counter: mesonpy._util.CLICounter,
         origin: Path,
         destination: pathlib.Path,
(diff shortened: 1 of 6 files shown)


mesonpy/_wheelfile.py
```diff
@@ -0,0 +1,99 @@
...
+    def timestamp(mtime: Optional[float] = None) -> Tuple[int, int, int, int, int, int]:
+        timestamp = int(os.environ.get('SOURCE_DATE_EPOCH', mtime or time.time()))
+        timestamp = max(timestamp, MIN_TIMESTAMP)
+        return time.gmtime(timestamp)[0:6]
+
+    @staticmethod
+    def hash(data: bytes) -> str:
+        return 'sha256=' + _b64encode(hashlib.sha256(data).digest()).decode('ascii')
+
+    def writestr(self, zinfo_or_arcname: Union[str, zipfile.ZipInfo], data: bytes) -> None:
```
dnicolodi, line 48: `mypy` is not happy with this type annotation
```
mesonpy/__init__.py:505: error: Argument 2 to "writestr" of "WheelFile" has incompatible type "_lru_cache_wrapper[bytes]"; expected "bytes"  [arg-type]
```
but my Python typing experience is not enough to decide what's the correct fix. Ideas?

mesonpy/_wheelfile.py
```diff
@@ -0,0 +1,99 @@
...
+
+class WheelFile:
+    """Implement the wheel package binary distribution format.
+
+    https://packaging.python.org/en/latest/specifications/binary-distribution-format/
+    """
+    def __new__(cls, filename: Path, mode: str = 'r', compression: int = zipfile.ZIP_DEFLATED) -> 'WheelFile':
+        if mode == 'w':
+            return super().__new__(WheelFileWriter)
+        raise NotImplementedError
```
dnicolodi, line 36: I decided for this indirection to keep open the possibility for implementing a `WheelFileReader` too. I haven't investigated the possibility of modifying a wheel file in place. Maybe all this is over engineered and can be simplified.

mesonpy/_wheelfile.py
dnicolodi: I solved this in the second commit in this PR. I think that this is a bug in `mypy` but the code in `meson-python` was a bit unusual without reason, so I changed it.

mesonpy/__init__.py
```diff
@@ -792,9 +790,8 @@ def version(self) -> str:
         assert isinstance(version, str)
         return version
 
-    @property
-    @functools.lru_cache(maxsize=1)
-    def metadata(self) -> bytes:  # noqa: C901
+    @cached_property
```
FFY00, line 793: IIRC we didn't have the `cached_property` wrapper, so I just did it this way.

FFY00 (approved):
Thank you @dnicolodi!

FFY00:
All tests are passing, so I am gonna go ahead and merge.

FFY00:
Actually, it seems you forgot to remove the dependency from `get_requires_for_build_wheel`.

dnicolodi:
> Actually, it seems you forgot to remove the dependency from `get_requires_for_build_wheel`.

Could very well be. Let me fix this.

henryiii:
Closes #192 should be added to the description?

dnicolodi:
> Closes #192 should be added to the description?

It's in the commit message. Should be enough for GitHub to do the right thing. Updated the description too.

PR #218, opened 2022-11-19
ENH: use dash in config options instead of an underscore

FFY00:
This feels more natural.


FFY00:
I discussed this with @rgommers, we both agree using a dash is better.

rgommers (approved):
LGTM, thanks @FFY00!

PR #219, opened 2022-11-19
Check for invalid config keys

FFY00:


mesonpy/__init__.py
```diff
@@ -948,6 +959,10 @@ def _validate_string_collection(key: str) -> None:
                 ))
             )
 
+    for key in config_settings:
+        if key not in ('dist_args', 'setup_args', 'compile_args', 'install_args'):
```
rgommers, line 963: Missing `builddir` here

mesonpy/__init__.py
```diff
@@ -704,6 +706,15 @@ def _validate_metadata(self) -> None:
                     f'expected `{self._metadata.requires_python}`'
                 )
 
+    def _check_for_unknown_config_keys(self) -> None:
```
rgommers, line 709: Design thing: this kind of code tends to be easier to understand if the dict/keys to validate are passed in explicitly, rather than magically fetched. 

I was originally making this comment for `self._meson_args`, however `(args,)` two lines below is worse - that's neither passed in nor created in this function, which is a worse problem (you need to go look for where it came from, plus you cannot safely refactor this code).

rgommers (commented):
Thanks for working on this @FFY00. A few comments inline.

mesonpy/__init__.py
FFY00: I did some refactoring, making this

```python
self._check_for_unknown_config_keys({
    'args': typing_get_args(MesonArgsKeys),
 })
```

Where `typing_get_args(MesonArgsKeys)` are the valid `args` keys -- they are extracted from the `MesonArgsKeys` type, so that we have a single source of truth for them.

Let me know what you think.

mesonpy/__init__.py
```diff
@@ -944,18 +947,21 @@ def _validate_string_collection(key: str) -> None:
...
 
-    _validate_string_collection('dist-args')
-    _validate_string_collection('setup-args')
-    _validate_string_collection('compile-args')
-    _validate_string_collection('install-args')
+    meson_args_keys = typing_get_args(MesonArgsKeys)
+    meson_args_cli_keys = tuple(f'{key}-args' for key in meson_args_keys)
+
+    for key in config_settings:
+        if key not in ('builddir', *meson_args_cli_keys):
```
dnicolodi, line 969: It does not matter much, but this could be a set instead than a tuple.

mesonpy/__init__.py
```diff
@@ -601,6 +601,10 @@ def __init__(
             self._meson_args[key] = args_from_config + tuple(self._meson_args[key])
         # XXX: We should validate the user args to make sure they don't conflict with ours.
 
+        self._check_for_unknown_config_keys({
+            'args': typing_get_args(MesonArgsKeys),
+        })
```
dnicolodi, line 606: I'm not sure I understand the signature of this method. Can the passed dict have any other key than `args`? Where does `args` come from?

mesonpy/__init__.py
```diff
@@ -704,6 +708,17 @@ def _validate_metadata(self) -> None:
                     f'expected `{self._metadata.requires_python}`'
                 )
 
+    def _check_for_unknown_config_keys(self, valid_args: Mapping[str, Collection[str]]) -> None:
```
dnicolodi, line 711: I would have written this as static method (or a function) accepting the config settings (the result of `self._config.get('tool', {}).get('meson-python', {})`) and the description of the valid configuration keys.

dnicolodi: Something like
```python
@staticmethod
def _check_config(config: dict, schema: Type):
    valid_args = typing_get_args(schema)
    ...
```
```

File: mesonpy/__init__.py (hunk 4 above)
Comment (FFY00) [reply mesonpy/__init__.py:606]: Yep, it could be something other than `args`, if we add more config settings in the future.

Eg.

```python
self._check_for_unknown_config_keys({
    'args': typing_get_args(MesonArgsKeys),
    'other-section': ('setting1', 'setting2'),
})
```

The args key, validates the args table of the config

```
[tool.meson-python.args]
dist = []
setup = []
compile = []
install = []
invalid = []
```

The first 4 keys will be okay, but an error will be generated for `invalid`.

Similarly, according to the snippet above, the following config will be valid

```
[tool.meson-python.other-section]
setting1 = []
setting2 = []
```

mesonpy/__init__.py
FFY00: That requires us to change the implementation when we add new keys, so I am not really a fan, besides we are just passing around data that the class would already have access to.

FFY00: This is essentially what we have, but without it being a static method. The class would already have access to the config anyways, and we are not parsing different configs, so I don't see the point in doing it like this :sweat_smile:

mesonpy/__init__.py
FFY00: Yeah, sure, I tend to use tuples in these situations as they are less expensive (not like it would have any effect in this case in specific anyway :rofl:, it's just my general rule) and we are not handling any user data, so if there's duplication, it's incorrect code.

mesonpy/__init__.py
dnicolodi: I agree. In my opinion either this is made generic to check a configuration dict against a scheme, or it is made to do not take any arguments at all. Now it is an in-between that IMHO does not make much sense.

mesonpy/__init__.py
dnicolodi: I see. I was confused by this thing taking arguments at all. See below.

mesonpy/__init__.py
FFY00: So, essentially, you want this to also check the types, instead of just the key structure?

dnicolodi: That is not the main point I was trying to make, but if it makes sense, why not?

FFY00: I think it makes sense, but might be worth doing in a future PR. I really wanted to merge this and then finally tag a release :blush:

dnicolodi: Sure. Can we have the documentation updates back up and running before the release? Otherwise the published documentation will be outdated. We also need documentation for the `pyproject.toml` settings.

FFY00: The docs are working, they just aren't reporting back to Github for some reason I haven't managed to figure out :confused: 

https://readthedocs.org/projects/meson-python/builds/

rgommers: This can happen after moving repos. Often removing and reinstalling the GitHub hooks/integration helps.

Issue #197, opened 2022-11-08
Rework documentation


FFY00:
The documentation is not super accessible to new users. I would like to rewrite it so that it is more accessible to new users, while still presenting the required information for advanced users.

FFY00:
@CAM-Gerlach, if you have time, would you like to have a look at our documentation and provide feedback?

CAM-Gerlach:
Sure, thanks for asking. I'll keep my initial feedback and suggestions mostly fairly high level, focusing on your goals above (making the docs more accessible to beginners, while enhancing the key information for more advanced users) and the [docs](https://meson-python.readthedocs.io/en/latest/index.html?highlight=meson.build) you have so far (maybe there are more than this, somewhere else besides the example?). For some context, this comes from the perspective of both a fairly experienced documentarian familiar with the [Diataxis principles](https://diataxis.fr/) (that the NumPy, SciPy etc. docs also follow), but also a new user having only minimal familiarity with Meson and MesonPy.


## Initial impressions and feedback

Below, I present what I noticed and the questions I had as a first time user of the site. I've embedded it in a details block in case you want to skip right to the suggestions/recommendations.

<details>

<summary>Initial impressions and feedback</summary>

* Overall most of the basic pieces of information are there, once I dig around for them, but the structure and organization doesn't feel that intuitive as someone not already familiar with it, which both marks it harder to find information I'm looking for, and makes it feel like I don't have a complete, coherent picture of what Meson is and how to use it. Re-organizing it to better serve readers' needs, using the Diataxis framework as a guid, should help here.
* The main page contains a brief "How to use section" (good), which served as a jumping off point to the Getting Started, which I figured would be a good beginner entry point. However, it then dives right into details of how Mesonpy works with sdists and wheels, which seem like it would be much better suited for a separate page. The (fairly significant) existing limitations are then only listed at the bottom (which is good mainpage content, but obscured by the bulk of the technical details in the "How does it work" section.
[...]

PR #221, opened 2022-11-20
CI tweaks

dnicolodi:
The main goal here is to cut down the total number of jobs we run from 41 to to something more sensible.

I got them down to:

- CPython 3.7 and 3.11, and on Windows, macOS, and Ubuntu: 6 jobs
- CPython 3.8, 3.9, 3.10, and PyPy 3.9 on Ubuntu: 4 jobs
- CPython 3.9 in Cygwin: 1 job
- Pyston 3.8 on Ubuntu: 1 job
- CPython 3.7 and 3.11 on Homebrew: 2 jobs
- a mypy job,
- the precommit tests,
- the Sage build job and the 8 Sage test jobs.

That's 25 jobs total. 

I've experimented with running the Sage workflow only of the tests workflow succeeds but that has the annoying consequence of removing the Sage workflow test results from the PR view. The only way I know to avoid this is to have the Sage jobs defined in the same workflow as the other tests, but it does not seem very tidy.

### DIFF
--- a/.github/workflows/sage.yml
+++ b/.github/workflows/sage.yml
@@ -64,6 +64,8 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
+    outputs:
+      systems: ${{ steps.push-to-main-systems.outputs.systems }} ${{ steps.pull-request-systems.outputs.systems }}
     steps:
       - name: Check out ${{ env.SPKG }}
         uses: actions/checkout@v3
@@ -83,7 +85,34 @@ jobs:
           && echo "sed -i.bak \"/pushdef.*LT_VERSION/s/3[0-9.]*/4/\" ../build/pkgs/python3/spkg-configure.m4" >> upstream/update-pkgs.sh \
           && echo "sed -i.bak \"/export.*proxy/d\" ../build/bin/sage-spkg" >> upstream/update-pkgs.sh \
           && ls -l upstream/
-      - uses: actions/upload-artifact@v2
+      - name: System factors for pull requests
+        # See https://github.com/sagemath/sage/blob/develop/.github/workflows/docker.yml for available systems
+        id: pull-request-systems
+        if: ${{ github.event_name == 'pull_request' }}
+        run: >-
+          echo 'systems=[
+          "archlinux-latest",
+          "conda-forge",
+          ]' >> $GITHUB_OUTPUT
+      - name: System factors for push to main
+        # See https://github.com/sagemath/sage/blob/develop/.github/workflows/docker.yml for available systems
+        id: push-to-main-systems
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+        run: >-
+          echo 'systems=[
+          "debian-bullseye",
+          "fedora-36",
+          "centos-stream-8-python3.9",
+          "gentoo-python3.9",
+          "archlinux-latest",
+          "opensuse-15.4-gcc_11-python3.10",
+          "opensuse-tumbleweed-python3.10",
+          "conda-forge",
+          "archlinux-latest",
+          "conda-forge",
+          ]' >> $GITHUB_OUTPUT
+      - name: Upload artifact
+        uses: actions/upload-artifact@v2
         with:
           path: upstream
           name: upstream
@@ -96,19 +125,9 @@ jobs:
       # Reduced list of systems to test: Only those that bring python >= 3.8
       # Full list at https://github.com/sagemath/sage/blob/develop/.github/workflows/docker.yml
       # See also https://github.com/sagemath/sage/blob/develop/tox.ini
-      tox_system_factors: >-
-          ["debian-bullseye",
-           "fedora-36",
-           "centos-stream-8-python3.9",
-           "gentoo-python3.9",
-           "archlinux-latest",
-           "opensuse-15.4-gcc_11-python3.10",
-           "opensuse-tumbleweed-python3.10",
-           "conda-forge",
-           ]
+      tox_system_factors: ${{ needs.build.outputs.systems }}
       tox_packages_factors: >-
-          ["minimal-develop",
-           ]
+        ["minimal-develop",]
       # Extra packages to install as system packages
       extra_sage_packages: "liblzma bzip2 libffi libpng python3 ninja_build"
       # Sage distribution packages to build
--- a/.github/workflows/sage.yml
+++ b/.github/workflows/sage.yml
@@ -112,7 +112,7 @@ jobs:
           "conda-forge",
           ]' >> $GITHUB_OUTPUT
       - name: Upload artifact
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           path: upstream
           name: upstream
--- a/.github/workflows/sage.yml
+++ b/.github/workflows/sage.yml
@@ -62,7 +62,7 @@ env:
 
 jobs:
 
-  dist:
+  build:
     runs-on: ubuntu-latest
     steps:
       - name: Check out ${{ env.SPKG }}
@@ -88,7 +88,7 @@ jobs:
           path: upstream
           name: upstream
 
-  linux:
+  test:
     # https://github.com/sagemath/sage/blob/develop/.github/workflows/docker.yml
     # Use branch u/mkoeppe/numpy_1_23_x__scipy_1_9_x for a fix for uppercase github repo names (FFY00)
     uses: sagemath/sagetrac-mirror/.github/workflows/docker.yml@u/mkoeppe/numpy_1_23_x__scipy_1_9_x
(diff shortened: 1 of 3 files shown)


FFY00:
Would it be possible to run everything on the `main` branch, but a fewer jobs on PRs? Sometimes there are version specific issue, that we still want to catch, but as they are not common enough it might be hard to justify doing that in PRs.

dnicolodi:
Sure, but you would most likely end up enabling the skipped tests in the PR that tries to fix the issues founds on main. Not a big deal, if infrequent enough. We could have the Sage tests run on push to main only. These are the ones that take the most time and are unlikely to catch anything not caught by the other tests.

FFY00:
I think the Sage tests are useful in PRs, as they test real work scenarios, but I don't we need these many in PRs. We could have only archlinux and conda-forge, for eg.

dnicolodi:
Can do that. Not running the gentoo test on PRs is already a big saving.

dnicolodi:
The workflow definition look a bit convoluted, but it achieves what we want. I'm surprised that the Sage jobs take so long, but that's a fight for another time. @FFY00 Unless there is something you don't like, I'll like to merge this soon as it is a big improvement upon the current status.

.github/workflows/sage.yml
```diff
@@ -76,40 +78,60 @@ jobs:
...
-           "conda-forge",
-           ]
+      tox_system_factors: ${{ needs.build.outputs.systems }}
       tox_packages_factors: >-
-          ["minimal-develop",
-           ]
-      # Extra packages to install as system packages
+        ["minimal-develop",]
+      # Extra system packages to install
       extra_sage_packages: "liblzma bzip2 libffi libpng python3 ninja_build"
```
mkoeppe, line 135: May want to add that these package names refer to https://github.com/sagemath/sage/tree/develop/build/pkgs

dnicolodi: Sure. By the way, do you know why we need these packages? I'm temped to remove them and see what happens.

mkoeppe: `python3` is needed because otherwise Sage will build its own copy of Python instead of the system python that is intended to be tested here.
`liblzma bzip2 libffi libpng` are needed because they are library dependencies of Python, and Sag will build its own copy of Python if they are not coming from the system.

dnicolodi: Also, do we still need to use your branch for the GitHub action or can we go back to the upstream one? What is the canonical source for it?

mkoeppe: See https://github.com/sagemath/sage/blob/develop/build/pkgs/python3/spkg-configure.m4#L23 (although I don't see libpng there)

dnicolodi: > `liblzma bzip2 libffi libpng` are needed because they are library dependencies of Python

Good to know. I thought that Sage did some dependency tracking. I'll add a comment.

dnicolodi: > (although I don't see libpng there)

Do we have a way to inspect if everything still works as intended if I remove `libpng` from the list?

mkoeppe: At the end of the `configure` run, you see lines such as this:
```
python3:                        using system package; SPKG will not be installed
```
https://github.com/mesonbuild/meson-python/actions/runs/3509068860/jobs/5877914584#step:11:1314

If the system python is no longer accepted, this line will change.

mkoeppe: > do we still need to use your branch for the GitHub action or can we go back to the upstream one?

Still need the branch; this will change it a few days.

mkoeppe: > > `liblzma bzip2 libffi libpng` are needed because they are library dependencies of Python
> 
> Good to know. I thought that Sage did some dependency tracking.

It does, but this goes in the opposite direction: If Sage does not find system `libffi` for example, it builds its own copy; and then it does not accept system `python3` (because that is bringing its own copy of `libffi`)

dnicolodi: I see. A quick note in passing: it would be nice for the Sage test action to upgrade to `actions/upload-artifact@v3` as the `v2` version emits a warning about `node` 12 going away soonish.

dnicolodi: `libpng` does not seem to be required indeed. Do you know if there is a way to cut down some of the time that goes into the image preparation? IIUC, there are is a lot of stuff that we don't need being searched for and compiled.

mkoeppe: > Do you know if there is a way to cut down some of the time that goes into the image preparation?

I've opened https://trac.sagemath.org/ticket/34765 to allow us to use system `meson`.

dnicolodi: I don't think that Meson is the biggest contender. Looking at the Arch job I see 224 SPKG packages installed.

mkoeppe: > upgrade to `actions/upload-artifact@v3` as the `v2` version emits a warning about `node` 12 going away soonish

Yes, I'll do this in https://trac.sagemath.org/ticket/34766

mkoeppe: > I don't think that Meson is the biggest contender. Looking at the Arch job I see 224 SPKG packages installed.

If you're referring to what `configure` reports as "will be installed" -- that's what would be installed if a full `make build` was done. You can see what is actually being installed in lines such as
https://github.com/mesonbuild/meson-python/actions/runs/3509347178/jobs/5878504732#step:11:1751
```
[meson-0.63.3] installing. Log file: /sage/logs/pkgs/meson-0.63.3.log
```

dnicolodi: I see. Thank you. I'm trying to understand what exactly in the setup phase takes 6 minutes for Arch and 14 minutes for Conda and see if there is a way to cut down this time.

mkoeppe: Just for context, a full build of Sage (`make build`) takes between 3 and 6 hours on GH Actions, depending on what system package configuration is used.

dnicolodi: Sure, but all we need is a regular distribution image with Python installed. Maybe the Sage test infrastructure is not the right place where to start...

mkoeppe: I have no objections if you want to replace the Sage CI by something that you maintain by yourself

FFY00:
@dnicolodi yes, sounds good. I am gonna tag a release, feel free to merge afterwards.

dnicolodi:
I would appreciate having the documentation for the new options before the release. But I can live without that.

FFY00:
I did not do that because we really needed a release, and we are planning a documentation rewrite anyway (#197)  :sweat_smile:

.github/workflows/tests.yml
```diff
@@ -253,8 +253,8 @@ jobs:
       - name: Run mypy
         run: mypy -p mesonpy
 
-  tests-pass:
```
webknjaz, line 261: @dnicolodi moreover — this implementation was broken anyway.

https://github.com/marketplace/actions/alls-green#why

henryiii: It was not faulty. It just gets marked as skip instead of fail, but it's required to merge, so that's the same result. It would be better if it always ran, and reported the result (via alls-green) probably, but it still lets you do things like "merge when CI passes".

Why was this commented out? It doesn't slow the job by reducing parallism since it always runs at the end. So there's an extra at most 7 seconds or something tacked on the end. And it lets you makes the "passing" requirement in GitHub much simpler, and changes to it don't cause previous jobs to turn orange.

webknjaz: > but it still lets you do things like "merge when CI passes".

Not exactly. It also let's you do things "merge when CI fails" because skipped==pass in branch protection.

webknjaz: > It was not faulty.

Read the whole page at https://github.com/marketplace/actions/alls-green#why for explanation.

henryiii: Ahah, so that action is removing the "merge when CI fails" feature. :+1:

dnicolodi: > Why was this commented out? It doesn't slow the job by reducing parallism since it always runs at the end. So there's an extra at most 7 seconds or something tacked on the end. And it lets you makes the "passing" requirement in GitHub much simpler, and changes to it don't cause previous jobs to turn orange.

I commented it out because it is not used. Because all the projects within the mesonbuild organization share the GitHub Actions runners quota, it happened several times that the first batch of jobs from the meson-python is queued then some long-running jobs from the meson or wrapdb projects are queued and the summary job is left waiting for a long time for a runner just to execute an `echo`.

> And it lets you makes the "passing" requirement in GitHub much simpler

In general I don't think that having something enforcing the policy that a PR is to be merged only when the test suite passes is very useful. I think we can trust our pairs to behave. The test suite was failing on several configurations and these tests were simply excluded from the required set. The fact that merging was allowed made no one paid attentions to what was actually a bug. Just merge when all the tests are green is a much simpler and much more effective policy.

> changes to it don't cause previous jobs to turn orange

I don't know what you mean by this. I don't think the status of open PRs changes when the CI jobs definitions on the main branch change. If it would, that would be a welcome feature: PRs should be tested against the current main branch.

PR #227, opened 2022-11-23
ENH: include allowed keys in error message

tacaswell:
This adds the allowed keys to the error message (I was trying with `_` instead of `-` in a name, this would have cut my debugging time to seconds rather than a few minutes).

### DIFF
--- a/tests/test_pep517.py
+++ b/tests/test_pep517.py
@@ -54,9 +54,22 @@ def test_get_requires_for_build_wheel(monkeypatch, package, system_patchelf, nin
 
 
 def test_invalid_config_settings(package_pure, tmp_path_session):
-    raises_error = pytest.raises(mesonpy.ConfigError, match='Unknown config setting: invalid')
+    raises_error = pytest.raises(mesonpy.ConfigError, match="Unknown config setting: 'invalid'")
 
     with raises_error:
         mesonpy.build_sdist(tmp_path_session, {'invalid': ()})
     with raises_error:
         mesonpy.build_wheel(tmp_path_session, {'invalid': ()})
+
+
+def test_invalid_config_settings_suggest(package_pure, tmp_path_session):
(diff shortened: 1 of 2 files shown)


henryiii:
This could eventually become large, would it be better to show similar keys only, probably using difflib?

tacaswell:
This is better tested than I expected!  It was easier to just implement it with `difflib` than argue why it should not be done...will push as soon as I fix the tests.

tacaswell:
squashed to one commit, think this ready to go again @henryiii .

henryiii (approved):

tests/test_pep517.py
```diff
@@ -54,9 +54,19 @@ def run(cmd: List[str], *args: object, **kwargs: object) -> subprocess.Completed
...
+    raises_error = pytest.raises(mesonpy.ConfigError, match="Unknown config setting: 'invalid'")
 
     with raises_error:
         mesonpy.build_sdist(tmp_dir_session, {'invalid': ()})
     with raises_error:
         mesonpy.build_wheel(tmp_dir_session, {'invalid': ()})
+
+
+def test_invalid_config_settings_suggest(package_pure, tmp_dir_session):
+    raises_error = pytest.raises(mesonpy.ConfigError, match='setup-args')
```
FFY00, line 66: nitpick: Can you put the full message here?

FFY00 (approved):
Thanks! This is very helpful.

Note, we should use the `ENH` prefix when we merge.

tests/test_pep517.py
tacaswell: I opted to only search for the target I think we should have hit because if the number of options does grow as @henryiii is worried it will I expect that this message will change in a way that should not fail.  I changed the pattern to check more of the message.

FFY00: Thanks!

PR #231, opened 2022-11-24
ENH: raise SystemExit when calling meson fails

### LINKED ISSUE #228: raise `SystemExit` on build failure
Good idea from @henryiii in https://github.com/mesonbuild/meson-python/issues/222#issuecomment-1322635442: _When Meson fails, we should raise SystemExit (like setuptools usually does), not throw an exception. Pip/build report a mess when there's an uncaught exception from the build hooks, while SystemExit is handled gracefully without nearly as many extra reports about missing build hooks and changed exceptions. The actual failure is printed by Meson._

dnicolodi:
Fixes #228.

### DIFF
--- a/tests/test_pep517.py
+++ b/tests/test_pep517.py
@@ -56,23 +56,19 @@ def test_get_requires_for_build_wheel(monkeypatch, package, system_patchelf, nin
         assert set(mesonpy.get_requires_for_build_wheel()) == expected
 
 
-def test_invalid_config_settings(package_pure, tmp_path_session):
-    raises_error = pytest.raises(mesonpy.ConfigError, match="Unknown config setting: 'invalid'")
+def test_invalid_config_settings(capsys, package_pure, tmp_path_session):
+    for method in mesonpy.build_sdist, mesonpy.build_wheel:
+        with pytest.raises(SystemExit):
+            method(tmp_path_session, {'invalid': ()})
+        out, err = capsys.readouterr()
+        assert out.splitlines()[-1].endswith(
+            'Unknown configuration entry "invalid"')
 
-    with raises_error:
-        mesonpy.build_sdist(tmp_path_session, {'invalid': ()})
-    with raises_error:
-        mesonpy.build_wheel(tmp_path_session, {'invalid': ()})
 
-
-def test_invalid_config_settings_suggest(package_pure, tmp_path_session):
-    raises_error = pytest.raises(
-        mesonpy.ConfigError,
-        match='Did you mean one of: .*setup-args'
-    )
-
-    with raises_error:
-        mesonpy.build_sdist(tmp_path_session, {'setup_args': ()})
-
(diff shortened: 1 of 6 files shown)


dnicolodi:
This fixes the case when Meson errors out. However, the new configuration handling code raises `ConfigurationError` exceptions that are not handled gracefully. Maybe the pep517 hooks should be wrapped in something that turns these into nicer error messages. I don't know what is the recommended pep517 way of doing this, though.

tests/test_pep517.py
```diff
@@ -27,11 +27,12 @@ def which(prog: str) -> bool:
...
+    subprocess_run = subprocess.run
+
     def run(cmd: List[str], *args: object, **kwargs: object) -> subprocess.CompletedProcess:
-        if cmd != ['ninja', '--version']:
-            # smoke check for the future if we add another usage
-            raise AssertionError(f'Called with {cmd}, tests not expecting that usage')
-        return subprocess.CompletedProcess(cmd, 0, f'{ninja}\n', '')
+        if cmd == ['ninja', '--version']:
+            return subprocess.CompletedProcess(cmd, 0, f'{ninja}\n', '')
+        return subprocess_run(cmd, *args, **kwargs)
```
FFY00, line 35: The patch to raise `SystemExit` shouldn't be tripping this, right? What's the reasoning behind this change?

dnicolodi: It actually does. The patch changes a the meson execution from a `subprocess.check_call()` to `subprocess.run()`, which is intercepted by this monkeypatch and everything crumbles.

eli-schwartz:
What is the status of this PR? People are reporting the failures directly to Meson itself now.

As far as I can tell, this flawlessly solves the problem some of the time, and doesn't touch the code involved in some other times. It seems obviously correct (incremental improvements FTW) and solves real problems for people, so why delay in merging it?

FFY00:
> What is the status of this PR? People are reporting the failures directly to Meson itself now.

I am almost finished in #87, after that is in, I will merge this too and make a release.

dnicolodi:
On user input error (either command line or configuration file) we now raise `ConfigError` and nothing catches that. I wanted to fix that too before merging. But this can come at a later time.

rgommers:
> I am almost finished in https://github.com/mesonbuild/meson-python/pull/87, after that is in, I will merge this too and make a release.

@FFY00 I recommend leaving enough time pre and post merge of that PR, so we're sure it's in good shape when it is released. This PR doesn't have to wait for that one.

dnicolodi:
I added handling of the exceptions raised for configuration errors. Please have a look.

FFY00:
> @FFY00 I recommend leaving enough time pre and post merge of that PR, so we're sure it's in good shape when it is released. This PR doesn't have to wait for that one.

Yeah, it makes sense to release this first :+1:

PR #232, opened 2022-11-24
CI: test with different Meson versions

dnicolodi:


tests/test_wheel.py
```diff
@@ -160,13 +160,10 @@ def test_contents_license_file(wheel_license_file):
 def test_executable_bit(wheel_executable_bit):
     artifact = wheel.wheelfile.WheelFile(wheel_executable_bit)
 
-    # With Meson 0.63.x we see `data/bin/`, with master `scripts/`. The latter
-    # seems correct - see gh-115 for more details.
     executable_files = {
         'executable_bit-1.0.0.data/purelib/executable_module.py',
         'executable_bit-1.0.0.data/scripts/example',
         'executable_bit-1.0.0.data/scripts/example-script',
-        'executable_bit-1.0.0.data/data/bin/example-script',
```
FFY00, line 169: Can't we add this depending on the Meson version instead?

dnicolodi: There is no need. The minimum required meson version (which is not changed in this PR) behaves consistently with the test.

Issue #109, opened 2022-07-24
pip install . --user -> [Errno 2] No such file or directory: 'meson'


Nic30:
Hello,

I am trying to make package build by meson-python installable by pip using `pip install --user .` (a complete [build script](https://github.com/Nic30/circleci-orb-python-all-in-1/blob/master/src/commands/install-package-git.yml), [CI config](https://github.com/Nic30/hwtHls/blob/llvm_mir_integration/.circleci/config.yml))

Recent changes in pip broke this and mesonpep517 has the same issue
https://github.com/jeandet/mesonpep517/issues/1#issuecomment-1148288250

mesonpep517 has opened [pull request](https://gitlab.com/thiblahute/mesonpep517/-/merge_requests/18) which is fixing this,  however merging takes some time and I simply need this so I tried meson-python.

The result is the same:
```
...
        File "/tmp/pip-build-env-j8gehd_x/overlay/local/lib/python3.10/dist-packages/mesonpy/__init__.py", line 477, in _meson
          return self._proc('meson', *args)
        File "/tmp/pip-build-env-j8gehd_x/overlay/local/lib/python3.10/dist-packages/mesonpy/__init__.py", line 472, in _proc
          subprocess.check_call(list(args))
        File "/usr/lib/python3.10/subprocess.py", line 364, in check_call
          retcode = call(*popenargs, **kwargs)
        File "/usr/lib/python3.10/subprocess.py", line 345, in call
          with Popen(*popenargs, **kwargs) as p:
        File "/usr/lib/python3.10/subprocess.py", line 966, in __init__
          self._execute_child(args, executable, preexec_fn, close_fds,
        File "/usr/lib/python3.10/subprocess.py", line 1842, in _execute_child
          raise child_exception_type(errno_num, err_msg, err_filename)
      FileNotFoundError: [Errno 2] No such file or directory: 'meson'
```

There is a full log from [build](https://app.circleci.com/pipelines/github/Nic30/hwtHls/278/workflows/b27df647-9707-4f60-9c38-cceb750b2e2f/jobs/284)

**The problem is** that pip does modify the PATH and meson executable will be missing. Note that this happens only if meson is not installed in the advance.
[...]

PR #241, opened 2022-12-03
CI: add Cirrus CI jobs

dnicolodi:
Run the tests on a few different Linux and Python distributions. I think this setup is much more agile than the Sage CI based one, and it catches issues much more effectively. The test are currently failing on issue I identified in one of my projects: the Debian jobs fail because of #109, fixed in https://github.com/pypa/pip/pull/11623 and the manylinux Python 3.7 job is failing because of https://github.com/mesonbuild/meson/issues/11097, fixed in https://github.com/mesonbuild/meson/pull/11104. 

Once the container images have been created and cached, the tests run in just a bit more than a minute and we can run 16 parallel configurations. I added a few relevant Linux distributions and Conda on Linux. Sage CI is also run on openSUSE but I don't know anything about package management on Suse thus I don't know how to create the container images. If someone contributes the image setup definition, it is trivial to add new jobs. I'm definitely not a container expert and I didn't do any effort to minimize the size of the images. Given that they are cached on the CI infrastructure, I don't think it matters.

Someone will need to figure out how to add the Cirrus CI GitHub Application to the repository and allow access to the logs.


dnicolodi:
@jpakkane can you please install Cirrus CI for the meson-python repository? The instructions are here https://cirrus-ci.org/guide/quick-start/

jpakkane:
Installed.

FFY00:
Makes sense, and seems much faster. @mkoeppe do you have any thoughts?

FFY00:
@dnicolodi can you look into the failure?

dnicolodi:
The reason of the failure is detailed in the PR description 😃 "manylinux Python 3.7 job is failing because of https://github.com/mesonbuild/meson/issues/11097, fixed in https://github.com/mesonbuild/meson/pull/11104." I can find a way to have the job to use an unreleased Meson, but the Meson version fixing that should materialize soon so maybe we can just wait a bit.

FFY00:
Ah sorry, my dyslexia sometimes makes it hard to parse stuff, and the ADHD messes with my concentration, I forgot you had already mentioned that :upside_down_face:

Yep, we can wait. But IMO, we could just merge it right now.

FFY00 (approved):

PR #242, opened 2022-12-09
TST: remove ninja from test dependencies

dnicolodi:
Now that meson-python can use the system provided ninja there is no need to depend on the PyPI package.


dnicolodi:
This clearly needs some CI adjustment to install ninja, but I would like to know if there is consensus toward removing the dependency before doing that.

PR #245, opened 2022-12-13
CI: test with Meson 1.0.0rc1

dnicolodi:
I don't expect any surprise, but release candidates are made to be tested, thus do our good citizens duty.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,7 +37,7 @@ jobs:
           - '3.7'
           - '3.11'
         meson:
-          -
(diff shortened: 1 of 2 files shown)


rgommers (approved):
All seems happy as expected, nice. Seems fine to me to merge this, and update to 1.0.0 final once that is released.

dnicolodi:
This run the tests only on Python 3.7 and 3.11 on Ubuntu (I changed the Python version manually). I was entertaining the idea of running the tests also on other platforms.

FFY00 (approved):

PR #251, opened 2022-12-28
ENH: drop importlib_resources dependency

FFY00:

### DIFF
--- a/mesonpy/_compat.py
+++ b/mesonpy/_compat.py
@@ -3,6 +3,7 @@
 # SPDX-FileCopyrightText: 2021 Filipe Laíns <lains@riseup.net>
 
 import functools
+import importlib.resources
 import os
 import pathlib
 import sys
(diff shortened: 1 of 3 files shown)


mesonpy/_compat.py
```diff
@@ -37,6 +37,17 @@
     cached_property = lambda x: property(functools.lru_cache(maxsize=None)(x))  # noqa: E731
 
 
+if sys.version_info >= (3, 9):
+    import importlib.resources
+
+    def read_binary(package: str, resource: str) -> bytes:
+        return importlib.resources.files(package).joinpath(resource).read_bytes()
+else:
+    import importlib.resources
```
rgommers, line 46: This is the same import twice in both `if` and `else` branches, so it can just be at the top of the file?

FFY00: Ah, of course :sweat_smile:

Issue #248, opened 2022-12-21
Make `typing-extensions` a test only dependency


dnicolodi:
In #231 I need to make use to of `typing` definitions added only in Python 3.10, thus making `typing-extensions` a requirement for most of the supported Python version. It is quite annoying to have an extra dependency only for type checking. I think the dependency should be made a test only dependency.

rgommers:
I agree with that sentiment. Adding a runtime dependency that is never for anything at runtime, and used for one type annotation object, feels wrong. As dependencies go, `typing-extensions` is pretty light weight, but still.

henryiii:
If anyone uses the types downstream and typing-extensions is missing, this adds a hidden undeclared dependency. There's a pretty good chance something else depends on typing-extensions, but you technically are broken if a user don't have it. (Not sure this even has py.typed, but this is a bug I've seen for libraries).

IMO, _any_ backport dependency should be treated as a 0-cost dependency. If someone complains about it, they can upgrade their Python and it goes away. As long as 3.10 or 3.11+ doesn't require the dep, I'd say it's totally fine to depend on it on older versions. Same is true for tomllib, exceptiongroup, or any other backport. The older versions eventually won't even matter. This happened with dataclasses, and now code prevously using it got to drop the dep and now has code that looks like it was designed for 3.7+. Maybe it's even less than 0-cost.

rgommers:
> If anyone uses the types downstream and typing-extensions is missing,

We don't have a Python API, so how is this possible?

henryiii:
> We don't have a Python API, so how is this possible?

That's what I meant by:

> Not sure this even has py.typed, but this is a bug I've seen for libraries

And that's why I also said it could be done if you really wanted too.

rgommers:
I had a closer look: this was already a dependency, but now it's a much stronger one. This dependency was/is missing in conda-forge, which now matters. The metadata also needs a tweak in all other downstream packaging systems. The change isn't listed in our Changelog, so that will probably go wrong in most places.

tl;dr this is certainly not a zero cost dependency

There is zero usage of static typing downstream, plus extra dependencies have a small overhead at runtime (downloads, build time). So I think @dnicolodi's proposal is the right one - best to ensure we don't have this dependency at all except for meson-python development purposes.

eli-schwartz:
> Adding a runtime dependency that is never for anything at runtime, and used for one type annotation object, feels wrong.

[...]

PR #255, opened 2022-12-28
ENH: drop typing-extensions dependency

FFY00:
~~On top of #251, as it conflicts, and I don't want to have to rebase the PR after that gets merged.~~

As suggested in #248

Pros

- Drop the `typing-extensions` runtime dependency

Cons

- Unknown future, as we rely on `from __future__ import annotations`
- More complexity
- More unnecessary boilerplate
	- This can be avoided by limiting `TYPE_CHECKING` to only when necessary

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -31,17 +31,13 @@ import textwrap
 import typing
 import warnings
 
-from typing import (
-    Any, Callable, ClassVar, DefaultDict, Dict, List, Optional, Sequence, Set,
-    TextIO, Tuple, Type, TypeVar, Union
-)
+from typing import Dict
 
 
 if sys.version_info < (3, 11):
     import tomli as tomllib
 else:
     import tomllib
-
 
 import mesonpy._compat
 import mesonpy._elf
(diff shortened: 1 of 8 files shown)


henryiii:
A lot of the complexity can be reduced by using the builtins or `collections.abc` versions of things now that `from __future__ import annotations` is used. I think it's a net positive.

There should be some clear path forward for `from __future__ import annotations` (hopefully it either becomes a no-op or stays for years), but regardless, there will be some at least simi-automated way forward. Many (most?) 3.7+ libraries rely heavily on it today.

dnicolodi:
Making the type annotations effective only when type checking, another opportunity for simplification is to use an higher Python minimum version for the type checking than for normal execution. Using Python 3.11 as the base version for type checking would allow to remove quite a bit of Python version compatibility code. As we don't have any (or extremely little) Python version dependent code, this should not impact the effectiveness of the type checking.

eli-schwartz:
BTW see e.g. https://github.com/NixOS/nixpkgs/pull/210878

PR #257, opened 2022-12-30
BUG: Don't reconfigure if the build directory does not exist

lithomas1:
~~I am erroring out on macOS if I pass a build directory that doesn't exist.~~
EDIT: It's more complicated than this. Somehow the build directory is getting corrupted, and I can only repro this sometimes.

This is what the corrupted builddir looks like
e.g.
```
.mesonpy
   - build
       - .hgignore
       - .gitignore
```

This is the error
```
  + meson setup --reconfigure --prefix=/Users/thomasli/opt/miniconda3/envs/pandas-dev /Users/thomasli/pandas /Users/thomasli/pandas/builddir --native-file=/Users/thomasli/pandas/.mesonpy-native-file.ini -Ddebug=false -Doptimization=2
  Directory does not contain a valid build tree:
  /Users/thomasli/pandas/builddir
```


lithomas1:
Can't reproduce this anymore.

lithomas1:
Actually, this still happens sometimes, maybe related to build dir being corrupted.

Maybe a good middle ground is to check for the existence of build.ninja.

tests/test_wheel.py
```diff
@@ -249,3 +250,22 @@ def test_editable(
         assert venv.python('-c', 'import plat; print(plat.foo())').strip() == 'something else'
     finally:
         plat.write_text(plat_text)
+
+def test_editable_broken_non_existent_build_dir(
```
rgommers, line 254: This PR looks like a useful robustness improvement. It'd be good to add a more explicit comment here about a few of the ways that this can happen (e.g. user does `rm -rf .mesonpy`) and that what needs to happen then is a full rebuild.

lithomas1: Unfortunately, I don't have a way to consistently reproduce this. Sometimes, I can still get it to work.

Also, this method is not foolproof. If I delete all the meson folders(meson is somehow resilient to only one folder being deleted) in the build dir, by hand, then we can still fail to regenerate the builddir, even with build.ninja present.

(We probably don't care about that case, though, since I don't see how the build dir could get corrupted like that).

lithomas1: Maybe @eli-schwartz has a better idea of what to check for.

eli-schwartz: Meson internally checks to see if the file `meson-private/coredata.dat` exists inside the build directory.

- There are too many reasons an empty directory might exist (including several ways that all files, but no directories, can be deleted by forces outside of Meson).
- The build.ninja isn't suitable, since that doesn't tell us whether it was Meson that created it.
- Internally, we decided to use coredata.dat as the "existence check", so correspondingly, we also arrange to have this file be created as late as we can, or deleted if something went wrong at the last minute.

lithomas1: Awesome, this is exactly what I've been looking for.

Also kind of explains, why meson-python fails sometimes when the setup process fails.

lithomas1:
I can now consistently reproduce this error in the test case (it fails without my changes).

lithomas1:
gentle ping @rgommers

mesonpy/__init__.py
```diff
@@ -750,8 +750,15 @@ def __init__(  # noqa: C901
...
-        self._configure(reconfigure=bool(build_dir) and not native_file_mismatch)
+        # Don't reconfigure if build directory doesn't have meson-private/coredata.data
+        # (means something went wrong)
+        # See https://github.com/mesonbuild/meson-python/pull/257#discussion_r1067385517
+        reconfigure = bool(build_dir)
+        if reconfigure:
+            if not os.path.exists(os.path.join(self._build_dir, 'meson-private/coredata.dat')):
+                reconfigure = False
+        reconfigure = reconfigure and not native_file_mismatch
+        self._configure(reconfigure=reconfigure)
```
FFY00, line 761:
```suggestion
        has_valid_build_dir = bool(build_dir) and not self._build_dir.joinpath('meson-private', 'coredata.dat')
        self._configure(reconfigure=has_valid_build_dir not native_file_mismatch)
```

FFY00 (approved):
Thanks @lithomas1, this looks good to me. I just have a suggestion to simplify the logic.

lithomas1:
Thanks for the reviews. Gonna self-merge this now since we need this for pandas.

(The one CI failure looks unrelated)

PR #260, opened 2023-01-14
Fix linking against libraries from Meson project on macOS

pastewka:
This PR fixes linking against libraries created during the Meson build on macOS. It uses `otool` and `install_name_tool` to update the LC_RPATH load command in the dylib. (This is identical to how the library is patched on Linux.)

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -41,6 +41,7 @@ else:
     import tomllib
 
 import mesonpy._compat
+import mesonpy._dylib
 import mesonpy._elf
 import mesonpy._introspection
 import mesonpy._tags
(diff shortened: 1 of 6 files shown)


rgommers:
This looks promising, thanks for tackling this topic @pastewka!

The `fix-pip` step seems on the Cirrus CI Debian jobs seems to be failing consistently, probably because https://github.com/pypa/pip/pull/11623 was merged a few days ago. @dnicolodi would you mind having a look?

dnicolodi:
@rgommers Yes, upgrading pip to a version that fixes a known issue we are hitting on Debian fails because the merge branch has been deleted. Fixing that is on my todo list. I'll do that as soon as a find a little bit of time.

pastewka:
@rgommers @dnicolodi Happy to contribute, meson-python is a fantastic project that has made my life a whole lot easier. It looks like all tests pass now, except the Debian tests which as discussed seem unrelated to this PR.

dnicolodi:
Please squash the commits

pastewka:
Done

mesonpy/__init__.py
```diff
@@ -558,7 +568,8 @@ def build(self, directory: Path) -> pathlib.Path:
 
                 # install bundled libraries
                 for destination, origin in self._wheel_files['mesonpy-libs']:
-                    assert platform.system() == 'Linux', 'Bundling libraries in wheel is currently only supported in POSIX!'
+                    assert platform.system() == 'Linux' or platform.system() == 'Darwin', \
+                        'Bundling libraries in wheel is currently only supported in POSIX!'
```
HDembinski, line 572: Sorry for the intrusion, but I also looked into this code and wanted to ask a question.

Why is the assert placed in line 571; this is remote from the location where the code actually fails to handle the platform. It makes the code a bit harder to understand than necessary. I would do an `elif` in line 529 and then a final `else`  branch and put the assert there.

pastewka: This makes a lot of sense, I've moved the failure to the point you suggested. Note that is is still an assert, should this become an exception?

pastewka:
I've also updated the README to reflect status on macOS.

mesonpy/__init__.py
rgommers: Good question. I agree that an exception is better than an `assert`. Asserts are for things that are not supposed to happen, and may be stripped out on install.

pastewka: A `NotImplementedError` would probably be appropriate. I'll update the code accordingly.

pastewka: I changed this to a `NotImplementedError`

pastewka:
Is there anything else here that I should change? What is the process for getting this into a release?

FFY00 (approved):
@pastewka sorry for the delay! I rebased this PR on `main`, we should now be able to merge as soon as the CI runs. Thank you so much for the contribution!

rgommers:
Awesome, great to see this land - thanks @pastewka and @FFY00!

pastewka:
Thanks for merging this, happy to contribute!

PR #261, opened 2023-01-15
BUG: fix macOS support for cross-compiling C++

eli-schwartz:
In commit 13c6748a122bf7b5e887451165f66d33618dbe3b support was added for setting up cross file support based on the presence of $ARCHFLAGS, which is something that setuptools supported specifically for macOS.

It writes a Meson cross file out which is mostly correct, but it defines the C++ compiler in terms of the C preprocessor, which doesn't work. Fix this to use the correct program.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -682,7 +682,7 @@ class Project():
                     cross_file_data = textwrap.dedent(f'''
                         [binaries]
                         c = ['cc', '-arch', {arch!r}]
-                        cpp = ['cpp', '-arch', {arch!r}]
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Fixes https://github.com/scikit-image/scikit-image/issues/6667, probably.

dnicolodi:
Ups. Why have tests when you can have embarrassing bugs? Thanks @eli-schwartz

eli-schwartz:
Honestly, I really feel like Meson made a big mistake back in the day by calling the language identifier "cpp" instead of "cxx". It causes confusion like this more than once...

Issue #68, opened 2022-06-05
Controll about what is installed into binary wheel


peter-urban:
Currently "pip wheel -w dist ." installs also all headers and static libraries into the binary wheel for my project. 
This is unnecessary and also lets the builds fail on windows ('installing libraries (in my case a pkg-config file) is only supported in POSIX so far').

Is it possibly to control this behavior?

In "meson install" I can control what is install using tags e.g. meson install --tags python-runtime. 
Maybe supporting this option would be a solution?

If you share your thoughts on this and give me some directions I'd be willing to work on such a feature.

Thanks for this great Project!
Peter

rgommers:
@peter-urban thanks, this is interesting. Is your project public? Would be nice to see it. Otherwise, could you expand on how your `meson.build` files look like for the feature that's triggering this check, and why you don't need it installed?

> Is it possibly to control this behavior?

IIUC, not right now. But it sounds like it should be possible.

peter-urban:
@rgommers: thx for your response
The project is public here: https://github.com/themachinethatgoesping/themachinethatgoesping
It is very new, we are currently building up it's infrastructure.
The relevant meson.build are within this subproject: https://github.com/themachinethatgoesping/tools
I simply marked the targets I want to be installed in my python environment using mesons "install_tag : 'python-runtime'"

Admittedly I am new python packaging. So it is possible that I misunderstand some of the concepts. 
But to me it seems unnecessary clutter to install c++ headers and static libraries within a wheel that contains a build c++ binary module that works without these files?

Otherwise my request is triggered by a current practical problem:
Mesonpy fails with 'Bundling libraries in wheel is currently only supported in POSIX!' if it tries to copy e.g. a librarie's pkg-config file into the wheel on windows or mac.

FFY00:
Hi @peter-urban, thank you for raising this issue. I think this use-case could be meet by allowing users to specify Meson options, which is something we do want to do, but the design and UX around such settings is complex and so, I haven't got around to settling on a design.

rgommers:
> Admittedly I am new python packaging. So it is possible that I misunderstand some of the concepts.
> But to me it seems unnecessary clutter to install c++ headers and static libraries within a wheel that contains a build c++ binary module that works without these files.

So this is the `meson.build` file that contains things you don't want in your wheel: https://github.com/themachinethatgoesping/tools/blob/a11c2b18274386919c5d00c95458b40df0c2b974/src/tools/meson.build

[...]

PR #267, opened 2023-01-17
Add install-tags filtering for wheel generation

Brishen:
This is a PR for #68 comprising entirely of @peter-urban's work.


mesonpy/__init__.py
```diff
@@ -876,8 +876,12 @@ def build_commands(self, install_dir: Optional[pathlib.Path] = None) -> Sequence
...
     def build(self) -> None:
         """Trigger the Meson build."""
-        for cmd in self.build_commands():
-            self._meson(*cmd[1:])
+        self._meson('compile', *self._meson_args['compile'],)
+        if self._meson_args['install-tags']:
+            install_tags = '--tags=' + ','.join(self._meson_args['install-tags'])
+            self._meson('install', '--destdir', os.fspath(self._install_dir), install_tags, *self._meson_args['install'],)
+        else:
+            self._meson('install', '--destdir', os.fspath(self._install_dir), *self._meson_args['install'],)
```
FFY00, line 884: I think this may have happened because of the merge, but this code now needs to go into `build_commands`.

mesonpy/__init__.py
```diff
@@ -631,7 +631,7 @@ def build_editable(self, directory: Path, verbose: bool = False) -> pathlib.Path
         return wheel_file
 
 
-MesonArgsKeys = Literal['dist', 'setup', 'compile', 'install']
+MesonArgsKeys = Literal['dist', 'setup', 'compile', 'install', 'install-tags']
```
FFY00, line 634: I wonder it wouldn't just be better to parse this from the `install` args.

mesonpy/__init__.py
```diff
@@ -903,6 +907,12 @@ def _info(self, name: str) -> Dict[str, Any]:
 
     @property
     def _install_plan(self) -> Dict[str, Dict[str, Dict[str, str]]]:
+        install_plan = self._info('intro-install_plan').copy()
+
+        for files in install_plan.values():
+            for file, details in list(files.items()):
+                if details['tag'] not in self._meson_args['install-tags']:
+                    del files[file]
```
FFY00, line 915: I'm not sure if we should do it here.

mesonpy/__init__.py
peter-urban: True, this is a merge error, my code was still based on a very old version of meson-python

mesonpy/__init__.py
peter-urban: good point, I'll implement this
```suggestion
MesonArgsKeys = Literal['dist', 'setup', 'compile', 'install']
```

mesonpy/__init__.py
peter-urban: if install-tags is not separated from the "install" args, this whole block can be reverted to
`for cmd in self.build_commands():
    self._meson(*cmd[1:])`

mesonpy/__init__.py
peter-urban: i updated this block to parse the "install" args instead of using an extra 'install-tags' field
```suggestion
    def _install_plan(self) -> Dict[str, Dict[str, Dict[str, str]]]:
        """Meson install_plan metadata."""
    
        # copy the install plan so we can modify it
        install_plan = self._info('intro-install_plan').copy()

        # parse install args for install tags (--tags)
        install_tags = []
        for arg in self._meson_args['install']:
            if arg.strip().startswith('--tags='):
                install_tags = arg.split('=', 1)[1].split(',')
                break
        else:
            return install_plan
            
        # filter out files that do not fit the install tags
        for files in install_plan.values():
            for file, details in list(files.items()):
                if details['tag'].strip() not in install_tags:
                    del files[file]

        return install_plan
```

mesonpy/__init__.py
FFY00: Thanks! This can be done with `argparse` and [`parse_known_args`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_known_args).

peter-urban:
@Brishen I implemented the above made code suggestions in my fork. Could you update your PR? :-)

mesonpy/__init__.py
peter-urban: @FFY00 This is just a suggestion (it was easy to implement it here). 
If you give me some hints or thoughts on your doubts I could look for a better place to put this filter.

FFY00: This is a bit fragile, and I'd prefer not having to maintain it. I think it should be pretty easy to use `argparse`, as I mentioned in https://github.com/mesonbuild/meson-python/pull/267#discussion_r1072979001. What do you think?

peter-urban: Oh ja, I didn't see the comment 🙂 I'll update the suggestion tomorrow

FFY00: No worries. Thank you :blush:

peter-urban: Adapted suggestion

```suggestion
    """Meson install_plan metadata."""
    
        # copy the install plan so we can modify it
        install_plan = self._info('intro-install_plan').copy()

        # parse install args for install tags (--tags)
        parser = argparse.ArgumentParser()
        parser.add_argument('--tags')
        args, _ = parser.parse_known_args(self._meson_args['install'])
        
        # filter the install_plan for files that do not fit the install tags
        if args.tags:
            install_tags = args.tags.split(',')
                
            for files in install_plan.values():
                for file, details in list(files.items()):
                    if details['tag'].strip() not in install_tags:
                        del files[file]

        return install_plan
```

naturally this needs `import argparse` at the beginning

peter-urban:
@Brishen is it possible for you to update this pull request to adapt it to FFY00s comments? 
You could again merge the it again against the main branch of my fork. :-) 
Cheers

peter-urban:
Hi @Brishen, it seems that you have no time to update this PR. I grew a bit impatiens and, since this PR is based on my changes anyways, took the liberty to create a new, updated PR (https://github.com/mesonbuild/meson-python/pull/286) 

Hope that is no problem for you :-)

FFY00:
Closed in favor of #288.

Issue #233, opened 2022-11-24
Document meson-python requirements on how `import('python').find_installation()` is called


dnicolodi:
meson-python invokes meson and meson may be run by a different Python interpreter than the one used to run meson-python via pip or build or other pep517 frontend. meson-python uses a native file to pass the python interpreter to build for to meson. AFAIK, this works only if a path to a Python interpreter is not explicitly passed as an argument to `import('python').find_installation()`. This should be documented.

kif:
+1:
I was debugging an issue with multiple interpreter  (python 3.10 + 3.11) on a "remote" architecture (i386 chroot on amd64 or armv7 on aarm64) outside of any virtual-env (for debian packaging) thus specifying the interpreter is very important to me.
Having it documented is equally important.

dnicolodi:
What issue did you encounter? meson-python works in the same way as other Python build tools work: the extensions are built for the interpreter running the build process.

kif:
I think debian is part of the problem since they did not yet move completely away from calling `setup.py`. The DD in charge of my packages (@picca) asked me to keep the `distutils` part for the year to come to be able to continue packaging them. Debian did neither package scipy 1.9.

That said, I don't think illogical to have a way to say `python=/usr/bin/pyhton310` or `python=/usr/bin/python311`. The same way I encountered a similar bug with cython not found since on debian, `cython` refers to the cython for python2 and one has to call `cython3`. Once again those are remarks for the documentation when packaging in the PEP517 spirit, and probably not bugs specific to `meson-python`.

rgommers:
> Debian did neither package scipy 1.9.

That was not because they can't use any other build system I believe. It was because (a) SciPy 1.9.0 wasn't quite robust enough, so I recommended downstream packagers to use the `setup.py`-based build for that release, and (b) Debian has an issue where they can only use a single `setuptools` version, and they unwisely chose the most recent release, leaving any projects that have issues with `setuptools` breaking compatibility with `distutils` out in the cold.

This is a different issue though than the original topic of this issue.

eli-schwartz:
> The same way I encountered a similar bug with cython not found since on debian, `cython` refers to the cython for python2 and one has to call `cython3`.

Since Meson 0.63.0, if you use the builtin cython language support it will search for both names: https://github.com/mesonbuild/meson/commit/ebbe4425e7a1c1c16281ef163e443fc1c0a8b209

Issue #224, opened 2022-11-22
Document that Meson and thus meson-python require MSVC compilers to be found in `$PATH`


dnicolodi:
setuptools has its magic to find the compilers but Meson does not thus this may come as a surprise especially to people like me that compiles on Windows only on CI jobs. Because many CI environments have images where the GCC compilers are available via MSYS2 or similar, Meson picks up those by default and things break in unexpected ways because the right command line arguments should be passed to these compilers to correctly build Python extensions. 

One more related issue is that `cibuildweels` (I don't know how popular but definitely convenient way to build distributable wheels) most prominently way to build wheels in GitHub Actions is using a reusable action that does not play nicely with adding the MSVC compilers in `$PATH`. The solution is to use `pipx run cibuildwheel` directly. @henryiii can probably fill in the details.

All this should be documented somewhere.

eli-schwartz:
`meson setup --vsenv` will force Meson to run the `vcvars*.bat` script and acquire its information on finding MSVC before any other compiler. Without that option, Meson does indeed pick an msys2 compiler from $PATH.

henryiii:
I've looked into updating the action to run the native pwsh backend on Windows. Unfortunately, it seems powershell doesn't like empty args, that is, `--flag ""` complains about flag not getting an argument. We use this for a couple of arguments, so that's where I'm stuck at the moment.

henryiii:
Here's the run if you are curious: https://github.com/pybind/scikit_build_example/actions/runs/3519251284/jobs/5899006227

```
Run henryiii/cibuildwheel@henryiii/fix/action
with:
  package-dir: .
  output-dir: wheelhouse
env:
  CIBW_ARCHS_MACOS: auto universal2
Run actions/setup-python@v4
with:
  python-version: 3.7 - 3.11
  update-environment: false
  check-latest: false
  token: ***
env:
  CIBW_ARCHS_MACOS: auto universal2
Successfully set up CPython (3.11.0)
Run pipx run --python "C:\hostedtoolcache\windows\Python\3.11.0\x64\python.exe" --spec "D:\a\_actions\henryiii\cibuildwheel\henryiii/fix/action" cibuildwheel "." --output-dir "wheelhouse" --config-file "" --only ""
[...]

Issue #159, opened 2022-09-29
Grab version from __version__ attribute


lithomas1:
meson-python should provide a way to use the __version__ attribute of a package as its version instead of using the version from meson.

rgommers:
Thanks @lithomas1. xref https://github.com/pandas-dev/pandas/pull/47081#issuecomment-1257147672 for more context.

There are multiple things folks want to do with version names in wheel files, see e.g. `versioneer`, `setuptools_scm` and SciPy's [custom code](https://github.com/scipy/scipy/blob/main/tools/version_utils.py) to produce a version. It doesn't seem like a good idea to support all that directly in `meson_python`. Allowing to use the `__version__` attribute string seems like a way to avoid that.

One issue with `__version__` is that it will not work for cross-compiling, because the produced wheel is in general not runnable. Also, I think it would require installing and running the wheel, which is too expensive. So a slight tweak on this proposal is using a hook of some sort to let the package author indicate to `meson-python` what the version is. Either as standalone runnable Python code, or by writing the version string to a file.

I am +1 on adding the feature, because this is something a lot of folks want - and want it enough to go through a _lot_ of effort to get it.

eli-schwartz:
I don't think that it's a problem for cross compiling. setuptools has long supported (via setup.cfg but not setup.py, IIRC) reading versions from an attribute of a source file, commonly `__version__` -- it does so using:
https://github.com/pypa/setuptools/blob/646c71f9af7e9ba6677203acde4ccd0a478dfbf8/setuptools/config/expand.py#L62-L86

The tl;dr is that it takes `foo.bar.baz` and chops off the last bit to treat as a module attr, then uses the rest to find a suitable `.py` file which it parses via the `ast` module, searching for an assignment to that attr and parsing its value with `ast.literal_eval`. This is explicitly designed to do no arbitrary code execution at all, and only supports simple statements that *can* be literal_eval'ed, fortunately that is all that stuff like versioneer or setuptools_scm usually do.

Setuptools actually does fall back to directly importing the file if literal_eval fails. That's not precisely required, and obviously also has cross-compile concerns.

rgommers:
Thanks @eli-schwartz, interesting, I didn't know `setuptools` had such a feature. I gave it a spin anyway, it's a little fiddly:

```python
>>> spec = _find_spec('scipy', None)
>>> spec
[...]

PR #276, opened 2023-01-27
DOC: major documentation rewrite

### LINKED ISSUE #138: Document support for installing python scripts
How can I install python scripts with `meson-python` replacing setuptool's `console_scripts` `enty_point` or `[project.scripts]` in `pyproject.toml`? If there is a vision of how this should look like I'm happy to try to work on the support.

### LINKED ISSUE #197: Rework documentation

### LINKED ISSUE #230: config-settings support needs docs
This is a follow-up issue for gh-167. The most important thing to add is docs. The invocations are verbose and argument quoting can be difficult, so we need a good number of one-liners as usage examples (there are some near the bottom of gh-167).

Also, it's possible that there's still an issue with functionality. `--config-settings` doesn't take multiple arguments at once, so if you'd want to pass multiple arguments to `meson setup`, it seems like there's no good way to do it. Something like:
```
python -m pip install . --no-build-isolation --config-settings='setup-args=-Dblas=blas' --config-settings='setup-args=-Dlapack=lapack'
```
seems to just use one of the two arguments. @tacaswell and I tried for a while, and didn't get it to work. It's possible we missed something.

### LINKED ISSUE #253: Add the PyPA discord as an official project communication channel
I have created a `meson-python` channel in the "Packaging Related" (meaning it's affiliated with the PyPA) category of the PyPA discord. Is everyone okay in documenting this as one of the official communication channels for the project?

https://discord.gg/pypa

### LINKED ISSUE #264: Consider dropping direct references to PEP 517
See https://discuss.python.org/t/name-for-pyproject-toml-builds-and-peps/1659 for the rationale.

Basically, we've been moving to a world where "pyproject.toml-based builds" is a much more consistent and better way to refer to such build-backends. We don't need to point users to dense design documents to understand how stuff works.

FFY00:
Closes #197, closes #230, closes #253, and closes #264
Sets us up to fix later #138, #233, and #224


FFY00:
Sorry for the big PR, I'd recommend you having a look to see if there is anything major wrong. If there isn't anything major wrong, we should probably merge, and then we can send followup PRs to improve it.

docs/tutorials/introduction.rst
```diff
@@ -0,0 +1,307 @@
...
+
+   If you are not familiar with the TOML configuration language or need some
+   help, be sure to check out
+   `their documentation <https://toml.io/en/v1.0.0>`__.
+
+
+Generating the distribution artifacts
+=====================================
+
+Now that we have a valid Python project, we can release it.
```
rgommers, line 175: There is one issue here I think (also related to the FIXME 3 paragraphs down): you go straight from metadata to a release, without local use first. How about inserting a section _"Building and installing our package"_ first? Touching on `pip install .`, mentioning that builds a wheel for the current Python interpreter and then installs it. Show that it's installed with `pip list`, then import it. Conclude that we now have a working package. Next step: a release.

docs/tutorials/introduction.rst
```diff
@@ -0,0 +1,307 @@
...
+.. admonition:: Check the pypa/build documentation
+   :class: seealso
+
+   To learn more about `pypa/build`_, you can check
+   `their documentation <https://pypa-build.readthedocs.io/>`__.
+
+
+.. admonition:: Building wheels for many systems
+   :class: tip
+
```
rgommers, line 240: I'd say here that for a pure Python package, all one needs is one sdist and one wheel for a release. So we'd be good at this point to upload to PyPI. For a package with an extension module, one needs multiple wheels - they are specific to the OS, hardware, and (unless one uses the stable ABI) the Python minor version. So `.. warning::` this must be done with `cibuildwheel`.

docs/how-to-guides/add-dependencies.rst
```diff
@@ -0,0 +1,13 @@
+.. _how-to-guides-add-dependencies:
+
+*******************
+Adding dependencies
```
rgommers, line 4: Is this page needed? At least it should start by stating that this is all standardized, and nothing is specific to `meson-python`. It looks like the Python Packaging User Guide has no good section on this - maybe we should fix that (after finishing up this PR) so we can link to it for more details.

docs/how-to-guides/build-directory.rst
```diff
@@ -0,0 +1,14 @@
+.. _how-to-guides-build-directory:
+
+*******************************
+Use a permanent build directory
+*******************************
+
+
+.. todo::
+
+   - Explain why you want to use the ``builddir`` option
```
rgommers, line 10: This includes things that are hard to pass through a front end's config-settings I think. Would be good to point out that this works as expected when you, for example, run `meson setup --cross-file ...`, and `meson-python` will reconfigure only what's needed.

docs/how-to-guides/config-settings.rst
```diff
@@ -0,0 +1,45 @@
...
+
+On `pip`_, they can be used via the ``--config-settings`` option.
+
+
+.. code-block:: console
+
+   $ pip install . --config-settings key=value
+
+
+Please check the `pypa/build documentation`_ for more information.
```
rgommers, line 39: typo: `pip` docs, not `build` docs

docs/how-to-guides/editable-installs.rst
```diff
@@ -0,0 +1,79 @@
...
+This will install the project in the current directory in editable mode. When a
+``meson-python`` package is installed in editable mode, it will be re-built the
+first time it is imported. That means, the first time you import a module from
+your package, it will take a bit longer.
+
+
+.. admonition:: What is the scope of editable installs?
+   :class: seealso
+
+   The editable installs feature only targets the Python modules of the project.
```
rgommers, line 25: This sounds like pure Python. It's worth pointing out that, unlike in any other build backend, files for an extension module (C/C++/Cython/whatever) are also editable and will be automatically rebuilt.

docs/how-to-guides/editable-installs.rst
```diff
@@ -0,0 +1,79 @@
...
+you simply need to set the ``MESONPY_EDITABLE_VERBOSE`` environment variable
+to any non-null value.
+
+
+.. code-block:: console
+
+   $ MESONPY_EDITABLE_VERBOSE=1 python
+
+
+Enable it permanently
```
rgommers, line 54: Just deleting these section headers will make the page look better. If it's one sentence + one line of code, that's a sign that it's too little for its own section.

docs/how-to-guides/editable-installs.rst
```diff
@@ -0,0 +1,79 @@
...
+
+
+.. admonition:: ``MESONPY_EDITABLE_VERBOSE`` will have no effect here
+   :class: attention
+
+   Please note that the ``MESONPY_EDITABLE_VERBOSE`` environment variable will
+   not have **any** effect on the installation, it simply forces verbose mode to
+   be enabled **after** the package has been installed.
+
+
```
rgommers, line 76: The limitations when working with an IDE (breakpoints, running tests from the UI) are worth pointing out here. And saying work is planned to improve this.

docs/how-to-guides/executable-with-internal-dependencies.rst
```diff
@@ -0,0 +1,20 @@
+.. _how-to-guides-executable-with-internal-dependencies:
+
+*************************************
+Executable with internal dependencies
+*************************************
+
+
+.. todo::
+
```
rgommers, line 9: The macOS PR for this is about ready to merge, and we can probably sort Windows out soon. So maybe the static library explanation isn't needed?

docs/how-to-guides/editable-installs.rst
rgommers: This applies to other pages as well. For example, the _Use build config settings_ does not need section headers for `pypa/build` and `pip`.

docs/how-to-guides/meson-args.rst
```diff
@@ -0,0 +1,17 @@
...
+
+***************************
+Passing arguments to Meson_
+***************************
+
+
+.. todo::
+
+   - Explain how you can add extra Meson arguments via ``pyproject.toml``
+   - Explain how you can add extra Meson arguments via the build conifig
```
rgommers, line 11: minor: typo in last word here

docs/reference/limitations.rst
```diff
@@ -0,0 +1,50 @@
+.. _reference-limitations:
+
+***********
+Limitations
+***********
+
```
rgommers, line 6: The issue with the `license` key in `pyproject.toml` is probably worth pointing out.

rgommers (commented):
Thanks @FFY00! This looks like a good structure. I added some comments from a first read.

Content suggestions I have:
- a how-to for controlling what files get included in an sdist (with `.gitattributes`) and that a change to a vcs-tracked file gets included but a new file does not. It's mentioned briefly in the section on sdist, but could use more detail and an example `.gitattributes file
- a tutorial page on cross-compiling (I can fill it in, I am dealing with bits and pieces of this in multiple places already)
- a how-to page on including a VCS-generated version (e.g. with `fs.exists` snippet in https://github.com/mesonbuild/meson-python/issues/159#issuecomment-1265962676)
- A _Who uses meson-python_ page, linked from the front page.

rgommers:
> Sorry for the big PR, I'd recommend you having a look to see if there is anything major wrong. If there isn't anything major wrong, we should probably merge, and then we can send followup PRs to improve it.

It seems like a good strategy to merge when the structure is right (it matches the suggestion in gh-197; let's give everyone a couple of days to have at least a high-level look at this) and what is there in terms of text is in decent shape. So we get the improved docs for the issues that this already closes. And then we can parallelize filling in the most important blanks.

rgommers:
By the way, on updating this PR: it would be quite useful to not rewrite history as you go. That makes it impossible to check what changed for reviewers. Rather, push new commits as updates, and then squash commits right at the end.

FFY00:
> By the way, on updating this PR: it would be quite useful to not rewrite history as you go. That makes it impossible to check what changed for reviewers. Rather, push new commits as updates, and then squash commits right at the end.

Yeah, sorry. Github isn't a great git review platform. It would be awesome if they added support for rebase-merges with autosquash, that way I could push fixup commits for specific commit, and not have to squash the whole history. Currently, to update a specific commit I need to force-push, with the only alternative being committing to squash the whole history on merge. FYI you can actually click in "force-pushed" in the message to see what changed.

I want to push some of your feedback as separate commits, so instead of force-pushing, I can just push the fixup commits without squashing for review, and then when it comes time to merge, I will squash them and merge. Would that work for you? The other, more annoying, alternative would be to just push those changes as a separate PR.

docs/tutorials/introduction.rst
```diff
@@ -0,0 +1,307 @@
...
+we want it to build, and how to do it.
+
+
+.. code-block:: meson
+   :caption: meson.build
+
+   project('purelib-and-platlib', 'c')
+
+   py_mod = import('python')
+   py = py_mod.find_installation()
```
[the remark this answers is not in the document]
FFY00: From https://mesonbuild.com/Python-module.html#find_installation

>  `pure`: On some platforms, architecture independent files are expected to be placed in a separate directory. However, if the python sources should be installed alongside an extension module built with this module, this keyword argument can be used to override the default behavior of `.install_sources()`. since 0.64.0 

Hummm... I do not follow. AFAICT this will just force sources to be placed in platlib.

docs/tutorials/introduction.rst
```diff
@@ -0,0 +1,307 @@
...
+
+.. admonition:: Writing TOML
+   :class: seealso
+
+   If you are not familiar with the TOML configuration language or need some
+   help, be sure to check out
+   `their documentation <https://toml.io/en/v1.0.0>`__.
+
+
+Generating the distribution artifacts
```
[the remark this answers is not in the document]
FFY00: > Maybe this is too technical a title. How about _"Creating a first release of your package"_?

I changed it to "Creating your first release" :+1:

docs/how-to-guides/add-dependencies.rst
FFY00: I didn't want to get into it in the tutorial, but wanted to have somewhere to point to quickly learn this. I'd say keep it for now, try to fix the Packaging User Guide, and then update this based on the outcome.

docs/how-to-guides/build-directory.rst
FFY00: Could you elaborate? I don't think I follow :sweat_smile:

docs/how-to-guides/editable-installs.rst
FFY00: Yeah :sweat_smile:. I wanted to use smaller headers, but rST doesn't allow that.

docs/how-to-guides/executable-with-internal-dependencies.rst
FFY00: Well, it's already written :rofl:. We can just keep it and remove when it's fixed, I don't think it'd be much trouble.

docs/how-to-guides/meson-args.rst
FFY00: Yeah, I also caught that locally when I added spellchecking to the docs, which I will push in a separate commit. I was originally thinking of doing a different PR, but it'll have conflicts and it's pretty simple anyway, so I will probably push it here instead.

FFY00:
I fixed the smaller stuff you mentioned, I will do the rest monday :+1:

docs/who-uses-meson-python.rst
```diff
@@ -0,0 +1,26 @@
+.. _who-uses-meson-python:
+
+**************************
+Who uses ``meson-python``?
+**************************
```
dnicolodi, line 5: I don't like adding a list of projects using meson-python to the documentation, or anywhere else. As we don't have any need for advertisement (and I'm pretty sure we don't want to be advertising any other project) it does not serve any purpose and it is going to be a burden to maintain. For the users, it is much better to have a short and curated list of projects that follow best practices and that can be used as examples.

Unless the listed projects are known to following best practices, the list is not going to be useful for the users. I don't think we have resources to review all the projects submitted to be listed, and to keep our review current.

I would much prefer it this would be something like:
```
 Curated list of projects using ``meson-python``
===================================

The following projects use Meson as their build system and meson-python to produce Python wheels. They are roughly listed in order of complexity:

- `siphash24 <https://github.com/dnicolodi/siphash24>`_ A very simple project. It demonstrates how Meson makes it trivial to compile a CPython extension written in Cython via a simple tempate engine and link it to a library compiled from a Meson subproject. Also an example of how to use `cibuildwheel`_ to produce Python wheels for several platforms.

- (i don't know the peculiarities of the other projects)

- `SciPy <https://github.com/scipy/scipy>`_ Probably the most complex Python project using Meson and ``meson-python``. Staring at the build definition too long may cause headaches. 

```

docs/tutorials/introduction.rst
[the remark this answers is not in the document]
rgommers: Which is what you want, since you have C code in your package. This is a safety measure; otherwise you have to plaster `pure: false` everywhere you use `py.install_sources`, and if you forget it just once you end up with a broken package. That's why it was added: https://mesonbuild.com/Release-notes-for-0-64-0.html#pythonfind_installation-now-accepts-pure-argument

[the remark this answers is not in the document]
dnicolodi: Why do you end up with a broken package? Is there an assumption somewhere that all the code in a package needs to be installed either in `{platlib}` or in `{purelib}`? meson-python does not do this assumption, and actually has code to handle this case. If this is the case, it would be nice to detect the circumstance and raise an error in meson-python. However, I don't see why this is the case. Also, in most platforms, `{platlib}` and `{purelib}` are the same path.

docs/how-to-guides/build-directory.rst
rgommers: I've come across this multiple times already, for cross-compilation in particular. See for example https://github.com/scipy/scipy/blob/1c836efe5ff37ffa4490756269b060a464690e62/.github/workflows/wheels.yml#L180-L188.

Also for conda-forge, you need in `build.sh` something like this for cross-compiling (the first line is how the conda-forge build tooling configures automatically for cross-compilation):
```
meson ${MESON_ARGS} builddir
python -m build -Cbuilddir=builddir
```

docs/tutorials/introduction.rst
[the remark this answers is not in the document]
rgommers: > Is there an assumption somewhere that all the code in a package needs to be installed either in `{platlib}` or in `{purelib}`?

Yes of course. Say you have a `foo.py` file and `bar.so` extension module, and you do a relative import `from .bar` in `foo.py`, that only works if the two files are in the same directory. Which is the kind of thing you only see is broken when some niche Linux distro doesn't have purelib and platlib pointing to the same path (not Ubuntu, which is typically the one thing people test in their CI). This whole purelib/platlib concept is just awful.

> meson-python does not do this assumption, and actually has code to handle this case. If this is the case, it would be nice to detect the circumstance and raise an error in meson-python.

Not every build goes only through `meson-python`. If it's broken for plain `meson setup/install`, that's still bad. Raising an error SGTM.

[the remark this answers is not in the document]
dnicolodi: Ah, that's obvious, but I can have a wheel that installs multiple top-level modules. And those can be installed some in `{platlib}` and some in `{purelib}`. Saying that if you don't pass `pure: false` you get a broken wheel is a simplification that I don't think we should make.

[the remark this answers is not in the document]
rgommers: Multiple top-level modules is a corner case that we're not dealing with here. I'm not asking to say _" if you don't pass `pure: false` you get a broken wheel "_. I am asking to please use the robust `pure: false` as the default everywhere, instead of the very fragile state of not doing that. Which I'm quite sure is the right thing to do (for more details, please see https://github.com/mesonbuild/meson/issues/10523).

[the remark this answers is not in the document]
eli-schwartz: Setting this on the installation object will flip the default value on the install_sources() and get_install_dir() functions. This is the correct, sensible, sane, and *only* reasonable thing to do if you have a top-level module with, nested inside it, a mixture of pure python and compiled extensions: it means that by default, you install pure python together with the compiled extension.

It doesn't prevent you from passing `pure: true` to individual calls where you do intend to install some freestanding top-level pure-python modules to purelib anyways. And I agree, that's a valid use case -- although I expect it's also *incredibly* rare.

Failing to place pure python files in the platlib when they are nested inside a package together with platlib extensions, means that you end up with one package being a namespace package and the other not, with corresponding... abuses... of the importlib search path. You are not going to end up happy if that happens, because at least one user is going to be on a distro which splits the two, and then disaster is going to happen and it's not going to be pretty.

We should absolutely document the default recommended approach, as the one that doesn't break people. :)

FFY00:
The force-push was from a rebase to solve conflicts. The commits being reviewed haven't been changed other than just solving conflicts.

As mentioned previously, updates to the original commits are being done via fixup commits. They will be squashed after approval, just before merging.

docs/who-uses-meson-python.rst
FFY00: Are the new updates good for you, or do you change the wording further?

docs/tutorials/introduction.rst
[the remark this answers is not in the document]
FFY00: I think it's reasonable to set that, but IMO it would be best to not set it, and add an admonition below pointing to a page that explains this in-depth.

> Failing to place pure python files in the platlib when they are nested inside a package together with platlib extensions, means that you end up with one package being a namespace package and the other not, with corresponding... abuses... of the importlib search path. You are not going to end up happy if that happens, because at least one user is going to be on a distro which splits the two, and then disaster is going to happen and it's not going to be pretty.

We should try to detect these cases and warn users.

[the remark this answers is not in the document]
FFY00: I added this

```rst
.. admonition:: Mixing pure and native code
   :class: caution

   If you are mixing pure (``.py``) and native modules in your project, please
   check out :ref:`reference-quirks-mixing-purelib-and-platlib`, which explains
   the common foot-guns when doing so, and how to make sure you install your
   modules correctly.
```

FFY00:
@rgommers I think I've address all your comments, could you do a second pass?

docs/tutorials/introduction.rst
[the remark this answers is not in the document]
rgommers: @FFY00, I disagree - this is not a choice between two equivalent alternatives. `pure: false` is so far better, it must be used. You have no idea how much time this has already cost me in subtle bugs, and how tricky it is to debug. Please just accept the comment. Keeping a bad default and then adding cognitive load for the user with a note explaining why they should not follow your example code because of footguns is not useful.

docs/explanations/design.rst
```diff
@@ -0,0 +1,26 @@
...
+*******************************
+
+
+.. todo::
+
+   - Explain how ``meson-python`` maps files to the wheel
+
+     - Link to :ref:`tutorials-executable` in the relevant section
+
+   - Explain why the heuristics are necessary and how they work
```
rgommers, line 18: Aren't we about to remove those heuristics? Or are these not the ones for `{moduledir_shared}` & co?

docs/how-to-guides/build-directory.rst
```diff
@@ -0,0 +1,19 @@
...
+Use a permanent build directory
+*******************************
+
+
+.. todo::
+
+   - Explain why you'd want to use the ``builddir`` option
+     - Mention it works properly when using ``--cross-file`` (eg. https://github.com/scipy/scipy/blob/1c836efe5ff37ffa4490756269b060a464690e62/.github/workflows/wheels.yml#L180-L188)
+   - Explain how ``builddir`` works
+   - Warn that user-configured build directories are not "use at your own risk"
```
rgommers, line 17: I think the intent here was "are" rather tha" are not" (at your own risk)? That said, I'd rephrase as "Explain the constraints that the user must adhere to when using a user-configured build directory". That also takes care of the second warning the line below.

docs/reference/limitations.rst
```diff
@@ -0,0 +1,76 @@
...
+interpreter instances at the same time is not supported.
+
+
+Using editable installs with IDEs
+=================================
+
+Currently, with editable installs, setting breakpoints via an IDE or similar
+tool will not work.
+
+We have work planed to fix this issue.
```
rgommers, line 38: minor: typo in "planed"

docs/tutorials/introduction.rst
```diff
@@ -0,0 +1,355 @@
...
+
+As you might have figured out by the name, *sdists* contain the source code of
+the project, and *wheels* contain a compiled [#pure-wheels]_ version of the
+project, ready to be copied to the file-system.
+
+If your project has compiled code, its *wheels* will be platform-dependent, and
+if it has native Python modules, the *wheels* will also be Python version
+specific, as the native Python modules depend on the Python ABI.
+FIXME(this seems *a bit* convoluted, it isn't awful, but I feel like it could
+be simplified or explained in a more natural way)
```
rgommers, line 238: This still needs addressing. I'd start with the common case:
```
If your project uses Python extension modules, your *wheels* will be specific to
both the platform and the Python version (footnote: unless you use CPython's stable ABI).
```

I would not even bother with a case where there is native code, but no extension modules - this is quite unlikely to happen (it'd require using `ctypes` or `cffi` to do anything with the native code).

docs/tutorials/introduction.rst
```diff
@@ -0,0 +1,355 @@
...
+
+Strictly speaking, you don't need to distribute *wheels*, however they make the
+user experience much nicer. Unless you have any reason not to, we highly
+recommend you distribute *wheels* for at least the most common systems. When
+*wheels* are not available for a system, the project can still be installed, be
+it needs to be build from the *sdist*, which involves fetching all the build
+dependencies and going through the likely expensive build.
+FIXME(strictly speaking -- this doesn't sound great, how do we highlight the
```
rgommers, line 246: How about: 
```
While distributing *wheels* is not mandatory, they make the
user experience much nicer. Unless you have any reason not to, we highly
```

docs/who-uses-meson-python.rst
```diff
@@ -0,0 +1,34 @@
...
+
+Here's a curated list of projects using ``meson-python``.
+
+
+.. list-table::
+   :widths: 20 80
+
+   * - `SciPy <https://github.com/scipy/scipy>`_
+     - Probably the most complex project using Meson and ``meson-python``.
+       Staring at the build definition too long may cause headaches.
```
rgommers, line 19: I'm not sure if I should confirm that this is true, or be slightly offended:)

docs/who-uses-meson-python.rst
```diff
@@ -0,0 +1,34 @@
...
+
+   * - `scikit-image <https://github.com/scikit-image/scikit-image>`_
+     - Another complex project using ``meson-python``.
```
rgommers, line 22: I suggest: "A project of medium complexity, with a lot of Cython code"

(no need to repeat "using meson-python" - that's above the table already)

rgommers (commented):
This is pretty close. I suggested some rewordings for the FIXMEs.

docs/who-uses-meson-python.rst
eli-schwartz: Anyone who gets headaches from looking at SciPy's build definition should definitely not look at the systemd build definition, is all I can say. :)

docs/explanations/design.rst
FFY00: We are, this was written before that PR.

docs/how-to-guides/build-directory.rst
FFY00: Yes. Thank you for catching that.

docs/tutorials/introduction.rst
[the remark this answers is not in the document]
FFY00: I changed it.

IMO this is the kind of issue we can, and should be trying to detect.

rgommers (approved):
LGTM now, thanks @FFY00

docs/who-uses-meson-python.rst
FFY00: @dnicolodi please let me know if you'd like further changes, I think this is the only blocker for merging this.

docs/how-to-guides/editable-installs.rst
```diff
@@ -0,0 +1,84 @@
...
+Verbose mode
+============
+
+It might be useful for you to see the output of Meson_ when the package is being
+re-built. We provide a *verbose mode* that does this, and can be enabled either
+temporarily, after the install, or permanently, in the install.
+
+
+To enable the verbose mode on any existing ``meson-python`` editable install,
+you simply need to set the ``MESONPY_EDITABLE_VERBOSE`` environment variable
```
dnicolodi, line 51: It is not related to this PR but it came to my attention scrolling through the added documentation: I think that the `mesonpy` name should not be used and that we should stick with `meson-python` everywhere. Thus I would prefer if we standardize on `MESON_PYTHON_` as the prefix for environment variable names.

docs/who-uses-meson-python.rst
dnicolodi: I didn't expect it to be taken seriously when I wrote that sentence. Unless everyone is happy with the pun, I would replace with something a bit more useful. Something like: "It combines CPython extensions and libraries written in C, C++, Fortran, Cython, Pythran, different ASM variants, ..." maybe something about CPU feature detection or BLAS libraries. I'm sure @rgommers can suggest something.

docs/who-uses-meson-python.rst
```diff
@@ -0,0 +1,34 @@
...
+
+**************************
+Who uses ``meson-python``?
+**************************
+
+Here's a curated list of projects using ``meson-python``.
+
+
+.. list-table::
+   :widths: 20 80
```
dnicolodi, line 15: I haven't see how this is rendered, but I don't know if a table is the best choice. Maybe a simple itemized list is better.

docs/who-uses-meson-python.rst
dnicolodi: I'm not completely happy with the section name. I would prefer something like "Projects using meson-python".

docs/who-uses-meson-python.rst
```diff
@@ -0,0 +1,34 @@
...
+
+   * - `scikit-image <https://github.com/scikit-image/scikit-image>`_
+     - A project of medium complexity, with a lot of Cython_ code.
+
+   * - `siphash24 <https://github.com/dnicolodi/python-siphash24>`_
+     - A very simple project. It demonstrates how Meson makes it trivial to
+       compile a `CPython extension`_ written in `Cython`_ via a simple template
+       engine and link it to a library compiled from a Meson sub-project. Also
+       an example of how to use `cibuildwheel`_ to produce Python wheels for
+       several platforms.
```
dnicolodi, line 29: I used this little project of mine just because I know it, I'm not sure it is a good example. Feel free to pick another one.

dnicolodi (approved):
I had a look at that section only. I have some suggestions on the wording, but nothing major. I have a comment unrelated to this PR that we can discuss separately.

docs/who-uses-meson-python.rst
FFY00: I tried that, but preferred the list.

docs/who-uses-meson-python.rst
FFY00: There aren't *that* many projects out there right now. I originally added the main ones that seemed alright to me. Anyway, it is good to have a simpler one as an example.

FFY00:
I pushed a final update, [here](https://github.com/mesonbuild/meson-python/pull/276/commits/8ab7fbdfcae8b0d01b72a5191a9d6eef359c32f6). I am not rebase and merge now.

@rgommers please do send a PR re-writing the SciPy description if you feel anything can be improved there.

rgommers:
> @rgommers please do send a PR re-writing the SciPy description if you feel anything can be improved there.

Okay, will do.

PR #277, opened 2023-01-28
ENH: move to ruff

FFY00:

### DIFF
--- a/mesonpy/_introspection.py
+++ b/mesonpy/_introspection.py
@@ -20,8 +20,8 @@ def debian_python() -> bool:
         import distutils
         try:
             import distutils.command.install
-        except ModuleNotFoundError:
-            raise ModuleNotFoundError('Unable to import distutils, please install python3-distutils')
+        except ModuleNotFoundError as exc:
+            raise ModuleNotFoundError('Unable to import distutils, please install python3-distutils') from exc
         return 'deb_system' in distutils.command.install.INSTALL_SCHEMES
     except ModuleNotFoundError:
         return False
--- a/mesonpy/_tags.py
+++ b/mesonpy/_tags.py
@@ -60,13 +60,13 @@ def get_abi_tag() -> str:
     # https://github.com/pypa/packaging/pull/607.
     try:
         empty, abi, ext = str(sysconfig.get_config_var('EXT_SUFFIX')).split('.')
-    except ValueError:
+    except ValueError as exc:
         # CPython <= 3.8.7 on Windows does not implement PEP3149 and
         # uses '.pyd' as $EXT_SUFFIX, which does not allow to extract
         # the interpreter ABI.  Check that the fallback is not hit for
         # any other Python implementation.
         if sys.implementation.name != 'cpython':
-            raise NotImplementedError
+            raise NotImplementedError from exc
         return _get_cpython_abi()
 
     # The packaging module initially based his understanding of the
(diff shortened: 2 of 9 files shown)


FFY00:
This actually caught a bad test!

tests/test_wheel.py
```diff
@@ -98,7 +98,7 @@ def test_contents(package_library, wheel_library):
         re.escape('library-1.0.0.dist-info/RECORD'),
         re.escape('library-1.0.0.dist-info/WHEEL'),
         re.escape('library.libs/libexample.so'),
-    ]):
+    ], strict=True):
```
dnicolodi, line 101: `zip` has a `strict` argument only starting from Python 3.10.

mesonpy/_introspection.py
```diff
@@ -20,8 +20,8 @@ def debian_python() -> bool:
         import distutils
         try:
             import distutils.command.install
-        except ModuleNotFoundError:
-            raise ModuleNotFoundError('Unable to import distutils, please install python3-distutils')
+        except ModuleNotFoundError as exc:
+            raise ModuleNotFoundError('Unable to import distutils, please install python3-distutils') from exc
```
dnicolodi, line 24: I don't get the purpose of this change. AFAIK, within the handling of an exception, the `raise ... from` form is exactly the same as the one omitting the `from`. If anything, here I would use `... from None`.

mesonpy/_tags.py
```diff
@@ -60,13 +60,13 @@ def get_abi_tag() -> str:
...
         empty, abi, ext = str(sysconfig.get_config_var('EXT_SUFFIX')).split('.')
-    except ValueError:
+    except ValueError as exc:
         # CPython <= 3.8.7 on Windows does not implement PEP3149 and
         # uses '.pyd' as $EXT_SUFFIX, which does not allow to extract
         # the interpreter ABI.  Check that the fallback is not hit for
         # any other Python implementation.
         if sys.implementation.name != 'cpython':
-            raise NotImplementedError
+            raise NotImplementedError from exc
```
dnicolodi, line 69: Same here.

mesonpy/_introspection.py
FFY00: From https://github.com/PyCQA/flake8-bugbear#opinionated-warnings

> **B904**: Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling. See [the exception chaining tutorial](https://docs.python.org/3/tutorial/errors.html#exception-chaining) for details.

This is actually very helpful.

tests/test_wheel.py
FFY00: Thanks.

mesonpy/_introspection.py
FFY00: TLDR: It changed the message from

```
...
...
```

to

```
...
The above exception was the direct cause of the following exception:
...
```

dnicolodi: I never pick up on the different messages. Thanks for the explanation.

FFY00:
@dnicolodi do you still have any questions, or are you okay with merging? I want to open another PR that conflicts with this, so it'd be good to get this in.

dnicolodi:
I'm not big on linters and I don't know ruff, thus I don't have an opinion. Everything else looks good to me.

rgommers:
Everyone who I've seen try `ruff` seems to like it, so +1 from me.

FFY00:
No one said anything against it, so I am gonna merge it.

If there's anything that you think we might have not considered, feel free to open an issue to discuss it.

PR #280, opened 2023-01-29
ENH: remove fallback needed only with now unsupported Meson versions

dnicolodi:
I think we can safely drop this now. The CI should agree.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -128,6 +128,44 @@ _EXTENSION_SUFFIX_REGEX = re.compile(r'^\.(?:(?P<abi>[^.]+)\.)?(?:so|pyd|dll)$')
 assert all(re.match(_EXTENSION_SUFFIX_REGEX, x) for x in _EXTENSION_SUFFIXES)
 
 
+# Maps wheel installation paths to Meson installation path placeholders.
+# See https://docs.python.org/3/library/sysconfig.html#installation-paths
+_SCHEME_MAP = {
+    'scripts': ('{bindir}',),
+    'purelib': ('{py_purelib}',),
+    'platlib': ('{py_platlib}', '{moduledir_shared}'),
+    'headers': ('{includedir}',),
+    'data': ('{datadir}',),
+    # our custom location
+    'mesonpy-libs': ('{libdir}', '{libdir_shared}')
+}
+
+
+def _map_meson_destination(destination: str) -> Tuple[Optional[str], pathlib.Path]:
+    """Map a Meson installation path to a wheel installation location.
+
+    Return a (wheel path identifier, subpath inside the wheel path) tuple.
+
+    """
+    parts = pathlib.Path(destination).parts
+    for folder, placeholders in _SCHEME_MAP.items():
+        if parts[0] in placeholders:
+            return folder, pathlib.Path(*parts[1:])
+    warnings.warn(f'Could not map installation path to an equivalent wheel directory: {destination!r}')
+    return None, pathlib.Path(destination)
+
+
+def _map_to_wheel(sources: Dict[str, Dict[str, Any]]) -> DefaultDict[str, List[Tuple[pathlib.Path, str]]]:
+    """Map files to the wheel, organized by wheel installation directrory."""
+    wheel_files = collections.defaultdict(list)
+    for group in sources.values():
+        for src, target in group.items():
+            directory, path = _map_meson_destination(target['destination'])
+            if directory is not None:
+                wheel_files[directory].append((path, src))
+    return wheel_files
+
+
 def _showwarning(
     message: Union[Warning, str],
     category: Type[Warning],
@@ -180,40 +218,25 @@ class MesonBuilderError(Error):
 class _WheelBuilder():
     """Helper class to build wheels from projects."""
 
-    # Maps wheel scheme names to Meson placeholder directories
-    _SCHEME_MAP: ClassVar[Dict[str, Tuple[str, ...]]] = {
-        'scripts': ('{bindir}',),
-        'purelib': ('{py_purelib}',),
-        'platlib': ('{py_platlib}', '{moduledir_shared}'),
-        'headers': ('{includedir}',),
-        'data': ('{datadir}',),
-        # our custom location
-        'mesonpy-libs': ('{libdir}', '{libdir_shared}')
-    }
-
     def __init__(
         self,
         project: Project,
         metadata: Optional[pyproject_metadata.StandardMetadata],
         source_dir: pathlib.Path,
-        install_dir: pathlib.Path,
         build_dir: pathlib.Path,
         sources: Dict[str, Dict[str, Any]],
-        copy_files: Dict[str, str],
     ) -> None:
         self._project = project
         self._metadata = metadata
         self._source_dir = source_dir
-        self._install_dir = install_dir
         self._build_dir = build_dir
         self._sources = sources
-        self._copy_files = copy_files
 
         self._libs_build_dir = self._build_dir / 'mesonpy-wheel-libs'
 
     @cached_property
     def _wheel_files(self) -> DefaultDict[str, List[Tuple[pathlib.Path, str]]]:
-        return self._map_to_wheel(self._sources, self._copy_files)
+        return _map_to_wheel(self._sources)
 
     @property
     def _has_internal_libs(self) -> bool:
@@ -396,38 +419,6 @@ class _WheelBuilder():
             return True
         return False
 
-    def _map_from_scheme_map(self, destination: str) -> Optional[Tuple[str, pathlib.Path]]:
-        """Extracts scheme and relative destination from installation paths."""
-        parts = pathlib.Path(destination).parts
-        for scheme, placeholders in self._SCHEME_MAP.items():
-            if parts[0] in placeholders:
-                return scheme, pathlib.Path(*parts[1:])
-        return None
-
-    def _map_to_wheel(
-        self,
-        sources: Dict[str, Dict[str, Any]],
-        copy_files: Dict[str, str],
-    ) -> DefaultDict[str, List[Tuple[pathlib.Path, str]]]:
-        """Map files to the wheel, organized by scheme."""
-        wheel_files = collections.defaultdict(list)
-        for files in sources.values():  # entries in intro-install_plan.json
-            for file, details in files.items():  # install path -> {destination, tag}
-                # try mapping to wheel location
-                meson_destination = details['destination']
-                install_details = self._map_from_scheme_map(meson_destination)
-                if install_details:
-                    scheme, destination = install_details
-                    wheel_files[scheme].append((destination, file))
-                    continue
-                # not found
-                warnings.warn(
-                    'File could not be mapped to an equivalent wheel directory: '
-                    '{} ({})'.format(copy_files[file], meson_destination)
-                )
-
-        return wheel_files
-
     def _install_path(
         self,
         wheel_file: mesonpy._wheelfile.WheelFile,
@@ -522,7 +513,7 @@ class _WheelBuilder():
                     self._install_path(whl, counter, origin, destination)
 
                 # install the other schemes
-                for scheme in self._SCHEME_MAP:
+                for scheme in _SCHEME_MAP:
                     if scheme in (root_scheme, 'mesonpy-libs'):
                         continue
                     for destination, origin in self._wheel_files[scheme]:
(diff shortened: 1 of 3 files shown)


mesonpy/__init__.py
```diff
@@ -398,73 +398,13 @@ def _is_native(self, file: Union[str, pathlib.Path]) -> bool:
...
-            if destination.startswith(placeholder):
-                relative_destination = pathlib.Path(destination).relative_to(placeholder)
-                return scheme, relative_destination
-        return None  # no match was found
+        """Extracts scheme and relative destination from installation paths."""
+        parts = pathlib.Path(destination).parts
+        for scheme, placeholders in self._SCHEME_MAP.items():
+            if parts[0] in placeholders:
+                return scheme, pathlib.Path(os.path.join(*parts[1:]))
+        return None
```
rgommers, line 407: Under what circumstances is returning `None` the correct answer? If it's never the right answer, then should this raise an exception?

rgommers:
Thanks @dnicolodi. I'm a little nervous about this one, because we've had so many problems with this logic before. I think all Meson issues are fixed indeed. However, it's hard for me to verify that we definitely won't need this fallback code anymore. IIRC this was a problematic case that we may not have a test for:

```meson
custom_target('file_with_random_extension',
  input : ['codegen.c.in'],
  output : ['codegen.c'],
  command : [custom_script, '@INPUT@', '@OUTPUT@'],
  install : true,  # still needed because of meson#7372 - can't pass generated targets to py.install_sources
  install_dir : py.get_install_dir() / 'pkgname' / 'submod'  # this may work now, not sure if that covers everything
```

Or are we sure that by design there is no way to hit that heuristics code path anymore?

mesonpy/__init__.py
dnicolodi: Where this function is called a warning is emitted if the return value is `None`. I think we can turn this into an error, but I didn't want to modify the current code too much.

dnicolodi:
I just verified that the case above works correctly.

rgommers:
I tested this PR with SciPy, and all works as expected for a regular wheel build on Linux. So that looks good. I'll leave further review to @FFY00, because he's more familiar with this code.

FFY00:
Yeah, I am a bit uneasy about this. IMO we should wait until after the editable installs release to merge this, unless there's any dependence on this[^1].

[^1]: The `Traversable` approach would have a dependence on this, or at least, would greatly benefit from this because it needs to map the files. But I'd prefer to also wait to merge that change.

FFY00:
Even though newer Meson versions have it fixed, several users will still be using older versions, shipped by their distros. I'd really prefer to do this change after we release editable installs, so that we don't have to deal with breakage coming from several places.

> If we want to err on the safe side, we can check that all installation paths start with a placeholder and error out if they do not.

Yes, we should do that.

FFY00:
Please note that this change has a possibility of breaking normal installs. I just want to wait a little bit.

dnicolodi:
No, they cannot use older Meson shipped by their distros, we depend on Meson >= 0.63.3: https://github.com/mesonbuild/meson-python/blob/c9438b0c94fac9672173714c6de2d88a99c4e2e4/pyproject.toml#L31 We test this version an all sebsequent minor versions in the CI.

We already emit a warning if we cannot map the file. We can turn it into an error if the path does not start with a placeholder.

dnicolodi:
I'm not in an hurry to merge this, but for what should we wait exactly? To say that we need to wait without specifying what we need to wait for seem like postponing indefinitely.

What we should have done, was to add comments to this code specifying in which cases it was required, and ideally we should have added corresponding tests to the test suite. Now we would not be debating whether this work-around is still needed or not. But "del senno di poi son piene le fosse".

FFY00:
> No, they cannot use older Meson shipped by their distros, we depend on Meson >= 0.63.3:

I forgot about that.

mesonpy/__init__.py
FFY00: The reason for that was that we had two functions to try to map the files (via the direct scheme mapping, and heuristics). Now, we can move the warning here.

---

What was the reason for changing the mapping logic here? I mean, I am not gonna block it, but next time please keep keep code improvements separate from logical changes.

mesonpy/__init__.py
```diff
@@ -398,73 +398,13 @@ def _is_native(self, file: Union[str, pathlib.Path]) -> bool:
...
-        ]:  # scheme name, scheme path (see self._SCHEME_MAP)
-            if destination.startswith(placeholder):
-                relative_destination = pathlib.Path(destination).relative_to(placeholder)
-                return scheme, relative_destination
-        return None  # no match was found
+        """Extracts scheme and relative destination from installation paths."""
+        parts = pathlib.Path(destination).parts
+        for scheme, placeholders in self._SCHEME_MAP.items():
+            if parts[0] in placeholders:
+                return scheme, pathlib.Path(os.path.join(*parts[1:]))
```
FFY00, line 406:
```suggestion
                return scheme, pathlib.Path(*parts[1:])
```

Also, please keep the docstring.

FFY00 (changes requested):

mesonpy/__init__.py
dnicolodi: Oh, neat. I'm actually quite sure I tried this syntax and that it didn't work. I must have forgot the `*` or something.

mesonpy/__init__.py
dnicolodi: I had to read the previous version three times to make sure of what it does. Having nested iterator comprehensions in a `for` loop is just too much for me. So I decided to write in a way that I find more immediate to read.

it is a separate commit. Do you mean I should open a separate PR next time?

dnicolodi: Cool. We broke GitHub interface with this comments. They do not show up in the main PR page...

mesonpy/__init__.py
FFY00: Yeah, the constructor can't take a sequence as the first arguments, so you need to unpack it.

mesonpy/__init__.py
FFY00: No, separate commit is fine. I thought you were planning to squash-merge because the commit message was so small :sweat_smile:. Next time, use something like `simplify _map_from_scheme_map`.

FFY00:
> Cool. We broke GitHub interface with this comments. They do not show up in the main PR page...

@dnicolodi did you delete this? Because I actually do not see your comment at all :rofl:. I just got the email notification.

dnicolodi:
> @dnicolodi did you delete this? Because I actually do not see your comment at all 🤣. I just got the email notification.

Yes, I did. The comments are actually there. Just in a weird order. Sometimes the GitHub interface confuses me.

rgommers:
> > No, they cannot use older Meson shipped by their distros, we depend on Meson >= 0.63.3:
> 
> I forgot about that.

We're missing `meson_version: '>= 0.63.3' in the `project()` call in our own `meson.build` file, so this is right now not quite correct for non-isolated builds. Can we please add that to ensure we don't use an older Meson version? Just add a comment that the minimum version is implemented in both `pyproject.toml` and `meson.build` - that's unavoidable, and updating in two places is a very small effort.

rgommers:
(sorry, wrong button)

dnicolodi:
> We're missing `meson_version: '>= 0.63.3'` in the `project()` call in our own `meson.build` file, so this is right now not quite correct for non-isolated builds. Can we please add that to ensure we don't use an older Meson version? Just add a comment that the minimum version is defined in both `pyproject.toml` and `meson.build` - that's unavoidable, and updating in two places is a very small effort.

Adding something to the meson-python's `meson.build` enforces the minimum version only when building meson-python.

However, this is a good point. It is not impossible to conceive a way to use the system meson while installing meson-python in a virtual environment. I think maybe we should add a run-time check for the meson version. I think it would cost very little. Going down this line we could even move meson to a build dependency of the project being built obtained via `get_requires_for_build_wheel()` as we did for ninja, only if the meson executable is not found or is too old.

rgommers:
> However, this is a good point. It is not impossible to conceive a way to use the system meson while installing meson-python in a virtual environment. I think maybe we should add a run-time check for the meson version.

You are right, good point. +1 for a check.

eli-schwartz:
> Just add a comment that the minimum version is defined in both `pyproject.toml` and `meson.build` - that's unavoidable, and updating in two places is a very small effort.

> Going down this line we could even move meson to a build dependency of the project being built obtained via `get_requires_for_build_wheel()` as we did for ninja, only if the meson executable is not found or is too old.


It is theoretically possible to parse `meson.build` for the `meson_version` once you have some version of Meson installed, to fine-tune the required version number. Not sure how helpful that is in practice?

```python
import json
import subprocess

stdout = subprocess.check_output(['meson', 'introspect', '--ast', 'meson.build'])
j = json.loads(stdout)

for kw in j['lines'][0]['args']['kwargs']:
    if kw['key']['value'] == 'meson_version':
        meson_version = kw['val']['value']
        break
```

dnicolodi:
We don't install the meson-python's `meson.build` thus extracting the minimum required meson version from there at runtime is not that useful. Unfortunately, I think that having the minimum version both in `pyproject.toml` and somewhere in the code is the only way to do the check at run time (with reasonable complexity).

dnicolodi:
Is there a way to make the output of pre-commit-check more informative? pre-commit-check is failing with this error
```
ruff.....................................................................Failed
- hook id: ruff
- files were modified by this hook

Found 1 error (1 fixed, 0 remaining).
```
which tells me exactly nothing about what is wrong.

dnicolodi:
Why can't this be merged as is? It just removes a fallback for a bug in version of Meson that are not supported by meson-python anyway. I don't think that waiting will make anyone more confident that the bugs have actually been resolved. Do we need more tests? What configurations not covered by the tests already in place do we need to test? I would like to merge this and move on to fix #317.

mesonpy/__init__.py
```diff
@@ -1107,29 +1018,43 @@ def _project(config_settings: Optional[Dict[Any, Any]]) -> Iterator[Project]:
         yield project
 
 
+def _parse_version_string(string: str) -> Tuple[int, ...]:
+    """Parse version string."""
+    try:
+        return tuple(int(x) for x in string.split('.')[:3])
```
FFY00, line 1024:
```suggestion
        return tuple(map(int, string.split('.')))
```

mesonpy/__init__.py
```diff
@@ -1197,6 +1123,7 @@ def build_wheel(
     metadata_directory: Optional[str] = None,
 ) -> str:
     _setup_cli()
+    _check_meson_version()
```
FFY00, line 1126: This is better placed in `Project`, like ninja.

mesonpy/_introspection.py
```diff
@@ -68,6 +68,5 @@ def sysconfig_paths() -> Mapping[str, str]:
 
 
 __all__ = [
-    'DEBIAN_PYTHON',
```
FFY00, line 71: You can remove the rest of the Debian introspection code too.

FFY00 (changes requested):

mesonpy/__init__.py
```diff
@@ -126,6 +128,43 @@ def _init_colors() -> Dict[str, str]:
...
+        for src, target in group.items():
+            dest = target['destination']
+            directory, path = _map_meson_destination(dest)
+            if directory is not None:
+                wheel_files[directory].append((path, src))
+            else:
+                warnings.warn(f'Installation path {dest!r} for {src!r} could not be mapped to an equivalent wheel directory.')
+                if not re.match(r'^{\w+}$', path.parts[0]):
+                    raise RuntimeError('Meson installation path {dest!r} does not start with a placeholder. Meson bug!')
+    return wheel_files
```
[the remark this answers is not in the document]
dnicolodi: > Please keep these functions in `_WheelBuilder`.

Why? They don't need access to any member of `_WheelBuilder`.

> IMO it'd be cleaner to move the warnings to `_map_meson_destination`.

For doing that, either the src path needs to be passed to `_map_meson_destination` just for the purpose of maybe emitting the warning this information needs to be removed from the warning.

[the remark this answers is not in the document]
FFY00: > Why? They don't need access to any member of `_WheelBuilder`.

Because its whole purpose is to support code in `_WheelBuilder`, and this way it also can access internal state, for example the source paths you referred in the next comment.

> For doing that, either the src path needs to be passed to `_map_meson_destination` just for the purpose of maybe emitting the warning this information needs to be removed from the warning.

Or you can have it in `_WheelBuilder` and be able to access it directly :sweat_smile:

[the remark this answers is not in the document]
dnicolodi: Please have a look at how the code is written. Having it a method of `_WheelBuilder` does not help in any way.

rgommers:
I did some more testing with the merged version of this in SciPy's CI. We have a specific job for Debian there which used to be an issue. All good now it looks like. Glad to see this code gone!

PR #282, opened 2023-01-29
BUG: Deleting a file from an editable install

lithomas1:
I meant to block #87 on this, but forgot.

This PR should block release of meson-python with editable wheels.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -621,6 +621,8 @@ class _WheelBuilder():
                     hook_name={_as_python_declaration(hook_module_name)},
                     project_path={_as_python_declaration(self._source_dir)},
                     build_path={_as_python_declaration(self._build_dir)},
+                    install_dir={_as_python_declaration(install_path)},
+                    uninstall_old = True,
                     import_paths={_as_python_declaration(import_paths)},
                     top_level_modules={_as_python_declaration(self.top_level_modules)},
                     rebuild_commands={_as_python_declaration(rebuild_commands)},
@@ -882,7 +884,6 @@ class Project():
             (
                 'meson',
                 'install',
-                '--only-changed',
                 '--destdir',
                 os.fspath(install_dir or self._install_dir),
                 *self._meson_args['install'],
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -1,6 +1,7 @@
 import functools
 import importlib.abc
 import os
+import shutil
 import subprocess
 import sys
 import warnings
@@ -63,6 +64,8 @@ class MesonpyFinder(importlib.abc.MetaPathFinder):
         hook_name: str,
         project_path: str,
         build_path: str,
+        install_dir: str,
+        uninstall_old: bool,
         import_paths: List[str],
         top_level_modules: List[str],
         rebuild_commands: List[List[str]],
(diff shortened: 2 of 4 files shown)


lithomas1:
Anyone up for a review?

dnicolodi:
I don't think piling hacks on top of this solution will bring a polished solution. I think the approach in #279 is much more robust. It currently handles python modules (including modules appearing and disappearing during rebuild, except at the top level). I'm working on resource file support, but piecing together what is needed from the documentation of the `importlib` standard lib module takes a while. I would appreciate if you could test it for your use case.

lithomas1:
> I don't think piling hacks on top of this solution will bring a polished solution. I think the approach in #279 is much more robust. It currently handles python modules (including modules appearing and disappearing during rebuild, except at the top level). I'm working on resource file support, but piecing together what is needed from the documentation of the `importlib` standard lib module takes a while. I would appreciate if you could test it for your use case.

I would rather prefer merging this first, since I don't know when how far #279 is yet, and fixing this will allow for a release with editable wheels support (so pandas can finally adopt meson).

Hopefully, this PR shouldn't mess up too much of what your working on, since its a small ~10 LOC patch.

mesonpy/__init__.py
```diff
@@ -882,7 +884,6 @@ def build_commands(self, install_dir: Optional[pathlib.Path] = None) -> Sequence
             (
                 'meson',
                 'install',
-                '--only-changed',
```
rgommers, line 885: Can we keep this? It's quite important, at least verbose mode is completely unusable without it.

lithomas1: It doesn't do anything anymore as of this PR. 

This is because we nuke the install dir before hand, so the destination file will never exist.

For reference, see https://github.com/mesonbuild/meson-python/pull/87#discussion_r1056438367.

lithomas1: It's unfortunate that we have to do this, but the only other way, I think, is to patch meson, and I don't think we should leave editable installs broken in the meantime.

rgommers: To me, making the verbose mode effectively unusable seems worse than not cleaning up some now-unused old file (the chance of actual breakage is quite small here). So I'm ambivalent about this change.

That said, I don't use editable installs and there's further improvements in the pipeline, so I'll just register as neutral on this one; fine if others want to merge this.

lithomas1: To clarify, this old file is still importable if the directory had an __init__.py, which can be confusing at the very least, and maybe cause problems(I'm thinking maybe if someone did a star import?).

mesonpy/_editable.py
```diff
@@ -115,6 +120,9 @@ def _proc(self, command: List[str]) -> None:
     @functools.lru_cache(maxsize=1)
     def rebuild(self) -> None:
         self._debug(f'{{cyan}}{{bold}}+ rebuilding {self._project_path}{{reset}}')
+        if self.uninstall_old:
+            if os.path.exists(self.install_dir):
+                shutil.rmtree(self.install_dir)
```
FFY00, line 125: I don't like this approach. IMO saving the list of previous installed files and manually deleting the ones that are not installed by Meson anymore would be a better way to deal with this.

PR #283, opened 2023-01-30
ENH: add REUSE compliance

FFY00:
https://reuse.software/


FFY00:
Thanks to @tristan957 for making me aware of REUSE, and by proxy the `reuse` tool. Maintaining SPDX headers has always been a pain.

dnicolodi:
Do we really need the copyright line in all files?
```
SPDX-FileCopyrightText: 2023 meson-python developers
```
seems generic enough to me to be close to useless.

tristan957:
@FFY00 I suggest also using the github action.

https://github.com/fsfe/reuse-action

@dnicolodi REUSE currently requires copyright to be denoted either in the file itself or in `.reuse/dep5`

FFY00:
Yeah, the point of it is to be present on the file when looked at standalone, the same reasoning we have the license header. It also allows contributors who have made significant changes to the file to note their copyright if they so wish.

FFY00:
> @FFY00 I suggest also using the github action.

We already do the compliance check via pre-commit though, so I don't think it's needed. Having it in pre-commit also flags issues right away if you have the git hook installed.

tristan957:
Ah, I didn't know about pre-commit :)

dnicolodi:
Sure, but what is the MIT license is known, and is specified by the SPDX "standard". Who the "meson-python developers" are is not very well specified. It is pretty much as precise as to write "somebody". I think that if the tool requires us to add useless boiler plate to all source files, we should reconsider using the tool, or fix the tool before using it. I mean, there are plenty of projects out there that use SPDX without the copyright lines.

dnicolodi:
At the very minimum, can we remove the empty line in between the license line and the copyright line?

tristan957:
This is pulled from the FAQ.

https://reuse.software/faq/#many-copyright-statements

> Some files are edited by many people and would have an extremely long list of copyright holders in the header. This may be aesthetically unpleasing, but is not incorrect.
> 
> If you would rather not deal with having so many copyright notices, some projects such as Chromium circumvent this problem by using “Copyright (c) 2013 The Chromium Authors” as their copyright tag. You may consider doing this, but then you should keep a list of copyright holders and authors in a separate file in your project.

FFY00:
IMO this is the kind of thing which can be slightly annoying a first, but that doesn't really hurt anyone, and provides significant improvements -- it makes things easier for downstream distributors, and makes sure the license and copyright information is present if the file is taken out of the project. I think the price to pay is really mild, especially considering that we have tooling to handle everything for us.

dnicolodi:
> *but then you should keep a list of copyright holders and authors in a separate file in your project.*

Where is such list for meson-python?

dnicolodi:
Also, can we omit the copyright year, at least from the line used everywhere, so we don't have to go around once a year to update all files?

tristan957:
@dnicolodi you don't need to update the copyright year. Just leave it as the year of original creation. cURL already went through this exercise earlier this year.

tristan957:
https://reuse.software/faq/#years-copyright

> Generally, there are four options for you to choose:
>
> The year of initial publication.
>
> The year of the latest publication.
>
> All years of publications, either as range (e.g., 2017-2019) or as separate entries (e.g., 2017, 2018, 2019).
>
> Do not include any year.
> 
> Which option you choose is ultimately up to you.

dnicolodi:
For most files the year specified is not the year of first creation. The year in the copyright declaration line has no purpose, can we just omit it?

tristan957:
> The year in the copyright declaration line has no purpose, can we just omit it?

Why do you think it has no purpose?

> Notice identifies the year of first publication, which may be used to determine the term of 
copyright protection in the case of an anonymous work, a pseudonymous work, or a work 
made for hire.

https://www.copyright.gov/circs/circ03.pdf

dnicolodi:
> IMO this is the kind of thing which can be slightly annoying a first, but that doesn't really hurt anyone, and provides significant improvements -- it makes things easier for downstream distributors, and makes sure the license and copyright information is present if the file is taken out of the project. I think the price to pay is really mild, especially considering that we have tooling to handle everything for us.

Please don't take me wrong, I'm extremely happy we are doing this. Just, if we do it, I would like that a minimum of though goes into the process, instead than copy-and-pasting some header into all files because a linter or another told us we need to do it.

In this sense, `SPDX-FileCopyrightText: 2023 meson-python developers` does not make anyone reading the source file any wiser regarding who the copyright holders are.

dnicolodi:
> Why do you think it has no purpose?

It has no purpose if you put a random year in it.

The REUSE FAQ says that the year can be omitted altogether. I don't understand why "2023" (which is most definitely not the year of first publication, and if it the year of last publication we are bound to update it every year) is in your opinion better than not specifying a year.

FFY00:
> In this sense, `SPDX-FileCopyrightText: 2023 meson-python developers` does not make anyone reading the source file any wiser regarding who the copyright holders are.

It does... if anyone takes the file out of the project it absolutely does.

tristan957:
> I don't understand why "2023" (which is most definitely not the year of first publication, and if it the year of last publication we are bound to update it every year) is in your opinion better than not specifying a year.

This is not my opinion :).

dnicolodi:
> > I don't understand why "2023" (which is most definitely not the year of first publication, and if it the year of last publication we are bound to update it every year) is in your opinion better than not specifying a year.
> 
> This is not my opinion :).

Can we then get rid of the year, please?

tristan957:
To be perfectly clear, I believe that you should leave the year, but use Git history to find the original date of the file's creation.

This is what the FSFe does as you can see here. https://github.com/fsfe/reuse-action/blob/master/Dockerfile#L1

dnicolodi:
> > In this sense, `SPDX-FileCopyrightText: 2023 meson-python developers` does not make anyone reading the source file any wiser regarding who the copyright holders are.
> 
> It does... if anyone takes the file out of the project it absolutely does.

It does in the "moral" sense of things, but it does not in any other sense. Copyright can only be help by juridical entities, and "meson-python developers" is not a juridical entity (by the way, it should be `SPDX-FileCopyrightText: The meson-python developers` at least) nor a short-hand for a list that is defined elsewhere.

dnicolodi:
Anyhow, I'm in favor of adopting it, just please use
```
SPDX-FileCopyrightText: YEAR The meson-python developers.
```
where `YEAR` is the year of first publication, as @tristan957 suggested. If you give me a couple of minutes I give you a short script that spits out the first publication date for all files.

dnicolodi:
Probably not the most efficient way of doing it, but the repository is fortunately small:
```
git ls-files -z | xargs -0 -n1 -I{} -- sh -c "git log --reverse --format='%as {}' {} | head -1"
```

rgommers (changes requested):
This seems like a good thing to do in principle. Regarding the year, I agree with the "year of first creation" comments.

The questionable part of this PR is moving the MIT license text out of `LICENSE`. This is a bad idea, because (a) it's nonstandard, and (b) it breaks license detections mechanisms like the one GitHub uses. You can verify this easily by pushing the PR branch to the `main` branch of your own fork. Then the displayed license in the right sidebar goes from MIT to "View license" (typically indicating a nonstandard license). So I'd much prefer to not include that part.

tristan957:
https://github.com/orgs/community/discussions/44922

Throw some love into that discussion.

> (a) it's nonstandard

I wouldn't call a set of heuristics a standard. This PR introduces following a standard that GitHub should probably learn how to deal with. Other forges do support this, like SourceHut.

rgommers:
> I wouldn't call a set of heuristics a standard.

That's not what I meant, they were two separate arguments. I mean it's _the_ standard way of doing things in open source projects to put your license text in a file called `LICENSE` or `LICENSE.txt`. Maybe after that a `COPYRIGHT` file. Almost never in `LICENSES/xxx`. I'd suggest to reconsider only if it becomes standard or if we have a reason to add multiple license files (because then it'd make sense).

tristan957:
If that is really important to you, you can see that cURL does both (COPYING AND LICENSES/): https://github.com/curl/curl

FFY00:
> The questionable part of this PR is moving the MIT license text out of `LICENSE`.

I thought the symlink I added would keep it working, but I just tested it and it appears that's not the case. I'll keep a copy of the license on the top level.

FFY00:
The force push was just a rebase on main, to fix pre-commit, no changes in the commits being reviewed.

dnicolodi:
Sorry to insist, but can we use `SPDX-FileCopyrightText: YEAR The meson-python developers` instead of `SPDX-FileCopyrightText: YEAR meson-python developers`, note the added `The`? It is nicer and more correct.

FFY00:
Yeah, no worries. I'll make that change.

I'll also rebase on main to solve the conflicts. No changes will be made to the commits in review (other than resolving conflicts ofc).

dnicolodi:
It looks good to me. Just remember to squash the commits before merging.

dnicolodi (approved):

dnicolodi:
I don't know what is up with the Debian unstable CI job. Almost certainly unrelated to this PR. I'll have a closer look later.

PR #286, opened 2023-01-31
ENH: Add install-tags filtering for wheel generation (updated)

peter-urban:
This is a PR for https://github.com/mesonbuild/meson-python/issues/68
It is a continuation of  (and should probably replace) @Brishen's PR https://github.com/mesonbuild/meson-python/pull/267 which used my code, but has not been updated (since two weeks) to include the suggestions/changes made in reaction to @FFY00's comments.

PR #288, opened 2023-02-01
ENH: add install-tags filtering for wheel generation

FFY00:
Replaced #267 and  #286


peter-urban:
Thanks for getting this merged. It will allow me to use stock meson-python 🥳

PR #289, opened 2023-02-02
Misc CI improvements

FFY00:
Just a couple of small things, it didn't seem worth to open separate PRs.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -7,7 +7,7 @@ test_template: &test
     - command -v python3
     - python3 --version
     - python3 -m pip install .[test]
-    - python3 -m pytest
+    - python3 -m pytest --showlocals -vv
 
 alpine-3_task:
   container:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -3,11 +3,13 @@
 # SPDX-License-Identifier: MIT
 
 test_template: &test
+  environment:
+    CODECOV_TOKEN: ENCRYPTED[e18180139fa0a70726bb6f8dfbbc9fd53da36e6a4d2b449e30f99a255fdfdab63863117e8b6665deb06dd99d95e5aaad]
   test_script:
     - command -v python3
     - python3 --version
-    - python3 -m pip install .[test]
-    - python3 -m pytest --showlocals -vv
+    - python3 -m pip install .[test] pytest-codecov
+    - python3 -m pytest --showlocals -vv --cov --cov-report=xml:coverage-$CIRRUS_TASK_NAME.xml --codecov
 
 alpine-3_task:
   container:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -104,6 +104,9 @@ known_first_party = 'mesonpy'
 disable_warnings = [
   'couldnt-parse',
 ]
+omit = [
+  'tests/packages/*'
+]
 
 [tool.coverage.html]
 show_contexts = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,6 +108,8 @@ source = ['.']
 omit = [
   'tests/packages/*',
 ]
+
+[tool.coverage.report]
 ignore_errors = true
 
 [tool.coverage.html]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,6 +111,11 @@ omit = [
 
 [tool.coverage.report]
 ignore_errors = true
+exclude_lines = [
+  '\#\s*pragma: no cover',  # we need this because this field overrides the default
+  '^\s*raise NotImplementedError\b',
+  '^\sif typing.TYPE_CHECKING',
+]
 
 [tool.coverage.html]
 show_contexts = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,6 +108,7 @@ source = ['.']
 omit = [
   'tests/packages/*',
 ]
+ignore_errors = true
 
 [tool.coverage.html]
 show_contexts = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -104,6 +104,7 @@ known_first_party = 'mesonpy'
 disable_warnings = [
   'couldnt-parse',
 ]
+source = ['.']
 omit = [
   'tests/packages/*',
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,7 +105,7 @@ disable_warnings = [
   'couldnt-parse',
 ]
 omit = [
-  'tests/packages/*'
+  'tests/packages/*',
 ]
 
 [tool.coverage.html]


.cirrus.yml
```diff
@@ -3,11 +3,13 @@
...
 test_template: &test
+  environment:
+    CODECOV_TOKEN: ENCRYPTED[e18180139fa0a70726bb6f8dfbbc9fd53da36e6a4d2b449e30f99a255fdfdab63863117e8b6665deb06dd99d95e5aaad]
   test_script:
     - command -v python3
     - python3 --version
-    - python3 -m pip install .[test]
-    - python3 -m pytest --showlocals -vv
+    - python3 -m pip install .[test] pytest-codecov
+    - python3 -m pytest --showlocals -vv --cov --cov-report=xml:coverage-$CIRRUS_TASK_NAME.xml --codecov
```
dnicolodi, line 12: What's the point of reporting coverage data to codecov if we don't do anything with it? I would just get rid of the coverage reporting altogether.

FFY00: We do, https://github.com/seantis/pytest-codecov is a plugin to upload the coverage data.

dnicolodi: I know, but where is the data visualized?

FFY00: https://app.codecov.io/gh/mesonbuild/meson-python/ and the badge in the README and docs. I'd like to improve the coverage, so IMO it's definitely useful.

.cirrus.yml
```diff
@@ -41,7 +41,7 @@ debian-unstable_task:
     cpu: 1
   fix-pip_script:
     # see https://github.com/pypa/pip/pull/11623
-    - python3 -m pip install --upgrade "pip @ git+https://github.com/pypa/pip.git@a057c9bc99ad7eab2e3b1539e79eee7d9a923f4f"
+    - python3 -m pip install --upgrade pip>=23.0
```
dnicolodi, line 44: Nope. pip 23 is known to be broken on Debian. I tried my best, but failed: https://github.com/pypa/pip/pull/11740

Also, for some reason that I don't fully understand, it is not possible to upgrade the distribution provided pip on Debian unstable. I haven't had time yet to look into it.

Ah, and this would need some quotes, otherwise the shell will write the output of `python3 -m pip install --upgrade pip` into a file named `=23.0`.

.cirrus.yml
```diff
@@ -17,7 +17,7 @@ alpine-3_task:
     cpu: 1
   fix-pyproject-metadata_script:
     # see https://github.com/FFY00/python-pyproject-metadata/pull/44
-    - python3 -m pip install "pyproject-metadata < 0.7.0"
+    - python3 -m pip install "pyproject-metadata != 0.7.0"
```
dnicolodi, line 20: Just drop the work-around. There is no need to keep it. The latest release will be installed anyhow.

.cirrus.yml
FFY00: Thanks, I was trying to fix the Debian unstable CI too but failed :confused:

.cirrus.yml
dnicolodi: I mean to say that you should just drop the `fix-pyproject-metadata_script` block.

.cirrus.yml
```diff
@@ -30,16 +32,13 @@ debian-11_task:
...
+    - python3 -m pip install --upgrade "pip>=23.0"
   << : *test
 
 debian-unstable_task:
   container:
     dockerfile: ci/debian-unstable.docker
     cpu: 1
-  fix-pip_script:
-    # see https://github.com/pypa/pip/pull/11623
-    - python3 -m pip install --upgrade "pip @ git+https://github.com/pypa/pip.git@a057c9bc99ad7eab2e3b1539e79eee7d9a923f4f"
```
dnicolodi, line 42: The fact that this fixed itself is very surprising. The only explanation I have is that Debian started distributing a pip with a downstream fix for the bug. I'll need to take a closer look.

dnicolodi: Yup, that is it. From the `python3-pip` Debian package changelog:
```
python-pip (22.3.1+dfsg-2) unstable; urgency=medium

  * Patch: Use the default sysconfig scheme, in isolated environments.
    (Closes: #1019293)

 -- Stefano Rivera <stefanor@debian.org>  Fri, 20 Jan 2023 15:26:49 -0400
```

.cirrus.yml
dnicolodi: There must be something wrong with the codevcov.io: it does not point to the source code in this repository. and looking at the details for any given file just shows "There was a problem getting the source code from your provider. Unable to show line by line coverage." 

I don't expect that running the code on different flavors of Linux will exercise different code paths, and this doubles the CI execution time. I prefer fast feedback loops. Are we sure we really need this? Does codecov provide a measure of how much each "measurement" from each test run contributes to the total coverage?

dnicolodi:
@FFY00 I pushed on main the obvious fixes to the CI. Please rebase. Let me know if rebasing looks to tedious and I'll do it for you.

FFY00:
Next time please just push to my branch, it's easier and I don't mind 👍

dnicolodi:
Sure. I'm never sure whether people minds me messing with their branches. Force pushed now.

FFY00:
I think we should be okay to merge this now, right @dnicolodi?

dnicolodi:
I still think that reporting coverage from the different Linux flavors we are testing on is completely useless. There is no code in meson-python that executes only on conditions that are affected on which Linux distribution it is running (well, with the exception of a few lines in `mesonpy._introspection` but that code is going away soon). Doubling the time it takes to run the CI for no reason really sucks.

dnicolodi:
Also, there are still commits that need to be squashed.

pyproject.toml
```diff
@@ -111,6 +111,11 @@ omit = [
 
 [tool.coverage.report]
 ignore_errors = true
+exclude_lines = [
+  '\#\s*pragma: no cover',  # we need this because this field overrides the default
+  '^\s*raise NotImplementedError\b',
```
dnicolodi, line 116: Why exclude `raise NotImplementedError`? Error conditions are supposed to be tested as well.

FFY00:
> Doubling the time it takes to run the CI for no reason really sucks.

I don't recall it being that bad when I checked, but we should definitely look at it before merging then. Thanks for pointing it out :+1:

pyproject.toml
FFY00: Yes but `NotImplementedError` is for something that is supposed to work eventually, but doesn't currently, so it's different from something where we do actually intent on throwing an error. Writting tests to see if something raises `NotImplementedError` doesn't seem very productive, I'm not really sure what actual value it brings on most situations.

FFY00:
Well, this was triggered because of the custom Debian code. For example, *right now* it still provides us valuable information, like in #280, where we are planning to remove the custom Debian code, it will tell us if any code is actually being hit at the moment, which we assume will be none, but it's good to have data showing it.

I think due to the nature of the project, it being very dependent on the Python it runs on, it makes sense to have the coverage running in Cirrus *if* it's not a big performance hit. Before, with the Debian specific code, I'd be okay with a moderate performance it, but since that's not the case anymore, the value drops, and so the sensible tradeoff.

FFY00:
Yeah, sure, but there might be edge cases that are not triggered by any test. This just gives you more visibility over the code and what's actually happen.

Anyway, like I said, right now I think this only makes sense if it doesn't have a big performance hit, which I think it's reasonable, no?

dnicolodi:
> Yeah, sure, but there might be edge cases that are not triggered by any test.

If the edge cases are not convered by any test, the test coverage of that code would be exactly zero. Therefore, I don't see how coverage data can inform any decision: whatever you do to that code, the fact that it is not tested does not change.

FFY00:
> If the edge cases are not convered by any test, the test coverage of that code would be exactly zero. Therefore, I don't see how coverage data can inform any decision: whatever you do to that code, the fact that it is not tested does not change.

Yes, but we don't have a 100% test coverage policy, therefore having the extra coverage information does provide extra information we might get with the tests alone.

> My $2c: I'd personally not both with code coverage on >1 CI job - it's not worth it.

That depends on the project, but sure.

> Also Codecov has a very bad track record, both security and reliability wise (often numbers are just nonsensical, especially for diffs because it compares PRs to main incorrectly). Adding it to multiple CI providers is effort _and_ increases risk for very limited (if any) gains.

I don't really think there's any risk, the only additional thing that could be exposed is the codecov key, which only lets you upload coverage reports.

Regarding reliability, yes, codecov has been quite bad, but we are already using it anyway.

Increased effort would be the main argument, but I already spent the time implementing it, and it shouldn't really require much maintenance, and if it does, we can just rip it out anyway.

dnicolodi:
> > If the edge cases are not convered by any test, the test coverage of that code would be exactly zero. Therefore, I don't see how coverage data can inform any decision: whatever you do to that code, the fact that it is not tested does not change.
> 
> Yes, but we don't have a 100% test coverage policy, therefore having the extra coverage information does provide extra information we might not get with the tests alone.

This statement does not follow any logic: if some code is not tested it has zero test coverage, by definition. How can the lack of test coverage provide any information, other than the obvious one that it is not tested? If it is tested, the test coverage stays the same whether the code works as intended or not. The only thing that test coverage tells you is, unsurprisingly, the amount of code covered by the test suite. This can only be used to inform decision about which tests to write.

FFY00:
> This statement does not follow any logic

Please refrain for making this kind of statement in the future if you can, please. It's not constructive or beneficial, the only thing it can do is raise tensions. Thanks :heart:

> if some code is not tested it has zero test coverage, by definition. How can the lack of test coverage provide any information, other than the obvious one that it is not tested?

The issue is that the coverage might only be hit on certain systems, like Debian. Having the coverage metrics can tell us if the code used to be run, but isn't anymore, hinting that it might not be needed anymore, for eg.

Anyway, this PR is closed. No need to keep wasting energy on it, I think we all have many other things to deal with that are more beneficial.

PR #301, opened 2023-02-08
Sundry fixes around `meson setup` handling

dnicolodi:


mesonpy/__init__.py
```diff
@@ -754,25 +754,11 @@ def __init__(  # noqa: C901
...
-            except OSError:
-                # if there are permission errors or something else in the source
-                # directory, put the native file in the working directory instead
-                # (this won't survive multiple calls -- Meson will have to be reconfigured)
-                self._meson_native_file = self._working_dir / '.mesonpy-native-file.ini'
-                self._meson_native_file.write_text(native_file_data)
-
-        # Don't reconfigure if build directory doesn't have meson-private/coredata.data
-        # (means something went wrong)
-        # See https://github.com/mesonbuild/meson-python/pull/257#discussion_r1067385517
```
lithomas1, line 773: Can you keep the comment linking to the Github comment?

lithomas1 (approved):
LGTM. Just 1 small comment.

mesonpy/__init__.py
dnicolodi: I would prefer to paraphrase Eli's comment in the code comment so it is not necessary to click through the link for the explanation (and we don't rely on the link working in perpetuity). Would that be ok with you?

lithomas1: That works for me.

dnicolodi:
@FFY00 do you want to have a close look at this (and other recent PRs) or can I merge it? PS: if you still need help with laser goggles and optical filters shoot me an email

mesonpy/__init__.py
```diff
@@ -754,18 +754,18 @@ def __init__(  # noqa: C901
...
-        self._configure(reconfigure=has_valid_build_dir and not native_file_mismatch)
+        self._meson_native_file.write_text(native_file_data)
+
+        # reconfigure if we have a valid Meson build directory. Meson
+        # uses the presence of the 'meson-private/coredata.dat' file
+        # in the build directory as indication that the build
+        # directory has already been configured and arranges this file
+        # to be created as late as possible or deleted if something
+        # goes wrong during setup.
+        reconfigure = self._build_dir.joinpath('meson-private/coredata.dat').is_file()
```
FFY00, line 765: Does Meson not require us to reconfigure if the native file is different?

dnicolodi: All changes to configuration require `--reconfigure`. That's why we should always add the `--reconfigure` option when we have a valid build directory.

rgommers: That sounds reasonable. I was still a bit fuzzy on how this worked exactly, for example if we add a cross file in the initial `meson setup` step and then reconfigure with the default `meson-python` native file but not the cross file. I think/hope that that will give us a cross build _plus_ the build machine's Python interpreter in the native file.

dnicolodi: > if we add a cross file in the initial `meson setup` step and then reconfigure with the default `meson-python` native file but not the cross file. I think/hope that that will give us a cross build _plus_ the build machine's Python interpreter in the native file.

Why would that be desirable? I cannot think of a situation where this would make sense. If you don't want to change the configuration, don't run `meson setup`. If you reconfigure without a cross file, you don't get a cross build, why would you? `--reconfigure` means "please repeat the configuration step even if the build directory already exist".

rgommers: The problem is still that `pip install . --config-settings` doesn't take duplicate `setup-args` keys. So until that is solved, you must run `meson setup` separate before invoking `pip`. And some packagers refuse to switch from `pip install .` to using `pypa/build` to build a wheel and then installing that with `pip`.

rgommers: > If you reconfigure without a cross file, you don't get a cross build, why would you?

We just dropped the cross compilation for macOS arm64 in favor of a native build on Cirrus CI, but we did do exactly this and it worked: https://github.com/scipy/scipy/blob/maintenance/1.9.x/.github/workflows/wheels.yml#L180-L188.

dnicolodi: I see the problem. But this works if we never reconfigure an existing build directory. I see arguments for not reconfiguring a valid build directory and for always doing it. This PR was an attempt at being consistent in always doing it. However, there isn't a way to implement a "do what I mean" mode where configuration options are applied selectively.

Frankly, if someone insists in using a broken tool, they deserve to obtain a broken result. However, if they want to insist in using `pip`, they can just place the setup arguments in `pyproject.toml`.

dnicolodi: Is there an issue where this problem with pip is reported? What's the opinion of the `pip` maintainers on this?

rgommers: > Is there an issue where this problem with pip is reported? What's the opinion of the pip maintainers on this? 

Yep, this is https://github.com/pypa/pip/issues/11681. The PEP 517 update for this was merged (https://github.com/python/peps/pull/2996), and @FFY00 will open a PR to implement this in `pip` in the same way as done in `build`.

> We should probably have a test project with meson options in meson-python to check for this king of issues in the packaging stack.

Yes indeed, that'd be useful.

> However, there isn't a way to implement a "do what I mean" mode where configuration options are applied selectively.

Fair enough, it's always or never.



> However, if they want to insist in using `pip`, they can just place the setup arguments in `pyproject.toml`.

That's an option now, but (a) this option didn't exist yet when we had to implement the cross build for SciPy, and (b) it doesn't help folks who aren't the package authors (packagers can carry a patch though, for end users it's a bigger pain). Anyway, it should be resolved by the next `pip` release hopefully.

dnicolodi: > We just dropped the cross compilation for macOS arm64 in favor of a native build on Cirrus CI, but we did do exactly this and it worked: https://github.com/scipy/scipy/blob/maintenance/1.9.x/.github/workflows/wheels.yml#L180-L188.

I don't see how that can work. meson-python before this PR runs `meson setup --reconfigure` if there is a difference between the current native file and an existing one. Because no native file exists when meson-python is called by `pip  wheel` this should have resulted in a reconfiguration of the project without the cross file definitions. Are you sure the cross compilation does not work only because the flags specified via environment variables here https://github.com/scipy/scipy/blob/1c836efe5ff37ffa4490756269b060a464690e62/.github/workflows/wheels.yml#L175-L178 ?

Anyhow, now we have better ways to make macOS cross build work.

rgommers: I'm not sure, no. And it's hard to verify outside of the wheel build CI setup.

> Anyhow, now we have better ways to make macOS cross build work.

Agreed. We shouldn't need this anymore for SciPy.

dnicolodi: > I'm not sure, no. And it's hard to verify outside of the wheel build CI setup.

I think it could be inferred from the build log, but I would prefer to do not do the digging, unless required.

dnicolodi: @rgommers to summarize, do you think this PR can be merged or there are use cases that need further consideration?

rgommers: No need I think; always reconfiguring seems reasonable. Maybe we'll find a use case for allowing a user to explicitly disabling that at some point, but I don't think we have one now.

rgommers: > @rgommers to summarize, do you think this PR can be merged or there are use cases that need further consideration?

Yes, I think my concern was addressed in enough detail, and this should be good to go.

FFY00 (approved):
Thanks! Sorry for the delay.

PR #303, opened 2023-02-08
BUG: align build and install dependencies, again

dnicolodi:


tests/test_consistency.py
```diff
@@ -0,0 +1,23 @@
...
+
+def test_pyproject_dependencies():
+    pyproject = pathlib.Path(__file__).parent.parent.joinpath('pyproject.toml')
+    with open(pyproject, 'rb') as f:
+        data = tomllib.load(f)
+    build_dependencies = data['build-system']['requires']
+    project_dependencies = data['project']['dependencies']
+    # verify that all build dependencies are project dependencies
+    assert not set(build_dependencies) - set(project_dependencies), \
+        'pyproject.toml is incosistent: not all "build-syste.requires" are in "project.dependencies"'
```
rgommers, line 23: This test looks like a good idea. Minor comment: typos in this message, in `incosistent` and `syste`

tests/test_consistency.py
```diff
@@ -0,0 +1,23 @@
+# SPDX-FileCopyrightText: 2021 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+import pathlib
+import sys
+
+
+if sys.version_info > (3, 11):
```
rgommers, line 9: `>= (3, 11)` I think?

pyproject.toml
```diff
@@ -29,7 +29,7 @@ classifiers = [
 dependencies = [
   'colorama; os_name == "nt"',
   'meson>=0.63.3',
-  'pyproject-metadata>=0.6.1', # not a hard dependency, only needed for projects that use PEP 621 metadata
```
rgommers, line 32: agreed with removing this, it just raises a "why is this here and why do I need to care" question mark.

rgommers (commented):
These changes LGTM, thanks @dnicolodi. Only two minor comments.

tests/test_consistency.py
dnicolodi: Indeed. I wonder which of our dependencies brings in `tomli` on Python 3.11

PR #304, opened 2023-02-09
BUG: improve validation of pyproject.toml meson-python configuration

### LINKED ISSUE #293: BUG: Meson-python crashes when specifying an incomplete [tool.meson-python.args] section in pyproj.toml
Meson-python allows to specify arguments for meson dist, meson setup, meson compile and meson install using a ```tool.meson-python.args``` section in pyproject.toml.

However if this section is created, meson-python will fail if not all four options (dist, setup, compile and install) are explicitly specified in this section.

e.g. the following works:
```
[tool.meson-python.args] 
dist = []
setup = []
compile = []
install = ["--tags=runtime"]
```

the following crashes (``meson-python: error: Unknown configuration key "tool.meson-python.args.setup"``):
```
[tool.meson-python.args]
dist = []
compile = []
install = ["--tags=runtime"]
```

dnicolodi:
Switch from an incomplete (an bugged) ad hoc validation to a scheme based validation strategy. The scheme is defined in the function _validate_pyproject_config() as nested dictionaries where the keys are configuration field names and valued are validation functions. Unknown fields result in an error.

Fixes #293.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -654,6 +654,36 @@ class _WheelBuilder():
         return wheel_file
 
 
+def _validate_pyproject_config(pyproject: Dict[str, Any]) -> Dict[str, Any]:
+
+    def _table(scheme: Dict[str, Callable[[Any, str], Any]]) -> Callable[[Any, str], Dict[str, Any]]:
+        def func(value: Any, name: str) -> Dict[str, Any]:
+            if not isinstance(value, dict):
+                raise ConfigError(f'Configuration entry "{name}" must be a table')
+            table = {}
+            for key, val in value.items():
+                check = scheme.get(key)
+                if check is None:
+                    raise ConfigError(f'Unknown configuration entry "{name}.{key}"')
+                table[key] = check(val, f'{name}.{key}')
+            return table
+        return func
+
+    def _strings(value: Any, name: str) -> List[str]:
+        if not isinstance(value, list) or not all(isinstance(x, str) for x in value):
+            raise ConfigError(f'Configuration entry "{name}" must be a list of strings')
+        return value
+
+    scheme = _table({
+        'args': _table({
+            name: _strings for name in _MESON_ARGS_KEYS
+        })
+    })
+
+    table = pyproject.get('tool', {}).get('meson-python', {})
+    return scheme(table, 'tool.meson-python')
+
+
 class Project():
     """Meson project wrapper to generate Python artifacts."""
 
@@ -662,7 +692,7 @@ class Project():
     ]
     _metadata: Optional[pyproject_metadata.StandardMetadata]
 
-    def __init__(  # noqa: C901
+    def __init__(
         self,
         source_dir: Path,
         working_dir: Path,
@@ -712,11 +742,13 @@ class Project():
                     self._meson_cross_file.write_text(cross_file_data)
                     self._meson_args['setup'].extend(('--cross-file', os.fspath(self._meson_cross_file)))
 
-        # load config -- PEP 621 support is optional
-        self._config = tomllib.loads(self._source_dir.joinpath('pyproject.toml').read_text())
-        self._pep621 = 'project' in self._config
+        # load pyproject.toml
+        pyproject = tomllib.loads(self._source_dir.joinpath('pyproject.toml').read_text())
+
+        # package metadata
+        self._pep621 = 'project' in pyproject:
         if self.pep621:
-            self._metadata = pyproject_metadata.StandardMetadata.from_pyproject(self._config, self._source_dir)
+            self._metadata = pyproject_metadata.StandardMetadata.from_pyproject(pyproject, self._source_dir)
         else:
             print(
                 '{yellow}{bold}! Using Meson to generate the project metadata '
@@ -727,14 +759,10 @@ class Project():
         if self._metadata:
             self._validate_metadata()
 
-        # load meson args
-        for key in self._get_config_key('args'):
-            self._meson_args[key].extend(self._get_config_key(f'args.{key}'))
-        # XXX: We should validate the user args to make sure they don't conflict with ours.
-
-        self._check_for_unknown_config_keys({
-            'args': _MESON_ARGS_KEYS,
-        })
+        # load meson args from pyproject.toml
+        pyproject_config = _validate_pyproject_config(pyproject)
+        for key, value in pyproject_config.get('args', {}).items():
+            self._meson_args[key].extend(value)
 
         # meson arguments from the command line take precedence over
         # arguments from the configuration file thus are added later
@@ -767,14 +795,6 @@ class Project():
         # set version if dynamic (this fetches it from Meson)
         if self._metadata and 'version' in self._metadata.dynamic:
             self._metadata.version = self.version
-
-    def _get_config_key(self, key: str) -> Any:
-        value: Any = self._config
-        for part in f'tool.meson-python.{key}'.split('.'):
-            if not isinstance(value, Mapping):
-                raise ConfigError(f'Configuration entry "tool.meson-python.{key}" should be a TOML table not {type(value)}')
-            value = value.get(part, {})
-        return value
 
     def _run(self, cmd: Sequence[str]) -> None:
         """Invoke a subprocess."""
@@ -833,17 +853,6 @@ class Project():
                     f'Unsupported Python version {platform.python_version()}, '
                     f'expected {self._metadata.requires_python}'
                 )
-
-    def _check_for_unknown_config_keys(self, valid_args: Mapping[str, Collection[str]]) -> None:
-        config = self._config.get('tool', {}).get('meson-python', {})
-
-        for key, valid_subkeys in config.items():
-            if key not in valid_args:
-                raise ConfigError(f'Unknown configuration key "tool.meson-python.{key}"')
-
-            for subkey in valid_args[key]:
-                if subkey not in valid_subkeys:
-                    raise ConfigError(f'Unknown configuration key "tool.meson-python.{key}.{subkey}"')
 
     @cached_property
     def _wheel_builder(self) -> _WheelBuilder:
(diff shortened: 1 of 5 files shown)


mesonpy/__init__.py
```diff
@@ -715,29 +745,13 @@ def __init__(  # noqa: C901
...
-        # load meson args
-        for key in self._get_config_key('args'):
-            self._meson_args[key].extend(self._get_config_key(f'args.{key}'))
-        # XXX: We should validate the user args to make sure they don't conflict with ours.
-
-        self._check_for_unknown_config_keys({
-            'args': _MESON_ARGS_KEYS,
-        })
+        # load meson args from pyproject.toml
+        pyproject_config = _validate_pyproject_config(pyproject)
```
FFY00, line 752: Since `pyproject-metadata` is no longer optional, we can use it's API for our own validation.

dnicolodi: Yup. Working on it. I'm busy fighting mypy...

dnicolodi: In particular
```
mesonpy/__init__.py:795: error: Argument "version" to "StandardMetadata" has incompatible type "str"; expected "Optional[Version]"  [arg-type]
mesonpy/__init__.py:804: error: Incompatible types in assignment (expression has type "str", variable has type "Optional[Version]")  [assignment]
```
where the second is a preexisting issue. I don't know why mypy in CI does not find it.

FFY00: You need to make it a `packaging.version.Version`.

dnicolodi: Sure, but I don't understand why it was not flagged before. I didn't change the code at line 804.

dnicolodi: BTW, which `pyproject-metadata` validation API are you referring too? Looking at the source code I don't see anything related.

FFY00: The `DataFetcher` class. We should replicate `StandardMetadata`, but for our metadata.

dnicolodi: I'm not sure I understand how it should be used. Does an example exist somewhere? Maybe I'll do that in a separate PR.

FFY00 (changes requested):

mesonpy/__init__.py
FFY00: > `DataFetcher` seems an adapter used by `StandardMetadata` to obtain the metadata field values. I don't see much about validation. Or do you mean to use it to populate the metadata from `meson.build` when there is no metsdata in `pyproject.toml`? It seems not worth the effort to get just the package name and the version.

When you fetch a key via the helper methods, it validates the type, that's what I meant by its validation. Unfortunately, it can't check for unknown keys yet, an initial implementation was provided in #334, but I will finish some of the details I am not happy about and push it `pyproject-metadata`.

mesonpy/__init__.py
```diff
@@ -654,15 +656,83 @@ def build_editable(self, directory: Path, verbose: bool = False) -> pathlib.Path
...
+            for key, val in value.items():
+                check = scheme.get(key)
+                if check is None:
+                    raise ConfigError(f'unknown configuration entry "{name}.{key}"')
+                table[key] = check(val, f'{name}.{key}')
+            return table
+        return func
+
+    def _strings(value: Any, name: str) -> List[str]:
+        if not isinstance(value, list) or not all(isinstance(x, str) for x in value):
```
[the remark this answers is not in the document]
dnicolodi: The cannot pass tuples, see PEP 517 https://peps.python.org/pep-0517/#config-settings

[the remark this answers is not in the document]
dnicolodi: Scratch that. This validates the configuration form `pyproject.toml` and the toml parser returns lists, not tuples.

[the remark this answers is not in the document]
FFY00: Right, sorry. I misread.

tests/test_pep517.py
```diff
@@ -67,7 +67,7 @@ def test_invalid_config_settings(capsys, package_pure, tmp_path_session):
             method(tmp_path_session, {'invalid': ()})
         out, err = capsys.readouterr()
         assert out.splitlines()[-1].endswith(
-            'Unknown configuration entry "invalid"')
+            'unknown option "invalid"')
```
FFY00, line 70: Can we keep the capitalization?

tests/test_pep517.py
```diff
@@ -76,4 +76,4 @@ def test_invalid_config_settings_suggest(capsys, package_pure, tmp_path_session)
             method(tmp_path_session, {'setup_args': ()})
         out, err = capsys.readouterr()
         assert out.splitlines()[-1].endswith(
-            'Unknown configuration entry "setup_args". Did you mean "setup-args" or "dist-args"?')
+            'unknown option "setup_args". did you mean "setup-args" or "dist-args"?')
```
FFY00, line 79: Same.

mesonpy/__init__.py
```diff
@@ -949,45 +994,18 @@ def _meson_version(self) -> str:
 
     @property
     def name(self) -> str:
-        """Project name. Specified in pyproject.toml."""
-        name = self._metadata.name if self._metadata else self._meson_name
-        assert isinstance(name, str)
-        return name.replace('-', '_')
+        """Project name."""
+        return str(self._metadata.name).replace('-', '_')
```
FFY00, line 998: `str` is not needed here. The assert was because mypy was complaining because of the import IIRC.

mesonpy/__init__.py
```diff
@@ -685,6 +686,44 @@ def _strings(value: Any, name: str) -> List[str]:
...
+    def _string_or_strings(value: Any, name: str) -> List[str]:
+        return list([value,] if isinstance(value, str) else value)
+
+    options = {
+        'builddir': _string,
+        'editable-verbose': _bool,
+        'dist-args': _string_or_strings,
+        'setup-args': _string_or_strings,
+        'compile-args': _string_or_strings,
+        'install-args': _string_or_strings,
```
FFY00, line 709: We should not allow strings here.

mesonpy/__init__.py
```diff
@@ -685,6 +686,44 @@ def _strings(value: Any, name: str) -> List[str]:
     return scheme(table, 'tool.meson-python')
 
 
+def _validate_config_settings(config_settings: Dict[str, Any]) -> Dict[str, Any]:
+    """Validate options received from build frontend."""
+
+    def _string(value: Any, name: str) -> str:
+        if not isinstance(value, str):
+            raise ConfigError(f'only one value for "{name}" can be specified')
```
FFY00, line 694: This error messages assumes it can only be a container, while in fact it can be anything. For most PEP 517 frontends, it will be a container of some sort, but that is not guaranteed.

FFY00 (changes requested):
Just some minor comments, then I think it is ready for merging. Though, as mentioned in the other PR, we will probably move this code to `pyproject-metadata` when it implements the unknown keys check functionality anyway.

mesonpy/__init__.py
dnicolodi: No, it can be only a string or a list of strings. It is written in the PEP

mesonpy/__init__.py
dnicolodi: Please make up your mind. A little while ago you said that we need to allow frontends to pass single arguments as strings and that I broke that while refactoring.

mesonpy/__init__.py
FFY00: The PEP only makes a *recommendation* when frontends are building the config settings from a user provided string mapping :sweat_smile:

In the PEP, you can see 3 words capitalized, "MUST", "SHOULD", and "MAY", they indicate obligation, recommendation, and optionality, respectively. So while the PEP does recommend for that to be the case, it does not mandate it. It also explicitly mentions that fronts can have an alternative mechanism to build the dictionary.

> Build frontends MAY also provide arbitrary other mechanisms for users to place entries in this dictionary.

It's a simple enough check anyway.

mesonpy/__init__.py
FFY00: Sorry, sorry, I forgot this code was for `config_settings`.

mesonpy/__init__.py
dnicolodi: Please, if you need to quote from the PEP, do no quote only the part that makes it look like it supports your claims:

> This argument, which is passed to all hooks, is an arbitrary dictionary provided as an “escape hatch” for users to pass ad-hoc configuration into individual package builds. Build backends MAY assign any semantics they like to this dictionary. Build frontends SHOULD provide some mechanism for users to specify arbitrary string-key/string-value pairs to be placed in this dictionary. For example, they might support some syntax like --package-config CC=gcc. In case a user provides duplicate string-keys, build frontends SHOULD combine the corresponding string-values into a list of strings. Build frontends MAY also provide arbitrary other mechanisms for users to place entries in this dictionary.

The dictionary is string-key, string-value. You can use arbitrary mechanism to add entries to the dictionary, but the dictionary has string keys and string values.

mesonpy/__init__.py
dnicolodi: mypy in CI disagrees with you https://github.com/mesonbuild/meson-python/actions/runs/4359109213/jobs/7620540144#step:5:13

FFY00: That field can only ever be a string.

https://pep621.readthedocs.io/en/latest/#pyproject_metadata.StandardMetadata.name

I don't think mypy is loading the type hint correctly, hence the assert.

dnicolodi: It's the version of mypy used in the CI, later versions of mypy do not complain

mesonpy/__init__.py
dnicolodi: Can you please take your time and read all the text?

> Build frontends SHOULD provide some mechanism for users to specify arbitrary string-key/string-value pairs to be placed in this dictionary.

dnicolodi: https://github.com/pypa/setuptools/blob/main/setuptools/build_meta.py#L142

PR #308, opened 2023-02-14
CI: run tests with Python 3.12

dnicolodi:


dnicolodi:
Interesting: Cython 0.29.3 generated code using floats does not compile with Python 3.12.

FFY00 (approved):
Thanks!

dnicolodi:
This was not really meant to be merged yet. I thought that we have the policy of not merging PRs with failing tests, thus I didn't mark it explicitly. As noted, the tests fail on Python 3.12 because of a problem in the released version of Cython.

FFY00:
Cython uses private API so the test is unstable by design, I don't want us to have to spend time investigating possible failures and block things while they play cat and mouse with the CPython upstream. We should mark it as xfail, and make sure we are not using Cython to test anything other than just Cython projects.

I removed all the pass check requirements when we were refactoring the CI because the plan was to have a single job that reports the success of the other ones, so that we don't have to keep updating the branch protection rules manually, but I don't think that ever went through. I should add them back.

Anyway, dev Python releases were never required to be passing, they're purely informational, to give us an heads up to possible breakages with new Python versions. Everything else should be passing, unless it's something known to be broken and we haven't fixed it yet, so that we are not blocking other work.

dnicolodi:
Cython has been using private CPython API for more than 20 years and it seems that most did not even notice. I don't know what all the recent noise about Cython using the private API is about. Anyhow, now that Python 3.12 is in beta and the ABI should be stable, I expect a Cython release fixing the issue very soon. A Meson release dropping the dependency on distutils on recent Python releases will also appear soonish. Thus the added dependency will probably need to be revised.

FFY00:
The [Faster CPython](https://github.com/faster-cpython/ideas) team has been making a lot of internal changes, which is why we got the big speedups in 3.11, and they are not slowing down, so I'd expect more Cython breakage if they keep using private APIs.

rgommers:
We should revert this. The job is failing, and the dependency on `setuptools` is not desired (also, it has a hard conflict with my numpy dev env, which has a hard requirement on `setuptools < 60`, which is annoying since I actively work on both `numpy` and `meson-python`).

The Cython private API/ABI discussion really is not relevant here. And no, it should not be marked as `xfail`. We should merge a job like this only when the issues have been fixed.

Let's be very clear here: the latest stable releases of CPython and Cython work together fine, and both are on beta 1 for their next feature release (>6 months away till actual release, and 5 months till Python 3.12rc1 which is the earliest time anyone should be uploading binaries to PyPI). If those conflict, we wait till it's fixed and we're not going around passing judgement on one of the two projects. Everyone fixes bugs and makes decisions as best as they can, with the limited resources they have.

> The [Faster CPython](https://github.com/faster-cpython/ideas) team has been making a lot of internal changes, which is why we got the big speedups in 3.11, and they are not slowing down, so I'd expect more Cython breakage if they keep using private APIs.

Nothing broke in any 3.11 patch releases, which is the relevant thing to look at.

dnicolodi:
The dependency is conditional to installing with Python 3.12, thus you should be fine, unless you use Python 3.12, in which case things are broken anyhow. I have the tests fixed in #331 using Cython 3.0b1. I think it is a reasonable solution.

FFY00:
> And no, it should not be marked as `xfail`.

> Nothing broke in any 3.11 patch releases, which is the relevant thing to look at.

But it broke in the pre-releases. Even though this issue is not a big deal for normal releases, as the CPython projects tries to avoid breaking the private APIs in patch releases, it definitely is for pre-releases, where while there's still care to avoid breaking private APIs, it's mostly limited to simply avoiding unnecessary API changes.

CPython releases should be backwards compatible, barring deprecations, so testing against newer pre-releases should really only break if there's any regression, which is what we want to be testing. I don't want to have to wait until Cython updates their private API usage, and tags a new release, to be able to test for actual regressions, that might affect us, in the new release. Even if we do wait, things can still break again in newer pre-releases as the CPython release gets into shape. On top of that, considering that Cython can break on CPython patch releases, even though it might not happen often, I don't think it's unreasonable to mark the tests as xfail **if** the reason they fail is Cython modules not compiling. I don't really understand what's the issue here?

After the new Cython release is out, we can just switch the tests to use `CYTHON_LIMITED_API`, making them stable. The only question there is that `CYTHON_LIMITED_API` might still be experimental, but if it works on our examples it shouldn't really break afterwards.

> We should merge a job like this only when the issues have been fixed.

For the -dev job we've never done this in this repo, as long as the fail is not caused by us, and we don't do it in other places like https://github.com/pypa/build either. Personally, I find it easier to just have the job there and check if there has been any progress in the upstream from time to time, so that I can ping anyone if needed or help resolving small blockers that might exist. Not merging it means that I have to trigger CI re-runs periodically, wait for it to run, and then remember to go check the results.

What we didn't do in this PR, but should probably be doing, is marking the tests that fail because of external issues as xfail on the pre-releases, so that we can still keep testing newer pre-releases.

rgommers:
Having CI green is an almost universal practice, and very useful. The job has been fixed now, but let's try to aim for that. Red crosses are a significant distraction for any contributor.

For the Cython parts, let's discuss tomorrow.

PR #309, opened 2023-02-15
Rename MACOS_DEPLOYMENT_TARGET back to MACOSX_DEPLOYMENT_TARGET

stefanv:
This is the environment variable name used by Apple (see compat(5)). It is also supported by cibuildwheel, CMAKE, etc.

It was renamed in 9e84bd993c359721c3011cddd7f2403e5b019f74

Closes https://github.com/pypa/cibuildwheel/issues/1419

### DIFF
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -37,11 +37,11 @@ def test_wheel_tag():
 @pytest.mark.skipif(platform.system() != 'Darwin', reason='macOS specific test')
 def test_macos_platform_tag(monkeypatch):
     for minor in range(9, 16):
-        monkeypatch.setenv('MACOS_DEPLOYMENT_TARGET', f'10.{minor}')
+        monkeypatch.setenv('MACOSX_DEPLOYMENT_TARGET', f'10.{minor}')
         assert next(packaging.tags.mac_platforms((10, minor))) == mesonpy._tags.get_platform_tag()
     for major in range(11, 20):
         for minor in range(3):
-            monkeypatch.setenv('MACOS_DEPLOYMENT_TARGET', f'{major}.{minor}')
+            monkeypatch.setenv('MACOSX_DEPLOYMENT_TARGET', f'{major}.{minor}')
             assert next(packaging.tags.mac_platforms((major, minor))) == mesonpy._tags.get_platform_tag()
 
 
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -5,7 +5,6 @@
 import os
 import pathlib
 import platform
-import re
 import sysconfig
 
 from collections import defaultdict
@@ -93,18 +92,3 @@ def test_tag_mixed_abi(monkeypatch):
         'platlib': [f'extension{ABI3SUFFIX}', f'another{SUFFIX}'],
     })
     assert str(builder.tag) == f'{INTERPRETER}-{ABI}-{PLATFORM}'
-
-
-@pytest.mark.skipif(platform.system() != 'Darwin', reason='macOS specific test')
-def test_tag_macos_build_target(monkeypatch):
-    monkeypatch.setenv('MACOS_BUILD_TARGET', '12.0')
-    builder = wheel_builder_test_factory(monkeypatch, {
-        'platlib': [f'extension{SUFFIX}'],
-    })
-    assert builder.tag.platform == re.sub(r'\d+\.\d+', '12.0', PLATFORM)
-
-    monkeypatch.setenv('MACOS_BUILD_TARGET', '10.9')
-    builder = wheel_builder_test_factory(monkeypatch, {
-        'platlib': [f'extension{SUFFIX}'],
-    })
-    assert builder.tag.platform == re.sub(r'\d+\.\d+', '10.9', PLATFORM)
--- a/docs/reference/environment-variables.rst
+++ b/docs/reference/environment-variables.rst
@@ -48,7 +48,7 @@ Environment variables
           Please check our :ref:`how-to-guides-editable-installs` guide for more
           information on how to use it.
 
-   * - ``MACOS_DEPLOYMENT_TARGET``
+   * - ``MACOSX_DEPLOYMENT_TARGET``
      - Specify the target macOS version.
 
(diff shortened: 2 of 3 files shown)


stefanv:
There is one other similar occurrence, which is `MACOS_BUILD_TARGET`, which occurs only in `tests/test_tags.py`.

eli-schwartz (approved):

henryiii (approved):

henryiii:
Ideally, it would be nice to get this into a patch release, since it is a pretty bad regression that breaks the ability to make wheels for older macOS versions (unless you know to set `MACOS_BUILD_TARGET` as well as `MACOSX_BUILD_TARGET` :) ).

dnicolodi:
Ups, this typo is mine. I remember cross checking the variable name with the one used by other tools and convincing myself that all was good. I also tested this and it seemed to work fine, for example here https://github.com/dnicolodi/python-siphash24/actions/runs/3553666516/jobs/5969311668 cibuildwheel does not complain. Anyhow, the fix is obviously correct.

The `MACOS_BUILD_TARGET` thing in the tests is broken: the test does not test what it looks like it tests (setting this environment variable has no effect on the computed wheel tags but the test is broken). It is probably a left over from when the wheel tags where computed differently that I forgot to clean up. I'm having a look at what the story is there.

dnicolodi:
Nope, I added the broken `MACOS_BUILD_TARGET` test and I've looked everywhere, but it really seem to be something I made up. I think the test can be simply removed. Does anyone know how I could have come up with this environment variable name?

dnicolodi:
@stefanv I don't find anything about environment variables in the `comapt` man page on my macOS system with Xcode 14.2. Where did you find it documented?

stefanv:
> @stefanv I don't find anything about environment variables in the `comapt` man page on my macOS system with Xcode 14.2. Where did you find it documented?

Thanks for reviewing, @dnicolodi!

Here's the relevant manpage (I don't own a mac, so can't say exactly where this was captured):

https://www.manpagez.com/man/5/compat/

Czaki:
> The `MACOS_BUILD_TARGET` thing in the tests is broken

It was broken, because meson, in contrast to the wheel, does not check the platform tag in binary files, but only read values from the environment variable.

jarrodmillman:
> Ideally, it would be nice to get this into a patch release, since it is a pretty bad regression that breaks the ability to make wheels for older macOS versions (unless you know to set `MACOS_BUILD_TARGET` as well as `MACOSX_BUILD_TARGET` :) ).

@henryiii A 0.13.0rc1 release would be great, since we are already using 0.13.0rc0 for scikit-image.

dnicolodi:
> > The `MACOS_BUILD_TARGET` thing in the tests is broken
> 
> It was broken, because meson, in contrast to the wheel, does not check the platform tag in binary files, but only read values from the environment variable.

Would you like to elaborate? Meson is not responsible for generating Python wheels. Also, which piece of software react to the `MACOS_BUILD_TARGET` environment variable? I haven't found any. meson-python only checks the ` MACOSX_DEPLOYMENT_TARGET` to produce the wheel tags as it trusts the fact that if the environment variable is set, the toolchain has done the right thing and build a binary compatible with the specified macOS version. I don't see where `MACOS_BUILD_TARGET` enters in this process.

dnicolodi:
> Here's the relevant manpage (I don't own a mac, so can't say exactly where this was captured):
> 
> https://www.manpagez.com/man/5/compat/

Thanks. The linked man page is the same as the one on my system. I just overlooked the mention of the `MACOSX_DEPLOYMENT_TARGET` environment variable in it.

Czaki:
The test that was removed in this PR and does not fails even if a typo in the variable because there is no validation if files are created correctly. 


I do not play with meson, but when using another build system, it happens to have build polluted by the previous run (ex. make, cmake will not rebuild all files because the env variable is changed, only this that sources files were changed) or part of build process may be the usage of binary blob give by a vendor (like in this package https://github.com/tlambert03/nd2). 

Both scenarios will lead to an incorrectly tagged wheel file. First may be minimalized by using CI, but still user could enable cache operations for speedup build.

dnicolodi:
> The test that was removed in this PR and does not fails even if a typo in the variable because there is no validation if files are created correctly.

This is not what was wrong with the test. The typo was in an environment variable name not touched by the removed test.

tests/test_tags.py
```diff
@@ -93,18 +92,3 @@ def test_tag_mixed_abi(monkeypatch):
         'platlib': [f'extension{ABI3SUFFIX}', f'another{SUFFIX}'],
     })
     assert str(builder.tag) == f'{INTERPRETER}-{ABI}-{PLATFORM}'
-
-
-@pytest.mark.skipif(platform.system() != 'Darwin', reason='macOS specific test')
-def test_tag_macos_build_target(monkeypatch):
-    monkeypatch.setenv('MACOS_BUILD_TARGET', '12.0')
```
Czaki, line 100: I see here that is touched

dnicolodi: Sure, but who reads it? Which piece of software reads this environment variable? The environment variable expected by the toolchain is `MACOSX_DEPLOYMENT_TARGET`. I suspect that `MACOS_BUILD_TARGET` is something I made up.

Czaki: Yes. but when this test will set variable to 10.9 and the tag calculation will check files then this test will fail, as the test will in both runs return 12.0 (as the wrong variable was set).

dnicolodi: @Czaki you don't make any sense. I'm giving up.

Czaki: Your assumption that after correctly setting the environment variable, the toolchain will produce the correct files is wrong. It will often be true, but not always. 

Nad I only point the test that, if checking the platform tag will examine files, the test will find a bug in the variable name.

dnicolodi: Except that the test is not compiling any code an there is no file to examine.

Czaki: After reading the code I was pretty sure that it compile. I'm sorry for bothering you.

PR #311, opened 2023-02-17
solved syntax errors

akshaya-hub:
syntax errors  are solved


rgommers:
@akshaya-hub this is nonsense, please don't submit random PRs like this.

dnicolodi:
This is not Python. It is a Meson build definition file, and the syntax is correct: https://mesonbuild.com/Syntax.html#if-statements

Issue #29, opened 2022-01-21
UX for pinning build-time dependencies


rgommers:
There is a practical UX problem with `pyproject.toml` that we need to solve. A short description would be:
- Any project that builds against the `numpy` C API has to deal with API and ABI compatibility (see [the numpy docs on that](https://numpy.org/devdocs/user/depending_on_numpy.html?highlight=abi#adding-a-dependency-on-numpy) for details). 
- as a consequence, the dependency specification for local development is different from the one needed for uploading an sdist to PyPI:
  - for local development, you want loose dependencies, e.g. `numpy >= 1.19.5` (both a build and runtime dep)
  - in sdist released to PyPI you want: `oldest-supported-numpy` (build dep metapackage, contains `==` pins per platform) and `numpy>='buildtime_dep,<1.25.0'  # should be `N+3` where `N is current minor version.
 
That is obviously a complex set of dependencies to express in `pyproject.toml`, and we can't have two sets in there at the same time. The solution @FFY00 and I just discussed is:

1. On the `main` branch, use the loose development dependencies (`numpy>=1.19.5`)
2. In the CI job(s) that produce artifacts for release to PyPI, ensure that the _lowest_ supported dependencies (which may be platform-dependent) are installed.
3. In the _release_ branch, replace the dependencies with the build-time deps needed for uploading to PyPI.
4. Have support in `mesonpy` for the various kinds of pinnings people need for releases and API/ABI compatibility.

A rough sketch of how (4) could look:

```toml
[build-system]
requires = [
  "numpy >= 1.19.5",  # replace with (e.g.) `oldest-supported-numpy` in a release branch
]

[tool.mesonpy]
runtime_pinnings = [
  # if build version detected is 1.20.3, replace with `>=1.20.3` (while preserving the upper bound,
  # so end result would be for example `'>=1.20.3,<1.25.0')
  "numpy = 'pin_compatible'",
  "pythran = 'bugfix_only'",  # if build version detected is 0.10.0, override the build dependency with `'>=0.10.0,<0.11.0'`
  "torch = 'exact'",  # runtime version must equal build-time version (PyTorch has no ABI stability guarantees)
]
```

[...]

PR #312, opened 2023-02-17
CI: add job to build wheels and upload them when a release is tagged

dnicolodi:
I don't know if we want this. The main goal is to potentially distribute the duty of cutting releases and reduce the bus factor. We may not want to build the wheel and the sdist for pull requests, in that case the action triggers may be changed. The current choice is only to make sure that the action works properly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,22 @@
+# SPDX-FileCopyrightText: 2022 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+name: build
+on:
+  - push
+  - pull_request
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: python -m pip install build twine
+      - run: python -m build
+      - run: if [[ "${{ github.ref }}" =~ [0-9]+\.[0-9]+\.[0-9]+(\..*)? ]]; then echo "UPLOAD=1" >> $GITHUB_ENV; fi
+      - run: python -m twine upload dist/*
+        env:
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.PYPI }}
+        if: ${{ env.UPLOAD }}


FFY00:
While it shouldn't be a big issue for this project, I'd prefer if releases were made locally. This reduces targeting the CI as an attack vector, and allows the artifacts to be PGP signed.

dnicolodi:
Sure, I understand. Can you please tag a 0.13.0.pre1 release as requested in #309 ?

dnicolodi:
By the way, does PyPI support uploading GPG signed wheels?

FFY00:
> Sure, I understand. Can you please tag a 0.13.0.pre1 release as requested in #309 ?

Sure, is it worth to tag now, before #279 is in? I am in the middle of reviewing it, and was hoping to get it in early next week.
I am also working on #29, which I was hoping to merge next week.

But we definitely tag a pre-release now, and another one later.

> By the way, does PyPI support uploading GPG signed wheels?

Yep, it doesn't show the signatures in the artifact list, which is fair, but you can download them. I used to also think it didn't, but it does, which is nice.

rgommers:
Maybe just cherry-picking the gh-309 fix and doing a `0.12.1` release is the way to go there? That was a fairly annoying regression, that both scikit-image and SciPy have already burned some time on working around. That way, there's no time pressure on `0.13.0`.

FFY00:
Yeah, that makes sense. Will do.

dnicolodi:
> > By the way, does PyPI support uploading GPG signed wheels?
> 
> Yep, it doesn't show the signatures in the artifact list, which is fair, but you can download them. I used to also think it didn't, but it does, which is nice.

This is nice. But how do you upload the signatures? Does twine take care of it? Does pip verify the signatures? I don't think so. Maybe I should go hunting for some documentation. This feature should maybe be advertised more.

dnicolodi:
> Sure, is it worth to tag now, before #279 is in? I am in the middle of reviewing it, and was hoping to get it in early next week.

I was under the impression that not everyone is sold on the implementation of editable wheels in #279.

I was asking for another pre-release because it seems some projects need it, but I don't know why scikit-image cannot use meson-python 0.12

dnicolodi:
> That was a fairly annoying regression, that both scikit-image and SciPy have already burned some time on working around.

Yeah. I'm sorry for the stupid regression. To my defense, that PR was reviewed in detail and no one spotted the typo.

rgommers:
Yes, I know - such typos happen and are easy to make and to overlook in review. No worries at all there.

> I was under the impression that not everyone is sold on the implementation of editable wheels in https://github.com/mesonbuild/meson-python/pull/279.

For me it's more that I don't have the time to really wrap my head around that stuff - happy to rely on you and @FFY00 here, if you both agree on the implementation and the use cases are addressed, then I think we'll be in a good place.

FFY00:
My goal was always to shift to a better approach, like the one in #279. I think it's just a matter of if/when we are able to get it to work reliably.

dnicolodi:
I think #279 works reliably and has extensive tests coverage. However, there may be use cases that I didn't think about and that are not covered by the tests. It would be nice if someone that uses editable wheels on complex projects tries it.

FFY00:
> But how do you upload the signatures? Does twine take care of it?

Yep, you just need to pass the `--sign`/`-s` flag to it.

> Does pip verify the signatures?

No, it doesn't have the architecture to verify them in a way that actually makes sense.

PR #313, opened 2023-02-17
Document releases

FFY00:


docs/security.rst
```diff
@@ -0,0 +1,32 @@
...
+.. SPDX-License-Identifier: MIT
+
+
+.. _security:
+
+********
+Security
+********
+
+Distributors are recommended to fetch release sources via the Git_ tags ir our
```
dnicolodi, line 12: s/ir/or/

FFY00: It's actually meant to be "on".

PR #319, opened 2023-02-23
ENH: add support for dynamic dependencies computation

dnicolodi:


FFY00 (changes requested):
This approach is not enough. You need to be able to match versions with more than 3 parts on the *release* field.

Let's discuss this on the issue first. I had written a reply to you and Ralf, but I've been trying to put all my thoughts down regarding this in a document, so I haven't submitted it yet. Some more discussion happened, so I'll probably have to rewrite my reply, but please just wait for my reply there.

dnicolodi:
Easy fix. Added. You can use `{v.major}.{v.minor}.{v.micro}.{v[3]}.{v[4]}`. It is not pretty, but that's what you get for having crazy version numbers.

FFY00:
I converted the PR to draft while we wait for a resolution in #29, so that the state of PR log is easier to understand. @dnicolodi I hope that is okay.

Once we get a decision on the issue, we can go ahead with a proper review and merge of this PR :)

PR #322, opened 2023-02-24
ENH: do not check extension module file name against expected suffix

dnicolodi:
Checking that the extensions modules filenames end with a suffix accepted by the current Python interpreter assumes that the wheel being assembled is for the same Python platform. This is not true when cross-compiling. The check protects against a very unlikely occurrence and makes life harder for tools that allow cross-compiling Python wheels. Remove it.

See #321.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -334,13 +334,9 @@ class _WheelBuilder():
                 match = re.match(r'^[^.]+(.*)$', path.name)
                 assert match is not None
                 suffix = match.group(1)
-                if suffix not in _EXTENSION_SUFFIXES:
-                    raise ValueError(
-                        f'Extension module {str(path)!r} not compatible with Python interpreter. '
-                        f'Filename suffix {suffix!r} not in {set(_EXTENSION_SUFFIXES)}.')
                 match = _EXTENSION_SUFFIX_REGEX.match(suffix)
-                assert match is not None
-                abis.append(match.group('abi'))
+                if match:
+                    abis.append(match.group('abi'))
 
         stable = [x for x in abis if x and re.match(r'abi\d+', x)]
         if len(stable) > 0 and len(stable) == len(abis) and all(x == stable[0] for x in stable[1:]):


rgommers:
+1 for doing this. We can either remove the check completely, or remove it when we know we're cross-compiling. Which is something that I think we anyway need to know - we already do know on macOS, and outside of macOS it's a matter of checking for a cross file in config-settings.

FFY00 (approved):
> We can either remove the check completely, or remove it when we know we're cross-compiling.

I'd prefeer if we just kept the check, but limited to when we are not cross-compiling.

dnicolodi:
I added the check, but I came to the conclusion that it is not a good idea. Among other things, it prevents installing a data file in `{platlib}`, which I think is an acceptable thing to do.

FFY00:
> it prevents installing a data file in `{platlib}`

A data file that ends with `.so`... but yeah, that's reasonable.

dnicolodi:
> A data file that ends with `.so`... but yeah, that's reasonable.

Oh, you are right. I forgot about the double extension check. However, if someone goes into the trouble of putting something that looks like a Python extension module for the wront interpreter version in `{platlib}` they win the right of having thing break if they try to import it.

doronbehar:
Thanks @rgommers for linking this PR from #321, I missed it somehow when browsing #321. While thinking my self how can I fix this issue with a patch applied upon the `0.12.1` tag, I noticed a few things which perhaps were missing from this PR?

First of all (easy), the docstring wasn't trimmed:

```diff
diff --git i/mesonpy/__init__.py w/mesonpy/__init__.py
index 78db5d8..5832d4e 100644
--- i/mesonpy/__init__.py
+++ w/mesonpy/__init__.py
@@ -341,10 +341,6 @@ class _WheelBuilder():
         and .so on other platforms) and, if they all share the same
         PEP 3149 filename stable ABI tag, return it.
 
-        All files that look like extension modules are verified to
-        have a file name compatibel with what is expected by the
-        Python interpreter. An exception is raised otherwise.
-
         Other files are ignored.
 
         """
```

Secondly, this `if`:

https://github.com/mesonbuild/meson-python/blob/961fb56b0ca70610597f8da912e3d494896d0927/mesonpy/__init__.py#L351-L355

Doesn't assume that the "Host" machine and "Build" / "Target" machine have the same "shared object" file extension (`so` usually)? Maybe this is too much to ask - that `meson-python` will support cross compiling from Linux to Windows for example, but perhaps at least a comment above this conditional would be nice.

rgommers:
Good points, thanks @doronbehar. 

> Maybe this is too much to ask - that `meson-python` will support cross compiling from Linux to Windows for example, but perhaps at least a comment above this conditional would be nice.

It's not too much to ask - I'm hoping we can make that work at some point. But in practice it's not that common to do that, so it may take a while before we need to work around this one. In the meantime, there's some bigger cross-compilation fish to fry, in Meson itself and in dealing with `sysconfig` and other things that everyone has to monkeypatch right now.

Issue #316, opened 2023-02-19
What to do when `$NINJA` points to an invalid ninja executable?


dnicolodi:
I think something is off with the code that verifies the ninja version in `get_requires_for_build_wheel()` and when the `NINJA` environment variable is set: if the variable points to a path that does not exist or to an executable that reports a too old version, we ask the `ninja` package to be installed. In `build_wheel()` the `NINJA` environment variable is updated only if not set, thus it still points to the invalid `ninja` and Meson will fail when meson-python executes `meson setup`.

I think that failing when the `NINJA` environment variable is set incorrectly is the right thing to do. However, we could fail a bit sooner, in particular without installing the `ninja` package that is never going to be used.

FFY00:
Yeah, that makes sense.

cc @henryiii

henryiii:
Agree, it was an oversight, we should not request `ninja` if `NINJA` is set.

eli-schwartz:
> In `build_wheel()` the `NINJA` environment variable is updated only if not set, thus it still points to the invalid `ninja`

Hmm, but why does it need to be set at all? The code being used to set it is the same code used to detect if it exists, which is also modeled on the same code Meson uses.

Specifically:
- if $NINJA is set, it is always a no-op:
  - and valid, it is setdefault'ed to the path it already is
  - and invalid, it is not setdefault'ed to anything
- if it is unset, it gets set to the detected version on $PATH, if any

The code only takes effect if the variable is unset. But in that case what you really care about is that meson running in the isolated env can pick up the relevant meson from either the system PATH or the pip-installed build_requires. Which Meson handles without setting this variable.

dnicolodi:
I think it was "defensive coding" against the ninja discovery algorithm used in meson-python and the one used in Meson diverged. I agree that it is not necessary.

eli-schwartz:
Hmm, fair enough :) it should be harmless either way and unrelated to this problem.

Perhaps _env_ninja_command() should check at the very bottom, right before falling out of the function with `return None`, if `env_ninja` is actually something. If it's something, but the version check failed, then raise an error instead of returning None.

PR #328, opened 2023-02-28
ENH: do not use the 'meson compile' indirection to build the project

### LINKED ISSUE #324: Should compilation backends different than `ninja` be supported?
meson-python injects a build dependency on `ninja` if it detects that `ninja` is not available. Thus it seems that it implicitly assumes that the meson build backed used is exclusively `ninja`. This is fine. However, the project compilation is initiated with the `meson compile -C $builddir` command, which is the backend-agnostic way offered by Meson to compile a project. Should this command be replaced by `ninja -C $builddir`? There isn't much difference, except that invoking `ninja` directly is potentially much faster. This is relevant for editable wheel installations, where the project rebuild is triggered on import.

dnicolodi:
meson-python already assumes that the meson build backed is ninja, thus there is no advantage in going through the backend agnostic meson compile command. Invoking ninja directly saves about 200 ms on an average system. This is relevant for editable wheel installations, where the project rebuild is triggered on import.

Fixes #324.


mesonpy/__init__.py
```diff
@@ -1227,7 +1227,7 @@ def get_requires_for_build_wheel(
 ) -> List[str]:
     dependencies = []
 
-    if _env_ninja_command() is None:
+    if os.environ.get('NINJA') is None and _env_ninja_command() is None:
```
eli-schwartz, line 1230: Changed per #316

I was trying to figure out why this wasn't an issue if we just don't add it as a build-requires. The answer is it depends on the previous commit raising a `ConfigError()` that is logically part of *this* commit.

FFY00 (commented):
Why isn't mypy failing with the first commit? The type hint was not updated.
I don't really get the reason for the second commit.

Unfortunately, this is a breaking change, because we allow users to pass arguments to `meson compile`. I mean, we can add a `ninja-args` setting and backfill it by parsing `compile-args` and looking at `--ninja-args`, but it's a bit painful, and also I don't feel great about deprecating `compile-args`, in case the status quo changes :sweat_smile: 

> meson-python already assumes that the meson build backed is ninja, thus there is no advantage in going through the backend agnostic meson compile command.

Hum, really? Why does it take `--vs-args` and `--xcode-args` then?

> Invoking ninja directly saves about 200 ms on an average system.

Hum, that's relevant. How did you benchmark this?
We may be able to optimize this in Meson itself, btw.

dnicolodi:
> I don't know how @dnicolodi benchmarked it,

I didn't. I trusted the number @rgommers reported in the linked issue.

dnicolodi:
> Why isn't mypy failing with the first commit? The type hint was not updated.

Our CI does not test individual commits but the whole PR. I put the hunk updating the type annotation in the wrong patch when I committed the changes. I'll fix it.

> I don't really get the reason for the second commit.

Less levels of indirection are nice, and, most importantly, now one of the commands that need to be executed is not `meson`, thus having the wrapper makes it much more verbose rather than less.

> Unfortunately, this is a breaking change, because we allow users to pass arguments to `meson compile`. I mean, we can add a `ninja-args` setting and backfill it by parsing `compile-args` and looking at `--ninja-args`, but it's a bit painful, and also I don't feel great about deprecating `compile-args`, in case the status quo changes

As @eli-schwartz explained, the command lines arguments accepted by `meson compile` (except for the ones specific to each compilation back-end) are mostly the same as the ones accepted by ninja, and there aren't many ninja command line argument that make sense to pass to meson-python, anyway.

mesonpy/__init__.py
dnicolodi: I'm not sure I follow. Taking this commit in isolation, before this commit, if the `$NINJA` environment variable pointed to an invalid ninja executable, we would still require the `ninja` package but we would have left the environment variable pointing to the wrong executable and meson would have reported that it cannot find ninja. With this commit nothing changes in this respect, we just do not request to install a ninja package that would not be used anyway (because meson obeys the `$NINJA` environment variable).

FFY00 (approved):
Please fix the commit before merging. Thanks!

dnicolodi:
> Please fix the commit before merging.

Do you mean moving the type hint change to the right patch? This is already done.

PR #336, opened 2023-03-09
ENH: format errors nicely on multiline error strings

FFY00:
Before
```
meson-python: error: A single line error message
```
```
meson-python: error: A multiline error message. Eg. X went wrong in the fields:
- aaa
- bbb

And Y went wrong in:
- ccc
- ddd
```

After
```
meson-python: error: A single line error message
```
```
meson-python: error:
>
> A multiline error message. Eg. X went wrong in the fields:
> - aaa
> - bbb
> 
> And Y went wrong in:
> - ccc
> - ddd
> 
```

With the latest commit, I kept single-line error the same, following the review, keeping the change for multi-line errors only.

Note, the error messages get nicely colorized too. The `>` line prefix is red, and the following is reset text.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -1023,7 +1023,7 @@ def _pyproject_hook(func: Callable[P, T]) -> Callable[P, T]:
     def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
         try:
             return func(*args, **kwargs)
-        except Error as exc:
+        except (Error, pyproject_metadata.ConfigurationError) as exc:
             print((
                 '{red}meson-python: error:{reset}\n'
                 '{red}>{reset}\n' +
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -1024,7 +1024,12 @@ def _pyproject_hook(func: Callable[P, T]) -> Callable[P, T]:
         try:
             return func(*args, **kwargs)
         except Error as exc:
-            print('{red}meson-python: error:{reset} {msg}'.format(msg=str(exc), **_STYLES))
+            print((
+                '{red}meson-python: error:{reset}\n'
+                '{red}>{reset}\n' +
+                textwrap.indent(str(exc).strip(), prefix='{red}>{reset}  ') + '\n' +
+                '{red}>{reset}\n'
+            ).format(**_STYLES))
             raise SystemExit(1) from exc
     return wrapper
 
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -1024,12 +1024,16 @@ def _pyproject_hook(func: Callable[P, T]) -> Callable[P, T]:
         try:
             return func(*args, **kwargs)
         except (Error, pyproject_metadata.ConfigurationError) as exc:
-            print((
-                '{red}meson-python: error:{reset}\n'
-                '{red}>{reset}\n' +
-                textwrap.indent(str(exc).strip(), prefix='{red}>{reset}  ') + '\n' +
-                '{red}>{reset}\n'
-            ).format(**_STYLES))
+            error_msg = '\n{red}meson-python: error:{reset}'
+            if '\n' in str(exc):
+                error_msg += (
+                    '{red}>{reset}\n' +
+                    textwrap.indent(str(exc).strip(), prefix='{red}>{reset}  ') + '\n' +
+                    '{red}>{reset}\n'
+                )
+            else:
+                error_msg += f' {exc}'
+            print(error_msg.format(**_STYLES))
             raise SystemExit(1) from exc
     return wrapper
 
--- a/tests/test_pep517.py
+++ b/tests/test_pep517.py
@@ -66,7 +66,7 @@ def test_invalid_config_settings(capsys, package_pure, tmp_path_session):
         with pytest.raises(SystemExit):
             method(tmp_path_session, {'invalid': ()})
         out, err = capsys.readouterr()
-        assert out.splitlines()[-1].endswith('Unknown option "invalid"')
+        assert 'Unknown option "invalid"' in out
 
 
 def test_invalid_config_settings_suggest(capsys, package_pure, tmp_path_session):
@@ -74,7 +74,7 @@ def test_invalid_config_settings_suggest(capsys, package_pure, tmp_path_session)
         with pytest.raises(SystemExit):
             method(tmp_path_session, {'setup_args': ()})
         out, err = capsys.readouterr()
-        assert out.splitlines()[-1].endswith('Unknown option "setup_args". Did you mean "setup-args" or "dist-args"?')
+        assert 'Unknown option "setup_args". Did you mean "setup-args" or "dist-args"?' in out
 
 
 def test_validate_config_settings_invalid():


rgommers:
@FFY00 thanks. It would be helpful if you could add examples of both the old and new output in the PR description, so it's clear what this PR changed.

mesonpy/__init__.py
```diff
@@ -1138,7 +1138,12 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
...
             return func(*args, **kwargs)
         except Error as exc:
-            print('{red}meson-python: error:{reset} {msg}'.format(msg=str(exc), **_STYLES))
+            print((
+                '{red}meson-python: error:{reset}\n'
+                '{red}>{reset}\n' +
+                textwrap.indent(str(exc).strip(), prefix='{red}>{reset}  ') + '\n' +
+                '{red}>{reset}\n'
+            ).format(**_STYLES))
             raise SystemExit(1) from exc
```
dnicolodi, line 1147: This makes all errors printed by meson look like:
```
meson-python: error:
> 
>  Error message
>
```
turning one line error messages into four lines error messages wasting three lines for unnecessary decoration and makes meson-python errors harder to grep for. I don't expect that meson-python will be emitting many very long error messages, thus this format does not seem optimal.

I would prefer to use the same prefix for all lines, without decoration:
```
meson-python: error: Error message
```
with multi-line messages (which I guess are emitted by newer versions of pyproject-metadata, as meson-python itself does not emit any multi-line error messages), formatted like this:
```
meson-python: error: Error message first line
meson-python: error: Error message second line
meson-python: error: Even more lines...
```
This is more compact, immediate to grep for, and works equally well for reporting multiple unrelated errors and single error conditions with details spanning multiple lines.

FFY00: > I don't expect that meson-python will be emitting many very long error messages, thus this format does not seem optimal.

We will once `pyproject-metadata` is able to report all validation errors at once (see https://github.com/FFY00/python-pyproject-metadata/issues/45).

> I would prefer to use the same prefix for all lines, without decoration:
> 
> ```
> meson-python: error: Error message
> ```
> 
> with multi-line messages (which I guess are emitted by newer versions of pyproject-metadata, as meson-python itself does not emit any multi-line error messages), formatted like this:
> 
> ```
> meson-python: error: Error message first line
> meson-python: error: Error message second line
> meson-python: error: Even more lines
> ```

This looks like multiple errors, and looks weird on big messages.

> This is more compact, immediate to grep for,

But the error is the last thing we will print after exiting, and the `meson-python: error:` prefix isn't used anywhere here... I don't understand why people would be grepping, or at least why that'd be a use-case we care about.

> and works equally well for reporting multiple unrelated errors and single error conditions with details spanning multiple lines.

But that's not something we do, or need, and I don't think we will, I can't really think of a reason why we'd have multiple different kinds of errors.

---

I am okay with keeping the single line format on single line errors, and only opting for this one when we have multi-line error descriptions. What do you think?

dnicolodi: So we are making the logging of an error message with a single line of text looks ridiculous to optimize for an hypothetical future in which one of our dependencies will report multiple errors?

> I am okay with keeping the single line format on single line errors, and only opting for this one when we have multi-line error descriptions. What do you think?

Let's wait till there is the need to report multi-line errors to add this. At the moment it is not even clear in which form multiple validation errors will be reported by `pyproject-metadata`.

eli-schwartz: > Let's wait till there is the need to report multi-line errors to add this.

If the logic is going to if/else how to print it based on the presence of newlines in the message, then it should be fine to optimistically add support for multi-line errors without changing single-line errors. Even if the code is then never actually used, it's just "building infrastructure for the future". :D

FFY00: > So we are making the logging of an error message with a single line of text looks ridiculous to optimize for an hypothetical future in which one of our dependencies will report multiple errors?

> If the logic is going to if/else how to print it based on the presence of newlines in the message, then it should be fine to optimistically add support for multi-line errors without changing single-line errors. Even if the code is then never actually used, it's just "building infrastructure for the future". :D

`_pyproject_hook` currently only triggers on our errors, which we do control, but it should actually be triggering on *all* expected errors. As such, we should not assume the errors are single line.

> Let's wait till there is the need to report multi-line errors to add this.

I really don't understand the issue with keeping the format for single line errors and just handling multi-line errors better.

> At the moment it is not even clear in which form multiple validation errors will be reported by `pyproject-metadata`.

I don't think this is very relevant, but since I am the one implementing it, I can tell you it will raise an error something like this:

> Found several invalid keys in pyproject.toml:
>   - project.name: expecting a string, but got "None"
>   - project.license: the license file was not found
>   -  ...

It may be an `ExceptionGroup` with a custom repr, or a normal exception (with the same repr).

rgommers:
I've done some testing here, by comparing the behavior when not specifying `project.name`. On `main` I get this (no color-coding):

```
Found ninja-1.11.1 at /home/rgommers/mambaforge/envs/scipy-dev/bin/ninja
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pep517/in_process/_in_process.py", line 351, in <module>
    main()
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pep517/in_process/_in_process.py", line 333, in main
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pep517/in_process/_in_process.py", line 249, in build_wheel
    return _build_backend().build_wheel(wheel_directory, config_settings,
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/mesonpy/__init__.py", line 1034, in wrapper
    return func(*args, **kwargs)
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/mesonpy/__init__.py", line 1097, in build_wheel
    with _project(config_settings) as project:
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/mesonpy/__init__.py", line 984, in _project
    with Project.with_temp_working_dir(
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/mesonpy/__init__.py", line 829, in with_temp_working_dir
    yield cls(source_dir, tmpdir, build_dir, meson_args, editable_verbose)
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/mesonpy/__init__.py", line 717, in __init__
    self._metadata = pyproject_metadata.StandardMetadata.from_pyproject(pyproject, self._source_dir)
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pyproject_metadata/__init__.py", line 224, in from_pyproject
    raise ConfigurationError('Field `project.name` missing')
pyproject_metadata.ConfigurationError: Field `project.name` missing

```

With this PR I get (with color-coding of `meson-python: error:` and the `>` symbols):
```
Found ninja-1.11.1 at /home/rgommers/mambaforge/envs/scipy-dev/bin/ninja
meson-python: error:
>
>  Field `project.name` missing
>


```

So the key improvement is hiding the large and uninformative traceback (color coding is a nicety too) for single-line errors.

Nothing changes when there's multiple errors in the build log as of now. It's not something all that important compared to the single-error logging improvement I'd say.

mesonpy/__init__.py
rgommers: So I think this is a very minor point in the end. I don't have a clear preference, but I suggest simply removing the two lines containing `'{red}>{reset}\n'` to avoid the discussion and the extra two lines in the output that @dnicolodi does not like. 

Right now the error for a single config mistake by the user is:
```
meson-python: error:
>
>  Field `project.name` missing
>
```
It could be changed back to:
```
meson-python: error: Field `project.name` missing
```

It really doesn't matter too much either way. The gain of this PR is the exception chaining, so the huge traceback gets hidden. I'd like to get that merged.

FFY00:
> With this PR I get (with color-coding of `meson-python: error:` and the `>` symbols):

This should not be the case, it's a bug in the PR. The `>` format should only be used when the errors are multiline, per the review above.

rgommers (approved):
This is now back to single-line errors and a clear improvement without any downsides. So I'll merge this .Thanks @FFY00 & reviewers.



```
Found ninja-1.11.1 at /home/rgommers/mambaforge/envs/scipy-dev/bin/ninja

meson-python: error: Field `project.name` missing

```

Compared to this mess when using `meson-python` from the `main` branch:

<details>

```
Found ninja-1.11.1 at /home/rgommers/mambaforge/envs/scipy-dev/bin/ninja
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pep517/in_process/_in_process.py", line 351, in <module>
    main()
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pep517/in_process/_in_process.py", line 333, in main
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pep517/in_process/_in_process.py", line 249, in build_wheel
    return _build_backend().build_wheel(wheel_directory, config_settings,
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/mesonpy/__init__.py", line 1025, in wrapper
    return func(*args, **kwargs)
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/mesonpy/__init__.py", line 1088, in build_wheel
    with _project(config_settings) as project:
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/mesonpy/__init__.py", line 975, in _project
    with Project.with_temp_working_dir(
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/mesonpy/__init__.py", line 820, in with_temp_working_dir
    yield cls(source_dir, tmpdir, build_dir, meson_args, editable_verbose)
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/mesonpy/__init__.py", line 712, in __init__
    self._metadata = pyproject_metadata.StandardMetadata.from_pyproject(pyproject, self._source_dir)
  File "/home/rgommers/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/pyproject_metadata/__init__.py", line 224, in from_pyproject
    raise ConfigurationError('Field `project.name` missing')
pyproject_metadata.ConfigurationError: Field `project.name` missing

```

</details>

rgommers:
I don't think those will arrive any time soon, so I decided to ignore that part. Let's cross that bridge if and when we get to it - it really isn't very interesting either way.

I updated the commit message to explain the actually relevant, and significant, improvement here. I'll also re-title this PR to reflect that.

dnicolodi:
This broke reporting error messages that contain curly brackets. There are several places where meson-python reports errors about user supplied configuration.If any of these configuration strings contains a curly bracket, the error message results in an exception. meson-python also emits an error when it is not able to map a Meson install prefix into a wheel installation path and Meson install prefixes are defined with curly brackets.

FFY00:
Thank you for noticing it. Feel free to submit a PR if you want, otherwise I will submit one this week.

PR #325, opened 2023-02-27
ENH: set NDEBUG for release builds

rgommers:
`NDEBUG` is normally expected to be set for release builds, however for historical reasons Meson does not (yet) toggle `NDEBUG` for non-debug builds (which we already default to). So set this as the default. Choosing 'if-release' so that if users are explicitly asking for a debug build via config-settings, they don't see a change in behavior.

xref https://github.com/numpy/numpy/issues/22546#issuecomment-1440771047

PR #291, opened 2023-02-02
DOC: add documentation for Meson arguments

peter-urban:
This is a draft for the (how-to) meson-args documentation. It includes an example for using meson install_tags using meson-python's install options (PR https://github.com/mesonbuild/meson-python/pull/288 and issue https://github.com/mesonbuild/meson-python/issues/68)

I could not yet answer all todos
(Mention the order in which Meson arguments are added, and how it affect Meson)
and
(Explain how you can add extra Meson arguments via the build config settings)

Feel free to comment on what you like/dislike or directly copy/modify it to you liking

PR #338, opened 2023-03-10
DOC: add details on default build options used by meson-python

rgommers:
This follows up on the `n_debug` default added in gh-325, documents all other default options and they are used, as well as improves the page on using `pyproject.toml` settings.

### DIFF
--- a/docs/explanations/default-options.rst
+++ b/docs/explanations/default-options.rst
@@ -1,0 +1,64 @@
+.. SPDX-FileCopyrightText: 2023 The meson-python developers
+..
+.. SPDX-License-Identifier: MIT
+
+.. _explanations-default-options:
+
+*********************
+Default build options
+*********************
+
+Meson offers many `built-in options <https://mesonbuild.com/Builtin-options.html>`__,
+and in the vast majority of cases those have good defaults. There are a couple
+of cases however where ``meson-python`` either needs to or chooses to override
+those with its own defaults. To view what those are for the version of
+``meson-python`` you have installed, look at the *User defined options* section
+of the output during the configure stage of the build (e.g., by running
+``python -m build --wheel``). This will look something like:
+
+.. code-block:: text
+
+    User defined options
+      Native files     : /home/username/code/project/.mesonpy-native-file.ini
+      debug            : false
+      optimization     : 2
+      prefix           : /home/username/mambaforge/envs/project-dev
+      python.platlibdir: /home/username/mambaforge/envs/project-dev/lib/python3.10/site-packages
+      python.purelibdir: /home/username/mambaforge/envs/project-dev/lib/python3.10/site-packages
+      b_ndebug         : if-release
+
+Let's go through each option and why they are used:
+
+- meson-python uses a native file, written to the build dir and named
+  ``mesonpy-native-file.ini``, in order to point Meson at the correct
+  ``python`` interpreter to use (the same one for which ``meson-python`` was
+  installed). This is necessary, because Meson may otherwise look for the first
+  Python interpreter on the PATH (usually the same one, but not always the
+  case). Users may use ``--native-file`` to pass a second native file to Meson;
+  Meson will merge contents of both native file, so as long as the
+  user-provided file does not try to pass a different path for the ``python``
+  binary, this will work without a conflict.
+- The ``prefix`` and ``platlibdir``/``purelibdir`` options also point Meson at
+  that same interpreter and the environment in which it is installed.
+- The ``debug``, ``optimization`` and ``b_ndebug`` options are overridden,
+  because Meson defaults to values that are appropriate for development, while
+  the main purpose of meson-python is to build release artifacts.
+
+It is possible to override these defaults, either permanently in your project
+or at build time. For example, to build a wheel with debug symbols, use:
+
+.. code-block:: console
+
+   $ python -m build -Csetup-args=-Ddebug=true
+
+And to override all debug and optimization settings permanently, add this to
+your ``pyproject.toml`` file:
+
+.. code-block:: toml
+
+   [tool.meson-python.args]
+   setup = ['-Ddebug=true', '-Doptimization=0', '-Db_ndebug=false']
+
+For more details on overriding build options at build time see the
+:ref:`reference-config-settings` page, and in ``pyproject.toml`` see the
+:ref:`reference-pyproject-settings` page.
--- a/docs/explanations/default-options.rst
+++ b/docs/explanations/default-options.rst
@@ -44,21 +44,6 @@ Let's go through each option and why they are used:
   because Meson defaults to values that are appropriate for development, while
   the main purpose of meson-python is to build release artifacts.
 
-It is possible to override these defaults, either permanently in your project
-or at build time. For example, to build a wheel with debug symbols, use:
-
-.. code-block:: console
-
-   $ python -m build -Csetup-args=-Ddebug=true
-
-And to override all debug and optimization settings permanently, add this to
-your ``pyproject.toml`` file:
-
-.. code-block:: toml
-
-   [tool.meson-python.args]
-   setup = ['-Ddebug=true', '-Doptimization=0', '-Db_ndebug=false']
-
-For more details on overriding build options at build time see the
-:ref:`reference-config-settings` page, and in ``pyproject.toml`` see the
-:ref:`reference-pyproject-settings` page.
+It is possible to override these defaults, either permanently in your
+``pyproject.toml`` or at build time via the build frontend CLI.
+See the :ref:`how-to-guides-meson-args` page for examples of both methods.
--- a/docs/reference/pyproject-settings.rst
+++ b/docs/reference/pyproject-settings.rst
@@ -31,3 +31,13 @@ This page lists the configuration settings supported by ``meson-python`` in the
 
    * - ``tool.meson-python.args.install``
      - Extra arguments to be passed to the ``meson install`` command.
+
+Usage example:
+
+.. code-block:: toml
+
+   [tool.meson-python.args]
+   setup = ['-Dwarning_level=2', '-Db_pie=true']
+   dist = ['--include-subprojects']
+   compile = ['-j4']
+   install = ['--quiet']
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -135,6 +135,7 @@ Contributors
    :hidden:
 
    explanations/design
+   explanations/default-options
    explanations/internal-dependencies
    explanations/editable-installs
 


docs/explanations/default-options.rst
```diff
@@ -0,0 +1,63 @@
...
+
+.. _explanations-default-options:
+
+*********************
+Default build options
+*********************
+
+Meson offers many `built-in options <https://mesonbuild.com/Builtin-options.html>`__,
+and in the vast majority of cases those have good defaults. There are a couple
+of cases however where ``meson-python`` either is forced or chooses to override
```
dnicolodi, line 13: Forced?

docs/explanations/default-options.rst
```diff
@@ -0,0 +1,63 @@
...
+
+Let's go through each option and why they are used:
+
+- meson-python uses a native file, named ``.mesonpy-native-file.ini`` in order
+  to point Meson at the correct ``python`` interpreter to use (the same one for
+  which ``meson-python`` was installed). This is necessary, because Meson may
+  otherwise look for the first Python interpreter on the PATH (usually the same
+  one, but not always the case).
+- The ``prefix`` and ``platlibdir``/``purelibdir`` options also point Meson at
+  that same interpreter and the environment in which it is installed.
```
dnicolodi, line 38: These are going away, see #279.

docs/reference/pyproject-settings.rst
```diff
@@ -31,3 +31,19 @@ This page lists the configuration settings supported by ``meson-python`` in the
...
+   setup = ['-Dwarning_level=2', '-Db_pie=true']
+   dist = ['--include-subprojects']
+   compile = ['-j4']
+   install = ['--quiet']
+
+.. note::
+
+   At the moment it is necessary to specify all four arguments (``dist``,
+   ``setup```, ``compile`` and ``install``). This limitation will go away in a
+   future version.
```
dnicolodi, line 49: This was fixed in #304. In general, I think it is a good idea to link to a bug report any time we say "in a future version" so that the user can check whether that means in practice.

docs/explanations/default-options.rst
```diff
@@ -0,0 +1,63 @@
...
+      debug            : false
+      optimization     : 2
+      prefix           : /home/username/mambaforge/envs/project-dev
+      python.platlibdir: /home/username/mambaforge/envs/project-dev/lib/python3.10/site-packages
+      python.purelibdir: /home/username/mambaforge/envs/project-dev/lib/python3.10/site-packages
+      b_ndebug         : if-release
+
+Let's go through each option and why they are used:
+
+- meson-python uses a native file, named ``.mesonpy-native-file.ini`` in order
```
dnicolodi, line 32: The native file will not named like this anymore in the next meson-python release. It will be in the build directory and names `meson-python-native-file.ini`. One thing that is not obvious is that multiple `--native-file` options can be passed and Meson will merge the contents. Therefore, if the user needs to pass any other configuration via a native file, they can add another `--native-file` argument, as long as they do not try to set the path for the `python` executable. I think it would be useful to mention this in the documentation.

docs/reference/pyproject-settings.rst
rgommers: Ah, great!

>  In general, I think it is a good idea to link to a bug report any time we say "in a future version" so that the user can check whether that means in practice.

Agreed. I looked for one but couldn't find it. I think it was discussed in other PR review.

docs/explanations/default-options.rst
rgommers: That's merged, but these options are still there. They're not specific to editable installs, so I'm not sure I understand what you meant here.

docs/explanations/default-options.rst
rgommers: I'll rephrase as "needs to" if that's better. Overrides for, e.g., `--prefix` are necessary.

docs/explanations/default-options.rst
rgommers: Agreed, thanks. Will update and add that info.

docs/reference/pyproject-settings.rst
rgommers: Removed this note, so resolving the comment.

docs/explanations/default-options.rst
rgommers: added now

docs/explanations/default-options.rst
FFY00: I think Daniele meant #340. I'll review that tomorrow, but we can either wait until that gets a resolution, or merge this and have that PR update it.

docs/explanations/default-options.rst
dnicolodi: Technically, they are not necessary. We could very well use the default prefix and use the `$DESTDIR` environment variable to install in a temporary directory. I would even argue that this would be better. Anyhow, it does not matter.

docs/explanations/default-options.rst
dnicolodi: Yes, sorry, I linked to the wrong PR.

docs/explanations/default-options.rst
FFY00: Yes, I think it was only needed for the Debian heuristics, I think we can drop it now that Meson uses directory placeholders for all paths.

docs/explanations/default-options.rst
FFY00: Because of #341 and #344, we cannot get rid of the these options at the moment, so they should stay for now.

docs/explanations/default-options.rst
```diff
@@ -0,0 +1,64 @@
...
+.. _explanations-default-options:
+
+*********************
+Default build options
+*********************
+
+Meson offers many `built-in options <https://mesonbuild.com/Builtin-options.html>`__,
+and in the vast majority of cases those have good defaults. There are a couple
+of cases however where ``meson-python`` either needs to or chooses to override
+those with its own defaults. To view what those are for the version of
```
FFY00, line 14: possible improvement:

We could split the paragraph here and also document why we might choose to override the defaults. I wrote about the reasoning in https://github.com/mesonbuild/meson-python/pull/325#issuecomment-1463902837.

We could also note that the options we do need to set are mostly for build/installation details that should not be relevant to the user. The only exception is the native file, but the user should be able to override the parts of it that make sense (IIRC everything but the Python interpreter).

docs/explanations/default-options.rst
```diff
@@ -0,0 +1,64 @@
...
+.. code-block:: text
+
+    User defined options
+      Native files     : /home/username/code/project/.mesonpy-native-file.ini
+      debug            : false
+      optimization     : 2
+      prefix           : /home/username/mambaforge/envs/project-dev
+      python.platlibdir: /home/username/mambaforge/envs/project-dev/lib/python3.10/site-packages
+      python.purelibdir: /home/username/mambaforge/envs/project-dev/lib/python3.10/site-packages
+      b_ndebug         : if-release
```
FFY00, line 28: possible improvement:

I'd group the options we need to set, with the ones we choose to override the default. Something like this:

```suggestion
      (we need override)
      prefix           : /home/username/mambaforge/envs/project-dev
      python.platlibdir: /home/username/mambaforge/envs/project-dev/lib/python3.10/site-packages
      python.purelibdir: /home/username/mambaforge/envs/project-dev/lib/python3.10/site-packages
      Native files     : /home/username/code/project/.mesonpy-native-file.ini
      (we chose to override)
      debug            : false
      optimization     : 2
      b_ndebug         : if-release
```

It might make sense to make the distinction, so that users know what they can and can't override.

docs/explanations/default-options.rst
```diff
@@ -0,0 +1,64 @@
...
+- The ``prefix`` and ``platlibdir``/``purelibdir`` options also point Meson at
+  that same interpreter and the environment in which it is installed.
+- The ``debug``, ``optimization`` and ``b_ndebug`` options are overridden,
+  because Meson defaults to values that are appropriate for development, while
+  the main purpose of meson-python is to build release artifacts.
+
+It is possible to override these defaults, either permanently in your project
+or at build time. For example, to build a wheel with debug symbols, use:
+
+.. code-block:: bash
```
FFY00, line 50:
```suggestion
.. code-block:: console
```

This makes it so that the $ is not selectable, excluding it from ctrl+a or the copy button from the sphinx extension.

docs/explanations/default-options.rst
```diff
@@ -0,0 +1,64 @@
...
+your ``pyproject.toml`` file:
+
+.. code-block:: toml
+
+   [tool.meson-python.args]
+   setup = ['-Ddebug=true', '-Doptimization=0', '-Db_ndebug=false']
+
+For more details on overriding build options at build time see
+:ref:`here <reference-config-settings>`, and in ``pyproject.toml`` see
+:ref:`here <reference-pyproject-settings>`.
```
FFY00, line 64: nitpick

```suggestion
For more details on overriding build options at build time see the
:ref:`reference-config-settings` page, and in ``pyproject.toml`` see
the :ref:`reference-pyproject-settings` page.
```

docs/reference/pyproject-settings.rst
```diff
@@ -31,3 +31,13 @@ This page lists the configuration settings supported by ``meson-python`` in the
 
    * - ``tool.meson-python.args.install``
      - Extra arguments to be passed to the ``meson install`` command.
+
+Usage example:
```
FFY00, line 35: possible improvement:

I think this may be better as sub-section.

```suggestion
Example
=====
```

FFY00 (approved):
@rgommers thanks! I have a one comment, and a couple of suggestions, but overall looks good.

docs/explanations/default-options.rst
dnicolodi: This is the output of the `meson setup` command. Presenting an edited version of it that does not match what the uses sees would be confusing.

docs/explanations/default-options.rst
dnicolodi: If we keep `meson install` I would much prefer if we would get rid of the `--prefix` and `--python.{plat,pure}lib` options and use the `$DESTDIR` environment variable pointing to somewhere in `/tmp` instead. This would have the additional benefit of putting the installation directory on tmpfs, with possible significant speedups when building large wheels.

docs/explanations/default-options.rst
rgommers: Agreed, I just Ctrl-C'd this from actual build output (replacing my own user name with `username` and package name with `project`) and would prefer to leave it as is.

docs/reference/pyproject-settings.rst
rgommers: This was on purpose - subsection heading look terrible in our Sphinx theme, and we should not over-use them.

docs/explanations/default-options.rst
rgommers: > We could split the paragraph here and also document why we might choose to override the defaults. I wrote about the reasoning in [#325 (comment)](https://github.com/mesonbuild/meson-python/pull/325#issuecomment-1463902837).

I checked it in more detail, and it's effectively the same as what I already have in this PR: _"because Meson defaults to values that are appropriate for development, while the main purpose of meson-python is to build release artifacts."_

docs/explanations/default-options.rst
rgommers: adopted, resolving

docs/explanations/default-options.rst
rgommers: thanks, adopted

docs/explanations/default-options.rst
rgommers: rejecting the suggested change - resolving.

dnicolodi:
It would be nice to have this combined somehow with the "Passing arguments to Meson" section. By the way, I just read that section and it contain several inaccuracies and some errors, for example all the pip command lines are wrong.

docs/explanations/default-options.rst
rgommers: The current text is good as is given the current state of the repo, I think we can resolve this comment.

rgommers:
> It would be nice to have this combined somehow with the "Passing arguments to Meson" section.

rgommers:
> It would be nice to have this combined somehow with the "Passing arguments to Meson" section.

Ah, that is new. I don't have more time to spend on this right now, maybe end of the week. It looks to me like only the end of the page overlaps. How about leaving the explanation here, and the two examples of overriding arguments can be merged into the other page (so everything after _"It is possible to override these defaults, either permanently in your project or at build time."_)?

docs/explanations/default-options.rst
FFY00: > If we keep `meson install` I would much prefer if we would get rid of the `--prefix` and `--python.{plat,pure}lib` options and use the `$DESTDIR` environment variable pointing to somewhere in `/tmp` instead.

I don't think this is even a question, is it :rofl:? Those options and `DESTDIR` serve two different purposes. We only set those options for the heuristics, so they can be removed now.

Regarding `DESTDIR`, we already set it to a subdirectory of `working_dir`, which currently when using the PEP 517 hooks is always set by `tempfile.TemporaryDirectory` (in `with_temp_working_dir`), so isn't that true already?

dnicolodi: The working dir is created inside the project directory, thus it is on the same filesystem as the project sources.

dnicolodi:
Seems appropriate. I just noticed that you placed this page in the "Explanations" section. Why? I think it would be better in the "Reference" section. Anyhow, I think we can merge it as it is and tweak it up later.

docs/explanations/default-options.rst
FFY00: Ah, you're right :sweat_smile:. IIRC I did that to make things work properly for packagers by default, where access to `/tmp` might be a bit tricky. We can consider changing the default, though.

docs/explanations/default-options.rst
```diff
@@ -0,0 +1,64 @@
...
+
+   $ python -m build -Csetup-args=-Ddebug=true
+
+And to override all debug and optimization settings permanently, add this to
+your ``pyproject.toml`` file:
+
+.. code-block:: toml
+
+   [tool.meson-python.args]
+   setup = ['-Ddebug=true', '-Doptimization=0', '-Db_ndebug=false']
```
FFY00, line 60: Since #291 is not merged, wouldn't it be better to link to that page?

https://meson-python.readthedocs.io/en/latest/how-to-guides/meson-args.html

FFY00:
I think the page as-is should be in the explanations section, as its main purpose is to explain the reason behind the options we set, but it might make sense to have a page in the reference section detailing the default options we override.

FFY00:
@rgommers feel free to apply any of the changes suggested, if you want, and then merge.

docs/explanations/default-options.rst
rgommers: Agreed, done.

docs/explanations/default-options.rst
rgommers: I'll leave this comment thread open, because it seems like useful info. It doesn't apply to this PR directly though; it's discussion of a potential future design change. If that materializes, the docs can simply be updated.

rgommers:
The pre-commit failure is unrelated to this PR, it complains about a missing `stacklevel`. I'll ignore here.

PR #339, opened 2023-03-11
Handle ARCHFLAGS set to the empty string

mkoeppe:
When `ARCHFLAGS` is set to the empty string, meson-python breaks; see https://github.com/sagemath/sage/pull/35082#issuecomment-1464890197

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -686,8 +686,8 @@ class Project():
 
         # setuptools-like ARCHFLAGS environment variable support
         if sysconfig.get_platform().startswith('macosx-'):
-            archflags = self._env.get('ARCHFLAGS')
-            if archflags is not None:
+            archflags = self._env.get('ARCHFLAGS', '').strip()
+            if archflags:
                 arch, *other = filter(None, (x.strip() for x in archflags.split('-arch')))
                 if other:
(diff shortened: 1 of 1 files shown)


dnicolodi:
We don't really validate the content of `$ARCHFLAGS` before trying to interpret it. There are probably many ways in which the environmental variable can be set to break meson-python. I'm not sure we should try too hard to handle setting the environmental variable to arbitrary values, especially considering that this is a legacy interface. I'm tempted to say that if setting `$ARCHFLAGS` to an empty string breaks meson-python, well, don't do that. I don't see any reason why having this environmental variable set to the empty string makes much sense. However, the fix is simple enough that we can just apply it.

FFY00:
I think the main argument for this is that in most shells you can't easily unset an environment variable for a command, so it's common practice to interpret empty variables the same way as the variable being unset.

FFY00 (approved):

mkoeppe:
> I don't see any reason why having this environmental variable set to the empty string makes much sense.

We are using this particular setting, for example, to work around some problems with Perl packages. https://github.com/sagemath/sage/blob/develop/build/pkgs/perl_term_readline_gnu/spkg-install.in#L10

PR #345, opened 2023-03-14
ENH: error on failure to map installed files into wheel directories

dnicolodi:
Turn the warning emitted so far into an hard error: warnings are easy to miss when building packages containing many files and the reason for missing files in the installed wheel can be hard to track down.

Remove the message suggesting that a missing installation path placeholder in a Meson installation path is a Meson bug: this can be caused by the user using absolute paths as installation directories, for example using `py.get_path()` instead of `py.get_install_dir()`.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -143,7 +143,7 @@ _SCHEME_MAP = {
 }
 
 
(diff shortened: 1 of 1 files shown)


FFY00 (approved):
Thanks @dnicolodi!

I noticed that the error did not subclass from our error type too, breaking our error handling, and was gonna submit a PR. You bet me to it :rofl:

PR #346, opened 2023-03-14
DOC: improve "Passing arguments to Meson" page

dnicolodi:
Fix command line examples for pip. Use realistic options in the first example. Extend the descriptions a bit. Fix reST markup.


dnicolodi:
I think it would make a lot of sense to merge the "Use build config settings" page with this one.

FFY00 (changes requested):
We are following https://diataxis.fr/, so IMO having one guide for config settings in general, and one for meson args makes sense.

The guides are also meant to be small and very task-oriented, only containing the most important information needed to achieve the goal, so I think your rewrite of the meson args guide is better as a tutorial instead.

See the [introductory tutorial](https://meson-python.readthedocs.io/en/latest/tutorials/introduction.html#tutorial-introduction) and [first project guide](https://meson-python.readthedocs.io/en/latest/how-to-guides/first-project.html#how-to-guides-first-project) for an example of the differences in scope.

dnicolodi:
Having the two pages separates is just repetitive or it requires to artificially split the topic. I think following diataxis for the three pages of documentation we have is not the most effective way to organize the material. It makes sense for larger documentation bases, but having more than one obvious place where to look for a piece of information in the table of contents is not a good user experience.

dnicolodi:
Anyhow, if you don't like it, don't accept the patch. I don't need this documentation and I'm not going to invest more time on this. Feel free to pick the parts you like.

Issue #344, opened 2023-03-14
We are not copying the files from the Meson install directory


FFY00:
While reviewing #340, I found that we are not actually using the Meson install directory to copy files to the wheel, and instead we are copying them from the build directory.

This has two problems I can identify:
- The files get installed with a wrong `RPATH`
  - This is because Meson sets the `RPATH` so that you can run stuff from the build directory
- Exclusion parameters to [`install_subdir`](https://mesonbuild.com/Reference-manual_functions.html#install_subdir) are ignored
  - This is because Meson does not export them in the metadata, so we just copy the whole directory, but we can only rely on copying from the install location, as that's the only place which will have the correct tree installed

Reproducible with all examples from the `install_subdir` docs:

```meson
# SPDX-FileCopyrightText: 2023 The meson-python developers
#
# SPDX-License-Identifier: MIT

project('install-subdir-exclude', version: '1.0.0')

py = import('python').find_installation()
purelib = py.get_path('purelib')

install_subdir(
  'foo',
  install_dir: purelib / 'subdir1',
  strip_directory : false,
)
install_subdir(
  'foo',
  install_dir: purelib / 'subdir2',
  strip_directory : false,
)
install_subdir(
  'foo/bar',
  install_dir: purelib / 'subdir3',
  strip_directory : false,
)
install_subdir(
  'foo/bar',
  install_dir: purelib / 'subdir4',
  strip_directory : true,
)
install_subdir(
  'new_directory',
  install_dir: purelib,
)
```

dnicolodi:
I added the information required to handle `exclude_files` and `exclude_directories` without having to look at the installed tree to the Meson introspection data https://github.com/mesonbuild/meson/pull/11432

dnicolodi:
> ```meson
> py = import('python').find_installation()
> purelib = py.get_path('purelib')
> ```

This is wrong. It should be `py.get_install_dir()`

rgommers:
> The files get installed with a wrong `RPATH`

For this part it'd be great if we had a test case, that's a lot easier to talk about then discussing what may happen in the absence of it. Would you be able to construct one @FFY00?

> I added the information required to handle `exclude_files` and `exclude_directories` without having to look at the installed tree to the Meson introspection data [mesonbuild/meson#11432](https://github.com/mesonbuild/meson/pull/11432)

[...]

Issue #317, opened 2023-02-20
BUG: install_subdir() in meson.build does not honor exlude_files/directories


dalcinl:
I'm trying to use `install_subdir()` to simplify the inclusion of Python source files and other data files in the final wheel.  However, the `install_subdir()` function is not honoring the  `exlude_files/directories` options. 

Core Meson is installing things properly and exclusions are honored. However, when `mesonpy` takes over, it ends up copying the subtree from the source directory, and not the install directory as I was expecting. That's the reason the exclusions are not honored: things are being copied from the wrong location. As Meson is all about out-of-tree builds, I would argue this is a bug.

The following trivial patch fixes the problem:
```diff
diff --git a/mesonpy/__init__.py b/mesonpy/__init__.py
index 40b9645..fe5055d 100644
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -486,7 +486,7 @@ class _WheelBuilder():
                 )
                 if install_details:
                     scheme, destination = install_details
-                    wheel_files[scheme].append((destination, file))
+                    wheel_files[scheme].append((destination, copy_files[file]))
                     continue
                 # not found
                 warnings.warn(
```
However, this patch makes some tests in `tests/test_wheel.py` break.
I'm not familiar enough with Meson and meson-python to triage these test failures. Is anyone able to help?

dnicolodi:
I think this is a bug in Meson: we should be able to map from source code and build artifacts to installation paths via the `intro-install_plan.json` introspection data only but the introspection data contains the wrong source path.

FFY00:
Considering that, I'd say it's a meson-python bug, Meson just does not provide a way for us to fix it.

dnicolodi:
Yes, indeed.

dalcinl:
@dnicolodi Sorry if this has been discussed elsewhere, but could you please elaborate in the rationale for avoiding the install step? Is is only about supporting editable installs? Messon favors and out-of-source approach, and if you try to workaround that, at some point some logics will break. This is exactly what happened when I tried to `install_subdir()` with exclude options :disappointed:. Editable installs could be easily supported by copying over built extensions from the install directory back to the source tree, pretty much as setuptools does it.

PS: To be fully open and fair, I'm using `install_subdir()` to workaround the fact that Meson/meson-python does not allow (out of the box) use of wildcards  to include py sources and data files. I would love seeing meson-python taking a more pragmatic approach (assuming core Meson would allow for it).

[...]

PR #348, opened 2023-03-17
BUG: copy wheel files from the Meson install directory

FFY00:
The first commit is the interesting one, the next two are related to installing directories, which can be fixed now that we are copying the files from the installation directory, and the last commit is a refactoring of `_install_path` that extracts the RPATH changing logic into an external function.

I can split the PR if required. I just did all here because the changes depend on the previous ones, and I thought that they were simple enough to not complicate the review much.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -418,7 +418,7 @@ class _WheelBuilder():
         self,
         wheel_file: mesonpy._wheelfile.WheelFile,
         counter: mesonpy._util.CLICounter,
-        origin: Path,
+        path_in_builddir: Path,
         destination: pathlib.Path,
     ) -> None:
         """"Install" file or directory into the wheel
@@ -427,6 +427,8 @@ class _WheelBuilder():
         Some files might need to be fixed up to set the RPATH to the internal
         library directory on Linux wheels for eg.
         """
+        builddir_to_installdir_mapping = self._project._copy_files
+        origin = pathlib.Path(builddir_to_installdir_mapping[os.fspath(path_in_builddir)])
         location = destination.as_posix()
         counter.update(location)
 
@@ -447,7 +449,7 @@ class _WheelBuilder():
                     # add .mesonpy.libs to the RPATH of ELF files
                     if self._is_native(os.fspath(origin)):
                         # copy ELF to our working directory to avoid Meson having to regenerate the file
-                        new_origin = self._libs_build_dir / pathlib.Path(origin).relative_to(self._build_dir)
+                        new_origin = self._libs_build_dir / destination
                         os.makedirs(new_origin.parent, exist_ok=True)
                         shutil.copy2(origin, new_origin)
                         origin = new_origin
@@ -860,6 +862,18 @@ class Project():
                         del files[file]
 
         return install_plan
+
+
+    @property
+    def _copy_files(self) -> Dict[str, str]:
+        """Files that Meson will copy on install and the target location."""
+        copy_files = {}
+        for origin, destination in self._info('intro-installed').items():
+            destination_path = pathlib.Path(destination).absolute()
+            copy_files[origin] = os.fspath(
+                self._install_dir / destination_path.relative_to(destination_path.anchor)
+            )
+        return copy_files
 
     @property
     def _meson_name(self) -> str:
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -414,6 +414,44 @@ class _WheelBuilder():
             return True
         return False
 
+    def _add_mesonpy_libs_to_rpath(self, path: pathlib.Path, destination: pathlib.Path) -> pathlib.Path:
+        """Adds .mesonpy.libs to the RPATH """
+        if platform.system() not in ('Linux', 'Darwin'):
+            # Internal libraries are currently unsupported on this platform
+            raise NotImplementedError(
+                "Bundling libraries in wheel is not supported "
+                f"on platform '{platform.system()}'"
+            )
+
+        if not self._is_native(path):
+            return path
+
+        # copy ELF to our working directory to avoid Meson having to regenerate the file
+        new_path = self._libs_build_dir / destination
+        os.makedirs(new_path.parent, exist_ok=True)
+        shutil.copy2(path, new_path)
+        path = new_path
+
+        mesonpy_libs_path = os.path.relpath(
+            f".{self._project.name}.mesonpy.libs",
+            destination.parent,
+        )
+
+        binary: Union[None, mesonpy._elf.ELF, mesonpy._dylib.Dylib] = None
+        libdir_path: Optional[str] = None
+        if platform.system() == 'Linux':
+            binary = mesonpy._elf.ELF(path)
+            libdir_path = f'$ORIGIN/{mesonpy_libs_path}'
+        elif platform.system() == 'Darwin':
+            binary = mesonpy._dylib.Dylib(path)
+            libdir_path = f'@loader_path/{mesonpy_libs_path}'
+
+        assert binary and libdir_path
+        if libdir_path not in binary.rpath:
+            binary.rpath = [*binary.rpath, libdir_path]
+
+        return path
+
     def _install_path(
         self,
         wheel_file: mesonpy._wheelfile.WheelFile,
@@ -438,34 +476,8 @@ class _WheelBuilder():
                     arcname = destination / child.relative_to(origin)
                     wheel_file.write(child, arcname.as_posix())
         else:
-            # fix file
             if self._has_internal_libs:
-                if platform.system() == 'Linux' or platform.system() == 'Darwin':
-                    # add .mesonpy.libs to the RPATH of ELF files
-                    if self._is_native(os.fspath(origin)):
-                        # copy ELF to our working directory to avoid Meson having to regenerate the file
-                        new_origin = self._libs_build_dir / destination
-                        os.makedirs(new_origin.parent, exist_ok=True)
-                        shutil.copy2(origin, new_origin)
-                        origin = new_origin
-                        # add our in-wheel libs folder to the RPATH
-                        if platform.system() == 'Linux':
-                            elf = mesonpy._elf.ELF(origin)
-                            libdir_path = \
-                                f'$ORIGIN/{os.path.relpath(f".{self._project.name}.mesonpy.libs", destination.parent)}'
-                            if libdir_path not in elf.rpath:
-                                elf.rpath = [*elf.rpath, libdir_path]
-                        elif platform.system() == 'Darwin':
-                            dylib = mesonpy._dylib.Dylib(origin)
-                            libdir_path = \
-                                f'@loader_path/{os.path.relpath(f".{self._project.name}.mesonpy.libs", destination.parent)}'
-                            if libdir_path not in dylib.rpath:
-                                dylib.rpath = [*dylib.rpath, libdir_path]
-                        else:
-                            # Internal libraries are currently unsupported on this platform
-                            raise NotImplementedError("Bundling libraries in wheel is not supported on platform '{}'"
-                                                      .format(platform.system()))
-
+                origin = self._add_mesonpy_libs_to_rpath(origin, destination)
             wheel_file.write(origin, location)
 
     def _wheel_write_metadata(self, whl: mesonpy._wheelfile.WheelFile) -> None:
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -887,7 +887,7 @@ class Project():
 
     @property
     def _copy_files(self) -> Dict[str, str]:
-        """Files that Meson will copy on install and the target location."""
+        """Files that Meson will copy on install and their target locations."""
         copy_files = {}
         for origin, destination in self._info('intro-installed').items():
             destination_path = pathlib.Path(destination).absolute()
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -419,7 +419,7 @@ class _WheelBuilder():
         if platform.system() not in ('Linux', 'Darwin'):
             # Internal libraries are currently unsupported on this platform
             raise NotImplementedError(
-                "Bundling libraries in wheel is not supported "
+                'Bundling libraries in wheel is not supported '
                 f"on platform '{platform.system()}'"
             )
 
@@ -433,7 +433,7 @@ class _WheelBuilder():
         path = new_path
 
         mesonpy_libs_path = os.path.relpath(
-            f".{self._project.name}.mesonpy.libs",
+            f'.{self._project.name}.mesonpy.libs',
             destination.parent,
         )
 
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -446,9 +446,23 @@ class _WheelBuilder():
             binary = mesonpy._dylib.Dylib(path)
             libdir_path = f'@loader_path/{mesonpy_libs_path}'
 
+        # add our .mesonpy.libs directory to RPATH
         assert binary and libdir_path
         if libdir_path not in binary.rpath:
             binary.rpath = [*binary.rpath, libdir_path]
+
+        # fix LC_LOAD_DYLIB entries on macOS to use the RPATH
+        if platform.system() == 'Darwin':
+            libs = {
+                destination.name: destination
+                for destination, _ in self._wheel_files['mesonpy-libs']
+            }
+
+            assert isinstance(binary, mesonpy._dylib.Dylib)
+            for entry in binary.load_dylib:
+                libname = entry.split('/')[-1]
+                if libname in libs:
+                    binary.replace_load_dylib(entry, f'@rpath/{libs[libname].as_posix()}')
 
         return path
 
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -432,18 +432,13 @@ class _WheelBuilder():
         location = destination.as_posix()
         counter.update(location)
 
-        # fix file
-        if os.path.isdir(origin):
-            for root, dirnames, filenames in os.walk(str(origin)):
-                # Sort the directory names so that `os.walk` will walk them in a
-                # defined order on the next iteration.
-                dirnames.sort()
-                for name in sorted(filenames):
-                    path = os.path.normpath(os.path.join(root, name))
-                    if os.path.isfile(path):
-                        arcname = os.path.join(destination, os.path.relpath(path, origin).replace(os.path.sep, '/'))
-                        wheel_file.write(path, arcname)
+        if origin.is_dir():
+            for child in origin.rglob('**/*'):
+                if child.is_file():
+                    arcname = destination / child.relative_to(origin)
+                    wheel_file.write(child, arcname.as_posix())
         else:
+            # fix file
             if self._has_internal_libs:
                 if platform.system() == 'Linux' or platform.system() == 'Darwin':
                     # add .mesonpy.libs to the RPATH of ELF files
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -426,12 +426,6 @@ class _WheelBuilder():
         if not self._is_native(path):
             return path
 
-        # copy ELF to our working directory to avoid Meson having to regenerate the file
-        new_path = self._libs_build_dir / destination
-        os.makedirs(new_path.parent, exist_ok=True)
-        shutil.copy2(path, new_path)
-        path = new_path
-
         mesonpy_libs_path = os.path.relpath(
             f'.{self._project.name}.mesonpy.libs',
             destination.parent,
--- a/mesonpy/_dylib.py
+++ b/mesonpy/_dylib.py
@@ -9,7 +9,7 @@ import typing
 
 
 if typing.TYPE_CHECKING:
-    from typing import Optional
+    from typing import List, Optional
 
     from mesonpy._compat import Collection, Path
 
@@ -19,6 +19,7 @@ class Dylib:
     def __init__(self, path: Path) -> None:
         self._path = os.fspath(path)
         self._rpath: Optional[Collection[str]] = None
+        self._load_dylib: Optional[Collection[str]] = None
         self._needed: Optional[Collection[str]] = None
 
     def _otool(self, *args: str) -> str:
@@ -27,20 +28,28 @@ class Dylib:
     def _install_name_tool(self, *args: str) -> str:
         return subprocess.check_output(['install_name_tool', *args, self._path], stderr=subprocess.STDOUT).decode()
 
+    def _find_tag(self, name: str, field: str) -> List[str]:
+        entries = []
+        # Run otool -l to get the load commands
+        otool_output = self._otool('-l').strip()
+        # Manually parse the output and find the tag entries
+        looking_at_tag = False
+        for line in map(str.split, otool_output.splitlines()):
+            if line[0] == 'cmd':
+                if len(line) < 2:
+                    raise ValueError('Unexpected "otool -l" output')
+                looking_at_tag = line[1] == name
+            elif looking_at_tag and line[0] == field:
+                if len(line) < 2:
+                    raise ValueError('Unexpected "otool -l" output')
+                entries.append(line[1])
+                looking_at_tag = False
+        return entries
+
     @property
     def rpath(self) -> Collection[str]:
         if self._rpath is None:
-            self._rpath = []
-            # Run otool -l to get the load commands
-            otool_output = self._otool('-l').strip()
-            # Manually parse the output for LC_RPATH
-            rpath_tag = False
-            for line in [x.split() for x in otool_output.split('\n')]:
-                if line == ['cmd', 'LC_RPATH']:
-                    rpath_tag = True
-                elif len(line) >= 2 and line[0] == 'path' and rpath_tag:
-                    self._rpath += [line[1]]
-                    rpath_tag = False
+            self._rpath = self._find_tag('LC_RPATH', 'path')
         return frozenset(self._rpath)
 
     @rpath.setter
@@ -53,3 +62,17 @@ class Dylib:
         for rpath in value:
             self._install_name_tool('-add_rpath', rpath)
         self._rpath = value
+
+    @property
+    def load_dylib(self) -> Collection[str]:
+        if self._load_dylib is None:
+            self._load_dylib = self._find_tag('LC_LOAD_DYLIB', 'name')
+        return frozenset(self._load_dylib)
+
+    def replace_load_dylib(self, old: str, new: str) -> None:
+        if old not in self.load_dylib:
+            raise KeyError(f'LC_LOAD_DYLIB entry "{old}" not found')
+        self._install_name_tool('-change', old, new)
+        assert isinstance(self._load_dylib, list)
+        self._load_dylib.remove(old)
+        self._load_dylib.append(new)
--- a/tests/packages/install-subdir-exclude/meson.build
+++ b/tests/packages/install-subdir-exclude/meson.build
@@ -41,5 +41,5 @@ install_subdir(
   'exclude',
   install_dir: purelib,
   exclude_directories: ['subdir2', 'subdir3'],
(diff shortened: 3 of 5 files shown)


tests/packages/install-subdir-exclude/meson.build
```diff
@@ -0,0 +1,37 @@
+# SPDX-FileCopyrightText: 2023 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+project('install-subdir-exclude', version: '1.0.0')
+
+py = import('python').find_installation()
+purelib = py.get_install_dir(pure: true)
+
```
rgommers, line 9: Why doesn't this test include the use of `exclude_directories` and `exclude_files`? That should be the main point, right? More interesting than the niche bug reported in meson-11539.

rgommers (commented):
Thanks @FFY00. `test_local_lib` shows a few failures, so that'll need a tweak. My initial comment is about testing: right now this doesn't exercise either the bug in gh-317 or the one in gh-344 about RPATH handling. I made an inline comment about using `install_subdir(..., exclude_*)` keywords, and would like to repeat my suggestion from https://github.com/mesonbuild/meson-python/issues/344#issuecomment-1467836049 about adding a test that would actually fail without the bug-fix change in this PR.

FFY00:
@rgommers, thanks. I should have opened this as a draft PR. I mostly wanted to trigger the CI to see if anything was failing, instead of having to spin up all my local machines :sweat_smile:

I'll address the comments and mark as ready to review later.

FFY00:
I updated `test_rpath` to check all the RPATH values, it now fails on main, but passes with this PR.

```diff
$ git diff
diff --git a/tests/test_wheel.py b/tests/test_wheel.py
index fc91faf..f41236a 100644
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -191,12 +191,15 @@ def test_rpath(wheel_link_against_local_lib, tmp_path):
     artifact = wheel.wheelfile.WheelFile(wheel_link_against_local_lib)
     artifact.extractall(tmp_path)

+    module = tmp_path / f'example{EXT_SUFFIX}'
     if platform.system() == 'Linux':
-        elf = mesonpy._elf.ELF(tmp_path / f'example{EXT_SUFFIX}')
-        assert '$ORIGIN/.link_against_local_lib.mesonpy.libs' in elf.rpath
-    else:  # 'Darwin'
-        dylib = mesonpy._dylib.Dylib(tmp_path / f'example{EXT_SUFFIX}')
-        assert '@loader_path/.link_against_local_lib.mesonpy.libs' in dylib.rpath
+        binary = mesonpy._elf.ELF(module)
+        libdir_origin = '$ORIGIN'
+    elif platform.system() == 'Darwin':
+        binary = mesonpy._dylib.Dylib(module)
+        libdir_origin = '@loader_path'
+
+    assert binary.rpath == {f'{libdir_origin}/.link_against_local_lib.mesonpy.libs'}


 @pytest.mark.skipif(platform.system() not in ['Linux', 'Darwin'], reason='Unsupported on this platform for now')
```

```
E       AssertionError: assert frozenset({'$ORIGIN/.link_against_local_lib.mesonpy.libs', '$ORIGIN/'}) == {'$ORIGIN/.link_against_local_lib.mesonpy.libs'}
E         Extra items in the left set:
E         '$ORIGIN/'
E         Full diff:
E         - {'$ORIGIN/.link_against_local_lib.mesonpy.libs'}
E         + frozenset({'$ORIGIN/.link_against_local_lib.mesonpy.libs', '$ORIGIN/'})
E         ? ++++++++++                                               ++++++++++++ +
```

FFY00:
macOS is failing and I was able to confirm it was due to the commit that copies the files from the install directory.

After some debugging, I suspect that our RPATH handling code on macOS is broken. I don't think `install_name_tool -add_rpath` is enough.

FFY00:
Here's the difference before (good) and after (bad):

```diff
--- test-rpath/bad/otool-l.txt	2023-03-17 10:10:26
+++ test-rpath/good/otool-l.txt	2023-03-17 10:10:19
@@ -176,8 +176,8 @@
   version 0.0
 Load command 11
           cmd LC_LOAD_DYLIB
-      cmdsize 120
-         name /opt/homebrew/opt/python@3.11/Frameworks/Python.framework/Versions/3.11/lib/libexample.dylib (offset 24)
+      cmdsize 48
+         name @rpath/libexample.dylib (offset 24)
    time stamp 2 Thu Jan  1 01:00:02 1970
       current version 0.0.0
 compatibility version 0.0.0
@@ -197,6 +197,10 @@
   dataoff 49648
  datasize 560
 Load command 15
+          cmd LC_RPATH
+      cmdsize 32
+         path @loader_path/ (offset 12)
+Load command 16
           cmd LC_RPATH
       cmdsize 64
          path @loader_path/.link_against_local_lib.mesonpy.libs (offset 12)
````

With the working one:

```console
$ test-rpath/venv/bin/python -c 'import example; print(example.example_sum(1, 2))'
3
$ install_name_tool -delete_rpath @loader_path/ test-rpath/venv/lib/python3.11/site-packages/example.cpython-311-darwin.so
$ test-rpath/venv/bin/python -c 'import example; print(example.example_sum(1, 2))'
3
```

So it still works after removing `@loader_path/` from RPATH. Leaving our otool diff to:

```diff
--- test-rpath/bad/otool-l.txt	2023-03-17 10:10:26
+++ test-rpath/artificial/otool-l.txt	2023-03-17 10:18:08
@@ -176,8 +176,8 @@
   version 0.0
 Load command 11
           cmd LC_LOAD_DYLIB
-      cmdsize 120
-         name /opt/homebrew/opt/python@3.11/Frameworks/Python.framework/Versions/3.11/lib/libexample.dylib (offset 24)
+      cmdsize 48
+         name @rpath/libexample.dylib (offset 24)
    time stamp 2 Thu Jan  1 01:00:02 1970
       current version 0.0.0
 compatibility version 0.0.0
```

Meaning we really just need to replace the library install path with `@rpath` in the `LC_LOAD_DYLIB` entry.

FFY00:
The macOS CI is happy now, but Windows is still broken.

```
FAILED tests/test_wheel.py::test_install_subdir_exclude - AssertionError: assert {'subdir2/bar/file1', 'install_subdir_exclude-1.0.0.dist-info/RECORD', 'install_subdir_exclude-1.0.0.dist-info/WHEEL', 'subdir1/foo/file2', 'subdir4/file1', 'exclude/subdir1/file1', 'exclude/subdir1/file2', 'install_subdir_exclude-1.0.0.dist-info/METADATA', 'exclude/file1', 'exclude/subdir1/file3', 'subdir3/bar/file1', 'subdir1/foo/bar/file1', 'subdir2/file2'} == {'subdir2/bar/file1', 'install_subdir_exclude-1.0.0.dist-info/RECORD', 'install_subdir_exclude-1.0.0.dist-info/WHEEL', 'subdir1/foo/file2', 'subdir4/file1', 'exclude/subdir1/file1', 'install_subdir_exclude-1.0.0.dist-info/METADATA', 'exclude/file1', 'subdir3/bar/file1', 'subdir1/foo/bar/file1', 'subdir2/file2'}
  Extra items in the left set:
  'exclude/subdir1/file3'
  'exclude/subdir1/file2'
  Full diff:
    {
     'exclude/file1',
     'exclude/subdir1/file1',
  +  'exclude/subdir1/file2',
  +  'exclude/subdir1/file3',
     'install_subdir_exclude-1.0.0.dist-info/METADATA',
     'install_subdir_exclude-1.0.0.dist-info/RECORD',
     'install_subdir_exclude-1.0.0.dist-info/WHEEL',
     'subdir1/foo/bar/file1',
     'subdir1/foo/file2',
     'subdir2/bar/file1',
     'subdir2/file2',
     'subdir3/bar/file1',
     'subdir4/file1',
    }
```

I suspect Meson simply isn't doing the path separator translation in the `exclude_files` field.

```meson
...
install_subdir(
  'exclude',
  install_dir: purelib,
  exclude_directories: ['subdir2', 'subdir3'],
  exclude_files: ['file2', 'file3', 'subdir1/file2', 'subdir1/file3'],
)
```

@eli-schwartz is this expected?

FFY00:
@rgommers this is now ready for review. There's just the Windows failure to deal with, but that's fairly simple, I just want to get some confirmation from the Meson maintainers, and then we can just adjust the test to work around the issue.

eli-schwartz:
> @eli-schwartz is this expected?

Looking at the debug commit, this isn't producing any values at all, anywhere. Looking at the point release, this change wasn't released in 1.0.1...

FFY00:
> Looking at the debug commit, this isn't producing any values at all, anywhere. Looking at the point release, this change wasn't released in 1.0.1...

Just to be clear, you mean the metadata, right? Anyway, it's super weird the behavior is different on Linux+macOS and Windows.

I am gonna mark this test as XFAIL on Windows to unblock this PR and open a new issue on the bug tracker to track this issue.

tests/test_wheel.py
```diff
@@ -191,12 +191,15 @@ def test_rpath(wheel_link_against_local_lib, tmp_path):
...
-    else:  # 'Darwin'
-        dylib = mesonpy._dylib.Dylib(tmp_path / f'example{EXT_SUFFIX}')
-        assert '@loader_path/.link_against_local_lib.mesonpy.libs' in dylib.rpath
+        binary = mesonpy._elf.ELF(module)
+        libdir_origin = '$ORIGIN'
+    elif platform.system() == 'Darwin':
+        binary = mesonpy._dylib.Dylib(module)
+        libdir_origin = '@loader_path'
+
+    assert binary.rpath == {f'{libdir_origin}/.link_against_local_lib.mesonpy.libs'}
```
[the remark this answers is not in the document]
rgommers: Testing on macOS shows a difference, there the test does pass within a conda env and there's only one RPATH entry with this PR:
```
(Pdb) binary.rpath
frozenset({'@loader_path/.link_against_local_lib.mesonpy.libs'})
```
while on `main` I get:
```
(Pdb) dylib.rpath
frozenset({'/Users/rgommers/mambaforge/envs/scipy-dev/lib', '@loader_path/.link_against_local_lib.mesonpy.libs', '@loader_path/'})
```
The output on `main` is correct here. I did a bit of digging, but it wasn't immediately clear what is happening in `_dylib.py` and the macOS-specific RPATH postprocessing that's not done for Linux.

rgommers (changes requested):
I looked at the RPATH related changes first and found two issues, one with the test change which is too strict, and one with the macOS change which lost an RPATH entry.

FFY00:
> It would be very useful for review to not mix unneeded code moves like `MAINT: refactor WheelBuilder._install_path ` with bug fixes and test changes in the future @FFY00. It makes the diff way larger than it has to be, and checking commit-by-commit is a pain.

Gotcha. I guess this depends on everyone's review approach :sweat_smile:. I'll add `XXX` comments just to note the possible improvements instead, to avoid mixing changes :+1:

tests/test_wheel.py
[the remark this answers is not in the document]
FFY00: > the path to the conda-forge env's `lib/` should be there

Why? This is a wheel, it's meant to be distributed, it shouldn't be leaking your local environment paths.

> . I did a bit of digging, but it wasn't immediately clear what is happening in `_dylib.py` and the macOS-specific RPATH postprocessing that's not done for Linux.

I assume you meant changing `LC_LOAD_DYLIB`, right? See https://github.com/mesonbuild/meson-python/pull/348#issuecomment-1473618719.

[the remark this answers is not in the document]
rgommers: To be more concrete, if I have a conda env and I type `pip install .`, I'm going through a wheel and I expect things to work without manually messing with `LD_LIBRARY_PATH`.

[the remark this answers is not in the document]
FFY00: > This is conceptually wrong, wheels have multiple purposes. See the paths at [pypackaging-native.github.io/meta-topics/build_steps_conceptual/#building-and-installing-or-uploading-artifacts](https://pypackaging-native.github.io/meta-topics/build_steps_conceptual/#building-and-installing-or-uploading-artifacts).

Fair enough.

> That is not correct, the RPATH entry is supposed to be there, and only when you're distributing you must use `auditwheel` to strip it. I'm sure we went over this in other issues.

I don't think we discussed stripping the RPATH specifically.

Is there any particular reason you think the RPATH should be set here?

> To be more concrete, if I have a conda env and I type `pip install .`, I'm going through a wheel and I expect things to work without manually messing with `LD_LIBRARY_PATH`.

Won't lots of other things break in that case? I was under the impression that you needed to enable a conda environment to use it, and setting up the library search path is one of the things enabling an environment is supposed to do, lots of things rely on that. Is this not the case?

Setting the RPATH to a user directory this is problematic for reproducible builds, so I'd like to avoid doing that if possible.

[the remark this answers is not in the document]
rgommers: No, conda envs are relocatable via RPATH rewriting. Spack envs too I believe. See, e.g., https://github.com/mesonbuild/meson-python/issues/125#issuecomment-1249241305 for details. That shows an example of how conda edits `LDFLAGS` on environment activation to insert `-Wl,-rpath,${CONDA_PREFIX}/lib`. That's why they are there in my binaries, and they should not be stripped by `meson-python`.

> Setting the RPATH to a user directory is problematic for reproducible builds, so I'd like to avoid doing that if possible.

This is not problematic. Reproducible builds are reproducible _for a given environment_, changing things like rpaths is completely normal. The rules here are simple:
1. don't change any RPATH entries in meson-python, except for inserting a `.mesonpy.libs` one in case we have an internal library
2. anyone wanting to distribute wheels must actually do the work to make them distributable by building in the right environment (e.g. a Docker manylinux container) and running a tool like `auditwheel` as appropriate.

You clearly cannot expect to take any random build env and then have `meson-python` spit out a wheel that is okay to upload to PyPI. 

This doesn't only apply to conda by the way, it will apply to any prefix dev env that's outside a standard loader search path (IIRC Nix also does something like this).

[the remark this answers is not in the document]
FFY00: > This is not problematic. Reproducible builds are reproducible _for a given environment_, changing things like rpaths is completely normal.

It isn't that simple :sweat_smile:

While it is possible to achieve reproducible builds when setting the RPATH, tools to validate reproducibility need to add support for arbitrary build locations to be able to handle these cases. That's why I said it is *problematic*, not that it makes reproducible builds impossible. https://github.com/kushaldas/asaman for eg. does not have support for this.

But you pointed out conda does this via `LDFLAGS`, which makes much more sense to me, and I agree we should be keeping that.

> don't change any RPATH entries in meson-python, except for inserting a `.mesonpy.libs` one in case we have an internal library

I slightly disagree. IMO we should be able to change RPATHs set by Meson directly if it doesn't match the behavior we want, but we should definitely keep anything that comes out of user configurations, like `LDFLAGS`.

> No, conda envs are relocatable via RPATH rewriting. Spack envs too I believe. See, e.g., [#125 (comment)](https://github.com/mesonbuild/meson-python/issues/125#issuecomment-1249241305) for details. That shows an example of how conda edits `LDFLAGS` on environment activation to insert `-Wl,-rpath,${CONDA_PREFIX}/lib`. That's why they are there in my binaries, and they should not be stripped by `meson-python`.

Thank you, this was very helpful to me :+1: 

> You clearly cannot expect to take any random build env and then have `meson-python` spit out a wheel that is okay to upload to PyPI.

I don't think this is an unreasonable expectation, but it definitely depends on the project. We should have more documentation regarding this to help users.

> This doesn't only apply to conda by the way, it will apply to any prefix dev env that's outside a standard loader search path (IIRC Nix also does something like this).

Yes.

IIRC Nix sets `LD_LIBRARY_PATH` on the fly to build very specific environments, but it's possible that they're using RPATH for something.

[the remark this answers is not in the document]
rgommers: This has now tripped us up several times. I think we should add a separate test for this. Maybe a good way of doing that is using the `install_rpath` argument to [executable](https://mesonbuild.com/Reference-manual_functions.html#executable). It's documented as: _A string to set the target's rpath to after install (but not before that). On Windows, this argument has no effect._. So if we set it to some unique string and check that that string gets preserved, that's probably a fairly robust test.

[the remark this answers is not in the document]
FFY00: Agreed, I'll update the PR tomorrow to adjust for the feedback and add tests for this behavior.

> Maybe a good way of doing that is using the `install_rpath` argument to [executable](https://mesonbuild.com/Reference-manual_functions.html#executable)

I think we should test both that, and setting the RPATH by passing linker arguments to Meson.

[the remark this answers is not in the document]
rgommers: > I slightly disagree. IMO we should be able to change RPATHs set by Meson directly if it doesn't match the behavior we want,

In general such things are Meson-internal implementation details that we don't want to have knowledge of within `meson-python`, so in practice it comes down to the same thing I'd say.

> I don't think this is an unreasonable expectation, but it definitely depends on the project. We should have more documentation regarding this to help users.

I can have a go at this. I'd like to keep it short and lean heavily on what I already wrote for pypackaging-native for concept, and on other tutorials and docs like those of `cibuildwheel` for how to actually go about releasing wheels.

eli-schwartz:
Maybe it's possible to just land those separately then? :)

FFY00:
They conflict with current PR, which is why I did not send them as separate PRs in the first place.

mesonpy/__init__.py
```diff
@@ -414,6 +414,44 @@ def _is_native(self, file: Union[str, pathlib.Path]) -> bool:
...
+            # Internal libraries are currently unsupported on this platform
+            raise NotImplementedError(
+                "Bundling libraries in wheel is not supported "
+                f"on platform '{platform.system()}'"
+            )
+
+        if not self._is_native(path):
+            return path
+
+        # copy ELF to our working directory to avoid Meson having to regenerate the file
```
dnicolodi, line 429: We are operating in the install directory, not the in the build directory, therefore if the file is modified here Meson does not need to regenerate it. The installation directory, is always a temporary directory, thus the file is copied from the build directory anyway at each execution. Coping it to another temporary location is not necessary.

dnicolodi (changes requested):

mesonpy/__init__.py
```diff
@@ -414,11 +414,57 @@ def _is_native(self, file: Union[str, pathlib.Path]) -> bool:
...
+            libs = {
+                destination.name: destination
+                for destination, _ in self._wheel_files['mesonpy-libs']
+            }
+
+            assert isinstance(binary, mesonpy._dylib.Dylib)
+            for entry in binary.load_dylib:
+                libname = entry.split('/')[-1]
+                if libname in libs:
+                    binary.replace_load_dylib(entry, f'@rpath/{libs[libname].as_posix()}')
```
dnicolodi, line 459: Uhm. I need to research this a bit more, but the impression is that there is something strange here. I'm not sure how to reconcile the different observations:

1. Meson adds an rpath during compilation pointing to the build directory, and that is sufficient to make things work,
2. if we move the libraries to  the `.mesonpy.libs` folder, add one rpath that we thing points to that folder and leave the rpath added by Meson in place (because we pick the files from the build directory and not from the install directory) things keep working,
3. if we do the same but we strip the rpath added by Meson (by picking the files from the install directory) things stop working

Unless there is a funny bug somewhere, I don't think that removing one rpath pointing to the wrong directory in 3. is what breaks things. Looking at what Meson does on install https://github.com/mesonbuild/meson/blob/5dc4fcae34ee3a5a3d47021f8ea8b2c5d3b14ea9/mesonbuild/scripts/depfixer.py#L410-L458 one thing that changes on install in the `-install_name` that gets rewritten from an `@rpath` relative path to an absolute path. May this be what needs to be reverted?

In particular, going from 2. where things work to 3. where things do not work, `LC_LOAD_DYLIB` entries did not change, thus I'm not convinced having to mess with these is the right solution.

I'll investigate more later.

mesonpy/_dylib.py
```diff
@@ -27,20 +28,28 @@ def _otool(self, *args: str) -> str:
...
+    def _find_tag(self, name: str, field: str) -> List[str]:
+        entries = []
+        # Run otool -l to get the load commands
+        otool_output = self._otool('-l').strip()
+        # Manually parse the output and find the tag entries
+        looking_at_tag = False
+        for line in map(str.split, otool_output.splitlines()):
+            if line[0] == 'cmd':
+                if len(line) < 2:
+                    raise ValueError('Unexpected "otool -l" output')
```
dnicolodi, line 40: I'm not sure being too picky about the exact format of the `otool` output is the right thing: it is not something we control and I don't know what are the backward compatibility guarantees for it, but I suspect not that many. I would write this to be as flexible as possible and to do not error out on unexpected input.

tests/packages/install-subdir-exclude/meson.build
```diff
@@ -0,0 +1,45 @@
...
+# We need two foo directories because of a bug in the Meson metadata generation.
+# https://github.com/mesonbuild/meson/issues/11539
+
+# Examples removed from https://mesonbuild.com/Reference-manual_functions.html#install_subdir
```
dnicolodi, line 13: "removed"? The examples are still in the linked Meson manual page. Maye "copied"? Although the example are not exactly the same, thus I would simply add a
```
# See https://mesonbuild.com/Reference-manual_functions.html#install_subdir
```
comment.

tests/packages/install-subdir-exclude/meson.build
```diff
@@ -0,0 +1,45 @@
...
+install_subdir(
+  'foo',
+  install_dir: purelib / 'subdir1',
+  strip_directory : false,
+)
```
dnicolodi, line 18: An empty line between `install_subdir()` calls makes reading this a bit easier.

tests/packages/install-subdir-exclude/meson.build
```diff
@@ -0,0 +1,45 @@
...
+install_subdir(
+  'foo2',
+  install_dir: purelib / 'subdir2',
+  strip_directory : true,
+)
+install_subdir(
+  'foo/bar',
```
dnicolodi, line 25: This reads a bit funny: the `foo` directory was already installed two `install_subdir()` invocations above. I would pick a different source directory for this example. And same thing for the example below.

mesonpy/__init__.py
```diff
@@ -427,48 +473,19 @@ def _install_path(
...
-                # Sort the directory names so that `os.walk` will walk them in a
-                # defined order on the next iteration.
-                dirnames.sort()
-                for name in sorted(filenames):
-                    path = os.path.normpath(os.path.join(root, name))
-                    if os.path.isfile(path):
-                        arcname = os.path.join(destination, os.path.relpath(path, origin).replace(os.path.sep, '/'))
-                        wheel_file.write(path, arcname)
+        if origin.is_dir():
+            for child in origin.rglob('**/*'):
```
dnicolodi, line 482: `pathlib.Path.rglob()` is like `pathlib.Path.glob()` but with `**/` prepended to the pattern. Why do we need another `**/` in the pattern here? Also, what does `pathlib.Path.rglob()` do with symbolic links? Is that what we need here?

mesonpy/__init__.py
```diff
@@ -427,48 +473,19 @@ def _install_path(
...
+                if child.is_file():
+                    arcname = destination / child.relative_to(origin)
```
dnicolodi, line 484: Aren't the paths returned by `pathlib.Path.rglob()` already relative to `origin`?

dnicolodi (changes requested):
I had a quick look to the combined diff. I'm not sure the fix for the macOS problem is the right one. 

I don't understand why the commits are organized the way they are. The PR would be much easier to review if the fixup commits (and the fixups of the fixups and the reverts of the fixups, and so on) were squashed to the commits they are fixing.

tests/test_wheel.py
```diff
@@ -215,6 +221,23 @@ def test_uneeded_rpath(wheel_purelib_and_platlib, tmp_path):
...
+    artifact = wheel.wheelfile.WheelFile(tmp_path / wheel_name)
+    artifact.extractall(tmp_path)
+
+    module = tmp_path / f'example{EXT_SUFFIX}'
+    if platform.system() == 'Linux':
+        binary = mesonpy._elf.ELF(module)
+    elif platform.system() == 'Darwin':
+        binary = mesonpy._dylib.Dylib(module)
+
+    assert 'custom-rpath' in binary.rpath
```
dnicolodi, line 238: This fails on macOS. I have some ideas on why, but I would need to verify.

FFY00:
> I don't understand why the commits are organized the way they are. The PR would be much easier to review if the fixup commits (and the fixups of the fixups and the reverts of the fixups, and so on) were squashed to the commits they are fixing.

I was asked to not force-push to existing commits to make review easier. This should be an easy enough thing for you to do locally, though.

mesonpy/__init__.py
FFY00: > In particular, going from 2. where things work to 3. where things do not work, `LC_LOAD_DYLIB` entries did not change, thus I'm not convinced having to mess with these is the right solution.

The `LC_LOAD_DYLIB` entry did change, and that why it breaks. Feel free to play with it locally.

https://github.com/mesonbuild/meson/blob/5dc4fcae34ee3a5a3d47021f8ea8b2c5d3b14ea9/mesonbuild/scripts/depfixer.py#LL447C49-L447C49

mesonpy/_dylib.py
FFY00: I disagree, IMO it's better to error out instead of keeping working with unexpected output, possibly causing obscure errors that are hard to debug. If this fails, we reproduce it, adjust the logic to expect some new format, and make a new (bugfix if needed) release.

tests/packages/install-subdir-exclude/meson.build
FFY00: This was poorly worded. I will update it.

tests/packages/install-subdir-exclude/meson.build
FFY00: Sure.

tests/packages/install-subdir-exclude/meson.build
FFY00: That requires to copy the entire tree again, that's why I didn't do it. I don't think it really matters either way.

dnicolodi:
Both the `install_subdir()` issue and the rpath issue has been an issue for as long as meson-python exist. I don't think there is anything that requires urgent attention and delaying a release here. `install_subdir()` with excludes keeps being broken in some configurations on Windows because a bug in Meson.

tests/test_wheel.py
```diff
@@ -215,6 +221,23 @@ def test_uneeded_rpath(wheel_purelib_and_platlib, tmp_path):
             assert 'mesonpy.libs' not in rpath
 
 
+@pytest.mark.skipif(platform.system() not in ['Linux', 'Darwin'], reason='Unsupported on this platform for now')
+def test_custom_rpath(package_link_against_local_lib, monkeypatch, tmp_path):
+    monkeypatch.setenv('LDFLAGS', '-Wl,-rpath,custom-rpath')
```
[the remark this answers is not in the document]
FFY00: Ping.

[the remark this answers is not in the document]
dnicolodi: Having the linker arguments added in `meson.build` allows to build the wheel package only once during testing. Having an extra rpath entry should not cause any problem for the other tests. I think that passing arguments to the build via environment variables is contrary to best practices. I don't think that it is the job of the meson-python test suite to validate all aspects of how a Meson project can be built, thus validating that linker arguments passed via environment variables are used by Meson is outside the scope here.

Therefore my preference is for adding the linker arguments to the test project via `meson.build`.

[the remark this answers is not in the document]
rgommers: > Having the linker arguments added in `meson.build` allows to build the wheel package only once during testing

That sounds fine to me, and should be equivalent. If not (which seems pretty unlikely) we can fix that when we find out.

FFY00:
Closing this as the PR is out of date, probably needs to be reconsidered facing the work done since it was opened, and I am not currently actively contributing this project anymore.

Issue #306, opened 2023-02-10
Remove "warning" when package metadata generate from `meson.build`


dnicolodi:
Currently a "warning" (some yellow text prefixed with an exclamation mark, which is very hard to read with my terminal emulator settings, but that is probably a problem of the default palette of the terminal emular) is emitted when the package metadata is synthesized from the project metadata in `meson.build`.

I think we should either decide that this is supported and make the warning simply an informative message, or decide that it is a bad idea and error out. As we can get the minimum required information to build a package out of `meson.build` we should support this and remove the warning. It seems its main purpose is to inform the user that they may have a typo in their `pyproject.toml` but the warning is very easy to miss within the rest of the output. The effects of not populating the metadata from `pyproject.toml` are probably much more visible.

Also, having all out test packages emitting what looks like a warning looks like poor form to me.

rgommers:
What can actually be taken from `meson.build`? Just name and version, or more? `license` is another keyword in Meson's `project` that matches, but that's going to be ambiguous once we support PEP 639.

dnicolodi:
Just name and version.

The PEP 621 `license` field is for a license text, while the `licence` parameter to the Meson's `project()` is a license identifier. It is recommended to be a SPDX license expression, but it is not enforced. I'm not very familiar with PEP 639, but my understanding is that it will allow to use SPDX license expressions in Python packages metadata. Therefore, once PEP 639 is settled, approved, and implemented we could map the Meson's project license to Python package metadata.

rgommers:
Given that a `pyproject.toml` file is required anyway and the gain of supporting a couple of fields from `meson.build` is very small, I'd personally go for requiring PEP 621 metadata. 

However, if others want to support using `name` and `version` from `meson.build`, that's fine with me too - in that case I do agree that the warning should be removed, and the extent of the support documented in our html docs.

eli-schwartz:
You *have* to use the version scraped from meson.build if you want to support git-based versions, for example by running `python -m setuptools_scm` to print the version and read that into meson.build. It's also potentially useful for buildsystem processing, so you may want to single-source it rather than defining it in multiple places including meson.build as well as pyproject.toml.

It's probably the single most common metadata field to be specified as dynamic.

Name cannot be dynamic one way or the other except by totally opting out of PEP 621 altogether, which seems unlikely. And it's questionable what other metadata fields are super interesting to define dynamically...

[...]

PR #358, opened 2023-03-22
DOC: improvements to "Passing arguments to Meson" and "Using config settings" pages

dnicolodi:


rgommers:
Thanks for this!

> I still think that the organization of the documentation is horrible. But everyone else seems to like it

I'm still a little on the fence, but I'd rather add the most important content we need (closing the open doc issues and some of the key TODOs), and then re-evaluate. Overall the framework works decently well in my other projects; `meson-python` doesn't have much of an API though, so the API reference part may not map with my previous experience.

dnicolodi:
I think the frameworks is great /for thinking about how to write documentation/. However, structuring documentation following the prescribed structure, just because it has a fancy name, does not make sense in most cases. It makes sense when there is a large amount of information to be conveyed, but meson-python has literally 5 configuration knobs. We are in the situation where we have two documentation pages listing what these knobs are, and I mean /only/ listing them without explaining what they do, two pages describing how to set them, and one page to tell what are the default.

docs/how-to-guides/config-settings.rst
```diff
@@ -9,41 +9,53 @@ Use build config settings
...
+	.. code-block:: console
 
+	   $ python -m pip wheel . \
+               --config-settings=setup-args="-Doption=disable" \
+               --config-settings=compile-args="-j6"
 
-Please check the `pypa/build documentation`_ for more information.
+This examples use the ``python -m pip wheel`` command to build a Python
+wheel that can be later installed or distributed. To build a package and
+immediately install it, just replace ``wheel`` with ``install``.
```
FFY00, line 43: I think this should be an admonition, also IMO the `python -m` bit of the command can be dropped in this text.

typo: "examples" > "example"

```suggestion
.. admonition:: Using ``pip install``
   :class: hint

   This example uses the ``pip wheel`` command to build a Python
   wheel that can be later installed or distributed. To build a package and
   immediately install it, just replace ``wheel`` with ``install``.
```

docs/how-to-guides/config-settings.rst
```diff
@@ -9,41 +9,53 @@ Use build config settings
...
 .. _pip documentation: https://pip.pypa.io/
+
+Please note that, while ``pypa/build`` concatenates arguments for the same
+key passed to the ``-C`` option, ``pip`` up to version 23.0.1 does not offer
+any way to set a build config setting to a list of strings: later values for
+the same key passed to ``--config-settings`` override earlier ones. This
+effectively limits the number of options that can be passed to each command
+invoked in the build process to one. This limitation is tracked in `pip
+issue #11681`_. This limitation should be removed in the next version of
+pip.
```
FFY00, line 59: I think this should be an admonition, also, the last two sentences can be combined. It might also make sense to split into two paragraphs if we're putting it in an admonition.

```suggestion
.. admonition:: pip <= 23.0.1 limitation
   :class: caution
   
   Please note that, while `pypa/build`_ concatenates arguments for the same
   key passed to the ``-C`` option, `pip`_ up to version 23.0.1 does not offer
   any way to set a build config setting to a list of strings: later values for
   the same key passed to ``--config-settings`` override earlier ones.
   
   This effectively limits the number of options that can be passed to each
   command invoked in the build process to one. This limitation is tracked
   in `pip issue #11681`_ and should be removed in the next version of pip.
```

docs/how-to-guides/meson-args.rst
```diff
@@ -4,77 +4,54 @@
 
 .. _how-to-guides-meson-args:
 
-***************************
-Passing arguments to Meson_
-***************************
+**************************
+Passing arguments to Meson
+**************************
```
FFY00, line 9: Why?

docs/how-to-guides/meson-args.rst
```diff
@@ -4,77 +4,54 @@
...
 
-.. todo::
+``meson-python`` invokes the ``meson setup``, ``ninja``, and ``meson
+install`` to build the files that will be included in the Python
```
FFY00, line 12: typo: missing the word "commands"

```suggestion
``meson-python`` invokes the ``meson setup``, ``ninja``, and ``meson
install`` commands to build the files that will be included in the Python
```

docs/how-to-guides/meson-args.rst
```diff
@@ -4,77 +4,54 @@
...
-   - Mention the order in which Meson arguments are added, and how it affect
-     Meson
-
-Advanced Meson options can be accessed by specifying extra arguments to the
-individual Meson commands:
-
-   - meson dist: https://mesonbuild.com/Commands.html#dist
-   - meson setup: https://mesonbuild.com/Commands.html#setup
-   - meson compile: https://mesonbuild.com/Commands.html#compile
-   - meson install: https://mesonbuild.com/Commands.html#install
```
FFY00, line 22: I think this is relevant to keep.

docs/how-to-guides/meson-args.rst
```diff
@@ -4,77 +4,54 @@
...
 
-    .. tab-item:: pip
-        :sync: key_pip
-
-		.. code-block:: console
+Or can be specified via configuration settings passed by the Python
+build front-end as ``setup-args``, ``compile-args``, ``install-args``,
+and ``dist-args`` :ref:`config settings <how-to-guides-config-settings>`.
+Configuration settings specified via the Python build front-end have
+precedence over and override the ones specified in ``pyproject.toml``.
```
FFY00, line 33: They don't override, they can be used to override though. As-is, the sentence can read as if you pass configs via the build frontend, we will ignore the settings specified in `pyproject.toml`.

```suggestion
Configuration settings specified via the Python build front-end have
precedence over, and be used to override, the ones specified in ``pyproject.toml``.
```

docs/how-to-guides/meson-args.rst
```diff
@@ -4,77 +4,54 @@
...
 
-			$ python -m pip --config-settings=dist-args="args"    \
-					--config-settings=setup-args="args"   \
-					--config-settings=compile-args="args" \
-					--config-settings=install-args="args" .
+``meson-python`` overrides some of the default Meson options, see
+:ref:`explanations-default-options`. User options specified via
+``pyproject.toml`` or via Python build front-end config settings
+override the ``meson-python`` defaults..
```
FFY00, line 38:
```suggestion
``pyproject.toml`` or via Python build front-end config settings
can be used to override the ``meson-python`` defaults.
```

docs/how-to-guides/meson-args.rst
```diff
@@ -85,33 +62,37 @@ To set this option temporarily at build-time:
...
-			 $ python -m build -Csetup-args="--default-library=static" .
+           $ python -m build -Csetup-args="--default-library=static" .
 
     .. tab-item:: pip
         :sync: key_pip
 
         .. code-block:: console
 
-			 $ python -m pip --config-settings=setup-args="--default-library=static" .
+	   $ python -m pip wheel . --config-settings=setup-args="--default-library=static" .
```
FFY00, line 72: There's still a tab here.

```suggestion
           $ python -m pip wheel . --config-settings=setup-args="--default-library=static" .
```

docs/how-to-guides/meson-args.rst
```diff
@@ -85,33 +62,37 @@ To set this option temporarily at build-time:
...
-2) Use Meson install_tags for selective installs
+Select the build targets to include in the wheel
 ------------------------------------------------
 
-Meson install_tags can be used (since ``meson-python`` >= 0.13) to select which
-targets are installed into the binary wheels. This example causes meson-python
-to only install targets tagged with ``runtime`` or ``python-runtime``) into the
-binary wheel (ignoring e.g. C++ headers).
+It is possible to include in the Python wheel only a subset of the
+installable files using Meson `installation tags`_ via the ``meson
```
FFY00, line 79:
```suggestion
installable files using Meson `installation tags`_, via the ``meson
```

docs/how-to-guides/meson-args.rst
```diff
@@ -85,33 +62,37 @@ To set this option temporarily at build-time:
...
+install``'s ``--tags`` command line option. When ``--tags`` is
+specified, only files that have been tagged with one of the tags are
+going to be installed. Meson sets predefined tags on some
+files. Custom installation tag can be set using the ``install_tag``
+keyword argument passed to the target definition function.  In this
+example only targets tagged with ``runtime`` or ``python-runtime`` are
+included in the Python wheel.
```
FFY00, line 86: I'd make a new paragraph here, also suggested rewording.

I'd skip the large tags explanation, though, and simply point the user to the Meson docs. This is supposed to just show you what you can do, not to be a full tutorial (we can make it a full tutorial, though).

typo: "Custom installation tag can" > "Custom installation tags can"

```suggestion
install``'s ``--tags`` command line option. When ``--tags`` is specified,
only files that have one of the specified tags are going to be installed.

Meson sets predefined tags on certain files. Custom installation tags can
be set using the ``install_tag`` keyword argument passed to the target
definition function.  In this example only targets tagged with ``runtime``
or ``python-runtime`` are included in the Python wheel.
```

docs/how-to-guides/meson-args.rst
```diff
@@ -122,30 +103,18 @@ To set this option temporarily at build-time:
 
         .. code-block:: console
 
-			 $ python -m build -install-args="--tags=runtime,python-runtime" .
+	   $ python -m build -install-args="--tags=runtime,python-runtime" .
```
FFY00, line 106: There's still a tab here.

```suggestion
           $ python -m build -install-args="--tags=runtime,python-runtime" .
```

docs/how-to-guides/meson-args.rst
```diff
@@ -122,30 +103,18 @@ To set this option temporarily at build-time:
...
 
     .. tab-item:: pip
         :sync: key_pip
 
         .. code-block:: console
 
-			$ python -m pip --config-settings=install-args="--tags=runtime,python-runtime" .
+	   $ python -m pip wheel . --config-settings=install-args="--tags=runtime,python-runtime" .
```
FFY00, line 113: Same.

```suggestion
           $ python -m pip wheel . --config-settings=install-args="--tags=runtime,python-runtime" .
```

docs/how-to-guides/meson-args.rst
```diff
@@ -170,14 +136,11 @@ To set this option temporarily at build-time:
 
         .. code-block:: console
 
-			 $ python -m build -Csetup-args="-Doptimization=3" .
+	   $ python -m build -Csetup-args="-Doptimization=3" .
```
FFY00, line 139: Same.

```suggestion
           $ python -m build -Csetup-args="-Doptimization=3" .
```

docs/how-to-guides/meson-args.rst
```diff
@@ -170,14 +136,11 @@ To set this option temporarily at build-time:
...
     .. tab-item:: pip
         :sync: key_pip
 
         .. code-block:: console
 
-			 $ python -m pip --config-settings=setup-args="-Doptimization=3" .
-
-
-.. _Meson: https://github.com/mesonbuild/meson
+	   $ python -m pip wheel . --config-settings=setup-args="-Doptimization=3" .
```
FFY00, line 146: Same.

```suggestion
           $ python -m pip wheel . --config-settings=setup-args="-Doptimization=3" .
```

docs/reference/config-settings.rst
```diff
@@ -8,9 +8,11 @@
...
 *********************
 
-This page lists the build configuration settings, that is, the settings you can
-pass when building the project. Please check our
-:ref:`how-to-guides-config-settings` guide for information on how to use them.
+This page lists the build configuration settings, that is, the
+settings you can pass when building the project. Please refer to
+:ref:`how-to-guides-config-settings` and
+:ref:`how-to-guides-meson-args` and for information on how to use
+them.
```
FFY00, line 15: The ref role will put the title there, so for it to read naturally we should say "see the :ref:(...) guide", instead of "see :ref:(...)".

typo: "and (...) and"

```suggestion
settings you can pass when building the project. Please refer to the
:ref:`how-to-guides-config-settings` and :ref:`how-to-guides-meson-args`
guides for information on how to use them.
```

FFY00 (changes requested):
Thanks @dnicolodi, this looks good overall, but I have a couple comments.

docs/how-to-guides/config-settings.rst
dnicolodi: The command executed is `python -m pip wheel`. Wheel is (was?) a command from the `wheel` package. Why should this an admonition? There is no reason to focus the attention to the reader to this paragraph, it is just a pretty boring explanation of what the command does. I'm not a fan of using heavy markup for no reason.

docs/how-to-guides/meson-args.rst
dnicolodi: Because it looks nicer and we don't need to make every mention to Meson a link to the Meson home page. We don't do it in the text and I don't see any reason to do it in titles.

docs/how-to-guides/meson-args.rst
dnicolodi: No, a comma there is grammatically incorrect.

docs/how-to-guides/meson-args.rst
dnicolodi: This text was in the document before. I just edited it be readable.

docs/how-to-guides/config-settings.rst
```diff
@@ -9,41 +9,53 @@ Use build config settings
...
 Build config settings are settings you can pass to ``meson-python`` when
-building your project. They can be use to customize the build in some aspect.
+building your project. They can be use to customize the build in some
+aspect.
 
-The way you specify them depends on which tool you are using, here are the most
-popular ones.
+Several Python build front-ends exist, with different ways to pass
+configuration settings to the build back-end. The most popular are
+``pypa/build``, which uses the ``-C`` command line option, and ``pip``,
```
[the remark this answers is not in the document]
FFY00: - It is the style the docs are currently following
- Gives users a clickable link to the project, which is relevant if they are using pip for eg. and didn't know about pypa/build
- There's no downside?

> Again, without motivating your objections, it seems that you are disagreeing for the sake of disagreeing

I am not trying to disagree, these are just suggestions. I spent a significant amount of time going over and reviewing this PR, which I did not have to do, I could simply have a high level look and said :+1:, but I care about the docs, so I took the time to note what I consider to be possible improvements.

Isn't this common? When I open PRs, at least, I expect there to be comments, nothing I write is perfect or beyond improvement, and all that is subjective anyway. The point of reviews is to get other people's perspective, and see how it can be improved.

Anyway, I don't think our communication is very healthy at the moment. Is there anything you think we can do to try to improve this?
I sent you an email a couple days ago, were you able to read it over?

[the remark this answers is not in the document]
dnicolodi: I expect feedback to come with explanation. Otherwise it seems that your opinion is more important just because it is your opinion. I think it is the fundamental rule of providing feedback: if you cannot motivate it, it is just a persona preference, and there is no reason why your personal preference should be more important than the preference of someone else.

I don't think there is a communication issue. I disagree with you sometimes and you don't seem to handle that very well. See for example the interaction we just had on #360. I don't think that what you wrote is the proper way to deal with disagreement on technical matters. In particular the "Tired" comment you used to retire your objections. It makes it look like you retired them only because I've worn you out repeating the same arguments over and over. Instead the issue was already discussed in #306 where you decided to do not contribute to the discussion and the decision settled. I tried to provide the reasoning motivating the outcome of the discussion and you, instead of replying in the merit, decided to act in IMHO an immature way and declare that you don't care and I can have it may way if I really insist. If you really don't care, why did you object in the first place?

> I care about the docs, so I took the time to note what I consider to be possible improvements

How did the document I'm amending find its way in the documentation then?

PR #359, opened 2023-03-22
MAINT: simplify

dnicolodi:
Inverting the direction of the map between wheel installation paths and Meson installation path placeholders makes the code much simpler and more efficient (one dictionary lookup per installed item vs iterating a dictionary and checking for tuple membership for each installed item).


mesonpy/__init__.py
```diff
@@ -133,40 +133,32 @@ def _init_colors() -> Dict[str, str]:
...
         for src, target in group.items():
-            directory, path = _map_meson_destination(target['destination'])
-            if directory is not None:
-                wheel_files[directory].append((path, src))
+            dest = target['destination']
+            parts = pathlib.Path(dest).parts
+            path = _INSTALLATION_PATH_MAP.get(parts[0])
+            if path is None:
+                raise BuildError(f'Could not map installation path to an equivalent wheel directory: {dest!r}')
+            wheel_files[path].append((pathlib.Path(*parts[1:]), src))
```
FFY00, line 161: I think choosing the variables a bit more carefully makes the code easier to read. Also, using `destination.relative_to(anchor)` instead of `pathlib.Path(*parts[1:])`, for eg., makes what we are trying to do a bit more explicit and quicker to understand.

```suggestion
            destination = pathlib.Path(target['destination'])
            anchor = destination.parts[0]
            scheme = _INSTALLATION_PATH_MAP.get(anchor)
            if scheme is None:
                raise BuildError(f'Could not map installation path to an equivalent wheel directory: {destination!r}')
            wheel_files[scheme].append((destination.relative_to(anchor), src))
```

mesonpy/__init__.py
```diff
@@ -133,40 +133,32 @@ def _init_colors() -> Dict[str, str]:
 assert all(re.match(_EXTENSION_SUFFIX_REGEX, x) for x in _EXTENSION_SUFFIXES)
 
 
-# Maps wheel installation paths to Meson installation path placeholders.
+# Map Meson installation path placeholders to wheel installation paths.
```
FFY00, line 136: It's maybe better if we refer to the wheel locations as schemes.

```suggestion
# Map Meson installation path placeholders to wheel installation schemes.
```

FFY00 (approved):
Thanks @dnicolodi. This looks good, I just have a couple of minor comments.

Also, a bit of a nitpick, but in the future could you try to use more meaningful commit names (eg. "simplify _map_to_wheel")? I can squash-merge and fix the naming on this PR, so you don't need to update the commit message here, but it'd just be good to keep in mind in the future.

mesonpy/__init__.py
dnicolodi: `destination.relative_to(anchor)` is wrong and does not work.

We are dealing with `sysconfig` concepts. In this context a "scheme" is the collections of concrete directories where the parts of a wheel are installed. I've even recently added a link to the documentation explaining this. As the `sysconfig` maintainer I'm surprised that you get the basic nomenclature wrong.

mesonpy/__init__.py
dnicolodi: Again, these are not schemes. See the `sysconfig` documentation, the wheel format documentation, and many other documents. In `sysconfig` nomenclature, "schemes" are the collections of the concrete paths where wheel components are installed. For example:
```
>>> sysconfig.get_scheme_names()
('nt', 'nt_user', 'nt_venv', 'osx_framework_user', 'posix_home', 'posix_prefix', 'posix_user', 'posix_venv', 'venv')
```
Each scheme defines "paths":
```
>>> sysconfig.get_path_names()
('stdlib', 'platstdlib', 'purelib', 'platlib', 'include', 'scripts', 'data')
```
"paths" is the thing we are dealing with here, not schemes.

PR #360, opened 2023-03-22
ENH: remove warning when package metadata generate from meson.build

### LINKED ISSUE #306: Remove "warning" when package metadata generate from `meson.build`

dnicolodi:
This use case is supported and extensively used in the meson-python test suite, there is no reason to emit a warning when a supported feature is used. This is accordance with PEP 621, which states that the lack of a "project" table in pyproject.toml implicitly means that the build back-end will dynamically provide all fields.

Fixes #306.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -721,10 +721,6 @@ class Project():
         else:
             self._metadata = pyproject_metadata.StandardMetadata(
                 name=self._meson_name, version=packaging.version.Version(self._meson_version))
-            print(
-                '{yellow}{bold}! Using Meson to generate the project metadata '
-                '(no `project` section in pyproject.toml){reset}'.format(**_STYLES)
-            )
         self._validate_metadata()
 
         # set version from meson.build if dynamic


FFY00 (approved):
It is not a good idea to use this. It is simply provided for quick tests/hacking, but it should really not be used by anyone in production. I am -0.5 on this.

If the warning bothers you on the test suite, we can have an env var to disable it, and set it on the tests.

FFY00 (dismissed):
Sorry, wrong button :sweat_smile:

dnicolodi:
Why it is not a good idea? PEP 621 allows it (or mandates it, the wording in all the packaging PEPs is so poor that is impossible to tell what the authors meant). We have been discussed this extensively in #306 and we could not identify any reason why this is not desirable.  In general, it would be nice for you to motivate your objections. Otherwise it seems that you object any change just for the sake of having a divergent opinion to express.

FFY00:
You should not be releasing projects with only the name and version fields to PyPI, and these are the only things this method supports. It is fine for doing quick tests, etc., but it should not be used in production.

dnicolodi:
meson-python is not used exclusively to build wheel to be uploaded to PyPI. `name` and `version` are the only required fields in `pyproject.toml` https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata And we don't do anything to enforce that the users provides more fields. Why should we deviate from what is mandated and recommended by the PEPs?

FFY00:
> meson-python is not used exclusively to build wheel to be uploaded to PyPI. `name` and `version` are the only required fields in `pyproject.toml` [packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata) And we don't do anything to enforce that the users provides more fields.

Yes, but the point is of this warning is for people who might have not realized that. When adding PEP 621 metadata, you will likely add the rest of the fields too, and if you don't, it's because you explicitly chose too, not because you didn't realize there was more metadata to add.

Anyway, have it your way. This isn't a big issue anyway, there's no point in arguing more about this.

> Why should we deviate from what is mandated and recommended by the PEPs?

What are you referring to? This warning has nothing to do with PEP 621.

dnicolodi:
> What are you referring to? This warning has nothing to do with PEP 621.

PEP 621 states that when a `project` section is missing in `pyproject.toml` it should be treated as an implicit request for all the metadata to be generated dynamically. I already linked multiple times to the section of the PEP with the exact wording, I'm not going to look it up once more.

rgommers:
For the record, I think this outcome (removing the warning) is fine. I'm not in favor of warnings for supported usage, except in case we really want to remove something but can't for backwards compatibility reasons. In general, we should offer well-supported features (with no warnings), or not support them at all. That's a general software design principle, and I see no reason to deviate from it here. We can use docs to guide newcomers to recommend best practices.

PR #362, opened 2023-03-23
DOC: reference guide: `ARCHFLAGS` docs, and various style and phrasing improvements

dnicolodi:
These mostly target the reference pages.


docs/reference/environment-variables.rst
```diff
@@ -8,86 +8,74 @@
 Environment variables
 *********************
 
+.. envvar:: ARCHFLAGS
 
-.. todo::
+   This environmental variable is used for supporting architecture cross
+   compilation on macOS in a way compatible with setuptools_. It is ignored on
+   all other platforms. It can be set to ``-arch arm64`` or to ``-arch
+   x86_64`` for compiling for the aarch64 and the x86-64 architectures
```
rgommers, line 16: minor: `arm64` is the better architecture name to use for macOS specifically I think, that's what it's called in most docs/resources.

docs/reference/environment-variables.rst
```diff
@@ -8,86 +8,74 @@
...
-          of tags that are currently accepted by, e.g., ``pip`` are:
-          ``macosx_10_9``, ``macosx_10_13``, ``macosx_11_0``, ``macosx_12_0``.
-
-       .. admonition:: Specifying the target version via the compiler
-          :class: note
-
-          Another way of specifying the target macOS platform is to explicitly
-          use ``-mmacosx-version-min`` compile and link flags when invoking
-          Clang_ or GCC_. It is not possible for ``meson-python`` to detect this
-          though, so it will not set the *wheel* tag to that flag automatically.
```
rgommers, line 82: Just to note: I agree with dropping this content.

docs/reference/pyproject-settings.rst
```diff
@@ -8,36 +8,23 @@
...
 
-Usage example:
-
-.. code-block:: toml
-
-   [tool.meson-python.args]
-   setup = ['-Dwarning_level=2', '-Db_pie=true']
-   dist = ['--include-subprojects']
-   compile = ['-j4']
-   install = ['--quiet']
```
rgommers, line 43: Just to note: this seems okay to remove too, there are equivalent examples in the pages cross-linked at the top of this reference page.

docs/reference/environment-variables.rst
```diff
@@ -8,86 +8,74 @@
...
 
-   - Document ``ARCHFLAGS`` (see `#226`_)
+   The macOS toolchain allows architecture cross compilation passing the
+   ``-arch`` flat to the compilers. ``meson-python`` inspects the content of
+   this environment variable and synthesizes a Meson `cross build definition
+   file`_ with the appropriate content, and passes it to ``meson setup`` via
+   the ``--cross-file`` option.
 
+   Support for this environment variable is maintained only for compatibility
+   with existing tools and is not the recommended solution for cross
```
rgommers, line 27: consider inserting, here or higher up where setuptools is mentioned: "`cibuildwheel` in particular" (I _think_ that's the only real reason)

rgommers (commented):
Thanks @dnicolodi. The style changes look like an improvement overall, and the added content is useful. A few inline comments, and a few questions:

-  We indeed use `ninja` instead of `meson compile`, so that change reflects the current code. Just to make sure: we are happy with that right, and have no interest in changing that back and allowing other backends to be used?
- Mind if I squash-merge this? There's a lot of commits given the diff, and they don't all seem orthogonal.

The [`.. option::` directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-option) seems appropriate to use indeed.

dnicolodi:
Thanks for the review @rgommers. I applied the suggested changes. Feel free to squash merge, if you like it better that way. I like to keep logically separated changes in individual commits also for documentation, but I agree that these are a lot of commits given the number of lines changed.

docs/reference/environment-variables.rst
rgommers: Ah yes, overlooked that. It's okay to leave as is - I'll resolve this comment.

rgommers (approved):
LGTM now, in it goes. Thanks @dnicolodi!

Issue #330, opened 2023-03-01
Verify user arguments


FFY00:
In #167, we added options to allow users to pass arguments to Meson, but that implementation allows them to pass options that may conflict with ours, which may mess things up (eg. overriding `--prefix`).

While I think we should try to detect conflicting options and either error out or issue a warning, which is tricky to implement, most issues I think can be mitigated by disallowing arguments that don't start with `-`, I checked and there's no positional argument we'd want users to be able to specify, and putting our options last in the call, other than just defaults that we want the user to be able to override. I think this was a big overlook from my part when we merged that PR, probably because I was meaning to do the argument validation.

dnicolodi:
If we move away from having to run `meson install` (see #279 and #280) we don't need to pass the `--prefix` option (or the --python.purelibdir` and `--python.platlibdir` options) to `meson setup` because we don't care anymore about the real installation paths, only about the installation paths relative to the placeholder directories in `intro-install_plan.json`.

The remaining command line arguments are: `$source_dir`, $build_dir`, `--native-file=$native_file`, `-Ddebug=false`, `-Doptimization=2`, `$user_args`. I think this ordering is the right one because it allows the user to override the default `-Ddebug` and `-Doptimization` (and in the near future `-Db_ndebug`) options. I would need to check what happens when more than one `--native-file` argument are present. It would be cool if Meson could merge the content of the two. If he does not, things are most likely not going to work very well if the user adds their own `--native-file` argument.

`meson setup` errors out if more than two positional command line arguments (arguments not starting with a `-`) are specified:
```
$ meson setup . _build foo
usage: meson [-h] {setup,configure,dist,install,introspect,init,test,wrap,subprojects,rewrite,compile,devenv,env2mfile,help} ...
meson: error: unrecognized arguments: foo
```
Therefore, the command line validation is already performed and I don't think meson-python needs to adds its own validation on top.

The only open question is about the `--native-file` argument. If Meson does not try to merge the two native files, the easy fix would be disallow passing this option via meson-python. The more elaborate fix would be to detect it and merge the user provided native file with the meson-python generated one.

dnicolodi:
I've verified that Meson merges multiple native files. The only thing to decide is whether we want the user `--native-file` option to take precedence over the one generated by meson-python or vice versa.

FFY00:
> If we move away from having to run `meson install` (see #279 and #280)

We run `meson install` because binaries in the build directory have a RPATH set, so that you can use them from there.

```
[...]

PR #364, opened 2023-03-26
ENH: remove the need for a working directory & other cleanups

### LINKED ISSUE #317: BUG: install_subdir() in meson.build does not honor exlude_files/directories

### LINKED ISSUE #355: Prevent users from overriding our required arguments
We pass several arguments to Meson calls, some of them are required, some of them are simply meant as defaults, which the user can override (see GH-329). We should pass the our defaults first, then the user arguments, and finally, the arguments we require to not be changed.

Supersedes GH-330

dnicolodi:
This brings back the simplifications I proposed in #340 in a way compatible with having the wheel generated from files installed by `meson install` in a temporary destination directory. This covers a very minimal part of the work done in #348 too. It should be trivial to rebase #348 on top of this.

Fixes #355 and #330 as there are no arguments the user can supply that would interfere with meson-python internals. It also fixes #317 in the same way as #348 does.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -18,7 +18,6 @@ import difflib
 import functools
 import importlib.machinery
 import io
-import itertools
 import json
 import os
 import pathlib
@@ -505,30 +504,25 @@ class _WheelBuilder():
         self._project.install()
 
         wheel_file = pathlib.Path(directory, f'{self.name}.whl')
-
         with mesonpy._wheelfile.WheelFile(wheel_file, 'w') as whl:
             self._wheel_write_metadata(whl)
 
-            with mesonpy._util.cli_counter(
-                len(list(itertools.chain.from_iterable(self._wheel_files.values()))),
-            ) as counter:
-                # install root scheme files
-                root_scheme = 'purelib' if self.is_pure else 'platlib'
-                for destination, origin in self._wheel_files[root_scheme]:
-                    self._install_path(whl, counter, origin, destination)
+            with mesonpy._util.cli_counter(sum(len(x) for x in self._wheel_files.values())) as counter:
 
-                # install bundled libraries
-                for destination, origin in self._wheel_files['mesonpy-libs']:
-                    destination = pathlib.Path(f'.{self._project.name}.mesonpy.libs', destination)
-                    self._install_path(whl, counter, origin, destination)
+                root = 'purelib' if self.is_pure else 'platlib'
 
-                # install the other schemes
-                for scheme in self._wheel_files.keys():
-                    if scheme in (root_scheme, 'mesonpy-libs'):
-                        continue
-                    for destination, origin in self._wheel_files[scheme]:
-                        destination = pathlib.Path(self.data_dir, scheme, destination)
-                        self._install_path(whl, counter, origin, destination)
+                for path, entries in self._wheel_files.items():
+                    for dst, src in entries:
+
+                        if path == root:
+                            pass
+                        elif path == 'mesonpy-libs':
+                            # custom installation path for bundled libraries
+                            dst = pathlib.Path(f'.{self._project.name}.mesonpy.libs', dst)
+                        else:
+                            dst = pathlib.Path(self.data_dir, path, dst)
+
+                        self._install_path(whl, counter, src, dst)
 
         return wheel_file
 
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -498,30 +498,33 @@ class _WheelBuilder():
     def build(self, directory: Path) -> pathlib.Path:
         # ensure project is built
         self._project.build()
-        # install the project
-        self._project.install()
 
-        wheel_file = pathlib.Path(directory, f'{self.name}.whl')
-        with mesonpy._wheelfile.WheelFile(wheel_file, 'w') as whl:
-            self._wheel_write_metadata(whl)
+        # install project in temporary destination directory
+        with tempfile.TemporaryDirectory() as destdir:
+            self._project.install(destdir)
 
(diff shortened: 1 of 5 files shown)


dnicolodi:
The macOS test failure are the same found in #348 and will have the same solution.

rgommers:
Thanks @dnicolodi. Trying to think of a merge strategy here - I think the status is:
- this seems to overlap with, and improve on, the first 3 commits of gh-348.
- this is blocked on the macOS RPATH test failures right now, which are addressed by ``gh-348`` (and that's the key tricky and unfinished thing in that PR). This PR could temporarily xfail them though?
- ``gh-348`` also has tests for `subdir(..., exclude_files)` that are useful and will pass in combination with this PR. But there's no issue in adding those later.

dnicolodi:
I see this more like preparation work for landing #348 with the resulting code being in better shape (if we agree that the resulting code after merging this PR is "better").

From a "build things from the ground up" point of view, I would merge this and rebase the fixes in #348 on top. Or we can do the other way around and have this merged after #348 to cleanup the implementation. I think merging this first results in less (or easier to resolve) rebase conflicts and will make looking the commit history easier, but we can go either way. This PR being ready to be merged, while the proper solution for the rpath issues in #348 are still being discussed, is one reason to have this merged first. If you want to merge with all green tests, I can mark the rpath test as `xfail`.

I can take care of rebasing #348 on top of this.

The tests for `exclude_files` probably found a bug in Meson on Windows, thus these are also marked as `xfail` on that platform. I started investigating that on the Meson side, but I didn't reach a conclusion yet.

dnicolodi:
PR #348 rebased on top of this is here https://github.com/dnicolodi/meson-python/tree/pr348

FFY00:
I removed the v0.13.0 milestone, as this getting in is not critical for it, and we should release 0.13.0 as soon as we are able to. If this is in shape to get it by the time we sort the release blockers, great, otherwise we can just merge it afterwards and have it in 0.14.0.

dnicolodi:
This is ready to be merged since a long time. It just needs a review. I don't understand why you think otherwise.

FFY00:
It conflicts with #348, which needs to get in for the new release.

dnicolodi:
I already posted the conflict resolution.

FFY00:
But the PR may get more changes before merging. If you have an updated PR by then, great, we can merge it, otherwise we just do it for the next release.

dnicolodi:
What's the problem with merging this now, rebase #348 on top, and take it from there?

FFY00:
It can unnecessarily delay the release. Like I said, this is not critical, unlike the other PR, so I am not merging this now.

FFY00:
This PR hasn't even been reviewed btw, so it might still require changes, etc. I am prioritizing the release blockers.

FFY00:
There are still unaddressed items in that PR that I need to deal with, and I haven't even pushed any changes from the feedback received. When I do not have any more actionable items there, I will review this.

dnicolodi:
@rgommers I left out the commits with behavior changes from this PR. This is now just refactoring and a change in the structure of temporary directories. It would be nice to have this merged as a first step toward fixing the linked issues.

rgommers (approved):
Thanks @dnicolodi. Restructuring this to have no behavior changes was helpful. Sorry it took so long to review. I verified this as well as I could, and tested with SciPy - it all looks good. The code is easier to understand after this - and the simplifications in the tests are a good sign too. So in it goes.

rgommers:
@dnicolodi WDYT about tagging a 0.14.0 release? The PR queue is pretty much empty, so now would be a good time. We can quickly do a 0.15.0 too in case the subsequent PR to this one arrives quickly.

dnicolodi:
> WDYT about tagging a 0.14.0 release?

I don't know how critical it is to push a release with the fixes that already landed in `main`. There are three known bugs related to the fact that meson-python takes the files for the wheel from the source directory instead than from the build directory: (1) the handling of `exclude_path` and `exclude_directory` for `install_subdir`, and (2) removing the RPATH pointing to the build directory. In addition, there is a bug (3) in how sub direcrtories of direcotries installed with `install_subdir` are handled. I have patches for all three, thus getting these fixes in should be fast. However, investigating (2) I found a bug in Meson, thus having correct RPATH handling would require a new Meson release.

rgommers:
It's not critical to do it right this minute, so let's wait for at least the fixes for the bugs that don't depend on a new Meson release?

I'm also going through the backlog now and see what can be closed or easily addressed. Some of the documentation issues are straightforward, I'll try to fix a couple after dinner.

Issue #347, opened 2023-03-16
Use an autoformatter


FFY00:
I'd swear we had [black](https://github.com/psf/black) setup, but I guess not??

Anyway, having an autoformatter makes sure the codestyle is consistent across the whole codebase. I think [black](https://github.com/psf/black) is the best option.

rgommers:
I've given up having strong opinions about style issues (life's too short:)). I'm happy to go with whatever other folks prefer.

rgommers:
It seems like there isn't consensus that this is a good idea or worth the code churn. We have two closed PRs that tried to deal with code style more, and we've already got some checkers in CI and the code is generally already in quite good shape. So let me close this one as wontfix.

PR #367, opened 2023-03-28
MAINT: add muon-fmt pre-commit hook

FFY00:
Similarly to #347, it'd also be good to have an autoformatter for the meson files, to ensure we have a consistent style.

The only worry is that muon currently only supports Meson 0.64.0, which will limit us to that syntax, but I don't think we need any fancy features anyway, so I am not worried about that.

### DIFF
--- a/docs/examples/spam/meson.build
+++ b/docs/examples/spam/meson.build
@@ -2,40 +2,28 @@
 #
 # SPDX-License-Identifier: MIT
 
-project(
-  'spam',
-  'c',
-)
+project('spam', 'c')
 
 py_mod = import('python')
 py = py_mod.find_installation()
 
 # Specify directories to be included - usually this is where your header files live.
 # Note that specifying the path to Python.h is not necessary with meson-python.
-includes = include_directories(
-  [
-    'src'
-  ]
-)
+includes = include_directories(['src'])
 
 srcs = [
-  'src/spammodule.c',
+    'src/spammodule.c',
 ]
 
 # If some files (such as .py files) need to be copied to site-packages,
 # this is where they get specified. Files get copied to
 # <python directory>/site-packages/<subdir>
-py.install_sources(
-  [
-    'src/__init__.py'
-  ],
-  subdir: 'spam'
-)
+py.install_sources(['src/__init__.py'], subdir: 'spam')
 
 py.extension_module(
-  '_spam',
-  srcs,
-  install: true,
-  subdir: 'spam',
-  include_directories: includes
+    '_spam',
+    srcs,
+    install: true,
+    subdir: 'spam',
+    include_directories: includes,
 )
--- a/tests/packages/configure-data/meson.build
+++ b/tests/packages/configure-data/meson.build
@@ -13,8 +13,6 @@ py = py_mod.find_installation()
 configure_file(
     input: 'configure_data.py.in',
     output: 'configure_data.py',
-    configuration: configuration_data({
-        'MSG': 'hello!',
-    }),
+    configuration: configuration_data({'MSG': 'hello!'}),
     install_dir: py.get_install_dir(),
 )
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,3 +38,8 @@ repos:
     pass_filenames: true
   - id: reuse
     name: reuse lint
+- repo: https://github.com/FFY00/muon-pre-commit
+  rev: 0.1.0-c6207ca1
+  hooks:
+  - id: muon-fmt
+    args: [-c, muon_fmt.ini]
--- a/meson.build
+++ b/meson.build
@@ -8,18 +8,18 @@ py_mod = import('python')
 py = py_mod.find_installation()
 
 if py.language_version().version_compare('< 3.6')
-  error('Invalid Python version, only >= 3.6 is supported.')
+    error('Invalid Python version, only >= 3.6 is supported.')
 endif
 
 py.install_sources(
-  'mesonpy/__init__.py',
-  'mesonpy/_compat.py',
-  'mesonpy/_dylib.py',
-  'mesonpy/_editable.py',
-  'mesonpy/_elf.py',
-  'mesonpy/_introspection.py',
-  'mesonpy/_tags.py',
-  'mesonpy/_util.py',
-  'mesonpy/_wheelfile.py',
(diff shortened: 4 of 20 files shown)


FFY00:
It seems pre-commit.ci doesn't support docker-based actions.

docs/examples/spam/meson.build
```diff
@@ -3,39 +3,30 @@
...
 # If some files (such as .py files) need to be copied to site-packages,
 # this is where they get specified. Files get copied to
 # <python directory>/site-packages/<subdir>
-py.install_sources(
-  [
-    'src/__init__.py'
-  ],
-  subdir: 'spam'
-)
+py.install_sources(['src/__init__.py'], subdir: 'spam')
```
dnicolodi, line 24: I don't like how muon-fmt likes to format the code. More about this later, but can it at least be made to be consistent with itself? For example, here it prefers to have a list of one element with the delimiters inline and additional arguments to follow...

dnicolodi (changes requested):
I really like the effort to make the style of the meson source code more uniform and more polished. Thank you. However, it seems that muon-fmt chooses a fairly poor style for some code. I don't think it is worth to have to manually format these and have markers to tell muon-fmt to leave them alone, it is clutter in the source code for very little gain, also because these files do not really change often, if ta all.

Unless there is a way to configure muon-fmt to do the right thing in these instances (I haven't check, maybe there is). I would commit the result of running muon-fmt, except the instances where it make things worst, and I would not add the pre-commit hook.

While we are at it, we may also consider turning instances of
```meson
py_mod = import('python')
py = py_mod.find_installation()
```
into
```
py = import('python').find_installation()
```
I like the latter much more.

docs/examples/spam/meson.build
```diff
@@ -3,39 +3,30 @@
 # SPDX-License-Identifier: MIT
 
 project(
-  'spam',
-  'c',
+    'spam',
+    'c',
```
rgommers, line 7: Meson's recommended style (what it uses in all its docs is 2 spaces), not 4. So the formatter seems to be in the wrong here.

FFY00: Being is used in the docs and being recommended are two very different things. I could not find anything relating to 2 vs 4 spaces in the actual [Meson style recommendations page](https://mesonbuild.com/Style-guide.html), so I would not say this style is "wrong".

At the end of the day, it doesn't really matter anyway.

FFY00:
All the objections here are purely subjective, and I agree with many of them, but AFAIK there's no alternative Meson formatter, so we don't really have any alternative anyway.

The question here is if your dislike for the style, which, again, is purely subjective, is enough to justify the not adapting the auto-formatter at all.

My opinion is that everyone has opinions about code-style, you will never make everyone happy, and this is something that at the end of the day is completely useless to progressing the project anyway.

As such, I am +1 on adopting muon fmt, but definitely happy to reconsider moving to a different formatter if one shows up.

I do not wish to argue any further about this, it will not be productive, so it is now up to the rest of the maintainers to make the final call.

docs/examples/spam/meson.build
dnicolodi: The indentation is configurable.

dnicolodi:
I don't find auto formatters useful and I think imposing strict stylistic rules is more a nuisance than anything else. Enforcing a code style that we don't like just for the sake of having a tool enforce it does not make much sense. I'm -1 on enforcing the coding style. I'm +1 on making sure that the `meson.build` are formatted with (reasonable) consistency. Most of the proposed changes here are improvements, with the exceptions highlighted above.

FFY00:
I increased the line length to 127, just like we do with Python code, and played with trailing comas to force calls to be expanded, or the opposite, and I think it now looks significantly better.

@dnicolodi, what do you think?

tests/packages/configure-data/meson.build
```diff
@@ -13,8 +13,10 @@ py = py_mod.find_installation()
...
     output: 'configure_data.py',
-    configuration: configuration_data({
-        'MSG': 'hello!',
-    }),
+    configuration: configuration_data(
+        {
+            'MSG': 'hello!',
+        },
+    ),
     install_dir: py.get_install_dir(),
```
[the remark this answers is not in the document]
FFY00: This was forced because we had a trailing comma there. The trick is to not have a trailing comma at all, so that it inlines the dictionary. black works the same way.

I think the reasoning is that when you are gonna add another argument to the function, it will have to change the style significantly by moving the dictionary brackets, so in these cases you simply remove the trailing comma to allow it to  shrink the dictionary definition.

FFY00:
> While we are at it, we may also consider turning instances of
> 
> ```meson
> py_mod = import('python')
> py = py_mod.find_installation()
> ```
> 
> into
> 
> ```
> py = import('python').find_installation()
> ```
> 
> I like the latter much more.

I agree. If we don't need to acess the Python module, that is better. Can you submit a PR after we get this in?

dnicolodi:
This resolves some of the issues I pointed out, but not all. Also, the new one-line style of the `project()` call is worst than the one with the languages on separate lines. Still, pre-commit-hook fails to run the formatter, so there does not seem to be a way to run the formatter in CI. What are you proposing to merge exactly?

FFY00:
That is because the pre-commit action is Docker-based, which pre-commit.ci doesn't seem to support. I'll fix that later, after the new meson-python release.

docs/examples/spam/meson.build
rgommers: I meant it's used by Meson everywhere, and also by us and other initial adopters like SciPy. So "felt" like a recommendation.

> At the end of the day, it doesn't really matter anyway.

Agreed. I'll resolve this comment - doesn't really matter.

rgommers:
I'm also not really a fan of adding a new auto-formatter, it's just one more hurdle to trip over when submitting a PR. We haven't had any real issues with style in `meson.build` files.

Here is what I propose to merge/close this PR: @dnicolodi to update the style changes he liked / agreed with, merging those, and dropping the rest.

dnicolodi:
I submitted #446 to apply uniform coding style to the `meson.build` files.

I applied some manual fixes before running `muon fmt` with this configuration:
```ini
max_line_len = 127
indent_by = '    '
```

As there is not consensus about enforcing the coding style in CI, I'm closing this PR.

PR #371, opened 2023-03-29
BUG: Use meson compile wrapper on Windows

lithomas1:
Necessary if e.g. the user passes requests vsenv from meson setup.
(I guess I could restrict this to only when vsenv is requested in setup, but this needs thought as to how it would work with existing builddirs. I don't remember OTOH if we reconfigure the build dir if it exists already).

With this, pandas should be getting close to greenish again.

I still need to add a test for this, but this should be the right idea.

### DIFF
--- a/docs/how-to-guides/meson-args.rst
+++ b/docs/how-to-guides/meson-args.rst
@@ -44,6 +44,20 @@ building a Python wheel. User options specified via ``pyproject.toml``
 or via Python build front-end config settings override the
 ``meson-python`` defaults.
 
+When building on Windows, ``meson-python`` invokes the ``ninja``
+command via the ``meson compile`` wrapper. When the GCC or the LLVM
+compilers are not found on the ``$PATH``, this activates the Visual
+Studio environment and allows ``ninja`` to use the MSVC compilers. To
+activate the Visual Studio environment unconditionally, pass the
+``--vsenv`` option to ``meson setup``, see this :ref:`example
+<vsenv-example>`. When using the ``meson compile`` wrapper, the user
+supplied options for the compilation command are passed via the
+``--ninja-args`` option. This ensures that the behaviour is
+independent of how the build is initiated. Refer to the `Meson
+documentation`__ for more details.
+
+__ https://mesonbuild.com/Commands.html#backend-specific-arguments
+
 
 Examples
 ========
@@ -152,3 +166,44 @@ To set this option temporarily at build-time:
         .. code-block:: console
 
 	   $ python -m pip wheel . --config-settings=setup-args="-Doptimization=3" .
+
+
+.. _vsenv-example:
+
+Force the use of the MSVC compilers on Windows
+----------------------------------------------
+
+The MSVC compilers are not installed in the ``$PATH``. The Visual
+Studio environment needs to be activated for ``ninja`` to be able to
+use these compilers. This is taken care of by ``meson compile`` but
+only when the GCC compilers or the LLVM compilers are not found on the
+``$PATH``. Passing the ``--vsenv`` option to ``meson setup`` forces
+the activation of the Visual Studio environment and generates an error
+when the activation fails.
+
+This option has no effect on other platforms thus, if your project
+requires to be compiled with MSVC, you can consider to set this option
+permanently in the project's ``pyproject.toml``:
+
+.. code-block:: toml
+
+   [tool.meson-python.args]
+   setup = ['--vsenv']
+
+To set this option temporarily at build-time:
+
+.. tab-set::
+
+    .. tab-item:: pypa/build
+        :sync: key_pypa_build
+
+        .. code-block:: console
+
+	   $ python -m build -Csetup-args="--vsenv" .
+
+    .. tab-item:: pip
+        :sync: key_pip
+
+        .. code-block:: console
+
+	   $ python -m pip wheel . --config-settings=setup-args="--vsenv" .
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: MIT
 
 import ast
-import os.path
+import os
 import platform
 import shutil
 import sys
@@ -219,3 +219,15 @@ def test_invalid_build_dir(package_pure, tmp_path, mocker):
     assert meson.call_args_list[0].args[1][1] == 'setup'
     assert '--reconfigure' not in meson.call_args_list[0].args[1]
     project.build()
+
+
+@pytest.mark.skipif(not os.getenv('CI') or platform.system() != 'Windows', reason='Requires MSVC')
+def test_compiler(venv, package_detect_compiler, tmp_path):
+    # Check that things are setup properly to use the MSVC compiler on
+    # Windows. This effectively means running the compilation step
+    # with 'meson compile' instead of 'ninja' on Windows. Run this
+    # test only on CI where we know that MSVC is available.
+    wheel = mesonpy.build_wheel(tmp_path, {'setup-args': ['--vsenv']})
+    venv.pip('install', os.fspath(tmp_path / wheel))
+    compiler = venv.python('-c', 'import detect_compiler; print(detect_compiler.compiler())').strip()
+    assert compiler == 'msvc'
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -2,6 +2,8 @@
 #
 # SPDX-License-Identifier: MIT
 
+import ast
+import os.path
 import platform
 import shutil
 import sys
@@ -60,12 +62,26 @@ def test_unsupported_python_version(package_unsupported_python_version):
 
 def test_user_args(package_user_args, tmp_path, monkeypatch):
     project_run = mesonpy.Project._run
-    call_args_list = []
+    cmds = []
+    args = []
 
     def wrapper(self, cmd):
         # intercept and filter out test arguments and forward the call
-        call_args_list.append(tuple(cmd))
-        return project_run(self, [x for x in cmd if not x.startswith(('config-', 'cli-'))])
+        if cmd[:2] == ['meson', 'compile']:
+            # when using meson compile instead of ninja directly, the
+            # arguments needs to be unmarshalled from the form used to
+            # pass them to the --ninja-args option
+            assert cmd[-1].startswith('--ninja-args=')
+            cmds.append(cmd[:2])
+            args.append(ast.literal_eval(cmd[-1].split('=')[1]))
+        elif cmd[:1] == ['meson']:
+            cmds.append(cmd[:2])
+            args.append(cmd[2:])
+        else:
+            # direct ninja invocation
+            cmds.append([os.path.basename(cmd[0])])
+            args.append(cmd[1:])
+        return project_run(self, [x for x in cmd if not x.startswith(('config-', 'cli-', '--ninja-args'))])
 
     monkeypatch.setattr(mesonpy.Project, '_run', wrapper)
 
@@ -79,19 +95,32 @@ def test_user_args(package_user_args, tmp_path, monkeypatch):
     mesonpy.build_sdist(tmp_path, config_settings)
     mesonpy.build_wheel(tmp_path, config_settings)
 
-    expected = [
+    # check that the right commands are executed, namely that 'meson
+    # compile' is used on Windows rather than a 'ninja' direct
+    # invocation.
+    assert cmds == [
         # sdist: calls to 'meson setup' and 'meson dist'
-        ('config-setup', 'cli-setup'),
-        ('config-dist', 'cli-dist'),
+        ['meson', 'setup'],
+        ['meson', 'dist'],
         # wheel: calls to 'meson setup', 'meson compile', and 'meson install'
-        ('config-setup', 'cli-setup'),
-        ('config-compile', 'cli-compile'),
-        ('config-install', 'cli-install'),
+        ['meson', 'setup'],
+        ['meson', 'compile'] if platform.system() == 'Windows' else ['ninja'],
+        ['meson', 'install']
     ]
 
-    for expected_args, call_args in zip(expected, call_args_list):
+    # check that the user options are passed to the invoked commands
+    expected = [
+        # sdist: calls to 'meson setup' and 'meson dist'
+        ['config-setup', 'cli-setup'],
+        ['config-dist', 'cli-dist'],
+        # wheel: calls to 'meson setup', 'meson compile', and 'meson install'
+        ['config-setup', 'cli-setup'],
+        ['config-compile', 'cli-compile'],
+        ['config-install', 'cli-install'],
+    ]
+    for expected_args, cmd_args in zip(expected, args):
         for arg in expected_args:
-            assert arg in call_args
+            assert arg in cmd_args
 
 
 @pytest.mark.parametrize('package', ('top-level', 'meson-args'))
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -489,7 +489,10 @@ class _WheelBuilder():
             )
 
     def build(self, directory: Path) -> pathlib.Path:
-        self._project.build()  # ensure project is built
+        # ensure project is built
+        self._project.build()
+        # install the project
+        self._project.install()
 
         wheel_file = pathlib.Path(directory, f'{self.name}.whl')
 
@@ -521,7 +524,8 @@ class _WheelBuilder():
         return wheel_file
 
     def build_editable(self, directory: Path, verbose: bool = False) -> pathlib.Path:
-        self._project.build()  # ensure project is built
+        # ensure project is built
+        self._project.build()
 
         wheel_file = pathlib.Path(directory, f'{self.name}.whl')
 
@@ -534,14 +538,13 @@ class _WheelBuilder():
 
             # install loader module
             loader_module_name = f'_{self.normalized_name.replace(".", "_")}_editable_loader'
-            build_cmd = [self._project._ninja, *self._project._meson_args['compile']]
             whl.writestr(
                 f'{loader_module_name}.py',
                 read_binary('mesonpy', '_editable.py') + textwrap.dedent(f'''
                    install(
                        {self.top_level_modules!r},
                        {os.fspath(self._build_dir)!r},
-                       {build_cmd},
+                       {self._project._build_command!r},
                        {verbose!r},
                    )''').encode('utf-8'))
 
@@ -787,18 +790,30 @@ class Project():
             self._install_plan,
         )
 
-    def build_commands(self) -> Sequence[Sequence[str]]:
+    @property
+    def _build_command(self) -> List[str]:
         assert self._ninja is not None  # help mypy out
-        return (
-            (self._ninja, *self._meson_args['compile'],),
-            ('meson', 'install', '--no-rebuild', '--destdir', os.fspath(self._install_dir), *self._meson_args['install']),
-        )
+        if platform.system() == 'Windows':
+            # On Windows use 'meson compile' to setup the MSVC compiler
+            # environment. Using the --ninja-args option allows to
+            # provide the exact same semantics for the compile arguments
+            # provided by the users.
+            cmd = ['meson', 'compile']
+            args = list(self._meson_args['compile'])
+            if args:
+                cmd.append(f'--ninja-args={args!r}')
+            return cmd
+        return [self._ninja, *self._meson_args['compile']]
 
     @functools.lru_cache(maxsize=None)
     def build(self) -> None:
-        """Trigger the Meson build."""
-        for cmd in self.build_commands():
-            self._run(cmd)
+        """Build the Meson project."""
+        self._run(self._build_command)
+
+    def install(self) -> None:
+        """Install the Meson project."""
+        destdir = os.fspath(self._install_dir)
+        self._run(['meson', 'install', '--no-rebuild', '--destdir', destdir, *self._meson_args['install']])
 
     @classmethod
     @contextlib.contextmanager
--- a/tests/packages/detect-compiler/detect_compiler.c
+++ b/tests/packages/detect-compiler/detect_compiler.c
@@ -1,0 +1,38 @@
+// SPDX-FileCopyrightText: 2023 The meson-python developers
+//
+// SPDX-License-Identifier: MIT
+
+#include <Python.h>
+
+#if defined _MSC_VER
+#  define _COMPILER "msvc"
+#elif defined __clang__
+#  define _COMPILER "clang"
+#elif defined __GNUC__
+#  define _COMPILER "gcc"
+#else
+#  define _COMPILER "unknown"
+#endif
+
+static PyObject* compiler(PyObject* self)
+{
+    return PyUnicode_FromString(_COMPILER);
+}
+
+static PyMethodDef methods[] = {
+    {"compiler", (PyCFunction)compiler, METH_NOARGS, NULL},
+    {NULL, NULL, 0, NULL},
+};
+
+static struct PyModuleDef module = {
+    PyModuleDef_HEAD_INIT,
+    "detect_compiler",
+    NULL,
+    -1,
+    methods,
+};
+
+PyMODINIT_FUNC PyInit_detect_compiler(void)
+{
+    return PyModule_Create(&module);
+}
--- a/tests/packages/detect-compiler/meson.build
+++ b/tests/packages/detect-compiler/meson.build
@@ -1,0 +1,13 @@
+# SPDX-FileCopyrightText: 2023 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+project(
+    'detect-compiler',
+    'c',
+    version: '1.0',
+)
+
+py = import('python').find_installation()
+
+py.extension_module('detect_compiler', 'detect_compiler.c', install: true)
--- a/tests/packages/detect-compiler/pyproject.toml
+++ b/tests/packages/detect-compiler/pyproject.toml
@@ -1,0 +1,7 @@
+# SPDX-FileCopyrightText: 2023 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+[build-system]
+build-backend = 'mesonpy'
+requires = ['meson-python']


rgommers:
Ah yes, now that I see this, I am relying on `--vsenv` too, and it seems like we switched from using `meson compile` to `ninja` sometime after the `0.12.0` release. In https://github.com/PyWavelets/pywt/pull/667 I have:
```
python.exe -m pip install . -v --config-settings=setup-args=--vsenv
```
and that worked quite nicely.

dnicolodi:
Isn't there anywhere a command that can be used like `env` that populates the required environment variables to use MSVC and calls its argument, like `vsenv python.exe -m pip install . -v`? That would be the most obvious solution, and a better UX IMHO, the `--config-settings=setup-args=--vsenv` is not really that obvious or discoverable.

lithomas1:
> Isn't there anywhere a command that can be used like `env` that populates the required environment variables to use MSVC and calls its argument, like `vsenv python.exe -m pip install . -v`? That would be the most obvious solution, and a better UX IMHO, the `--config-settings=setup-args=--vsenv` is not really that obvious or discoverable.

You would really only use the ``--vsenv`` feature in CI(so the UX doesn't matter that much) because typically you wouldn't have the messed up situation with different compilers being/not being on the PATH that Github Actions does.

Also I would argue, having to hardcode the path to vcvarsall.bat or add another dependency to do this for me, is worse than adding just`--config-settings=setup-args=--vsenv`.

lithomas1:
> Isn't there anywhere a command that can be used like `env` that populates the required environment variables to use MSVC and calls its argument, like `vsenv python.exe -m pip install . -v`? That would be the most obvious solution, and a better UX IMHO, the `--config-settings=setup-args=--vsenv` is not really that obvious or discoverable.
> 
> If it does not exist, someone should write it.

Right now the easisest way to do this is
``call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvarsall.bat"``

which doesn't work on bash and also not on powershell.

rgommers:
I agree with @lithomas1 on the UX point, the `--config-settings` way (which will soon be `-C`) is a lot nicer than the MSVC-native stuff with `.bat` script. The `vcvarsall.bat` paths are changing regularly and there aren't any guarantees for where it lives. This is what works in GHA/bash, and it makes me a bit sad:
```bash
          function Invoke-VSDevEnvironment {
            $vswhere = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe"
              $installationPath = & $vswhere -prerelease -legacy -latest -property installationPath
              $Command = Join-Path $installationPath "Common7\Tools\vsdevcmd.bat"
            & "${env:COMSPEC}" /s /c "`"$Command`" -arch=amd64 -no_logo && set" | Foreach-Object {
                  if ($_ -match '^([^=]+)=(.*)') {
                      [System.Environment]::SetEnvironmentVariable($matches[1], $matches[2])
                  }
              }
          }
          Invoke-VSDevEnvironment
```

rgommers:
I'm actually using it in Appveyor, so a GHA action isn't helpful. That said, I haven't looked under the hood to see what `--vsenv` does and if there's something sensible to replace it with. Ideally we find something nicer than cargo culting the above bash function. 

Perhaps adding native support in `meson-python` for a `--use-msvc <optional year/version>` flag or some such thing?

dnicolodi:
If the issue presents itself almost exclusively in CI setups, I don't understand the reason for the reluctance in using a simple external tool to solve the issue in a flexible way. Taking the implementation of `setup_vsenv()` from Meson and trowing it into a stand-alone script that looks up the required environment variables and executes `os.execvpe(sys.argv[1], sys.argv[1:], env)` seems a very good solution to me. And it just requires a `pip install vsenv` or something similar.

dnicolodi:
Here it is https://github.com/dnicolodi/vsenv No documentation and no tests, but it should be doing exactly what I was describing above.

rgommers:
> If the issue presents itself almost exclusively in CI setups, 

I don't think it does? Anyone wanting to use MSVC on Windows may run into this. The Visual Studio activation is run on every subcommand [say the docs](https://mesonbuild.com/Builtin-options.html#details-for-vsenv). The downside of that is timing (from gh-328: _meson compile, 1 second. ninja, 0.005 seconds_), the upside is that things work.

For non-editable builds, the 1 sec. gain is not interesting compared to the UX gains of things not mysteriously breaking when you, for example, run dev commands in a freshly opened Git Bash shell. _However_, I agree that for editable builds where `ninja` gets invoked all the time, the extra 1 sec. is annoying. Perhaps we can get the best of both worlds here by documenting that if folks want to use MSVC on Windows together with editable builds, they should do separate env activation in their shell themselves - and we give them a tool for that like the `vsenv` repo above. WDYT?

dnicolodi:
The fact that the issue is mostly encountered in CI context was reported as a justification for not having an optimal UX. I just repeated the statement. If this is something that users need to use, the `--config-settings=setup-args=--vsenv` incantation seems way too magical. Having different compilation commands depending on the platform, on which options are provided by the user, or on whether a regular wheel or an editable wheel is being built, is also too magical, too complex, fragile, and potentially source of inconsistencies.

Documenting that the MSVC compiler needs to be in the PATH (when other compiler suites are found on the PATH) is the correct solution. I already opened a bug about this a while ago, see #224.

Note that the `vsenv` tool I hacked together does not set the environment for the current shell. It is a wrapper that sets up the required environment variables and invokes the command specified as an argument, like `env` does.

I'm really not that keen to compromises motivated to fixup basic compilation environment setup requirements. If you want to run a compiler, the compiler needs to be found on the PATH. If your compiler vendor makes it ridiculously hard to have that setup properly, please go complain to your compiler or OS vendor.

dnicolodi:
Sorry, I misunderstood. I thought that the issue was only present with more than one compiler installer on the system. However, obviously, the issue is also when only MSVS is present as there is nothing that setups the environment also in this case when meson-python invoked `ninja`. To answer your question: this is indeed a regression compared to meson-python 0.12.0 as calling `meson setup build && ninja -C build` does not find the compiler, `meson setup build && meson compile -C build` works.

I still think that having the compilers in the PATH is a more than reasonable solution, and that vendors should be pressured to provide a solution to easily do that. However, indeed, I don't see this happening any time soon. It would be great if Windows and Visual Studio users would make their license money worth and demand things to work reasonably, but I have the impression that open source maintainers are easier to reach that the Microsoft technical support.

Let's revert back to use `meson compile` for compilation. Should we do this always or only on Windows? Have everyone pay the price of longer execution times sucks, but I really don't like having different behaviors for different systems. It is also a pain to document: every time we mention the compilation command we need to say that it is different on Windows and on other systems.

rgommers:
> Have everyone pay the price of longer execution times sucks, but I really don't like having different behaviors for different systems.

Yeah, there's a tradeoff between performance and complexity here. I don't have a clear preference - I think I agree with what you propose here, but I could see us going another way later perhaps. The third option is to default to `meson compile` everywhere, and let users opt in to skipping that and calling `ninja` instead. Also extra complexity, so may only worth doing later if it turns out to be a pain point?

One other thing I'd like to do is add a test with code that is MSVC-specific and only runs on Windows with MSVC installed. That way we're dogfooding whatever we do here.

dnicolodi:
I'm not sure I proposed a solution, I think I have more questions than answers on this issue. IMHO the reasonable options are:

1. always use `meson compile`
2. use `meson compile` on Windows and `ninja` everywhere else
3. use `meson compile` and allow the user to opt in to use `ninja`

I think 2. is the less bad of the options. I think the execution time penalty for editable installations on systems other than Windows is too large to pay, and having a different compilation command for editable and regular builds is much more surprising than a difference based on the platform.

I'll polish @lithomas1 patch and add a documentation patch on top.

rgommers:
Ah okay, I misunderstood what you meant. Yes, I think (2) seems like the way to go at the moment, +1 for that.

dnicolodi:
Another possible solution came to mind:

4. integrate code similar to `vsenv` linked above into meson-python and use it to setup the environment before calling `ninja`.

Drawbacks of this approach are that this is a little bit more code that needs to be maintained in meson-python and the code would have to de duplicated in every editable package loader, although a hundred lines of code do not really seem to be that significant in the broad picture.

rgommers:
@FFY00 we're already on the same page about this one - we all agree now it's a breaking change and we'll fix it before the release.

FFY00:
Okay, so, here's my opinion. Let's roll back the `meson compile` > `ninja` change globally, and open a new issue to discuss strategies to possibly get rid of the `meson compile` penalty on editable installs.

@lithomas1 could you update the PR to use `meson compile` all the time, instead keeping `ninja` and only selecting `meson compile` on Windows?

This is a release blocker, so I'd like to get this solved soon, so that we can finally have a stable release with editable installs support. And thank you very much for reporting the issue and opening the PR!

FFY00:
I opened #373 to track the editable installs speedup. Let's release with `meson compile` and then look at that and see what we can do to improve it more calmly.

FFY00:
Sorry, I misread the comments.

FFY00:
I'll probably be out until Monday. I am happy to go with whatever opinion y'all want. The only worry is the `compile-args`/`tool.meson-python.args.compile` semantics if we go with compile invocations. If you can define how that will work, I'll be happy to take a look monday. Thanks.

FFY00:
> @rgommers @lithomas1 and I already agreed on the way forward. Why is that decision being disregarded?

Just to clarify, I thought Ralf was agreeing, so I only very quickly skimmed the history above. My point with that message was to provide the steps forward, so that we could sort this as soon as possible, not to disregard any of the other maintainers. My "opinion" was on how to move forward, as I thought we wanted to use `meson compile` globally. I apologize for the misunderstanding.

eli-schwartz:
A couple points:
- I am "recommendation from the Meson maintainers", but to clarify, I do agree that you need meson compile if and only if you are using MSVC, which in terms of actual detection means "using Windows which might mean the detected compiler is MSVC"
- I advocate for option 2, use `meson compile` on Windows, and pay the performance price to ensure that things actually work
- "Alternatively, we could help with optimizing `meson compile` import times" -- this is a long-term goal of mine not limited to meson compile, I have a series of very fiddly patches to delay imports of e.g. the dependencies subsystem... this is a follow-up for my previous work on delayed imports for the compilers subsystem (dozens of files and ten thousand lines of code no longer imported at startup for *any* command)
- ultimately the real issue here is https://github.com/ninja-build/ninja/issues/1610 because this functionality should be supported by ninja so that people don't need to run slow wrapper scripts and it becomes impossible for people to accidentally do the wrong thing
- on the compile-args topic, you can invoke `meson compile --ninja-args ...` and have the raw args forwarded to ninja

dnicolodi:
> on the compile-args topic, you can invoke `meson compile --ninja-args ...` and have the raw args forwarded to ninja

I don't think this is necessary. The only arguments that makes sense to pass to `meson compile` are understood by `ninja` too (IIUC the former has been modeled on the command line interface of the latter). This is the reason why I don't think it is problematic to have different commands on different platforms. The only exception are arguments selecting the targets to build, which are interpreted differently by `meson compile` and `ninja`, but I don't know how this is relevant to out use case.

rgommers:
> The only arguments that makes sense to pass to `meson compile` are understood by `ninja` too (IIUC the former has been modeled on the command line interface of the latter). This is the reason why I don't think it is problematic to have different commands on different platforms.

There's a couple of random ones that don't seem supported, like `ninja`'s `-w` to tweak warnings. The CLI flags that are common do have the same meaning, but it seems safer to use `meson compile --ninja-args <users-compile-args>` on Windows, it should be slightly more robust, and doesn't seem to have a downside (right?). And then we can safely document that the `compile-args` input to `meson-python` is always for `ninja`.

docs/how-to-guides/meson-args.rst
```diff
@@ -44,6 +44,16 @@ building Python wheel. User options specified via ``pyproject.toml``
 or via Python build front-end config settings can be used to override
 the ``meson-python`` defaults..
 
+When building on Windows, ``meson-python`` invokes the ``ninja``
+command via the ``meson compile`` wrapper. This activates the Visual
+Studio environment and allows ``ninja`` to use the MSVC compilers. The
```
lithomas1, line 49: This is not necessarily true, it depends on what compiler meson selected at setup time (which depends compilers you have in the PATH.).
```suggestion
Studio environment (if that was the compiler meson selected at setup time) and allows ``ninja`` to find the MSVC compilers. The
```

dnicolodi: Thank for the review. I was trying to keep the explanation as simple as possible. I really wanted to refer to the Meson documentation for the details of the compilers discovery, but I haven't found the section of the Meson manual where it is described. I thought hat "allows ninja to use the MSVC compilers" didn't imply that these are not necessarily the ones that will be used. Maybe adding more detail is required. Let me try again.

dnicolodi: I've extended the documentation. Please take a look.

docs/how-to-guides/meson-args.rst
```diff
@@ -44,6 +44,21 @@ building Python wheel. User options specified via ``pyproject.toml``
 or via Python build front-end config settings can be used to override
 the ``meson-python`` defaults..
 
+When building on Windows, ``meson-python`` invokes the ``ninja``
+command via the ``meson compile`` wrapper. This activates the Visual
```
lithomas1, line 48: Can you delete the line ``This activates the Visual Studio environment ...``?

This contradicts the sentence right after, which should be the more accurate one (since activation of VS env is not guaranteed).

dnicolodi: Sure. It should be better now. PTAL

docs/how-to-guides/meson-args.rst
```diff
@@ -44,6 +44,20 @@ building Python wheel. User options specified via ``pyproject.toml``
 or via Python build front-end config settings can be used to override
 the ``meson-python`` defaults..
 
+When building on Windows, ``meson-python`` invokes the ``ninja``
+command via the ``meson compile`` wrapper. When the GCC or the LLVM
+compilers are not found on the ``$PATH``, this activates the Visual
+Studio environment and allows ``ninja`` to use the MSVC compilers. To
+activate the Visual Studio environment unconditionally, pass the
+``--vsenv`` option to ``meson setup``, see this :ref:`example
```
lithomas1, line 52: We should mention ``--setup-args`` here instead of ``meson setup``.

lithomas1 (commented):
Thanks, this generally LGTM.

Can you add a test for ``vsenv``? I don't think MSVC is being activated currently in the CI.

Ideally, maybe we can have a test C file, where we have an ifdef macro(maybe `_MSC_VER`) to check if we're being compiled by MSVC, and have something that fails to compile otherwise.

(The tricky part is probably skipping the test if MSVC is not installed)

docs/how-to-guides/meson-args.rst
dnicolodi: What's `--setup-args`? How to pass arguments to `meson setup` is described two paragraph above and repeated in the linked example. I don't think it is necessary to repeat it.

dnicolodi:
> Can you add a test for `vsenv`? I don't think MSVC is being activated currently in the CI.

Inded, we don't use MSVC in the CI. The tests are written to be happy with whatever compiler.

What so you mean with "a test for `vsenv`"? Checking if passing `--vsenv` does the right thing means checking if we get the MSVC compiler when the test runs. But from within the test we cannot know if MSVC is not found because it is not installed, or because `--vsenv` does not work. Also, a test for `--vsenv` better belongs to the Meson test suite. We already check that we propagate the options specified by the user to the `meson setup` command.

docs/how-to-guides/meson-args.rst
lithomas1: I was referring to the cmdline version of ``[tool.meson-python.args]``

Users don't have direct access to ``meson setup`` through meson-python, so I think mentioning it confusing.

Maybe we can reword the previous sentence as ``pass the --vsenv option as a setup argument`` followed by the link to the example.

dnicolodi: The diff does not make it obvious, but the page that being amended is title "Passing arguments to Meson". The man topic of the page is how to pass arguments to `meson setup`. It is explained in details two paragraphs above.

lithomas1: I don't have a strong opinion on this, so if you really feel like the current approach is fine, I'm OK I guess.

lithomas1:
> > Can you add a test for `vsenv`? I don't think MSVC is being activated currently in the CI.
> 
> Inded, we don't use MSVC in the CI. The tests are written to be happy with whatever compiler.
> 
> What so you mean with "a test for `vsenv`"? Checking if passing `--vsenv` does the right thing means checking if we get the MSVC compiler when the test runs. But from within the test we cannot know if MSVC is not found because it is not installed, or because `--vsenv` does not work. Also, a test for `--vsenv` better belongs to the Meson test suite. We already check that we propagate the options specified by the user to the `meson setup` command.

Maybe we can test instead for ``meson compile`` being used on Windows?

docs/how-to-guides/meson-args.rst
dnicolodi: What needs to be done is to pass an argument to `meson setup`, which is the consumer of the argument. The command line config setting or the `pyproject.toml` setting are the way you use to specify the argument when using `meson-python`. The way I would word this would be "pass the `--vsenv` option to `meson setup` via the `setup-args` config setting or the `tool.meson-python.meson.args.setup` setting in `pyproject.toml`." But this is just a repetition of what is written ten lines above.

dnicolodi:
> Maybe we can test instead for `meson compile` being used on Windows?

Our tests build a few packages with compiled code and we run the tests on Windows too. But we don't require MSVC. I'm not sure that we should make MSVC an hard dependency to run the tests. How do you know if MSVC is installed on Windows?

eli-schwartz:
You could check if `cc.get_id() == 'msvc'` and conditionally define the test in meson.build (or do something like meson's own `error('MESON_SKIP_TEST: reason')`).

This will tell you if `meson setup` wanted MSVC, which is what you want to know here.

dnicolodi:
Thanks @eli-schwartz. We don't run the meson-python tests as Meson `test()` instances. We could hook that up as a test for a package in the meson-python test suite, but the result would be a bit of a tautological test: to enable the test that checks whether the Visual Studio activation works we need to the Vistual Studio activation to work, otherwise `cc.get_id()` is not going to return `msvc`. It is not clear what @lithomas1 is proposing to test exactly.

lithomas1:
I think I can settle for the second option I mentioned, which is just to check that meson-python invokes ``meson compile`` on Windows. This does not depend on MSVC.

rgommers:
Using a package containing
```meson
if not cc.get_id() == 'msvc'
  error('This test is only supposed to run when MESON_PYTHON_TEST_MSVC is defined)
endif
```
and in the test file a `skipif(os.environ.get('MESON_PYTHON_TEST_MSVC') is None)` should work. Then setting that env var in a dedicated Windows CI job. 

That shouldn't have to be part of this PR though. But it forces us to write that CI job content to force the use of MSVC, which is a little painful right now. So once we see that in our own test suite, we can figure out how to simplify it.

dnicolodi:
@rgommers This makes the detection of whether MSVC is available or not a manual one: I set the env var only if I know MSVC is installed. We can do that in our CI. But I'm not sure that meson-python is where we need to test that Meson Visual Studio activation works as advertised. I think Meson is the best place where to do that. Maybe it makes sense as an integration test. I can add it, if the setup is not too cumbersome.

eli-schwartz:
As far as I understand, you're not actually worried about whether meson-python can set up a build using MSVC (at the `meson setup` stage), you're just worried about whether meson-python guarantees that, when a build uses MSVC, it will execute correctly (at the ninja stage) without errors from the operating system stating that it can't find a cl.exe binary.

So maybe it's okay to just trust meson.build to report via crafted message strings that the test should be skipped?

rgommers:
Indeed, it is intended to only be run in our CI, or "manual" locally in case one is working on that CI job. It does serve as a regression test for the problem that this PR addresses.

Agreed that that doesn't remove the need to do something in Meson itself, because that's where the actual code to do the activation lives. 

> you're just worried about whether meson-python guarantees that, when a build uses MSVC, it will execute correctly (at the ninja stage) without errors from the operating system stating that it can't find a cl.exe binary.

pretty much this

dnicolodi:
I've added a test checking that meson-python invokes `meson compile` on Windows. I've also added a more convoluted test that verifies that MSVC is actually used for the compilation when available. It is setup to run only on the CI and only on Windows as I don't see a reason to run it elsewhere. Thanks everyone for the suggestions about how to implement it. PTAL

dnicolodi:
@eli-schwartz I don't have a good idea on why the last commit is needed. The only thing I can think about is that somehow setting up the MSVC compiler sets an environment variable that affects the Python introspection data used by the Meson's python module to determine which python DLL to link with. But this seems very unlikely.

eli-schwartz:
https://github.com/mesonbuild/meson/blob/71487c876e6f0046cbb09fce23b09d8a38a8224f/mesonbuild/dependencies/python.py#L255

I can't see where the debug lib for python is being handled at all. In fact I thought that python either builds with debug enabled or not, but doesn't distribute both at the same time.

dnicolodi:
Indeed the debug lib is not found and the build fails. The build log for the failure is here https://github.com/mesonbuild/meson-python/actions/runs/4590489619/jobs/8106066122#step:9:160

dnicolodi:
Also, we already have `-Ddebug=false` thus as far as I understand, `-Dbuildtype=plain` should not be doing much.

dnicolodi:
Sorry, that is an unrelated test failure. This is the one when linking fails https://github.com/mesonbuild/meson-python/actions/runs/4590440580/jobs/8105987201#step:9:159

dnicolodi:
Looking at the link command and to the failure I'm even more puzzled:
```
"link"  /MACHINE:x64 /OUT:detect_compiler.cp311-win_amd64.pyd \
  detect_compiler.cp311-win_amd64.pyd.p/detect_compiler.c.obj \
  "/nologo" "/release" "/nologo" "/DLL" "/IMPLIB:detect_compiler.cp311-win_amd64.lib" \
  "C:\hostedtoolcache\windows\Python\3.11.2\x64\libs\python311.lib" \
  "kernel32.lib" "user32.lib" "gdi32.lib" "winspool.lib" "shell32.lib" "ole32.lib" \
  "oleaut32.lib" "uuid.lib" "comdlg32.lib" "advapi32.lib"

LINK : fatal error LNK1104: cannot open file 'python311_d.lib'
```
newlines added for readability.

Where is `python311_d.lib` coming from?

eli-schwartz:
Weirdly you can see that the command line which the linker is invoked with mentions python311.lib but then it errors out saying it cannot find python311_d.lib -- I don't understand how the linker decided on its own to do that without a command line argument telling it to...

dnicolodi:
Indeed. The funny thing is that it seem to depend on the optimization level: that should be the only difference between using `-Ddebug=false` and `-Ddebug=false -Dbuildtype=plain`.

dnicolodi:
One more data point: `-Ddebug=false -Doptimization=plain` still results in the link failure.

dnicolodi:
Either I've been staring at this too long or there is something unusual going on: the link command that fails and the one that succeeds are identical.

dnicolodi:
Found it! The issue does not depend on how the Visual Studio environment gets activated. It really depends on the buildtype option. I think there is still some confusion in the interplay of the debug, optimization, and buildtype options on the Meson side. The issue is caused by the compiler options used for `-Dbuildtype=debug` and for `-Dbuildtype=plain`. The fact that the compiler options cause a linker error made the failure look magical.

After removing the common options, the default `-Dbuildtype=debug` uses these compiler options: `"/MDd" "/O2"`, `-Dbuildtype=plain` uses instead: `"-DNDEBUG"`.

Clearly the `"/MDd"` option is the issue. I'm not entirely sure whether this behavior is correct or not.

eli-schwartz:
From https://mesonbuild.com/Builtin-options.html#base-options

> The default value of `b_vscrt` is `from_buildtype`.

That's what sets /MDd.

dnicolodi:
Thank, I was not aware of that option, but I don't know much of how compilation works on Windows either, and I find all this surprising. I really don't know what the best way forward is.

dnicolodi:
One thing that I don't understand is that `buildtype=plain` does not specify any runt time library flag. From reading about it here https://learn.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-170 I would expect that at least one of that flags needs to be specified.

dnicolodi:
It is also surprising that `buildtype=debugoptimized` does not use the debug version of the runtime library
https://github.com/mesonbuild/meson/blob/71487c876e6f0046cbb09fce23b09d8a38a8224f/mesonbuild/compilers/mixins/visualstudio.py#L381-L382

dnicolodi:
With #383 merged, this can be merged without the workaround for the linking problem.

dnicolodi:
Rebased once more. @rgommers is there any reason for not merging this?

rgommers:
> Rebased once more. @rgommers is there any reason for not merging this?

I don't think so - I'll make time for a final review(/merge) later today.

rgommers (approved):
The tests pass and this all looks quite good to me. Just a few very minor things I noticed (docs, typos). @dnicolodi please have a look and feel free to merge when those are addressed or if I'm wrong about them.

docs/how-to-guides/meson-args.rst
```diff
@@ -44,6 +44,20 @@ building a Python wheel. User options specified via ``pyproject.toml``
 or via Python build front-end config settings override the
 ``meson-python`` defaults.
 
+When building on Windows, ``meson-python`` invokes the ``ninja``
+command via the ``meson compile`` wrapper. When the GCC or the LLVM
```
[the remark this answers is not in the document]
eli-schwartz: In terms of the code implementation, we quite literally check for `shutil.which` with any of cc, gcc, clang, clang-cl. The remaining autodetected default compilers for C are icl and pgcc, but icl ships with its own cl.exe (super "helpful" as noted by the code comments in detect.py) and meson will prefer cl.exe over pgcc anyway, if it can.

tests/test_project.py
```diff
@@ -60,12 +62,26 @@ def test_unsupported_python_version(package_unsupported_python_version):
...
         # intercept and filter out test arguments and forward the call
-        call_args_list.append(tuple(cmd))
-        return project_run(self, [x for x in cmd if not x.startswith(('config-', 'cli-'))])
+        if cmd[:2] == ['meson', 'compile']:
+            # when using meson compile instead on ninja directly, the
+            # arguments needs to be unmarshalled from the form used to
+            # pass them to the --ninja-args option
+            assert cmd[-1].startswith('--ninja-args=')
+            cmds.append(cmd[:2])
+            args.append(ast.literal_eval(cmd[-1].split('=')[1]))
```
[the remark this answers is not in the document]
dnicolodi: What are you referring to? The not-so-nice part is marshaling the arguments into the format expected by Meson. Parsing them back is just a consequence of that. However, the code for either does not look that horrible.

docs/how-to-guides/meson-args.rst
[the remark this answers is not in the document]
dnicolodi: I haven't found detailed documentation for the compiler auto detection. I had a look at the code, which behaves as @eli-schwartz describes https://github.com/mesonbuild/meson/blob/eb472a133f45826a246b40c3d775eebf098fd6b1/mesonbuild/utils/vsenv.py#L43-L51 I don't know how much of this is appropriate to document on the meson-python side, given that it is not explicitly documented and that I don't know how much of it is susceptible of change in Meson. I thought that GCC and LLVM covered the most popular cased and anyone using a less common compiler would know what they are doing.

PR #377, opened 2023-04-02
ENH: detect packages installed both in purelib and platlib

dnicolodi:
I've tried to make the error message concise but informative. I'm not sure I succeeded.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -150,14 +150,28 @@ _INSTALLATION_PATH_MAP = {
 
 def _map_to_wheel(sources: Dict[str, Dict[str, Any]]) -> DefaultDict[str, List[Tuple[pathlib.Path, str]]]:
     """Map files to the wheel, organized by wheel installation directrory."""
-    wheel_files = collections.defaultdict(list)
+    wheel_files: DefaultDict[str, List[Tuple[pathlib.Path, str]]] = collections.defaultdict(list)
+    packages: Dict[str, str] = {}
+
     for group in sources.values():
         for src, target in group.items():
             destination = pathlib.Path(target['destination'])
             anchor = destination.parts[0]
+
             path = _INSTALLATION_PATH_MAP.get(anchor)
             if path is None:
                 raise BuildError(f'Could not map installation path to an equivalent wheel directory: {str(destination)!r}')
+
+            if path == 'purelib' or path == 'platlib':
+                package = destination.parts[1]
+                other = packages.setdefault(package, path)
+                if other != path:
+                    this = os.fspath(pathlib.Path(path, *destination.parts[1:]))
+                    that = os.fspath(other / next(d for d, s in wheel_files[other] if d.parts[0] == destination.parts[1]))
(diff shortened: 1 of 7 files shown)


dnicolodi:
Funny, this detected some instances of this issue in our test suite.

mesonpy/__init__.py
```diff
@@ -150,14 +150,26 @@ def _init_colors() -> Dict[str, str]:
...
-                raise BuildError(f'Could not map installation path to an equivalent wheel directory: {destination!r}')
+                raise BuildError(f'Could not map installation path to an equivalent wheel directory: {str(destination)!r}')
+
+            if path == 'purelib' or path == 'platlib':
+                package = destination.parts[1]
+                other = packages.setdefault(package, path)
+                if other != path:
+                    this = os.fspath(pathlib.Path(path, *destination.parts[1:]))
+                    that = os.fspath(other / next(d for d, s in wheel_files[other] if d.parts[0] == destination.parts[1]))
+                    raise BuildError(f'The {package} package is split between {path} and {other}: {this!r} and {that!r}')
```
rgommers, line 171: I think the reason is always going to be a missing `pure: false`, so maybe hint at the suggestion by appending:
```
, you may be missing a `pure: false` keyword.
```
?

dnicolodi: Or there is a `pure: false` too many. However, indeed, the most common case I believe is for a `pure: false` missing. I didn't add more information to the error message because it is already pretty long as it is and I don't know how many users would read past the part before the colons. Adding the hint however does not have any contraindication other than most likely making the error message to wrap around.

dnicolodi: I went for a slightly different wording and I used double quotes instead of slanted quotes because that is what we use for literal that are not file paths in other error messages. I also didn't add the full stop at the end, because the other error messages don't have it either. Maybe it should be added everywhere.

dnicolodi:
This does not catch the case in which an extension module and a python package with the same name are installed, or a package and a module with the same name are installed. However, both result in an obviously broken wheel, thus I don't think it is worth having complex detection code for these cases. Mixing `platlib` and `purelib` on the other hand works almost everywhere, thus is not easy to catch, but it is undefined behavior and should be avoided.

mesonpy/__init__.py
rgommers: LGTM. I don't mind the full stop or quotes either way, whatever seems best (probably easiest to ignore for now).

rgommers (approved):
LGTM and seems to work as advertised. This is a nice usability improvement.

ngoldbaum:
> I've tried to make the error message concise but informative. I'm not sure I succeeded.

I ran into the new error today and I was a little confused. It may help if you say that the `pure: false` needs to go into the `py.install_sources` call (or at least that's where I put it to fix this).

rgommers:
Thanks for the suggestion @ngoldbaum. 

The best place to put `pure: false` is in:
```meson
py = import('python').find_installation(pure: false)
```

PR #379, opened 2023-04-04
TST: add test demonstrating the ndebug issue

dnicolodi:
The meson-python `-Db_ndebug=if-release` default option does not have effect unless the project is also configured with `-Dbuildtype=release` or `default_options: ['buildtype=release']` is specified in `project()`. I believe this was not the desired effect.

### DIFF
--- a/tests/test_default_options.py
+++ b/tests/test_default_options.py
@@ -1,0 +1,15 @@
+import os
+import subprocess
+
+import pytest
+
+import mesonpy
+
+
+@pytest.mark.parametrize('args', [[], ['-Dbuildtype=release']], ids=['defaults', '-Dbuildtype=release'])
+def test_ndebug(package_scipy_like, tmp_path, args):
+    with mesonpy._project({'setup-args': args}) as project:
+        command = subprocess.run(
+            ['ninja', '-C', os.fspath(project._build_dir), '-t', 'commands', '../../mypkg/extmod.c^'],
+            stdout=subprocess.PIPE, check=True).stdout
+        assert b'-DNDEBUG' in command


tests/test_default_options.py
```diff
@@ -0,0 +1,15 @@
...
+import pytest
+
+import mesonpy
+
+
+@pytest.mark.parametrize('args', [[], ['-Dbuildtype=release']], ids=['defaults', '-Dbuildtype=release'])
+def test_ndebug(package_scipy_like, tmp_path, args):
+    with mesonpy._project({'setup-args': args}) as project:
+        command = subprocess.run(
+            ['ninja', '-C', os.fspath(project._build_dir), '-t', 'commands', '../../mypkg/extmod.c^'],
```
rgommers, line 13: The `^` is a typo I think?

rgommers (commented):
Thanks for adding this test. Looks like we should settle the `buildtype` change discussion in gh-381, and then this will be mergeable.

tests/test_default_options.py
dnicolodi: The `^` asks `ninja` to report the build command for the target having the file preceding the `^` as part of its inputs. See the second paragraph here https://ninja-build.org/manual.html#_running_ninja

rgommers: Ah interesting, didn't know that.

dnicolodi:
I didn't meant this to be merged but only as a reproducer of the issue I saw. It would need some adjustment if we want to merge it.

rgommers:
This can be closed now that gh-383 is merged I assume?

dnicolodi:
The purpose of #383 is also to fix this, thus this can be closed. However, there isn't a test that makes sure the `-DNDEBUG` thing does not break in the future.

PR #383, opened 2023-04-06
ENH: adjust default build options

### LINKED ISSUE #381: Defaults for Meson build options, linking issues with MSVC
Meson native default build options set `buildtype=debug` https://mesonbuild.com/Builtin-options.html#core-options This implies `debug=true`, `optimization=0` https://mesonbuild.com/Builtin-options.html#details-for-buildtype

meson-python sets `debug=false`, `optimization=2`. This was understood to result in a "release" kind of build. Note that `buildtype=release` would set `optimization=3` thus is not exactly the same. Also note that, despite what the Meson documentation says, setting `debug` and `optimization` separately has no effect on the `buildtype` configured for the build.

However, these settings do not result in the `buildtype` to change. Therefore, other build options whose values are derived from `buildtype` are set as for a `debug` build.

In particular, the default value for the `b_vscrt` build option https://mesonbuild.com/Builtin-options.html#base-options in particular is `from_buildtype` which means that on a `debug` build the MSVC compiler will link with the debug version of the VS runtime library, which in turn will make the linker look for the debug build of all the linked DLLs. This is problematic because the regular Python distribution for Windows does not contain a debug version of `python.dll` and linking fails. This was the reason for build issues in #371 and most likely for other build issue reported in unrelated issues.

This also result in the `b_ndebug` option that meson-python sets to `if-release` to do not work as intended (or, at least, not as I understood it was intended to work): `buildtype` different from `release` implies `b_ndebug=false` independent of the value assigned to the `debug` option. This is demonstrated in #379.

There may be other behavior of Meson that depends on the value assigned to `buildtype` regardless to the value assigned to the separate `debug` and `optimization` options.

While the `-DNDEBUG` issue is not critical and other consequences of having `buildtype=debug` may at worst be not-so-well optimized builds, the `b_vscrt` issue results in build failures and should be addressed somehow.

If we focus the attention to the `b_vscrt` option only, there are a few possible solutions: 

1. Do not change the build options and document that the default Meson options need to be adjusted by package authors to avoid link issue with the most common distribution of Python on Windows when compiling with MSVC.
2. Adjust the `buildtype` to something different from `debug` so that `b_vscrt` gets the correct value. This incurs in the minor issue that Meson emits a warning when `debug` or `optimization` in combination with `buildtype`.
3. adjust the value of `b_vscrt` to do the right thing regardless of the value of `buildtype`.

I don't think that (1) goes in the spirit of the defaults adopted for meson-python so far.

Solution (2) seems a bit heavy handed, but for most things there is no much of a practical difference between setting `debug=false`, `optimization=2` and setting `buildtype=release`. Another option could be to set `buildtype=debugoptimized`, `debug=false`, but this still presents the `b_ndebug` issue and having a debug build without debug symbols seems paradoxical.

Because setting the from VS runtime library compilation flags results in a link failure on most systems, maybe we should do (3) regardless. However, I'm not sure about what the right value for that option would be: does Python require a dynamically or statically linked VS runtime?

dnicolodi:
Switch from debug=false, optimization=2 to buildtype=release. Despite what the Meson documentation says, the former does not change the default buildtype=debug, which undesired effects on other settings derived from the value of the buildtype option.

Whit the default b_vscrt=from_buildtype option, Meson instructs the MSVC compiler to use the debug version of the VS runtime library when buildtype=debug. This causes the linker look for the debug build of all the linked DLLs. The Python distribution for Windows does not contain a debug version of the python.dll and linking fails. To avoid this issue when the user explicitly asks for a debug build, also set b_vscrt=mt.

The b_ndebug=if-release option set by meson-python also looks at the value of the buildtype and not to the value of the debug options. Setting buildtype=release has the desired effect of disabling assertions.

The prefix, python.purelibdir, and python.platlibdir build options are no more necessary after the heuristic for determining the location where installed files need to be placer from their installation location has been removed in #280. Remove them.

Fixes #381.


dnicolodi:
I'm updating the documentation. I wanted to check that this does not break anything before finishing the documentation.

dnicolodi:
Documentation updated. I still feel that the "Default build options" page belongs to "Reference" and not to "Explanations". I also feels that splitting the documentation is these sections is silly, but this is a longer story...

dnicolodi:
Running the tests with MSVC bumped into a bug in Meson https://github.com/mesonbuild/meson/pull/11655. I added a work around.

.github/workflows/tests.yml
```diff
@@ -104,6 +104,44 @@ jobs:
...
+      - name: Set up target Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python }}
+
+      - name: Install Ninja
+        run: python -m pip --disable-pip-version-check install ninja
+
+      - name: Setup MSVC
+        uses: bus1/cabuild/action/msdevshell@v1
```
rgommers, line 130: Since this seems to be from a GitHub user that we don't necessarily want to trust, I suggest changing this to:
```
uses: bus1/cabuild/action/msdevshell@e22aba57d6e74891d059d66501b6b5aed8123c4d  # v1
```
following https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

mesonpy/__init__.py
```diff
@@ -800,18 +784,11 @@ def _wheel_builder(self) -> _WheelBuilder:
...
             (self._ninja, *self._meson_args['compile'],),
-            (
-                'meson',
-                'install',
-                '--only-changed',
-                '--destdir',
-                os.fspath(install_dir or self._install_dir),
-                *self._meson_args['install'],
-            ),
+            ('meson', 'install', '--no-rebuild', '--destdir', os.fspath(self._install_dir), *self._meson_args['install']),
```
rgommers, line 791: Why is `--only-changed` dropped here?

mesonpy/__init__.py
```diff
@@ -742,26 +737,15 @@ def _run(self, cmd: Sequence[str]) -> None:
...
             '-Db_ndebug=if-release',
-            '-Doptimization=2',
-
-            # XXX: This should not be needed, but Meson is using the wrong paths
-            #      in some scenarios, like on macOS.
-            #      https://github.com/mesonbuild/meson-python/pull/87#discussion_r1047041306
-            '--python.purelibdir',
-            sys_paths['purelib'],
-            '--python.platlibdir',
-            sys_paths['platlib'],
```
rgommers, line 762: I'll note that the change to `--destdir` and these options no longer being needed was discussed in https://github.com/mesonbuild/meson-python/pull/338#discussion_r1135271706. Looks like we agreed on that, and it has a good rationale.

rgommers (commented):
This looks pretty good, thanks @dnicolodi. One suggestion for security, and one question on `--only-changed`.

.github/workflows/tests.yml
dnicolodi: The user in question is one of the maintainers of dbus-broker, among other things, and a known figure in the open source world. I trust him much more than, say, codecov. Howrver, I'll explicitly pin the action to the git revision.

mesonpy/__init__.py
dnicolodi: It is explained in the commit message. `destdir` is always an empty temporary directory, thus `--only-changed` doesn't do anything at best, most likely it at at least requires an extra `stat()` system call for each file copied.

dnicolodi: I extended the commit message with some more background.

rgommers: Ah yes, the thing that changed is that this command is no longer run by an editable install. That's why `--only-changed` was needed before. Resolving this comment.

.github/workflows/tests.yml
rgommers: Thanks. Yep, I'm also not a fan of codecov. I think trusting https://github.com/actions/ makes sense (and is more convenient than pinning), maybe a few other well-known orgs, and the rest needs pinning.

dnicolodi: Makes sense. It is pinned now.

rgommers (approved):
LGTM, let's give this a go! 

The added MSVC CI job is nice. There's no test that _only_ passes with MSVC, so this may in principle pass also if MSVC activation fails - but that's being dealt with in gh-371 already.

PR #384, opened 2023-04-09
ENH: specify meson-python native file as last option to meson setup

### LINKED ISSUE #330: Verify user arguments

dnicolodi:
This ensure that the python interpreter path specified in the meson-python native file overrides any other provided in user supplied native files.

Fixes #330.

### DIFF
--- a/docs/explanations/default-options.rst
+++ b/docs/explanations/default-options.rst
@@ -50,9 +50,10 @@ The options that ``meson-python`` specifies by default are:
 
    Additional ``--native-file`` options can be passed to ``meson
    setup`` if further adjustments to the native environment need to be
-   made. Meson will merge the contents of all machine files. The
-   user-provided machine files must not override the path for the
-   ``python`` binary.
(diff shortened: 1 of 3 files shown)


rgommers (approved):
This LGTM, thanks @dnicolodi.

I'll xref https://github.com/mesonbuild/meson-python/issues/330#issuecomment-1449730377, where it was discussed/agreed that this is the desired behavior.

PR #386, opened 2023-04-11
MAINT: Add black pre-commit hook

### LINKED ISSUE #347: Use an autoformatter

jarrodmillman:
Close #347.

I removed isort, since I believe ruff handles it. I removed `  - id: double-quote-string-fixer` from the pre-commit-hooks, since it conflicts with black. I also had to manually edit a couple lines to make ruff happy.


jarrodmillman:
@dnicolodi You may want to comment or close #347.

Issue #380, opened 2023-04-04
Parallel use of interpreters on editable installs


FFY00:
Currently, in editable installs, we just rebuild the project the first time to avoid mismatched versions of modules from bring loaded. No safeties are in place when running multiple at the same time, though, allowing module mismatches to happen if the project is rebuilt by a different interpreter.

My proposal would be for us to not rebuild the project when it is being used by other interpreter, and raise a warning informing the user of this.

rgommers:
Should be addressed well enough by gh-387, which adds the ABI tag to the install path.

PR #387, opened 2023-04-11
ENH: make the editable install build directory more discoverable

### LINKED ISSUE #370: Build directory for editable installations
The build directory for editable build is currently `.mesonpy/editable/build` in the source directory. I don't see a reason why the build directory is nested two levels deep into otherwise empty directories . A flatter directory structure could be used. 

While doing so, attaching a Python API tag to the build directory name would allow to have multiple parallel installations of the same package for different interpreters.

Isn't there prior art for choosing the name of the build directory? I don't like something specific to meson-python much. Python uses `__pycache__` for its cache. Is it allowed to add tool specific stuff in there? distutils uses a `build` folder, IIRC. What about other build backends?

It would be nice to have this decided before we release a version with the editable support.

EDIT: reworded for politeness

dnicolodi:
Change the editable install build directory to a subdirectory named after the ABI tag for the target Python interpreter placed in the `build` directory in the project source directory.

Add .gitignode and .hgignore files with catch all glob patterns to the `build` directory, if it is empty, to have the major VCS systems ignore it.

Fixes #370, #380.

This implements what has been discussed in #370. I decided to add a subdirectory named as the Python ABI tag as including the platform tag does not seem very useful. I went with `build` as that seems to be where the consensus seemed to be. I'm not very attached to this choice.


rgommers (approved):
LGTM and seems to work smoothly for me. Thanks @dnicolodi!

PR #390, opened 2023-04-14
DOC: update changelog

dnicolodi:
Update changelog for 0.13.0 release.  Add names of contributors for each release.  Fix some changelog entries.  Add full stops at the end of sentences.  Move release date to after the contributor names.


dnicolodi:
Thanks for fixing the typos, @rgommers 

PS: `[skip ci]` in the commit message doesn't do anything. We don't have a way to skip the CI jobs,

rgommers:
> PS: `[skip ci]` in the commit message doesn't do anything. We don't have a way to skip the CI jobs,

It does work for GitHub Actions, but indeed for Cirrus CI we'd need custom logic (IIRC there's an open feature request for it to support `skip ci`). Ah well, CI is fast enough to not bother with custom logic for now.

rgommers:
In it goes, thanks again @dnicolodi. Updated the date for today, so let's see if we can tag a release:)

PR #394, opened 2023-04-18
DOC: update readme

dnicolodi:
Remove badges: the README is used an description for the sdist and the wheel. Having the badges pointing to the status of the main branch on the PyPI page for specific releases is not very consistent. Furthermore loading external resources from the documentation is an issue for distributions, which for privacy reasons need to patch the resources out.

Remove link to contributing documentation page: linking to a page that has only a TODO entry does not look very nice.

Reword and add some more details for the use cases target by meson-python.


dnicolodi:
GitHub is back to normal, apparently. PR is updated.

rgommers (approved):
Thanks @dnicolodi. I did a test release on TestPyPI with this branch, and all links work and it looks pretty clean:

image


Compared to the version for 0.12.1:

image

The checkboxes of what does and doesn't work were more relevant at the start of this project, when the package was half-complete. Now we're at a stage where basically everything works, and the smaller exceptions and feature requests are best viewed on our issue tracker. So the changes in this PR seem useful to me.

dnicolodi:
The check-boxes were not removed in this PR. They were removed a while ago during the big documentation refactoring.

PR #400, opened 2023-04-19
CI: make ruff show the lint violations

dnicolodi:
I admit that I don't understand why we disable automatic fixup but then configure the linters to apply fixes. I'm going to leave this here till someone proposes an explanation.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,6 @@ repos:
   rev: v0.0.261
   hooks:
   - id: ruff
-    args: [--fix, --show-fixes, --format, grouped]
 - repo: https://github.com/fsfe/reuse-tool
   rev: v1.1.2
   hooks:


FFY00:
We don't. I don't want to be involved in another argument with you, so perhaps @henryiii could answer your questions, given that he has handled the ruff migration in multiple projects.

FFY00:
@dnicolodi why was my comment marked as off-topic?

dnicolodi:
Because the only part of it that is relevant to this PR is "We don't" and that is not at all helpful. If it refers to my assertion that "we disable automatic fixup but then configure the linters to apply fixes", I don't understand it. We disable automatic fixups: https://github.com/mesonbuild/meson-python/blob/55afbc0ed38a5f403ffb033b6fe8e2451c40ec27/.pre-commit-config.yaml#L5-L6 and then we ask ruff to apply automatic fixups https://github.com/mesonbuild/meson-python/blob/55afbc0ed38a5f403ffb033b6fe8e2451c40ec27/.pre-commit-config.yaml#L27-L31 resulting in the CI job to fail, when linter checks trigger, without giving any clear explanation of what is wrong. The `--show-fixes` option only shows the location where the fixes have been applied, not which fixes have been applied.

The configuration for pre-commit-hook in the repository is used to run the CI job. As we don't require contributors to install pre-commit-hook to submit patches, having the CI report what's wrong with a patch is a much higher priority than providing an out-of-the-box useful configuration of pre-commit-hook for developers that want to use it.

FFY00:
This is exactly the kind of discussion I wanted to avoid. I have no desire to engage with you here, so I pinged Henry to see if he could do it instead.

I don't think this is a good way to interact with contributors. I'll now be unsubscribing from this repo.

henryiii:
Please avoid escalating discussions. There's no need to be negative, and an issue asking why something was set up the way it was is usually a better place to start. Here are a few thoughts:

* Personally I'm a huge fan of the CI just jumping in and fixing things, it makes contributions easier. My preferred fix would the to remove the disabled auto-fixes. :)
* You can ask the CI to update it even though the default is for the CI not to help. Comment "pre-commit.ci autofix" and it will be fixed for you. No need to bother the user or to clone it yourself. I think this is much friendlier than asking contributors to manually update trivial things!
* Black also doesn't report what needs changes? This is not unique.
* This is not just a "out of the box useful configuration" - you can't pass flags to change it. You have to edit the pre-commit file to add these flags if you don't want to be bothered with recording imports, etc.
* There's a big difference between "not required" and not properly supporting developers who do use pre-commit. I would hope most developers (who are up to patching a build system!) will be using it, or would feel the desire to use it if they aren't. And you can (or should, didn't check) run it through tox/nox.
* I'm strongly against `--format, grouped`, by the way. It breaks jumping to the error in vscode, pycharm, iterm, and probably any other terminal or tool that supports jumping.

rgommers:
This does seem to be a case of "colliding workflows". This setup works very well for folks using pre-commit it seems, and less well when you're not using pre-commit. Ideally it works smoothly for both groups, but an improvement for one of them shouldn't be a regression for the other.

I also don't use pre-commit, and usually just edit the relevant commit if CI is unhappy. So far I've found that the CI error log is enough to get the fix right in one go. If that's not the case for `ruff` (which I haven't run into yet), that's a little annoying. `ruff` is easy to use locally though, so that should probably be fine to do and then apply the fix (either manually or automatically)? If not, then that'd be an issue. Is that the case or not @dnicolodi?


> Personally I'm a huge fan of the CI just jumping in and fixing things, it makes contributions easier. My preferred fix would the to remove the disabled auto-fixes. :)

That seems to collide with the strong emphasis on keeping a very clean linear history in this project. I'm personally also not of fan of a bot editing my code/commits.

> Comment "pre-commit.ci autofix" and it will be fixed for you.

Thanks, I did not know that. It does seem helpful in case your next step is to anyway squash-merge.

> I would hope most developers (who are up to patching a build system!) will be using it, or would feel the desire to use it if they aren't. And you can (or should, didn't check) run it through tox/nox.

This is not how things seem to work in practice. Also in other projects like scikit-learn and scipy I've seen a split, with some people being big fans of certain tools, and some never wanting to touch them. Which is fine - I the golden rules are:
1. Set things up so everyone can use their preferred workflow (whether that is pre-commit yes/no, or conda envs vs. virtualenvs, tox/nox yes/no, etc.)
2. Try to avoid discussions around _requiring_ adoption of a certain workflow that you like. This is almost never productive.

eli-schwartz:
> * You have to edit the pre-commit file to add these flags if you don't want to be bothered with recording imports, etc. And then fight git to keep the modification from being added.

TBH this is kind of why I don't really like pre-commit and would rather just run linter tools myself. Which my editor can even do automatically regardless of the presence of pre-commit, actually. :D All it needs is the config file to fine-tune results...

But anyways this entire thing seems like quite the tempest in a teapot. Unsure why this PR had to degenerate into offenses taken. I also think there's probably a hugely simpler solution: why not submit a feature request to ruff, asking for the badly formatted output of `--fix --show-fixes` to be improved (potentially via an option) to provide as much information for fixed violations as for unfixed ones?[^1]

In the meantime, do what it takes to make CI maximally useful, which may mean this PR or may mean adding a github workflow that executes ruff without jumping through pre-commit hoops, and instead controls its own options without impacting the local developer workflow.

[^1]: Or alternatively submit a feature request to pre-commit, asking for the ability to configure different options in CI vs. on developer machines -- that way, the CI could refrain from --fix, and the local developer workflow would use it. But I'm cynical about UX improvements to pre-commit being seen as desirable by the project, it is very "my way or the highway".

eli-schwartz:
Is it possible to run `ruff --fix --diff; ruff --fix` in pre-commit? The former will only print a patchdiff of the changes that it *could* make. The latter will silently make those changes, and print violation reports for all the issues that it couldn't fix. Seems like the best of both worlds, especially because if the hook overall fails then IMO the most useful thing you could do is print in-depth diffs of whatever improvements could be made.

Unfortunately there doesn't seem to be a ruff mode to both --diff and print additional non-fixable issues, so this does require the ability to run two commands in one hook. Maybe this non-existent mode is something else worth submitting a feature request to ruff for.

eli-schwartz:
> So far I've found that the CI error log is enough to get the fix right in one go. If that's not the case for `ruff` (which I haven't run into yet), that's a little annoying.

Ruff output looks like this, if I introduce a bunch of totally random errors to mesonpy:
```
$ ruff mesonpy
mesonpy/__init__.py:59:128: E501 Line too long (132 > 127 characters)
mesonpy/_tags.py:166:119: RUF100 [*] Unused blanket `noqa` directive
mesonpy/_tags.py:167:64: E702 Multiple statements on one line (semicolon)
mesonpy/_tags.py:167:97: E702 Multiple statements on one line (semicolon)
mesonpy/_tags.py:167:128: E501 Line too long (144 > 127 characters)
mesonpy/_tags.py:170:65: RUF100 [*] Unused blanket `noqa` directive
Found 6 errors.
[*] 2 potentially fixable with the --fix option.
```

If I try to autofix them, then this is the result:
```
$ ruff --fix --show-fixes mesonpy
mesonpy/__init__.py:59:128: E501 Line too long (132 > 127 characters)
mesonpy/_tags.py:167:64: E702 Multiple statements on one line (semicolon)
mesonpy/_tags.py:167:97: E702 Multiple statements on one line (semicolon)
mesonpy/_tags.py:167:128: E501 Line too long (144 > 127 characters)

Fixed 2 errors:
- mesonpy/_tags.py:
    2 × RUF100 (unused-noqa)

Found 6 errors (2 fixed, 4 remaining).
```

Figuring out *where* those "2 × RUF100 (unused-noqa)" were just from reading CI reports could end up being virtually impossible.

henryiii:
> why not submit a feature request to ruff, asking for the badly formatted output of --fix --show-fixes to be improved

This is exactly what was originally requested as part of adding `--show-fixes` to Ruff, and is not possible due to the design of Ruff. If one fix creates another (possibly fixable) issue, Ruff reruns and iterates. So the lines and code can change. And should pre-fix or post-fix line numbers be shown? (The way it works, in-the-middle-of-fixing line numbers is all you have).

I think there's a way to fix this, though, based on the suggestions above. We can add a non-pre-commit.ci job that runs Ruff without fixes, and reports this (there's likely even a way to get GH annotations or GH check reports with nicer output working!). Pre-commit then would keep the ability to fix Ruff (along with Black, etc), and work properly natively.

eli-schwartz:
> and is not possible due to the design of Ruff

Reality strikes again... :D



> there's likely even a way to get GH annotations or GH check reports with nicer output working!

I have heard of https://github.com/reviewdog/reviewdog but never tried it.

rgommers:
> I think there's a way to fix this, though, based on the suggestions above. We can add a non-pre-commit.ci job that runs Ruff without fixes

Seems like a pragmatic suggestion, +1 for this fix.

PR #401, opened 2023-04-19
ENH: implement rpath manipulation for ELF binaries

dnicolodi:
This allows to drop the dependency or patchelf.

For now it is only a proof of concept, the code can be cleaned up some more. The meat of the code is taken from Meson. There have not be functional changes in many years, thus the maintenance overhead of including it should not be much.


dnicolodi:
Well. It seems that the code in Meson is very simple because it does not allow arbitrary operations on the RPATH. Adding entries to the RPATH requires making space for them in the ELF file and that requires updating all the sections pointers. Not something that can be implemented so easily. Even more work is required to add an RPATH to ELF files that do not yet have one. It would have been nice to remove the dependency, but it seems to be harder than I anticipated.

mid-kid:
I've stumbled across this issue while wondering why patchelf was being used, but now I'm wondering, why is rpath being manipulated to add new paths in the first place?

PR #408, opened 2023-04-22
Simplify requirements to build the documentation

dnicolodi:


rgommers (approved):
Thanks @dnicolodi. This LGTM - I verified it doesn't change anything for `nox -s docs` while also making a plain `sphinx-build . _build` work in an environment with doc dependencies installed.

rgommers:
I'll also note that this removes a large amount of build warnings for me because I am no longer required to use the spell checker, so that's a good thing. My doc build is now completely clean.

PR #414, opened 2023-04-28
Prepare release 0.13.1

dnicolodi:


dnicolodi:
Ah! It is because the last patch is the same as a revert of the one but last patch, and thus all the changes introduced by the PR are documentation changes, and we don't run the tests on documentation only changes (well, looking at the workflow definition, this does not work as intended anymore, it broke after the documentation reorganization, but it work for patches that only touch the changelog). Everything is working as expected.

rgommers (commented):
Thanks @dnicolodi. A couple of things we may want to add to the changelog:
- Adding `--quiet` to `meson install` (commit 55afbc0)
- A minor fix to improve the consistency of the CLI output on terminal emulators (commit 3a18d932)

Also, would you consider merging gh-412 before this PR? If not, that's also fine, but it improves an error message we introduced in 0.13.0 and that we received feedback on.

dnicolodi:
Merged #412. I thought that the remaining fixes were too small to be added to the changelog. I added them now.

rgommers (approved):
LGTM now, in it goes. On to tagging the release.

PR #424, opened 2023-06-16
BUG: handle install plan with null tag

### LINKED ISSUE #421: Error when filtering install tags on install targets with no tag
```
mesonpy\__init__.py", line 860, in _install_plan
    if details['tag'].strip() not in install_tags:
       ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'strip'

```

It occurs because some targets in intall_plan have `null` tag.

For an example, it occurs when I try to build wheel for https://github.com/bruchar1/pypkgconf

bruchar1:
Fixes #421.

- Do not strip tags from install plan.
- Strip tags in --tags argument

meson install have this behavior.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -853,11 +853,10 @@ class Project():
 
         # filter the install_plan for files that do not fit the install tags
         if args.tags:
-            install_tags = args.tags.split(',')
-
+            install_tags = {tag.strip() for tag in args.tags.split(',')}
             for files in install_plan.values():
                 for file, details in list(files.items()):
-                    if details['tag'].strip() not in install_tags:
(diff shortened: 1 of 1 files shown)


mesonpy/__init__.py
```diff
@@ -853,11 +853,11 @@ def _install_plan(self) -> Dict[str, Dict[str, Dict[str, str]]]:
 
         # filter the install_plan for files that do not fit the install tags
         if args.tags:
-            install_tags = args.tags.split(',')
+            install_tags = [i.strip() for i in args.tags.split(',')]
```
dnicolodi, line 856: Can we make this a set, while at it?

PR #427, opened 2023-06-20
BUG: fix system name specified when cross compiling via $ARCHFLAGS

### LINKED ISSUE #426: host_machine.system incorrectly set to 'Darwin' instead of 'darwin' in meson-python-cross-file.ini
For non-cross-compilation builds meson returns `'darwin'` for `host_machine.system()` on macOS (see also https://mesonbuild.com/Reference-tables.html#operating-system-names).

With meson-python, when cross-compiling for amd64 macOS on macOS intel, `host_machine.system()` returns `'Darwin'` (with capital D). This makes custom OS-dependent checks in the `meson.build`, which rely on the lower case version to distinguish between systems fail.

I believe the wrong capitalization is because meson-python writes `system = 'Darwin'` to the `meson_cross_file`.

https://github.com/mesonbuild/meson-python/blob/78def82ff1e6466afda91c69d7630cc89bc5177d/mesonpy/__init__.py#L681

Since I can test this only in the GH action workflow for my project, I can't actually examine the generated meson_cross_file, but `message('Host machine system:', host_machine.system())` in my `meson.build` outputs `Host machine system: Darwin` when cross-compiling. 

https://github.com/SpotlightKid/python-rtmidi/actions/runs/5316348338/jobs/9625777672#step:4:403

dnicolodi:
Fixes #426.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -678,7 +678,7 @@ class Project():
                         c = ['cc', '-arch', {arch!r}]
                         cpp = ['c++', '-arch', {arch!r}]
                         [host_machine]
-                        system = 'Darwin'
+                        system = 'darwin'
                         cpu = {arch!r}
                         cpu_family = {family!r}
                         endian = 'little'


rgommers:
Thanks @dnicolodi!

Quick comment: I've been very much preoccupied with other stuff (both non-work and projects like changing NumPy over to Meson), but I should be getting back to reviewing your other PRs very soon. Looks like between this and gh-424 we may want to cut a quick bug fix release perhaps, before merging one of the larger PRs?

dnicolodi:
Hello @rgommers. No worries, I've also been busy with other things and I'll be for a few more weeks. I've found some time to fix these small issues, but not much more. Indeed, it may be a good idea to cut a release, unless there are other urgent things pending.

PR #432, opened 2023-06-27
ENH: implement --skip-subprojects install option

### LINKED ISSUE #423: meson `--skip-subprojects` option ignored when building wheel
My project use a subproject, and the subproject installs a static lib. I don't want to include the static lib, or any file installed by the subproject. Howerver, adding
```
[tool.meson-python.args]
install = ['--skip-subprojects']
```
to pyproject.toml does not prevent meson-python trying to install files from the subproject:

```
+ meson install --quiet --no-rebuild --destdir C:\Users\CHARLE~1.BRU\AppData\Local\Temp\build-via-sdist-03nfoz4f\pypkgconf-0.1.0\.mesonpy-tmvd08tf\install --skip-subprojects

meson-python: error: Could not map installation path to an equivalent wheel directory: '{libdir_static}\\libpkgconf.a'

```

For reference, my project is on https://github.com/bruchar1/pypkgconf

dnicolodi:
Requires introspection metadata added in upcoming Meson release 1.2.0.

Fixes #423.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -63,6 +63,10 @@ jobs:
           - os: ubuntu
             python: '3.11'
             meson: '~=1.0.0'
+          # Test with Meson master branch.
+          - os: ubuntu
+            python: '3.11'
+            meson: '@git+https://github.com/mesonbuild/meson.git'
 
     steps:
       - name: Checkout
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -842,24 +842,29 @@ class Project():
     @property
     def _install_plan(self) -> Dict[str, Dict[str, Dict[str, str]]]:
         """Meson install_plan metadata."""
+        install_plan = self._info('intro-install_plan')
 
-        # copy the install plan so we can modify it
-        install_plan = self._info('intro-install_plan').copy()
-
-        # parse install args for install tags (--tags)
+        # parse install args to extract --tags and --skip-subprojects
         parser = argparse.ArgumentParser()
         parser.add_argument('--tags')
+        parser.add_argument('--skip-subprojects', nargs='?', const='*', default='')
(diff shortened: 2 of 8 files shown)


rgommers:
Reading this PR first I was a little confused about how to use this switch, but the docs in https://mesonbuild.com/Release-notes-for-0-58-0.html#skip-subprojects-installation explain it clearly. Given that it's one of many of Meson's subcommand flags, there's no need for this option to be included in the `meson-python` docs - this PR just makes it work as advertised.

rgommers (approved):
This LGTM and works as advertised. Thanks @dnicolodi, in it goes.

One thing I noticed and may be good to discuss: `py.install_sources` in a subproject does not work as I'd expect. The `dep.py` file from the subproject gets mixed into the main package (see file listing in test in this PR). It's designed to be installed into `site-packages/<subproject-package-name>/`. The current behavior seems completely wrong, I can't think of a reason why anyone would want this. Of course, it's very tricky to do the "right" thing and it's not even clear what that would be. I suspect that not installing any files from the subproject into the main package will be the right thing to do in most cases.

 This behavior is already the case with Meson 1.1.1 and meson-python 0.13.2, this PR doesn't change that.

dnicolodi:
> `py.install_sources` in a subproject does not work as I'd expect. The `dep.py` file from the subproject gets mixed into the main package (see file listing in test in this PR). It's designed to be installed into `site-packages/<subproject-package-name>/`.

Why would you expect that? Subprojects in Meson are installed exactly how they would be installed if they were stand-alone projects. I find the current behavior perfectly reasonable. `install_sources()` behaves exactly the same whether it is invoked from the subprojet or from the main project. If the sources need to be installed in a package, the `subdir` argument to `install_sources()` needs to be used.

rgommers:
Ah okay, yes. You're right, this is fine. The use of non-package files is what confused me.

PR #433, opened 2023-06-27
CI: replace Python 3.7 with Python 3.8 on Homebrew

dnicolodi:
Homebrew dropped support for Python 3.7.


rgommers (approved):
Thanks! CI is running really slowly right now, but that must be a GitHub issue.

PR #435, opened 2023-06-28
DOC: update release process page, remove GPG-signing for PyPI

rgommers:
PyPI now discards uploaded GPG signatures and then emails you that it has done so.

This also cleans up the page a bit so the instructions are easier to follow.

### DIFF
--- a/docs/contributing/release-process.rst
+++ b/docs/contributing/release-process.rst
@@ -19,56 +19,38 @@ other maintainer.
 
 After that is done, you may release the project by following these steps:
 
-#. Release to the Git_ repository
+#. Release to the Git repository on GitHub:
     #. Create the release commit
         #. Bump the versions in ``meson.build`` and ``mesonpy/__init__.py``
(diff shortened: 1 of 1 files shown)


docs/contributing/release-process.rst
```diff
@@ -19,56 +19,38 @@ other maintainer.
...
 
-    #. Sign and push the artifacts to PyPI
+    #. Push the artifacts to PyPI
         .. code-block:: console
 
-           $ twine upload -s dist/*
+           $ twine upload dist/*
 
+       Note that there is no need to GPG-sign the artifacts; PyPI no longer
+       supports uploading GPG signatures.
```
eli-schwartz, line 50: But those could still be uploaded as assets to github.

(That's actually what meson always does, but I am not sure PGP signatures were ever uploaded to PyPI -- for the simple reason that PyPI campaigned to remove awareness that it was possible, so no wonder no one was using them lol.)

PR #436, opened 2023-06-28
TST: use symlinks in venv to fix Gentoo PyPy compatibility

### LINKED ISSUE #434: Test failures with (Gentoo) PyPy3
When running the test suite using the Gentoo install of PyPy3.10 7.3.2, I'm getting the following test failures:

```
ERROR tests/test_editable.py::test_editable_install - subprocess.CalledProcessError: Command '['/tmp/pytest-of-mgorny/pytest-0/mesonpy-test-venv0/bin/python', '-m', 'ensurepip', '--upg...
ERROR tests/test_editable.py::test_editble_reentrant - subprocess.CalledProcessError: Command '['/tmp/pytest-of-mgorny/pytest-0/mesonpy-test-venv1/bin/python', '-m', 'ensurepip', '--upg...
ERROR tests/test_examples.py::test_spam - subprocess.CalledProcessError: Command '['/tmp/pytest-of-mgorny/pytest-0/mesonpy-test-venv2/bin/python', '-m', 'ensurepip', '--upg...
ERROR tests/test_wheel.py::test_local_lib - subprocess.CalledProcessError: Command '['/tmp/pytest-of-mgorny/pytest-0/mesonpy-test-venv3/bin/python', '-m', 'ensurepip', '--upg...
```

The problem seem to boil down to the test virtualenv being broken:

```
$ /tmp/pytest-of-mgorny/pytest-0/mesonpy-test-venv3/bin/python
ModuleNotFoundError: No module named 'encodings'
debug: OperationError:
debug:  operror-type: ModuleNotFoundError
debug:  operror-value: No module named 'encodings'
$ ls -lh /tmp/pytest-of-mgorny/pytest-0/mesonpy-test-venv3/bin/python
-rwxr-xr-x 1 mgorny mgorny 57M 06-28 08:35 /tmp/pytest-of-mgorny/pytest-0/mesonpy-test-venv3/bin/python
```

So apparently PyPy executable is being copied into the venv rather than symlinked. Since PyPy uses the location of its executable (after resolving symlinks) to locate `lib_pypy`, it doesn't work after being copied elsewhere.

mgorny:
Pass `symlinks=True` when creating venv during testing to ensure that the virtualenv is created with symlinks to the Python interpreter (whenever possible) rather than copies of it.  This aligns the test suite with the defaults used by `python -m venv` CLI and fixes compatibility with Gentoo installations of PyPy.  The standard installation of PyPy uses the executable location to find its stdlib, and when the executable is copied (without copying the stdlib) it can't start.

Fixes #434


dnicolodi:
Thanks for finding the root cause and fixing the issue @mgorny I notice that we don't run the test suite on Gentoo. I would be happy to merge a patch that adds  a Gentoo image (with CPython and/or PyPy) to the Cirrus CI infrastructure.

mgorny:
I'll try to find a minute to look at it. I suppose we'll need to revive the images with Python preinstalled since I don't think you want to be building all Python versions from source on every CI run ;-).

dnicolodi:
The docker images used in the CI are cached, thus building Python as part of the image generation would not be a problem, but it would need to be done in the dockerfile, not as part of the test runner.

eli-schwartz:
This reminds me that @thesamesam wanted to add a Gentoo CI image for mesonbuild/meson -- a similar logic applies there, we have a weekly image builder to refresh CI images and then regular testing is performed by simply starting the prebuilt images.

PR #442, opened 2023-07-12
TST: add test for handling of the -DNDEBUG compilation flag

dnicolodi:
Verify that the meson-python default meson arguments and user setup arguments result in the -DNDEBUG compilation flat to be set as expected.


rgommers (approved):
This is a nice test, thanks @dnicolodi. I learned something, the ninja `filename.c^` introspection is useful.

PR #443, opened 2023-07-12
Documentation cleanup

dnicolodi:
Mostly removes pages that have not yet been written from the index, and reorganizes the index a bit.


rgommers:
This looks useful, thanks @dnicolodi. Not sure if you're still adding things? I'll review/merge if ready.

dnicolodi:
I think the documentation would need some more love, but this is IMHO already a nice step forward while being minimally invasive. I'll open another PR for further work. I'll be happy if you could review this PR in its current form.

.github/workflows/tests.yml
```diff
@@ -10,13 +10,6 @@ on:
     branches:
       - main
       - release-*
-    paths-ignore:
```
rgommers, line 13: I'd consider keeping this; even if it goes out of date I'd say it's better than triggering quite a bit of CI.

dnicolodi: I updated the paths filter. Let's see if it works as intended...

dnicolodi:
I've pushed two more commits with the suggested changes to the TODOs-only pages. PTAL.

.github/workflows/tests.yml
```diff
@@ -10,13 +10,14 @@ on:
...
+    paths:
       # This is spelt like this to do not ignore the docs/examples/ folder.
-      - docs/*
-      - docs/usage/*
-      - CHANGELOG.rst
-      - LICENSE
-      - README.md
+      - '**'
+      - '!docs/examples/**'
```
rgommers, line 16: This didn't quite go right I think. It now ignores the examples and triggers on other files in `docs/`.

rgommers:
Last two commits LGTM.

.github/workflows/tests.yml
dnicolodi: Indeed. Fixed, I think.

rgommers (approved):

PR #444, opened 2023-07-13
DOC: fix example for feature enabled

bruchar1:


rgommers (approved):
That looks right (https://mesonbuild.com/Build-options.html#features), thanks @bruchar1

PR #445, opened 2023-07-15
DOC: add License, Maintainers, and Funding sections

dnicolodi:
This replaces #354.

The left-hand-side index on the documentation pages gets again a bit too long, but I cannot think to an organization of these sections that makes sense and spares top level index entries. Suggestions are welcome.


dnicolodi:
@rgommers What's the copyright blurb we should use for the Quansight logo?

rgommers:
> What's the copyright blurb we should use for the Quansight logo?

Good question. I think the options are:
-leaving it out, or
- simply "Copyright Quansight", or
- something custom like "Copyright Quansight, approved for use by meson-python in its documentation, and including redistributions of meson-python without restrictions".

This has got to be a common question, because a ton of projects use logos like this. I'm not aware of a single canonical way to treat copyright statements though - I'll do some searching.

rgommers:
Common question, but there is no good answer. I think it's actually an annoyance for packagers to have this logo image included, because it is the kind of thing that has to be treated with care. For most projects this doesn't come up because logos go on a website, which is kept in a separate repo from the code and docs so there's no cross-talk to an sdist and other redistributable artifacts for the software package.

The easiest thing here is to get rid of it. That's too bad, but attribution in text is also fine.

docs/about.rst
```diff
@@ -0,0 +1,37 @@
...
+License 2.0`__.
+
+__ https://github.com/mesonbuild/meson-python/blob/main/LICENSE
+__ https://github.com/mesonbuild/meson/blob/master/COPYING
+
+
+Maintainers
+===========
+
+- `Ralf Gommers <https://github.com/rgommers>`_, Quansight
```
rgommers, line 21: Would you mind deleting the `, Quansight` here? it's already acknowledged further down, and my remaining a `meson-python` maintainer is not at all conditional on my job.

rgommers:
also, `pre-commit` has a nitpick

dnicolodi:
`pre-commit` complains about the Quansight logo SVG not terminating in a newline. Issue will go away if we renounce to include the logo. Indeed having third party logos in the source distribution is a pain, but it is also too bad to do not give visibility to companies that contribute to free software.

dnicolodi:
All fixed. I've also added one more commit on top. PTAL

docs/about.rst
```diff
@@ -31,3 +31,26 @@ Ralf Gommers contributes to the development of ``meson-python`` as
 part of their job at `Quansight`__.
 
 __ https://www.quansight.com/
+
+
+Packages
+========
+
+``meson-python`` packages are distriburted via `PyPI`_ as Python wheel
```
rgommers, line 39: minor: there's a typo in "distributed"

docs/about.rst
```diff
@@ -31,3 +31,26 @@ Ralf Gommers contributes to the development of ``meson-python`` as
...
+for Python 3.7 and later.
```
rgommers, line 40: can we say "all supported Python releases" to avoid having the minimum supported version go out of date here when we drop support for 3.7?

rgommers: also, maybe add "source distribution". Let me propose a full rephrase of this sentence:
```
``meson-python`` packages are distributed via `PyPI`_, as a source distribution and a
single pure Python wheel that works across all supported Python versions.
```

rgommers (approved):
This LGTM, thanks. Renaming Security to Packages sounds good to me, because it's not really about Security. For a section named Security, I'd expect a security vulnerability disclosure policy or some such thing.

Please consider my proposed rephrase - either way this is good to go in.

docs/about.rst
dnicolodi: I avoided to write "supported" because technically Python 3.7 is already end-of-life but I don't think we are planning to retire support for it. However, I don't think it is important to be overly precise here. I like the proposed wording. I'll update the commit.

PR #446, opened 2023-07-15
meson.build cleanup

dnicolodi:
This replaces #367.


rgommers (approved):
LGTM, these seem like clear improvements. Thanks @dnicolodi

PR #451, opened 2023-08-15
ENH: fix support for Python limited API / stable ABI wheels

dnicolodi:
Meson gained support for building Python extension modules targeting the Python limited API, therefore it is time for meson-python to properly support tagging the build wheels as targeting the stable ABI.

Unfortunately there isn't a reliable and cross-platform way to detect when extension modules are build for the limited API. Therefore, we need to add an explicit "limited-api" configuration option in the [tool.meson-python] section in pyproject.toml.


dnicolodi:
Documentation is still missing.

dnicolodi:
I've added the documentation.

tests/packages/limited-api/meson.build
```diff
@@ -0,0 +1,22 @@
+# SPDX-FileCopyrightText: 2023 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+project('limoted-api', 'c', version: '1.0.0')
```
rgommers, line 5: minor: typo in the project name

tests/packages/limited-api/meson.build
```diff
@@ -0,0 +1,22 @@
...
+
+py = import('python').find_installation()
```
rgommers, line 7: minor: probably good to add `pure: false` here. There's no `py.install_sources` so it doesn't technically matter here, but it would if a `.py` file was added.

rgommers (commented):
Thanks @dnicolodi, this looks quite good and this is a very nice new feature. 

My main question is about the added test case: it seems to demonstrate that the `limited_api` setting in `pyproject.toml` cannot be set correctly if there is a build option which has the effect of toggling between the limited API and the regular C API. This should be rare, but it is conceivable.

In such a case, I think the answer is that the setting in `pyproject.toml` should be overridden via a CLI flag in order to not have the build fail, so something like:
```
pip install . -C-Dextra=true -C--no-limited-api
```
However, we don't have a flag like `--no-limited-api` that `meson-python` will use to override the `pyproject.toml` setting. Do you think that's needed, or can it be left for later because no one needs this, and we may never need it? Or is the answer to just patch `pyproject.toml`?

eli-schwartz:
For a bit of context, during the development of the limited API support in meson, o recommended the addition of that option toggle based purely on the notion that it would be interesting to ISVs like a Linux distro that don't use wheels except as an unavoidable side effect of build backends requiring them. They manually construct their own distro package manager archive with fully resolved contents pointing to a exact python version and rebuild from source regardless of "limited API or regular extensions", so disabling the limited API define is a free optimization.

What are the downsides of an inaccurate wheel tag for purely local builds with an immediate install?

rgommers:
Ah, I also noticed for the first time that `meson_options.txt` has been renamed to `meson.options` (since Meson 1.1, as noted in https://mesonbuild.com/Build-options.html#build-options).

> What are the downsides of an inaccurate wheel tag for purely local builds with an immediate install?

As long as they are guaranteed to stay local, nothing I guess. But producing `-abi3` tags when the contents don't match that seems fishy. Sooner or later someone is going to upload them anyway.

eli-schwartz:
Hmm, then maybe you can just check the value of that setting from the json introspection files created in the builddir?

rgommers:
The Cygwin CI job isn't happy with the new `LIB_SUFFIX`:
```
    LIB_SUFFIX = {
E   KeyError: 'CYGWIN_NT-10.0-20348'
```

dnicolodi:
Cygwin is always funny, and given this result, I don't know what `platform.system()` is actually good for. It seems that we should standardize on using `sys.platform()` instead, which since a few Python releases actually has a well specified value for all major platforms.

rgommers (approved):
All green now and looks great, in it goes. Thanks @dnicolodi!

PR #457, opened 2023-08-22
DOC: remove mention of limitations that have been lifted

dnicolodi:


dnicolodi:
The first paragraphs about data files installed with `install_data()` in the Limitations page is actually incorrect. Data files installed in `{datadir}` are actually bundled in the wheel in the `data` install path (see list at point 3 here https://packaging.python.org/en/latest/specifications/binary-distribution-format/#file-contents) and installed correctly by pip. However, I don't know how they could be found at runtime. Anyhow, as it stands the paragraph is incorrect.

dnicolodi:
Fixed the paragraphs about the data files too.

rgommers (approved):
These changes look good to me, and 🎉 for the list of limitations being down to two items. Thanks @dnicolodi!

Issue #420, opened 2023-06-08
Mention that files need to be staged to VCS in tutorial 


FFY00:
In the [beginner tutorial](https://meson-python.readthedocs.io/en/latest/), we don't mention that files need to be staged to VCS for them to be included in the sdist. This is required for the [building step](https://meson-[python.readthedocs.io/en/latest/tutorials/introduction.html#building-the-project), so we should add it.

PR #464, opened 2023-09-01
DOC: clean up front page and intro tutorial, state files must be in VCS

rgommers:
Closes gh-420

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -9,8 +9,8 @@ meson-python
 
 .. highlights::
 
-   ``meson-python`` is a Python build backend built on top of the Meson_
-   build-system. It enables you to use Meson_ for your Python packages.
+   ``meson-python`` is a Python build backend built on top of the Meson
+   build-system. It enables you to use Meson for your Python packages.
 
 
 Want to look at examples in real projects? Check out our curated list of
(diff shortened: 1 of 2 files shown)


docs/tutorials/introduction.rst
```diff
@@ -151,35 +146,23 @@ self-explanatory by looking at our example.
...
 Testing the project
 -------------------
 
-Now we should have a valid Python project, let's test it.
-
-We will install it with pip_
+Now we should have a valid Python project, so let's test it. Before continuing,
+ensure you have committed the three files we created above to your Git
+repository - ``meson-python`` will only take into account the files that Git
+knows about.
```
dnicolodi, line 163: I'm not sure mentioning here that only committed files are taken into account to build an sdist is relevant. AFAIK, `pip` does not build wheels via an sdist, this is the default behavior of `build` but not of `pip`. I would move this sentence further down where creating sdists and wheels is mentioned.

docs/index.rst
```diff
@@ -10,7 +10,7 @@ meson-python
 .. highlights::
 
    ``meson-python`` is a Python build backend built on top of the Meson_
```
dnicolodi, line 12: I would remove the link also from the first mention of Meson in the colophon. It really does not look nice. I would maybe link the first mention of Meson in the text.

rgommers: Agreed. Done now. That first mention is in the intro tutorial rather than on this page, but that's fine.

docs/tutorials/introduction.rst
rgommers: Good point - moved down.

dnicolodi:
This is a very nice improvement. Merging.

PR #465, opened 2023-09-01
ENH: copy files into wheel from Meson install directory

dnicolodi:


dnicolodi:
We have another problem. Meson since 1.2.0 byte compiles installed Python source files. We don't want that for building wheels. More than the wasted compilation time, the problem is that picking up files from the installation directory, we also pick up the bite compiled files, which should not be in the wheel. There is no option to turn the byte compilation off.

We want to pick the files from the installation directory to have meson to remove the build directory from the rpath and handle the exclusions for `install_subdir()`. The rpath fixup is broken on macOS anyway, and since a while we have enough information in the metadata to handle the exclusions ourselves. I'm really tempted to give up the idea of picking up the files for the wheel from the installation directory.

To fix the rpath, we can simply remove all rpaths that point inside the build directory. Doing so we lose support for `install_rpath` unless we find a way to get that from the metadata. From some quick tests I did, support for `install_rpath` is anyhow broken on macOS, thus I don't think is is used much.

rgommers:
That sounds reasonable to me.

QuLogic:
> There is no option to turn the byte compilation off.

Isn't there [set `python.bytecompile` option to -1](https://mesonbuild.com/Builtin-options.html#python-module)?

dnicolodi:
@QuLogic Thanks. I missed it reading the documentation. However, turns out that avoiding the install step actually makes things easier, thus I think we will go that way anyway.

dnicolodi:
Superseded by #467 and #469.

Issue #461, opened 2023-08-25
Document how to get a debug build


dnicolodi:
See https://github.com/mesonbuild/meson-python/issues/459#issuecomment-1692971600

Add a note about conda adding environment variables adding optimization flags to the compilation flags.

Issue #207, opened 2022-11-14
Building with a debug python build should generate a debug build


ngoldbaum:
Currently meson-python will generate the same wheel for a debug python build and a regular python build. I think this violates the principle of least surprise, since if one is using a debug python build I think one expects the C extensions to also be debug builds. This is what setuptools currently does.

I'm working around this by [manually passing compiler flags to the meson build for my C extension](https://github.com/ngoldbaum/metadatadtype/blob/f34f414726dff77e854716be1bbac0e62a61eb0f/meson.build#L44) when I need a debug build.

gh-154 would enable setting flags from the command line, so I wouldn't need to edit `meson.build`, however I think it's still worth having better default behavior for debug builds.

lithomas1:
Hmmm. I think I would lean more towards keeping the default as a release build, with a command line option to toggle.

Wheels are meant to be distributed, and having debug symbols makes it quite a bit bigger. I think users would also expect to have optimized code within a wheel.

FFY00:
I think the proposal here is that the Python interpreter itself is a debug build, meson-python should default to a debug build in Meson, and if the interpreter is a release build, default to a release build.

The proposed default behavior seems a bit surprising to me, I don't think I would expect it, and I generally treat debug interpreters and building extensions with debug arguments as different things that'd I would change independently. However, I don't maintain a lot of Python extensions, so I can't really tell if these are normal expectations. Perhaps @rgommers could provide some better feedback.

rgommers:
The proposal makes sense to me. The only reason I'd want a debug build of Python is to do some actual debugging - at which point you do want debug builds of everything else as well (may be necessary, can't really hurt). Some Linux distros even distribute packages like that, e.g. Ubuntu 18.04 has `python-dbg` and then matching packages like `python-numpy-dbg`, `python-scipy-dbg`. I'd also say that a debug Python build is nothing special here (or, Python as a package is nothing special, only the limitations of PyPI make it different) - it's just another package. 

As for the details:
- I wouldn't use `c_args: ['-g', '-O0'],` like @ngoldbaum does, I'd just add `-Ddebug-true`. The `-g` is a GCC-specific way of writing `-Ddebug=true`. And I would not touch optimization settings; a typical default for Meson is `debugoptimized`.
- I don't know without looking in more detail how to tell whether a Python install is a debug build. And there's corner cases like the user passing in a cross file that need a simple answer/implementation.

ngoldbaum:
[...]

Issue #459, opened 2023-08-24
Debug build type still produces optimized extensions


WillAyd:
When building pandas, if I specify a command like:

```sh
pip install -ve . --no-build-isolation --config-settings=builddir="debug" --config-settings=setup-args="-Dbuildtype=debug"
```

Shared libraries / extensions still get built with `-O2`. compile_commands.json confirms that while `-O0` is passed in, it gets overridden later in the command:

```sh
  {
    "directory": "/home/willayd/clones/pandas/debug",
    "command": "/home/willayd/mambaforge/envs/pandas-dev/bin/x86_64-conda-linux-gnu-cc -Ipandas/_libs/tslibs/base.cpython-310-
x86_64-linux-gnu.so.p -Ipandas/_libs/tslibs -I../pandas/_libs/tslibs -I../../../mambaforge/envs/pandas-dev/lib/python3.10/site-packages
/numpy/core/include -I../pandas/_libs/include -I/home/willayd/mambaforge/envs/pandas-dev/include/python3.10 -fvisibility=hidden 
-fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -w -std=c99 -O0 -g -DNPY_NO_DEPRECATED_API=0 
-DNPY_TARGET_VERSION=NPY_1_21_API_VERSION -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong 
-fno-plt -O2 -ffunction-sections -pipe -isystem /home/willayd/mambaforge/envs/pandas-dev/include -DNDEBUG 
-D_FORTIFY_SOURCE=2 -O2 -isystem /home/willayd/mambaforge/envs/pandas-dev/include -fPIC -MD -MQ pandas/_libs/tslibs
/base.cpython-310-x86_64-linux-gnu.so.p/meson-generated_pandas__libs_tslibs_base.pyx.c.o -MF pandas/_libs/tslibs/base.cpython-
310-x86_64-linux-gnu.so.p/meson-generated_pandas__libs_tslibs_base.pyx.c.o.d -o pandas/_libs/tslibs/base.cpython-310-x86_64-
linux-gnu.so.p/meson-generated_pandas__libs_tslibs_base.pyx.c.o -c pandas/_libs/tslibs/base.cpython-310-x86_64-linux-gnu.so.p
/pandas/_libs/tslibs/base.pyx.c",
[...]

Issue #441, opened 2023-07-12
Explicitly setting a debug build also requires explicitly setting b_ndebug=false


ngoldbaum:
For whatever reason I need to have `setup = ["-Dbuildtype=debug", "-Db_ndebug=false"]` in my `pyproject.toml`, not just `setup = ["-Dbuildtype=debug"]`.

I think what's happening is the default meson settings `meson-python` sets *all* need to be explicitly overriden. If I just pass `-Dbuildtype=debug`, the `meson setup` command `meson-python` creates is:

```
+ meson setup --reconfigure /home/nathan/Documents/numpy-user-dtypes/stringdtype /home/nathan/Documents/numpy-user-dtypes/stringdtype/build -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md -Dbuildtype=debug --native-file=/home/nathan/Documents/numpy-user-dtypes/stringdtype/build/meson-python-native-file.ini
```

So I guess meson-python's settings take precedence because they are explicitly set, but then my custom settings for the build type override that, but since I don't have a custom `ndebug`, the previously set `if-release` choice doesn't get reset.

Maybe meson-python could check if there is a user-provided `buildtype` and not set `ndebug` or `buildtype` if so?

dnicolodi:
When reporting an issue it is desirable to explicitly state what you are trying to achieve, the solution you tried, how it did not work, preferably corroborated by some evidence. The description of this issue allows us only to guess what you are trying to do and why you thin it does not work as expected.

I guess you are trying to obtain a build with debug symbols and with assertions enabled. To achieve that you added
```toml
[tool.meson-python.args]
setup = ["-Dbuildtype=debug"]
```
However, in your opinion that is not sufficient in enabling assertions (ie compile without the `-DNDEBUG` compilation flag). Why do you think this is the case? The combination of `-Dn_debug=if-release -Dbuildtype=debug` should not result in the `-DNDEBUG` compilation flag being set.

ngoldbaum:
Sorry if the initial description wasn’t clear. You’re right that I’d expect the combination of `-Dn_debug=if-release` and `-Dbuildtype=debug` to not set `-DNDEBUG`. However, I’m finding that not to be the case, I need to explicitly set `-Dn_debug=false`.

dnicolodi:
How did you verify that? I just tested it and it works as expected, see #442

ngoldbaum:
[...]

PR #466, opened 2023-09-01
DOC: add a howto page on debug builds

rgommers:
Closes gh-461
Closes gh-207
Also xref gh-459 and gh-441 which are already closed but contain discussion relevant to debug builds.

### DIFF
--- a/docs/how-to-guides/debug-builds.rst
+++ b/docs/how-to-guides/debug-builds.rst
@@ -1,0 +1,49 @@
+.. SPDX-FileCopyrightText: 2023 The meson-python developers
+..
+.. SPDX-License-Identifier: MIT
+
+.. _how-to-guides-debug-builds:
+
+******************
+Using debug builds
+******************
+
+For development work on native code in your Python package, you may want to use
+a debug build. To do so, we need to use the ``buildtype`` option, which is
+equivalent to ``-Ddebug=true -Doptimization=0``, to ``meson setup``. In addition,
+it is likely most useful to set up an editable build with a fixed build
+directory. That way, the shared libraries in the installed debug build will
+contain correct paths, rather than paths pointing to a temporary build
+directory which ``meson-python`` will otherwise use. IDEs and other tools will
+then be able to show correct file locations and line numbers during debugging.
+
+We can do all that with the following ``pip`` invocation:
+
+.. code-block:: console
+
+    $ pip install -e . --no-build-isolation -Csetup-args=-Dbuildtype=debug -Cbuilddir=build-dbg
+
+This debug build of your package will work with either a regular or debug build
+of your Python interpreter. A debug Python interpreter isn't necessary, but may
+be useful. It can be built from source, or you may be able to get it from your
+package manager of choice (typically under a package name like ``python-dbg``).
+Note that using a debug Python interpreter does not yield a debug build of your
+own package - you must use ``-Dbuildtype=debug`` or an equivalent flag for that.
+
+.. warning::
+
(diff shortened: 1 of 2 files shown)


WillAyd:
Looks nice!

docs/how-to-guides/debug-builds.rst
```diff
@@ -0,0 +1,49 @@
...
+.. SPDX-License-Identifier: MIT
+
+.. _how-to-guides-debug-builds:
+
+******************
+Using debug builds
+******************
+
+For development work on native code in your Python package, you may want to use
+a debug build. To do so, we need to use the ``buildtype`` argument, which is
```
QuLogic, line 12: Isn't it?
```suggestion
a debug build. To do so, we need to use the ``buildtype=debug`` argument, which is
```

rgommers: I was calling `buildtype` the argument, and I'd consider `debug` its value. The full invocation is given on the `pip install` line further down.

I don't have a strong preference here - the way I have it now seems more concise to me, but it could also be `buildtype=debug` or `-Dbuildtype=debug`.

dnicolodi: I would write "buildtype option" (preferred) or "-Dbuildtype=debug argument"

docs/how-to-guides/debug-builds.rst
```diff
@@ -0,0 +1,49 @@
...
+equivalent to ``-Ddebug=true -Doptimization=0``, to ``meson setup``. In addition,
+it is likely most useful to set up an editable build with a fixed build
+directory. That way, the shared libraries in the installed debug build will
+contain correct paths, rather than paths pointing to a temporary build
+directory which ``meson-python`` will otherwise use. IDEs and other tools will
+then be able to show correct file locations and line numbers during debugging.
+
+We can do all that with the following ``pip`` invocation:
+
+.. code-block:: bash
```
dnicolodi, line 22: Using `.. code-block:: console` for command line examples formats them in a slightly nicer way.

dnicolodi:
I usually prefer the impersonal form for documentation, but it may just be a professional bias, thus it is ok this way too. Otherwise I just have a couple of minor comments.

docs/how-to-guides/debug-builds.rst
rgommers: adopted, thanks

docs/how-to-guides/debug-builds.rst
rgommers: Updated to "`buildtype` option"

rgommers:
> I usually prefer the impersonal form for documentation

I usually prefer this form, and I also think it's fine either way. As long as individual pages are consistent so it doesn't look too messy.

docs/how-to-guides/debug-builds.rst
QuLogic: Sorry, perhaps I was unclear; you have "the ``buildtype`` option, which is equivalent to ``-Ddebug=true -Doptimization=0``", which is not directly exactly true. It is `buildtype=debug` which is equivalent to those two settings; `buildtype` by itself is equivalent to many combinations of those options.

dnicolodi: Thanks. Indeed, the sentence as it reads now is not precise. I'll fix this.

rgommers: Good point @QuLogic, thanks. The fix is included as part of gh-476.

PR #467, opened 2023-09-01
BUG: fix handling of files installed via install_subdir

dnicolodi:


rgommers:
This looks happy. It has a bit of overlap with gh-465 - can I review and merge this one first?

dnicolodi:
If the strategy used here work, and moving RPATH handling to meson-python work, #465 is not going to happen. I've the patch for the RPATH handling almost ready. We can merge this and retire #465.

dnicolodi:
I've fixed the RPATH handling in #469 on top of this PR. It seems that we can do without running `meson install` without any drawback, thus we can move forward with the strategy in this PR.

rgommers (approved):
Thanks @dnicolodi! All LGTM, and I tested a bunch of things with SciPy and Pandas - everything is happy.

PR #472, opened 2023-09-02
MAINT: use sys.platform instead of platform.system()

dnicolodi:
The latter is not well defined on some systems.


dnicolodi:
I've checked that all the tests that should be executed are still executed. Merging.

Issue #473, opened 2023-09-02
Amend documentation to reflect that `meson install` is not run


dnicolodi:
Amend the documentation to reflect the fact that we don't actually run meson install but just re-implement it. This affects two Meson features: `install_rpath` support, and running scripts registered with `meson.add_install_script()`. Both things have no effect when building a package with `meson-python`.

dnicolodi:
Also explicitly list the `meson install` options recognized by `meson-python` in https://meson-python.readthedocs.io/en/latest/reference/pyproject-settings.html#cmdoption-arg-tool.meson-python.args.install and https://meson-python.readthedocs.io/en/latest/reference/config-settings.html#cmdoption-arg-install-args or find a way to do not have to repeat the content twice.

Most of the `meson install` options not handled by `meson-python` aren't relevant, but it's possible `--strip` may be or some new option gets added to `meson install` that we must start handling.

FFY00:
I ran into this today, where I explicitly needed to perform some artifact fixup that was not possible any other sane way in meson other than using `meson.add_install_script()`.

dnicolodi:
@FFY00 AFAIK the only feature offered to a script executed as a `meson.add_install_script()` vs a script executed as a `custom_target()` is the knowledge of the installation location via environment variables like `MESON_INSTALL_PREFIX` and `MESON_INSTALL_DESTDIR_PREFIX`. The installation path is hard to define for packages installed as wheels. Therefore, I am not able to imagine what requirement makes an installation script the right solution. Can you explain your use case in more detail?

eli-schwartz:
Stuff like auditwheel repair, maybe?

FFY00:
> Stuff like auditwheel repair, maybe?

Yes, exactly. In my particular case, I needed to perform a small edit to the binary format data of certain artifacts.

This is not possible in Meson AFAIK, as you can't pass a build target as the input for a custom target,

dnicolodi:
The Meson documentation is not very clear on this, but you can pass the object returned by `build_target()` as input to a `custom_target()`.

PR #474, opened 2023-09-02
DOC: prepare changelog for 0.14.0 release

dnicolodi:
I've started to put together the changelog entry for the next release.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,33 @@
 Changelog
 +++++++++
 
+0.14.0
+======
+
+- Do not run ``meson install`` to build a wheel. This was unnecessary
+  as files are added to the wheel from the build and source
+  directories. This does not affect the handling of ``meson install``
+  options, which are interpreted by ``meson-python`` itself.
+- Obey the ``--skip-subprojects`` when specified for the ``meson
+  install`` command.
+- Implement support for the ``exclude_directories`` and
+  ``exclude_files`` arguments to Meson ``install_subdir()`` function
+  and similar installation functions. This requires Meson version
+  1.1.0 or later.
+- Implement support for building wheels targeting the Python limited
+  API. Extension modules targeting the Python limited API can be
+  easily built starting with the upcoming Meson 1.3.0 release.
+- when ``pyproject.toml`` does not contain a ``version`` field and
+  ``version`` is not declared dynamic, raise an error instead of
+  silently using the version declared in ``meson.build``.
+- Fix the mtime of source files in the sdist tarball.
+- Add ``objc`` and ``objcpp`` compilers to the cross file generated
+  when the ``$ARCHFLAGS`` is set.
+- Extensive documentation improvements.
+
+Charles Brunet, Daniele Nicolodi, Henry Schreiner, Michał Górny, Ralf
+Gommers --- xx-09-2023
+
 
 0.13.2
 ======


CHANGELOG.rst
```diff
@@ -11,6 +11,28 @@
 Changelog
 +++++++++
 
+0.14.0
+======
+
+- Obey the ``--skip-subprojects`` option passed to the ``meson
+  install`` command.
+- Implement support for building wheels targerting the Python limited
```
rgommers, line 19: typo in "targeting"

CHANGELOG.rst
```diff
@@ -11,6 +11,28 @@
...
+  API. Extension modules targeting the Python limited API can be
+  easily build starting with the upcomin Meson 1.3.0 release.
```
rgommers, line 21: - build -> built
- upcomin -> upcoming

CHANGELOG.rst
```diff
@@ -11,6 +11,28 @@
...
+- Instead than silently using the version declared in ``meson.build``,
```
rgommers, line 22: - instead than -> instead of
- perhaps turn it around: _"When `pyproject.toml` does not contain a `version` field and ..."_. Might be a little easier to understand for the reader whether this applies to them.

CHANGELOG.rst
```diff
@@ -11,6 +11,28 @@
...
+  raise an error when ``pyproject.toml`` does not contain a
+  ``version`` field and ``version`` is not declared dynamic.
+- Fix the mtime of source files in the sdist tarball.
+- Fix the installation of files via ``install_subdir()``. For the
```
rgommers, line 26: I suggest extending the first sentence with _"when the `exclude_directories or `exclude_files keywords are used"_. The way it reads now, a user may think there was a broader problem.

CHANGELOG.rst
```diff
@@ -11,6 +11,28 @@
...
+  ``exclude_directories`` and ``exclude_files`` arguments to work,
+  Meson version 1.1.0 or later is required.
+- Add ``objc`` and ``objcpp`` compilers to the cross file generated
+  when the ``$ARCHFLAGS`` is set.
+- Extensive documentation improvements.
```
rgommers, line 31: How about adding here:

- `meson install` is no longer run as part of the wheel build process, resulting in less log being written to stdout.

rgommers (commented):
Thanks @dnicolodi! Looks good overall - a few comments and suggestions.

CHANGELOG.rst
dnicolodi: I'm not sure it is worth mentioning it, we just got rid of a single line looking something like `+ meson install --quiet --no-rebuild --allow-dirty`

CHANGELOG.rst
dnicolodi: I thought that there were broader problems with handling `install_subdir()` but looking at the old code again, I cannot identify what these problems may have been. I'll update this entry.

CHANGELOG.rst
rgommers: maybe not indeed. okay, resolving.

CHANGELOG.rst
rgommers: Nothing important I think - both Pandas and NumPy use it fairly extensively, and it didn't give them problems at least AFAIK.

dnicolodi:
I've implemented all suggestions. I also added an entry bout `meson install` after all, and reordered the entries to be grouped more logically, instead of being sorted by implementation date, which is irrelevant to the reader.

dnicolodi:
Good idea, but we describe a way to pass arguments to `meson install` also here https://meson-python.readthedocs.io/en/latest/reference/config-settings.html#cmdoption-arg-install-args

rgommers:
Ah forgot we already have an issue: gh-473. I'll comment there.

dnicolodi:
I've beaten you to it :)

PR #475, opened 2023-09-03
Assorted tweaks to `pyproject.toml`

dnicolodi:


pyproject.toml
```diff
@@ -26,8 +26,8 @@ maintainers = [
   { name = 'Thomas Li'},
 ]
 classifiers = [
-  'Development Status :: 4 - Beta',
```
rgommers, line 26: How about changing it to `5 - Stable` instead?

pyproject.toml
```diff
@@ -19,8 +19,11 @@ readme = 'README.rst'
 requires-python = '>= 3.7'
 license = { file = 'LICENSES/MIT.txt' }
 keywords = ['meson', 'build', 'backend', 'pep517', 'package']
-authors = [
-  { name = 'Filipe Laíns', email = 'lains@riseup.net' },
```
rgommers, line 23: If we're dropping Filipe's name here, then I think we should give him some credit as original author / emeritus maintainer on https://meson-python.readthedocs.io/en/latest/about.html.

rgommers (commented):
Thanks @dnicolodi - a few comments.

pyproject.toml
dnicolodi: Absolutely. Feel free to propose a wording. I'm not particularly good in this kind of things.

rgommers: I suggest changing the Maintainers and Funding section to this:

**Maintainer**

The current maintainers of `meson-python` are:

- [Ralf Gommers](https://github.com/rgommers)
- [Daniele Nicolodi](https://github.com/dnicolodi)
- [Henry Schreiner](https://github.com/henryiii)
- [Thomas Li](https://github.com/lithomas1)

Emeritus maintainers: [Filipe Laíns](https://github.com/FFY00).

**Funding**

Ralf Gommers contributes to the development of `meson-python` as part of their job at [Quansight](https://www.quansight.com/). Initial creation of `meson-python` by Filipe Laíns was also funded by Quansight.

dnicolodi: Thanks! Implemented

rgommers (approved):
LGTM now, thanks @dnicolodi

pyproject.toml
```diff
@@ -19,8 +19,11 @@ readme = 'README.rst'
...
+maintainers = [
+  { name = 'Ralf Gommers ', email = 'ralf.gommers@gmail.com' },
+  { name = 'Daniele Nicolodi', email = 'daniele@grinta.net' },
+  { name = 'Henry Schreiner', email = 'HenrySchreinerIII@gmail.com' },
+  { name = 'Thomas Li'},
```
dnicolodi, line 26: PyPI makes a mess of this. It shows Thomas as the only maintainer but links his name to Ralf's email.

Also the license appears again in a funny way, it now displays the first lines of `LICENSES/MIT.txt`

rgommers: Hmm, not sure what's up with the maintainers field. Possibly because email for Thomas is missing? I didn't notice that before. If that doesn't fix it, I guess we should remove this change for now.

License looks as expected to me. PyPI fixed the problem where that full license text would horribly overflow. It's much better than it was before. Of course, even better would be to finish and implement PEP 639, so we can see the SPDX classifier ....

dnicolodi: The PyPI page says "Maintaner" singular, thus I'm not sure PyPI has support for multiple maintainers listed in the package metadata. It is a bit annoying that the package metadata specification and the main cunsumer of it are so out of sync. I thought that the license field was fixed doing something like what Github does: match the license text and provide a short license identifier. I've must have dreamed of this.

rgommers: Re maintainers: maybe we should just comment it out now and list "The meson-python maintainers"? We have a full list in the docs already.

> I thought that the license field was fixed doing something like what Github does: match the license text and provide a short license identifier. I must have dreamed of this.

There's some setuptools-specific old way of doing this I believe, taking it from classifiers if the `license` field under `[project]` isn't used. Deleting a standardized PEP 621 entry to go back to something undocumented doesn't seem quite right though, so I have this one filed under "revisit when PEP 639 is done".

dnicolodi: > Re maintainers: maybe we should just comment it out now and list "The meson-python maintainers"? We have a full list in the docs already.

Sure. I listed the names because I thought that having an entity there with an associated email address may be useful. PyPI lists you as a maintainer, I guess because your user is associated with the project. I'm fine having just you listed as a maintainer in the metadata.

> There's some setuptools-specific old way of doing this I believe, taking it from classifiers if the `license` field under `[project]` isn't used. Deleting a standardized PEP 621 entry to go back to something undocumented doesn't seem quite right though, so I have this one filed under "revisit when PEP 639 is done".

Yes, indeed. Let's keep things as they are for now.

dnicolodi: Uhm. The way PyPI reports the maintainers list may be due to how we generate the metadata. This is how the metadata fields related to the maintainers look like:
```
Maintainer: Thomas Li
Maintainer-Email: Ralf Gommers  <ralf.gommers@gmail.com>, Daniele Nicolodi <daniele@grinta.net>, Henry Schreiner <HenrySchrein
```
I'm verifying what the specification prescribes, but at a first look that seems wrong.

dnicolodi: At least the `Maintainer-Email` field is wrong, as it is supposed to be an RFC-822 From header compliant format https://packaging.python.org/en/latest/specifications/core-metadata/#maintainer-email The specification does not specify how multiple maintainers should be listed

dnicolodi: Lovely, the specification for `pyproject.toml` contradicts the specification for the metadata file https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#authors-maintainers

PR #476, opened 2023-09-03
DOC: add page stating Meson compatibility

dnicolodi:


docs/reference/meson-compatibility.rst
```diff
@@ -0,0 +1,50 @@
...
+.. SPDX-License-Identifier: MIT
+
+.. _meson-compatibility:
+
+*******************
+Meson compatibility
+*******************
+
+``meson-python`` tightly integrates with Meson to produce Python
+wheels ans sdists. Therefore, correct operation depends on
```
rgommers, line 12: typo in "ans"

rgommers (approved):
LGTM, thanks @dnicolodi. This matches the way we treat the current Meson version support, and when we bump the minimum version in the future we can update this page as needed.

Maybe fix the typo before merging?

docs/reference/meson-compatibility.rst
dnicolodi: I really need to setup my editor to spell check `.rst` files...

PR #477, opened 2023-09-03
Add some content about tests to the contributing docs.

robtaylor:


docs/contributing/test-suite.rst
```diff
@@ -7,10 +7,36 @@
 **********
 Test suite
 **********
```
rgommers, line 9: minor: a blank line here would be useful, that's typically done below headers in .rst files.

docs/contributing/test-suite.rst
```diff
@@ -7,10 +7,36 @@
...
+
+   $ pip install nox
+
+To run the test suite, run this in your checkout of meson-python:
+
+.. code-block:: console
+
+   $ nox -s tests
+
+This will run the test suite for all the python versions you have installed. Some may fail if they aren't a whole installation. e.g. on Ubuntu 22.04, python3.9 will fail unless you've installed it from the relevant PPA.
```
rgommers, line 22: style: could you break these lines? at 80 characters is common in other docs files

rgommers (commented):
Thanks @robtaylor! 

I added a couple style comments. Also, the pre-commit CI job is failing because of trailing whitespace. The Arch Linux CI job failure will go away if you rebase on latest `main`.

I think the content here is correct for `nox` usage. I don't personally use it though, I prefer a single dev env and then simply run `pytest` or (for example) `pytest tests/test_wheel.py`. I'm probably not the only one who prefers such a setup, so we may want to either put a `.. note::` at the bottom or explain it in the text as an alternative. @dnicolodi any preference?

Finally, can you add this page to `docs/contributing/index.rst`? Otherwise it won't show up in the rendered html.

docs/contributing/test-suite.rst
```diff
@@ -7,10 +7,36 @@
 **********
 Test suite
 **********
+To run the the test suite you'll need nox__, an automated testing tool for Python.
```
dnicolodi, line 10: `nox` is not required to run the tests. I don't use it and I would not even notice if the `nox` configuration in the repository would break. I don't feel recommending it as the tool of choice to run the tests. Plain `pytest` is probably a more conservative choice.

Also, I don't know `nox`, but from the configuration file I deduce that it requires installing `meson-python` for every test invocation. This is not necessary and unnecessarily slow.

robtaylor: Shall I keep the nox content (i guess useful for testing against a number of python versions..?) and prefix it with standard pytest?

rgommers: That sounds good to me. I'll note that the "set up your development environment" part, which is a prerequisite for a basic:
```
$ pip install -e . --no-build-isolation
$ pytest
```
is also still empty. I suggest we discuss that in a separate issue though, because it may involve either longer install instructions or duplicating dependencies, and there's a bunch of ways of doing this.

rgommers: If that sounds good, I can open a new issue with a proposal.

dnicolodi: > That sounds good to me. I'll note that the "set up your development environment" part, which is a prerequisite for a basic:

There is no need to install `meson-python` to run the tests. I never do. Simply running `pytest` in the project source directory works for me.

dnicolodi: Oh, sure. Sorry, I think I misinterpreted your comment. But you would get a very similar error if you try to run `pip install -e .` with the `--no-build-isolation` option.

rgommers: Yes, agreed. It's a separate topic, but a prerequisite for running tests at all. That's why I suggested keeping it to a separate issue/PR. "setting up a dev env" tends to be a hotly debated topic:)

dnicolodi: > "setting up a dev env" tends to be a hotly debated topic

That's why I'm not sure about covering it in our documentation. I already been deemed not worthy to work on `meson-python` because I don't use `pre-commit-hook`. I don't want to have to read similar things because the way to setup a development environment we suggest in the documentation does not reflect what is trendy this week. 😆

rgommers: Yeah, I definitely don't want it to be prescriptive. As a conda/mamba user I'm often left out in the cold too by other projects. I'd more like a few helpful hints like "here is what you need, you can set it up like _this_, or _that_, or any way you like as long as you have all the needed dev dependencies".

eli-schwartz: "The testsuite uses pytest. The easiest way to run pytest is via the provided `nox` testing tool, which handles all the nitty gritty details for you including installation of dev dependencies and much more. [rest of description]"

Seems reasonable?

robtaylor:
Hi I’ve closed this as I’ve decided to not use or work on meson-python
going forward. Happy for someone else to pick it up.

Issue #236, opened 2022-11-25
Implement prepare_metadata_for_build_wheel


lithomas1:
We should implement this method, since if it is unavailable, pip will call build_wheel to find the metadata.

Since we don't support editable installs, if a project supports both meson and setuptools as a build system, the project will be built unnecessarily with meson before proceeding with an editable install with setup.py.

In order to do this, we just need to write .dist-info with only METADATA and WHEEL (no need for RECORD0.

dnicolodi:
Where is the fall-back from a pep517-build to a setuptools-build when the pep517 backend does not support editable install documented? At a quick look I don't see it in pep517. If the installation is going to proceed with setuptools, what's the purpose of obtaining metadata from a different pep517 backend? Isn't it going to be discarded anyway?

rgommers:
> if a project supports both meson and setuptools as a build system

I wanted this for SciPy, but it is impossible by design, `pip` doesn't allow it. You can make duplicate sets of `setup.py` and `meson.build` files in your source tree, but you need a one line patch to switch build backends. Using `pip install --no-pep-517` does not work.

rgommers:
> Since we don't support editable installs, if a project supports both meson and setuptools as a build system, the project will be built unnecessarily with meson before proceeding with an editable install with setup.py.

We should support it soon. But other than that, is this actually happening? That seems like a `pip` bug, since they have on purpose not even made `--no-pep-517` work.

dnicolodi:
The only way I know to make this work is to call `python setup.py develop` explicitly.

rgommers:
When we released SciPy 1.9.0 the sdist had `build-backend = 'mesonpy'`, and I told all distributors that still had trouble with building (e.g. due to less flexible BLAS support) to carry a patch to comment out that one line, and then things do work as expected with setuptools.

dnicolodi:
@rgommers Sure. I meant that calling `setup.py` directly is the only way without patching `pyproject.toml`. Of course if you change the build backend declaration, the right backed is picked up by the frontend.

lithomas1:
> > if a project supports both meson and setuptools as a build system
> 
> I wanted this for SciPy, but it is impossible by design, `pip` doesn't allow it. You can make duplicate sets of `setup.py` and `meson.build` files in your source tree, but you need a one line patch to switch build backends. Using `pip install --no-pep-517` does not work.

pip is actually smart enough to fall back to setup.py, when it can't find the build_editable hook.

[...]

PR #478, opened 2023-09-03
Fix editable install of meson-python for pip 23.2.1 and python3.11

robtaylor:
With python 3.11 and pip 23.2.1, an editable install of meson-python itself fails with:
`pip._internal.exceptions.InstallationSubprocessError: Checking if build backend supports build_editable exited with 1`

This is due to lacking an implementation of PEP660's prepare_metadata_for_build_editable.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,6 +42,9 @@ jobs:
         python:
           - '3.7'
           - '3.11'
+        editable:
+          - true
+          - false
         meson:
           -
         include:
@@ -75,7 +78,6 @@ jobs:
           - os: windows
             python: '3.11'
             meson: '@git+https://github.com/mesonbuild/meson.git'
-
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -103,6 +105,11 @@ jobs:
 
       - name: Install
         run: python -m pip --disable-pip-version-check install .[test]
+        if: ${{ ! matrix.editable }}
+
+      - name: Editable Install
+        run: python -m pip --disable-pip-version-check install -e .[test]
+        if: ${{ matrix.editable }}
 
       - name: Run tests
         run: >-
@@ -111,6 +118,8 @@ jobs:
       - name: Upload coverage report
         uses: codecov/codecov-action@v3
         if: ${{ always() }}
+
+
 
   msvc:
     runs-on: windows-latest
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -1099,3 +1099,25 @@ def get_requires_for_build_editable(
     config_settings: Optional[Dict[str, str]] = None,
 ) -> List[str]:
     return get_requires_for_build_wheel()
+
+@_pyproject_hook
+def prepare_metadata_for_build_editable(
+    metadata_directory: str,
+    config_settings: Optional[Dict[str, str]] = None,
+) -> str:
+
+    with _project(config_settings) as project:
+        wheelbuilder = project._wheel_builder
+
+        dist_info_dir = pathlib.Path(metadata_directory, f'{wheelbuilder.basename}.dist-info')
+        dist_info_dir.mkdir(parents=True, exist_ok=True)
+
+        with dist_info_dir.joinpath('METADATA').open(mode='ab') as f:
+            f.write(project.metadata)
+        with dist_info_dir.joinpath('WHEEL').open(mode='ab') as f:
+            f.write(wheelbuilder.wheel)
+
+        if wheelbuilder.entrypoints_txt:
+            with dist_info_dir.joinpath('entry_points.txt').open(mode='ab') as f:
+                f.write(wheelbuilder.entrypoints_txt)
+        return str(dist_info_dir)
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -39,7 +39,20 @@ else:
     class Traversable:
         pass
     class TraversableResources:
-        pass
+        def files(self):
+            pass
+
+        def open_resource(self, resource):
+            return self.files().joinpath(resource).open('rb')
+
+        def resource_path(self, resource):
+            raise FileNotFoundError(resource)
+
+        def is_resource(self, path):
+            return self.files().joinpath(path).is_file()
+
+        def contents(self):
+            return (item.name for item in self.files().iterdir())
 
 
 MARKER = 'MESONPY_EDITABLE_SKIP'
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -165,7 +165,7 @@ def pep518_wheelhouse(tmp_path_factory):
     packages = [
         meson_python,
     ]
-    cmd = [sys.executable, '-m', 'pip', 'wheel', '--no-build-isolation', '--wheel-dir', wheelhouse, *packages]
+    cmd = [sys.executable, '-m', 'pip', 'wheel',  '--wheel-dir', wheelhouse, *packages]
     subprocess.run(cmd, check=True)
     return wheelhouse
 


robtaylor:
Still getting an issue in my environment that _meson_python_editable_loader.py is created with the ninja execitable set to the one installed in the temporary build enviroment created by pip. 
Not sure why I'm hitting this but the tests aren't!

dnicolodi:
How did you determine that the lack of the `prepare_metadata_for_build_editable()` hook is the source of your problems?

Everything works as expected for me:
```
$ python --version
Python 3.11.4
$ python -m pip --version
pip 23.2.1 from /Users/daniele/src/meson-python/venv/lib/python3.11/site-packages/pip (python 3.11)
$ python -m pip install --no-build-isolation -e .
Obtaining file:///Users/daniele/src/meson-python
  Checking if build backend supports build_editable ... done
  Preparing editable metadata (pyproject.toml) ... done
Requirement already satisfied: meson>=0.63.3 in ./venv/lib/python3.11/site-packages (from meson-python==0.14.0.dev0) (1.2.1)
Requirement already satisfied: pyproject-metadata>=0.7.1 in ./venv/lib/python3.11/site-packages (from meson-python==0.14.0.dev0) (0.7.1)
Requirement already satisfied: packaging>=19.0 in ./venv/lib/python3.11/site-packages (from pyproject-metadata>=0.7.1->meson-python==0.14.0.dev0) (23.1)
Building wheels for collected packages: meson-python
  Building editable for meson-python (pyproject.toml) ... done
  Created wheel for meson-python: filename=meson_python-0.14.0.dev0-py3-none-any.whl size=17632 sha256=43e997401297472d0d9edb03577b91f7f40637b2c3e100f2bb9f475020a4e599
  Stored in directory: /private/var/folders/6f/4l4d5n9x1y71cxyrprf_x35c0000gn/T/pip-ephem-wheel-cache-s0rqdxog/wheels/91/8a/31/237de59d176b9ce63c1dcd55038958a1238f2fd5f1c0cd7506
Successfully built meson-python
Installing collected packages: meson-python
  Attempting uninstall: meson-python
    Found existing installation: meson-python 0.14.0.dev0
    Uninstalling meson-python-0.14.0.dev0:
      Successfully uninstalled meson-python-0.14.0.dev0
Successfully installed meson-python-0.14.0.dev0
```

dnicolodi:
> Still getting an issue in my environment that _meson_python_editable_loader.py is created with the ninja execitable set to the one installed in the temporary build enviroment created by pip.
> Not sure why I'm hitting this but the tests aren't!

As I already answered here https://github.com/mesonbuild/meson-python/discussions/425#discussioncomment-6897683, you are installing in editable mode without disabling build isolation, and you don't have a suitable `ninja` installed.

robtaylor:
> How did you determine that the lack of the `prepare_metadata_for_build_editable()` hook is the source of your problems?
> 
> Everything works as expected for me:
> 
> ```
> $ python --version
> Python 3.11.4
> $ python -m pip --version
> pip 23.2.1 from /Users/daniele/src/meson-python/venv/lib/python3.11/site-packages/pip (python 3.11)
> $ python -m pip install --no-build-isolation -e .
> Obtaining file:///Users/daniele/src/meson-python
>   Checking if build backend supports build_editable ... done
>   Preparing editable metadata (pyproject.toml) ... done
> Requirement already satisfied: meson>=0.63.3 in ./venv/lib/python3.11/site-packages (from meson-python==0.14.0.dev0) (1.2.1)
> Requirement already satisfied: pyproject-metadata>=0.7.1 in ./venv/lib/python3.11/site-packages (from meson-python==0.14.0.dev0) (0.7.1)
> Requirement already satisfied: packaging>=19.0 in ./venv/lib/python3.11/site-packages (from pyproject-metadata>=0.7.1->meson-python==0.14.0.dev0) (23.1)
> Building wheels for collected packages: meson-python
>   Building editable for meson-python (pyproject.toml) ... done
>   Created wheel for meson-python: filename=meson_python-0.14.0.dev0-py3-none-any.whl size=17632 sha256=43e997401297472d0d9edb03577b91f7f40637b2c3e100f2bb9f475020a4e599
>   Stored in directory: /private/var/folders/6f/4l4d5n9x1y71cxyrprf_x35c0000gn/T/pip-ephem-wheel-cache-s0rqdxog/wheels/91/8a/31/237de59d176b9ce63c1dcd55038958a1238f2fd5f1c0cd7506
> Successfully built meson-python
> Installing collected packages: meson-python
>   Attempting uninstall: meson-python
>     Found existing installation: meson-python 0.14.0.dev0
>     Uninstalling meson-python-0.14.0.dev0:
>       Successfully uninstalled meson-python-0.14.0.dev0
> Successfully installed meson-python-0.14.0.dev0
> ```

using pip install -vvv

If you try running the test changes in this patchset without the fixes, you'll see the issues...

dnicolodi:
I just tried `python -m pip install -vvv -e .` in the `meson-python` source directory and everything still works as expected, just with more output. Can you please be more specific about how that fails for you?

robtaylor:
> > Still getting an issue in my environment that _meson_python_editable_loader.py is created with the ninja execitable set to the one installed in the temporary build enviroment created by pip.
> > Not sure why I'm hitting this but the tests aren't!
> 
> As I already answered here [#425 (reply in thread)](https://github.com/mesonbuild/meson-python/discussions/425#discussioncomment-6897683), you are installing in editable mode without disabling build isolation, and you don't have a suitable `ninja` installed.

Ahh, ok I had definitely missed how-to-guides-editable-installs. 
The patch here seems to fix the need for this in projects *using* meson-python - at least that's what the test set shows, maybe there's a corner case i'm missing?

Still need --no-build-isolation for meson-python itself, as the pip environment results in the installed ninja failing to work (the stub fails on `from ninja import ninja`). I wonder if we can do something so folk don't get caught out, and add a note in `contributing`?

dnicolodi:
The tests work just because `ninja` is provided by the system on the test runners. It would fail the same with or without these patches if it would not be present, as yourself pointed out happens on your system. How to install an editable wheel for a project using `meson-python` is documented. I don't think we need to repeat that. Also, an editable (or otherwise) installation of meson-python is not required to work, debug, or test meson-python.

dnicolodi:
Can you please describe what this PR achieves and why we need it? PEP 660 states that the `prepare_metadata_for_build_editable` is optional and the front end should fall back to `build_editable` if it is missing. In the case of `meson-python`, providing accurate metadata requires building the project, thus there is no advantage in providing `prepare_metadata_for_build_editable` over `build_editable` as the former would need to do all the work the latter does, except packaging the build artifacts in a zip file, which is negligible work compared to the rest.

dnicolodi:
I was actually going to close that issue as wontfix over the weekend. To produce metadata we need to know if the wheel is pure or not (ie if it included platform or python version specific content) and for doing that we need to configure, and in some occasions, actually build the project. 

PEP 660 is not clear on what `prepare_metadata_for_build_editable` is used for by the front-end. I fear that some front-ends may execute `prepare_metadata_for_build_editable` before `build_editable`, just in case. This would result in `meson-python` doing work twice for no good reason. Lacking an use case requiring `prepare_metadata_for_build_editable` (or `prepare_metadata_for_build_wheel` as requested in #236), I much prefer not to implement these methods.

robtaylor:
> The tests work just because `ninja` is provided by the system on the test runners. It would fail the same with or without these patches if it would not be present, as yourself pointed out happens on your system. How to install an editable wheel for a project using `meson-python` is documented. I don't think we need to repeat that. Also, an editable (or otherwise) installation of `meson-python` is not required to work, debug, or test `meson-python`.

To check this, i've added the test set on top of main:
https://github.com/robtaylor/meson-python/actions/runs/6074445627

CF: test set with the two other commits:
https://github.com/robtaylor/meson-python/actions/runs/6071530369

So I do feel like I'm still not understanding the objection overall here - does feel like i've stepped in a hornets nest, and I'm not attached to whether this is closed DONTFIX! I could just be sleep deprived though. ><

dnicolodi:
No problem. I'm happy that we can agree that there is nothing to fix.

However, for the future, I recommend you to be mindful of the fact that the project maintainers have most likely spend much more time than you thinking about an issue and almost certainly have a better understanding of the context in which a project is developed. Providing a concise and punctual description of the issue you have encountered and of the reason why you are proposing a patch. Having to argue the position of the maintainers, especially when the opposing position is not well defined, is very tiring. This talk does a very good job at describing the issue https://www.youtube.com/watch?v=o_4EX4dPppA

dnicolodi:
> The patches that are useful here are the patch for [breakage on 3.7/3.8](https://github.com/mesonbuild/meson-python/pull/478/commits/06a4bfa6d970b4ab4df83b896d46714691903766) and
> [the test set for meson-python editable install](https://github.com/mesonbuild/meson-python/pull/478/commits/592bc925a8c4c37ed4df2114be555c6b270a1d52)

We have tests for the editable wheels functionality, for example https://github.com/mesonbuild/meson-python/blob/main/tests/test_editable.py#L170 They run also on Python 3.7 and 3.8 thus I doubt there is anything to fix there.

rgommers:
Thanks @robtaylor, glad there was no problem in the end.

> I was actually going to close that issue as wontfix over the weekend. [...]

Thanks for those thoughts @dnicolodi - I'll follow up on that issue now with a reply to your concerns.

PR #479, opened 2023-09-04
CI: update codecov action

dnicolodi:


dnicolodi:
@rgommers Do you know anything about the integration of codecov into the github project? I think that as we upload the coverage reports to codecov we may as well have codecov report back the coverage report as a status check on the pull requests. However, the status checks do not appear. I've been reading the codecov documentation. AFAIU, status checks should be enabled by default

rgommers:
It used to appear in status checks, see e.g. https://github.com/mesonbuild/meson-python/pull/165#issuecomment-1271509356. That also discusses `informational: true`. Codecov status checks are somewhat useful indeed on occasion, as long as their bot doesn't comment and it never fails.

It looks like it disappeared when we migrated the repo to the `mesonbuild` org. The GitHub App isn't installed on this repo. The status checks are visible on Meson PRs, so I don't think there's a security concern with it (despite Codecov not exactly being the most reliable provider - multiple security issues).

I don't have permissions to install the App, but I was able to send a request for it - once it's approved by a `mesonbuild` org admin the problem should be fixed.

dnicolodi:
Thanks @rgommers

rgommers (approved):
These cleanups LGTM, thanks.

dnicolodi:
I wanted to make sure that the codecov reports are sensible before merging. But we can tweak the configuration further later, if needed.

rgommers:
Yeah, I figured this is almost certainly right, and the cleanups are useful. We can tweak it later if needed.

cclauss:
There are a lot of `Codecov: Failed to properly create commit` in GitHub Actions runs like at the bottom right of
https://github.com/mesonbuild/meson-python/actions/runs/10213496455

rgommers:
https://github.com/codecov/codecov-action?tab=readme-ov-file#breaking-changes - a token is needed with v4 of the action. I'll have a look.

PR #480, opened 2023-09-04
MAINT: restructure the interface between the project class and the wheel builder

dnicolodi:
This is an attempt at making the interface between the wheel builder class and the class representing the meson project a bit easier to understand.  Ideally we should be able to just pass the manifest and the metadata to the wheel builder class.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -476,14 +476,15 @@ class _WheelBuilder():
 
         return wheel_file
 
-    def build_editable(self, directory: Path, verbose: bool = False) -> pathlib.Path:
+    def build_editable(self, directory: Path, source_dir: pathlib.Path, build_dir: pathlib.Path,
+                       build_command: List[str], verbose: bool = False) -> pathlib.Path:
+
         wheel_file = pathlib.Path(directory, f'{self.name}.whl')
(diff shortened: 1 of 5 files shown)


dnicolodi:
@rgommers This is a relatively large scale reorganization of the code, but I find the resulting code much easier to reason about. It is still work in progress, but I would like to know if you like the direction in which this goes before putting the final touches to it.

dnicolodi:
Also, I've tried to split the changes in small self-standing commits. I think this makes the changes easier to review. However, if you prefer a big patch, that would be easier to generate because I don't need to make sure that the tests pass for each intermediate step.

rgommers:
Thanks! This looks like a useful set of changes to me. Making this interface as clean as possible will be helpful to future maintenance. The individual commits are quite helpful for reviewing - I read the full diff first, but it's pretty hard to review that. Each individual commit looks pretty good.

dnicolodi:
Great. I'll finish this then.

dnicolodi:
This is now ready for review. There is room for further simplification, but this is already getting big.

rgommers (approved):
This LGTM, and things are indeed easier to understand this way I think. So in it goes. Thanks @dnicolodi!

PR #491, opened 2023-09-08
MAINT: simplify logging and use of ANSI color escapes

dnicolodi:


mesonpy/__init__.py
```diff
@@ -178,29 +142,40 @@ def _map_to_wheel(sources: Dict[str, Dict[str, Any]]) -> DefaultDict[str, List[T
     return wheel_files
 
 
-def _is_native(file: Path) -> bool:
-    """Check if file is a native file."""
+class style:
+    ERROR = '\33[31m',
+    WARNING = '\33[93m'
+    INFO = '\33[36m\33[1m'
```
rgommers, line 148: Typo in the color spec here? Looks like two colors concatenated.

mesonpy/__init__.py
```diff
@@ -178,29 +142,40 @@ def _map_to_wheel(sources: Dict[str, Dict[str, Any]]) -> DefaultDict[str, List[T
     return wheel_files
 
 
-def _is_native(file: Path) -> bool:
-    """Check if file is a native file."""
+class style:
+    ERROR = '\33[31m',
```
rgommers, line 146: It may be useful to a comment behind it what each color is.
```python
ERROR = '\33[31m',  # red
```

mesonpy/__init__.py
rgommers: On closer inspection, it looks like this is "light blue" + "bold". For some reason only INFO is bold - I haven't noticed it as a problem, so fine to leave it as it is.

rgommers (approved):
LGTM, thanks @dnicolodi. One minor suggestion about adding comments to make clear what the ANSI codes map to. Other than that, looks great.

mesonpy/__init__.py
dnicolodi: Yes, that's it. I'm not particularly in love with this choice of styles and I don't know why it was picked. I don't see any reason why what I classified as INFO messages need to be bold but not the others. I can change them to be all bold or not bold.

rgommers: I think the answer here may be because the red/yellow/green warnings in a regular build are coming from `meson` and `build`, and they make these colors bold for warnings and the final line:
```
Successfully built meson_python-0.15.0.dev0.tar.gz and meson_python-0.15.0.dev0-py3-none-any.whl
```
Either way, this looks pretty good as is, let's get this in. It can always be tweaked in the future if we notice inconsistencies. Or accessibility issues - the green isn't ideal probably for people with vision problems and especially in a white terminal, but a change would have to be across the multiple packages working together here probably.

dnicolodi: I'm not sure I follow what you are trying to say. The colors used by `meson-python` do not match the ones used by `meson` nor the one used by `build` AFAICT. However, it does not really matter. The yellow warnings are not really readable with the default settings of my terminal, but I can live with that.

rgommers: I had a quick peak in Meson, the red and cyan match, yellow is a bit different. From `mesonbuild/mlog.py`:

```python
def red(text: str) -> AnsiDecorator:
    return AnsiDecorator(text, "\033[1;31m")

def green(text: str) -> AnsiDecorator:
    return AnsiDecorator(text, "\033[1;32m")

def yellow(text: str) -> AnsiDecorator:
    return AnsiDecorator(text, "\033[1;33m")

def blue(text: str) -> AnsiDecorator:
    return AnsiDecorator(text, "\033[1;34m")

def cyan(text: str) -> AnsiDecorator:
    return AnsiDecorator(text, "\033[1;36m")
```

I meant that if we'd choose radically different colors from `meson` and `build`, the overall impression would probably be more messy.

PR #492, opened 2023-09-08
MAINT: bump minimum required Meson version on Python 3.12 

dnicolodi:


dnicolodi:
The meson branch works as advertised, `meson-python` requires small tweaks to drop the `setuptools` dependency when a sufficiently nee meson is used. The testing found an unrelated issue in `meson-python` #494 and two Cython bugs: https://github.com/cython/cython/issues/5692, fixed in https://github.com/cython/cython/pull/5693, and https://github.com/cython/cython/issues/5694, but the latter is just a documentation issue.

.github/workflows/tests.yml
```diff
@@ -41,20 +43,25 @@ jobs:
           - windows
         python:
           - '3.7'
+          - '3.8'
+          - '3.9'
+          - '3.10'
           - '3.11'
+          - '3.12-dev'
+          - 'pypy-3.9'
```
cclauss, line 51:
```suggestion
          - '3.12'
          - 'pypy-3.10'
```

cclauss: And then below add the last line to:
```diff
  name: Set up target Python
        uses: actions/setup-python@v4
        with:
          python-version: ${{ matrix.python }}
+         allow-prereleases: true
```
https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#allow-pre-releases

.github/workflows/tests.yml
```diff
@@ -41,20 +43,25 @@ jobs:
...
+        #   - os: ubuntu
+        #     python: 'pypy-3.9'
+        #   - os: ubuntu
+        #     python: '3.8'
+        #   - os: ubuntu
+        #     python: '3.9'
+        #   - os: ubuntu
+        #     python: '3.10'
+        #   - os: ubuntu
+        #     python: '3.12-dev'
```
cclauss, line 64:
```suggestion
        #     python: '3.12'
```

cclauss (commented):
Just some ideas...

.github/workflows/tests.yml
dnicolodi: Why do you suggest to drop testing with pypy-3.9 and test pypy-3.10 instead?

.github/workflows/tests.yml
dnicolodi: I don't think that modifying commented out configuration has any effect.

cclauss: The reason that `allow-prereleases` was created was because the upgrade to production releases is done automatically when they ship instead of having to do that manually.  There is no difference today but there is a difference when Python 3.12 ships.

.github/workflows/tests.yml
cclauss: To ensure that there are not any compatibility issues.

.github/workflows/tests.yml
dnicolodi: That line is commented out. It could very well specify `3.99`.

.github/workflows/tests.yml
dnicolodi: Do you think that `pypy-3.9` was picked at random? Why is it more important to test with `pypy-3.10` than with `pypy-3.9`? Is anything in the PyPy 3.10 release changelog making you believe that it is important to test with this release?

.github/workflows/tests.yml
cclauss: You are correct.  I did a find and replace.  Sorry for the oversight.

dnicolodi:
The required change to drop the use of `distutils` have been merged in meson. We need to decide what we want to do on the meson-python side. Up until now we require `setuptools` when running on Python 3.12 or later. What I've implemented in this PR is to use the `get_requires_for_build_wheel()` hook to request `setuptools` if running on Python 3.12 or later and the meson version is known to still require `distutils`. This works, but I don't like the added complexity. What we could do is simply bump the minimum required meson version with something like this in `pyproject.toml`:
```
dependencies = [
  'colorama; os_name == "nt"',
  'meson >= 0.63.3; python_version < "3.12"',
  'meson >= 1.2.3; python_version >= "3.12"',
  'pyproject-metadata >= 0.7.1',
  'tomli >= 1.0.0; python_version < "3.11"',
]
```
This has the drawback to require a very recent (at the moment not yet released) version of meson on Python 3.12. I don't know whether this may be an issue in practice. Of course we can do this only when a suitable meson version is actually released.

rgommers:
Changing to `'meson >= 1.2.3; python_version >= "3.12"',` as soon as Meson 1.2.3 is released seems like a good solution.

dnicolodi:
Meson 1.2.3 has been released dropping the requirement for `distutils` on recent Pythons. We don't need to pull in `setuptools` anymore on Python 3.12.

rgommers (approved):
Nice. This all LGTM, in it goes. Thanks @dnicolodi!

PR #493, opened 2023-09-09
TST: remove pep518 tests

dnicolodi:
These tests verify that build can build sdists and wheels using meson-python. The pep518 name is misleading as meson-python itself does not implement PEP 518. These tests verify that the PEP 517 hooks exported by meson-python work as build expects them to work.

These tests setup a local directory with wheel for the build dependencies. This is problematic for two reasons: it requires network access, thus distributions need to disable these tests with downstream patches, and it makes it very difficult to run the tests against a specific version of meson.

Creation of wheels and sdists is tested as part of the unit tests. The pep517 tests are unit tests for some of the PEP 517 hooks. The installation of meson-python itself in the CI jobs works as integration tests.

Remove these tests.


rgommers (approved):
Agreed with the rationale here, we don't really need these tests. So in it goes. Thanks @dnicolodi

dnicolodi:
Thanks. This became a pain when testing on Python 3.12 and trying to make the setuptools dependency conditional to the meson version and testing the changes in meson to do not require setuptools.

Issue #458, opened 2023-08-23
Adding support for `$MESON` or another way to point at a customized Meson


rgommers:
In NumPy we had to vendor Meson in order to include a major new feature (for SIMD support) that is not yet available in Meson itself. That is fine, that's how it's supposed to work, since Meson isn't extensible by design - the point is to keep things simple and upstream your needed features/changes.

However, _using_ that vendored Meson also forced us to vendor `meson-python`. The only reason for that was to be able to point at the correct `meson` executable. So ideally `meson-python` would have a way for a package to point at that custom Meson instead. This could be done via a `MESON` environment variable (similar to the `NINJA` one), and/or in some other way (a setting in `pyproject.toml` perhaps.

Part of what's needed here is being able to point at a `meson.py` (as [shipped with Meson itself](https://github.com/mesonbuild/meson/blob/master/meson.py)), and have `meson-python` translate that to `[sys.executable, 'path/to/meson.py']`. This has to do with it being a problem on Windows to ship an executable file `meson` and have it recognized. See  https://github.com/numpy/numpy/pull/24379#issuecomment-1673584945 and following comments for details.

This idea was also touched upon in gh-323. It's related but not the same, so I thought I'd open a separate issue instead. The implementation is probably quite simply, the main thing to decide is what the UX for this should look like.

dnicolodi:
On the surface, adding support for looking up the meson executable via a`$MESON` environment variable seems a good idea. However, I wonder how this would work if it is used to point meson-python to a meson that has features that are not included in any distributed version of meson.

In the Python packaging ecosystem there is the assumption that binary distributions can be compiled from source distributions with the tools specified in the package definition. There is even a proposed PEP that looks into extending this beyond Python packages and things that can be installed from PyPI. @rgommers I'm sure you are familiar with it.

This stops working if something not specified in `pyproject.toml` in the `[build-system]` is required to compile a package.

A solution I used a while back for a package of mine when there were still bugs to iron out in the Meson Python support, is something like:

```toml
[build-system]
build-backend='mesonpy'
requires=[
  'meson-python',
  'meson @ https://github.com/dnicolodi/meson.git',
]
```
This ensures that the specific version of Meson is installed in the build environment, and the `$PATH` should be set up properly to have it take precedence over the system one.

[...]

PR #495, opened 2023-09-10
ENH: make the Meson executable user-configurable 

rgommers:
Allow using a MESON environment variable, analogous to the NINJA one, or a string to an executable or Python script in the `[tool.meson-python]` section of `pyproject.toml`.

Addresses gh-458.

Package authors are likely going to need/prefer the `tool.meson-python` setting, as discussed in gh-458. For local development or user overrides the environment variable will be useful; it is implemented analogous to the NINJA one.

This will allow me to unvendor `meson-python` from `numpy`, as tested in [this branch](https://github.com/rgommers/numpy/tree/unvendor-mesonpython) with these changes:

<details>

```diff
 [build-system]
 build-backend = "mesonpy"
-backend-path = ['./vendored-meson/meson-python']
 requires = [
     "Cython>=3.0",
-    # All dependencies of the vendored meson-python (except for meson, because
-    # we've got that vendored too - that's the point of this exercise).
-    'pyproject-metadata >= 0.7.1',
-    'tomli >= 1.0.0; python_version < "3.11"',
-    'setuptools >= 60.0; python_version >= "3.12"',
-    'colorama; os_name == "nt"',
-    # Note that `ninja` and (on Linux) `patchelf` are added dynamically by
-    # meson-python if those tools are not already present on the system. No
-    # need to worry about those unless one does a non-isolated build - in that
-    # case they must already be installed on the system.
+    "meson-python @ git+https://github.com/rgommers/meson-python.git@meson-cli-configurable"
 ]
 
+[tool.meson-python]
+cli = "vendored-meson/meson/meson.py"
```

</details>

If this looks good, it will still need docs before it's ready.

### DIFF
--- a/docs/reference/environment-variables.rst
+++ b/docs/reference/environment-variables.rst
@@ -59,6 +59,12 @@ Environment variables
    possible for ``meson-python`` to detect this, and it will not set the
    Python wheel platform tag accordingly.
 
+.. envvar:: MESON
+
+   Specifies the ``meson`` executable or script to use. It overrides
+   ``tool.meson-python.meson``. See :ref:`reference-pyproject-settings` for
+   more details.
+
 .. envvar:: MESONPY_EDITABLE_VERBOSE
 
    Setting this environment variable to any value enables directing to the
@@ -68,7 +74,7 @@ Environment variables
 
 .. envvar:: NINJA
 
-   Specify the ninja_ executable to use. It can also be used to select
+   Specifies the ninja_ executable to use. It can also be used to select
    ninja_ alternatives like samurai_.
 
 .. _ninja: https://ninja-build.org
--- a/docs/reference/pyproject-settings.rst
+++ b/docs/reference/pyproject-settings.rst
@@ -26,6 +26,14 @@ use them and examples.
    ``-Dpython.allow_limited_api=false`` option is passed to ``meson
    setup``.
 
+.. option:: tool.meson-python.meson
+
+   A string specifying the ``meson`` executable or script to use. If it is a
+   path to an existing file with a name ending in ``.py``, it will be invoked
+   as a Python script using the same Python interpreter that is used to run
+   ``meson-python`` itself. It can be overrridden by the :envvar:`MESON`
+   environment variable.
+
 .. option:: tool.meson-python.args.dist
 
    Extra arguments to be passed to the ``meson dist`` command.
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -600,7 +600,7 @@ def _validate_config_settings(config_settings: Dict[str, Any]) -> Dict[str, Any]
 class Project():
     """Meson project wrapper to generate Python artifacts."""
 
-    def __init__(  # noqa: C901
+    def __init__(
         self,
         source_dir: Path,
         build_dir: Path,
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -528,7 +528,15 @@ def _validate_pyproject_config(pyproject: Dict[str, Any]) -> Dict[str, Any]:
             raise ConfigError(f'Configuration entry "{name}" must be a boolean')
         return value
 
+    def _string_or_path(value: Any, name: str) -> str:
+        if not isinstance(value, str):
+            raise ConfigError(f'Configuration entry "{name}" must be a string')
+        if os.path.isfile(value):
+            value = os.path.abspath(value)
+        return value
+
     scheme = _table({
+        'meson': _string_or_path,
         'limited-api': _bool,
         'args': _table({
             name: _strings for name in _MESON_ARGS_KEYS
@@ -607,7 +615,22 @@ class Project():
         self._meson_args: MesonArgs = collections.defaultdict(list)
         self._limited_api = False
 
-        _check_meson_version()
+        # load pyproject.toml
+        pyproject = tomllib.loads(self._source_dir.joinpath('pyproject.toml').read_text())
+
+        # load meson args from pyproject.toml
+        pyproject_config = _validate_pyproject_config(pyproject)
+        for key, value in pyproject_config.get('args', {}).items():
+            self._meson_args[key].extend(value)
+
+        # meson arguments from the command line take precedence over
+        # arguments from the configuration file thus are added later
+        if meson_args:
+            for key, value in meson_args.items():
+                self._meson_args[key].extend(value)
+
+        # determine command to invoke meson
+        self._meson = _get_meson_command(pyproject_config.get('meson'))
 
         self._ninja = _env_ninja_command()
         if self._ninja is None:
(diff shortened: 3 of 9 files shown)


mesonpy/__init__.py
```diff
@@ -919,6 +926,34 @@ def _parse_version_string(string: str) -> Tuple[int, ...]:
         return (0, )
 
 
+def _get_meson_command(
+        meson_tool: str | None = None, *, version: str = _MESON_REQUIRED_VERSION
+    ) -> List[str]:
+    """Return the command to invoke meson.
```
dnicolodi, line 932: I usually leave an empty line between the function description and the extended docstring.

tests/packages/vendored-meson/meson.build
```diff
@@ -0,0 +1,13 @@
...
+
+project('vendored-meson', version: '1.0.0')
+
+py = import('python').find_installation()
+
+py.install_sources('pure.py')
+
+if get_option('custom-cli-used')
+    py.install_sources('extra/pure.py', subdir: 'extra')
+endif
```
dnicolodi, line 13: I think we could avoid having to build the wheel for the test using the Meson's `error()` function:
```meson
if not get_option('custom-cli-used')
    error('Expected option "custom-cli-used" was not specified')
endif
```
then the test package can be empty except for the meson wrapper script.

mesonpy/__init__.py
```diff
@@ -919,6 +926,34 @@ def _parse_version_string(string: str) -> Tuple[int, ...]:
...
+    Uses the MESON env var if set, else the `cli` entry in pyproject.toml if given,
+    else the standard `meson` executable.
+    """
+    _meson = os.environ.get('MESON')
+    if _meson is None:
+        if meson_tool is not None:
+            _meson = str(pathlib.Path(meson_tool).resolve())
```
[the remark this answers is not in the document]
rgommers: `absolute()` doesn't work well on Windows until some Python 3.10.x version I believe - I found this out the hard way before. I'm not sure what the best single explanation is, but https://discuss.python.org/t/pathlib-absolute-vs-resolve/2573 and https://bugs.python.org/issue39090 cover some ground. `.resolve()` works across all our supported Python versions I believe.

rgommers:
Thanks for the review @dnicolodi. All you comments seem good, and I like the `meson ` instead of `cli` name and potential future extension to `ninja`/`patchelf`. I'll update the code and will add some docs.

mesonpy/__init__.py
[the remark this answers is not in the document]
dnicolodi: Thanks, I didn't know about this. As this is a string to start with anyway, we could even use the old and proven `os.path.absolute()` but it is fine either way.

dnicolodi:
@rgommers Thinking about it, there is one problem with this patch: `meson` is listed as a dependency of `meson-python` 
https://github.com/mesonbuild/meson-python/blob/b4d42400b06f0f23420048dfe46e71a66f7fe54d/pyproject.toml#L36
thus, even when this mechanism is used to point to an alternative meson implementation, the meson dependency is installed or required to be installed when not using build isolation. Should the meson dependency removed from the static dependencies and returned from `get_requires_for_build_wheel()` if `meson` is not specified in `tool.meson.python` or the env var is not set?

mesonpy/__init__.py
```diff
@@ -528,7 +528,13 @@ def _bool(value: Any, name: str) -> bool:
             raise ConfigError(f'Configuration entry "{name}" must be a boolean')
         return value
 
+    def _str(value: Any, name: str) -> str:
+        if not isinstance(value, str):
+            raise ConfigError(f'Configuration entry "{name}" must be a string')
+        return value
+
     scheme = _table({
+        'cli': _str,
```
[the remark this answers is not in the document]
rgommers: When implementing this, I realized that it may not have to be a path to a file name - it could be an executable name as well. So both my initial check and this one are too strict. The logic should be:
- if it's a file, convert it to an absolute path
- else, assume it's an executable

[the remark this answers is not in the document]
dnicolodi: Right. If the normalization to the absolute path should be done only for the value in `pyproject.toml` you can still implement it here:
```python
def _string_or_path(value: Any, name: str) -> str:
    if not isinstance(value, str):
        raise ConfigError(f'Configuration entry "{name}" must be a string')
    path = patlib.Path(value)
    if path.is_file():
        path = path.resolve()
    return str(path)
```
or maybe without round tripping through `pathlib.Path`:

```python
def _string_or_path(value: Any, name: str) -> str:
    if not isinstance(value, str):
        raise ConfigError(f'Configuration entry "{name}" must be a string')
    if os.path.isfile(value):
        value = os.path.abspath(value)
    return value
```

dnicolodi:
Keeping the `meson` dependency is fine for me. I just wanted to make sure that we were not forgetting about it and keeping it by accident instead than by design.

rgommers:
Hmm, not sure what to think of this pre-commit complaint:
```
mesonpy/__init__.py:
  509:5  C901 `_validate_pyproject_config` is too complex (13 > 12)
```
It looks pretty clean to me.

tests/packages/vendored-meson/meson.build
```diff
@@ -0,0 +1,13 @@
+# SPDX-FileCopyrightText: 2023 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+project('vendored-meson', version: '1.0.0')
+
+py = import('python').find_installation()
+
+#py.install_sources('pure.py')
```
dnicolodi, line 9: Forgotten commented out line.

tests/packages/vendored-meson/meson_options.txt
```diff
@@ -0,0 +1,5 @@
+# SPDX-FileCopyrightText: 2023 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+option('custom-cli-used', type: 'boolean', value: false)
```
dnicolodi, line 5: I would call the option `custom-meson-used` to reflect the renamed configuration entry, but it is ok like this too.

docs/reference/environment-variables.rst
```diff
@@ -59,6 +59,12 @@ Environment variables
    possible for ``meson-python`` to detect this, and it will not set the
    Python wheel platform tag accordingly.
 
+.. envvar:: MESON
+
+   Specify the ``meson`` executable or script to use. It overrides
```
dnicolodi, line 64: `s/Specify/Specifies/`

docs/reference/pyproject-settings.rst
```diff
@@ -26,6 +26,17 @@ use them and examples.
...
+
+   A string identifying a custom version of ``meson``. It may be an executable
+   name, or a path to an executable or script. If it's a path to a file that
+   exists and the string ends in ``.py``, then it will be invoked as a script
+   with the same Python interpreter that is used to run meson-python itself.
+
+   If this setting is used, it will be used for all invocation of ``meson``.
+   It can be overridden by a ``$MESON`` environment variable (see
+   :ref:`reference-environment-variables`).
+
```
dnicolodi, line 39: I would avoid to write "version of meson" as it makes it look like what is specified is the version and not the executable. I like the wording in the env var description, and I think we can keep this a bit simpler without loss of information:
```
A string specifying the ``meson`` executable or script to use. If is it is a path to
an existing file with a name ending in ``.py``, it will be invoked as a Python script
using the same Python interpreter that is used to run ``meson-python`` itself. It
can be overrrideen by the :envvar:`MESON` environment variable.
```
The ``:envvar:`MESON` `` thing should link to the right place in the documentation, if I got the syntax right.

dnicolodi:
LGTM! Thanks for the revision @rgommers. I have just a couple of non critical comments.

rgommers:
> I would be inclined in disabling the check all together.

Agreed, done in the last commit.

rgommers:
All other changes adopted, I'll resolve those comments.

dnicolodi:
Thanks @rgommers

PR #502, opened 2023-09-21
added some documentation on how to use entry points

### LINKED ISSUE #500: How to define entry points
The [python packaging specifications](https://packaging.python.org/en/latest/specifications/entry-points/) include _entry points_, most notably `console_scripts` for creating executable console scripts from python packages.  

Old tools, such as `setuptools` have the functionality to define entry points. For meson-python, however, I only found a [stub](https://meson-python.readthedocs.io/en/latest/tutorials/entrypoints.html) in the documentation.

Is there a way to define entry points using meson/meson-python? If yes, are there any pointers on how to use it?

daknuett:
This fixes #500 and discussion #501.

I have tested the example code.

Thanks to @dnicolodi .

### DIFF
--- a/docs/tutorials/entrypoints.rst
+++ b/docs/tutorials/entrypoints.rst
@@ -8,10 +8,98 @@
 Entry points
 ************
 
-
 .. todo::
 
-   - Mention what is the purpose of entry points
-   - Give an example of a console entry point
    - Give an example of a custom entry point
      - Mention pluggy maybe for an example use-case?
+
+Introduction
+============
+
+Entry points provide a mechanism to advertise components of an installed
+distribution to other code or users. Most notably, the ``console_scripts``
+entry points will create a command line wrapper. 
+
+See the `PyPA documentation on entry points <https://packaging.python.org/en/latest/specifications/entry-points/>`_.
+
+Using entry points with meson-python
+====================================
+
+Entry points are defined in the ``pyproject.toml`` `metadata specification
+<https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#entry-points>`_.
+No further configuration is required when using ``meson-python``.
+
+Console entry point
+-------------------
+
+To show the usage of console entry points we build a simple
+python script: 
+
+.. code-block:: python
+   :caption: src/simpleapp/console.py
+
+    """
+    Simple test application.
+
+    Usage: 
+        simpleapp --help 
+        simpleapp doc
+        simpleapp run [<file>] [options]
+
+    Options:
+        -h --help                                   display this help message
+        --workdir-path=<workdir-path>               directory in which to run [default: none]
+
+    """
+    import docopt
+
+
+
+    def main():
+        args = docopt.docopt(__doc__)
+        # Just print the arguments for now.
+        print(args)
+
+    if __name__ == "__main__":
+        main()
+
+This script will be part of a larger python package (called ``simpleapp``).
+The meson build file is very simple and only installs the python sources.
+
+.. code-block:: meson
+   :caption: meson.build
+
+    project('simpleapp', version:'0.0.1')
+
+
+    pymod = import('python')
+    python = pymod.find_installation('python3')
+    pydep = python.dependency()
+
+    python.install_sources(
+      'src/simpleapp/__init__.py', 'src/simpleapp/console.py'
+      , pure: true
+      , subdir: 'simpleapp'
+    )
+
+The entry points are then specified in the ``pyproject.toml`` metadata specification.
+
+
+.. code-block:: toml
+   :caption: pyproject.toml
+
+    [project]
+    name = "simpleapp"
+    description = "simple app"
+    requires-python = ">=3.6"
+    dependencies = ["docopt"]
+    version = "0.0.1"
+
+    [project.scripts]
+    simpleapp = "simpleapp.console:main"
+
+    [build-system]
+    requires = ["meson", "toml", "ninja", "meson-python"]
+    build-backend = 'mesonpy'
+
+


dnicolodi:
`docs/tutorials/entrypoints.rst` is not linked to the documentation index, thus the proposed changes have no effect. Anyhow, I find the proposed addition way to heavy to describe the use of entry points, for which I'm pretty sure we don't need a whole tutorial. I think that a simple paragraph somewhere stating that `meson-python` implements https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata is enough.

daknuett:
@FFY00 what are your thoughts? I just implemented parts of the stub.

aminzayer (approved):

rgommers:
Entrypoints were not completely trivial to figure out when I needed them for NumPy, and the upstream docs are not very good for a user (and painfully formatted). So a small worked example would be nice to have in my opinion. I'd be happy to review this PR and get it merged.

This is a bit long on boilerplate though, and too short on the different types of entry points. TOML is also a bit weird here in that there's not a single `[project.entry-points]` table but that you must repeat it. E.g., for `numpy`:
```toml
[project.scripts]
f2py = 'numpy.f2py.f2py2e:main'

[project.entry-points.array_api]
numpy = 'numpy.array_api'

[project.entry-points.pyinstaller40]
hook-dirs = 'numpy:_pyinstaller_hooks_dir'
```

I'll add an initial review.

docs/tutorials/entrypoints.rst
```diff
@@ -8,10 +8,96 @@
...
      - Mention pluggy maybe for an example use-case?
+
+Introduction
+============
+
+Entry points provide a mechanism to advertise components of an installed
+distribution to other code or users. Most notably, the ``console_scripts``
+entry points will create a command line wrapper.
+
+See the `PyPA documentation on entry points <https://packaging.python.org/en/latest/specifications/entry-points/>`_.
```
rgommers, line 23: Those docs are actually outdated and confusing, so only linking to the `entry-points` section of https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata is better I'd say.

docs/tutorials/entrypoints.rst
```diff
@@ -8,10 +8,96 @@
...
-   - Mention what is the purpose of entry points
-   - Give an example of a console entry point
    - Give an example of a custom entry point
      - Mention pluggy maybe for an example use-case?
+
+Introduction
+============
+
+Entry points provide a mechanism to advertise components of an installed
+distribution to other code or users. Most notably, the ``console_scripts``
```
rgommers, line 20: I'd avoid using `console_scripts`, because it's what `setuptools` uses and using that in `pyproject.toml` is an error (the table is called `project.scripts`. So I'd say something like "there are three type of entry-points: those that provide commands to execute in a shell, commands to launch a GUI, and discoverable (plugin style) entry-points.

docs/tutorials/entrypoints.rst
```diff
@@ -8,10 +8,96 @@
...
+
+Entry points are defined in the ``pyproject.toml`` `metadata specification
+<https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#entry-points>`_.
+No further configuration is required when using ``meson-python``.
+
+Console entry point
+-------------------
+
+To show the usage of console entry points we build a simple
+python script:
```
rgommers, line 36: I suggest adding "that simply prints the arguments it was called with", and then remove the docstring and slim down the options used.

docs/tutorials/entrypoints.rst
```diff
@@ -8,10 +8,96 @@
...
+        simpleapp --help
+        simpleapp doc
+        simpleapp run [<file>] [options]
+
+    Options:
+        -h --help                                   display this help message
+        --workdir-path=<workdir-path>               directory in which to run [default: none]
+
+    """
+    import docopt
```
rgommers, line 54: It'd be better to use `argparse`, because it's in the stdlib - more well-known, and no separate install is needed to try this.

docs/tutorials/entrypoints.rst
```diff
@@ -8,10 +8,96 @@
...
+
+
+
```
rgommers, line 57: minor style: 2 blank lines above and below `def main():` please

docs/tutorials/entrypoints.rst
```diff
@@ -8,10 +8,96 @@
...
+.. code-block:: meson
+   :caption: meson.build
+
+    project('simpleapp', version:'0.0.1')
+
+
+    pymod = import('python')
+    python = pymod.find_installation('python3')
+    pydep = python.dependency()
```
rgommers, line 77: These 3 lines aren't quite idiomatic. Please use:
```meson
py = import('python').find_installation(pure: true)
py_dep = py.dependency()
```

docs/tutorials/entrypoints.rst
```diff
@@ -8,10 +8,96 @@
...
+
+    python.install_sources(
+      'src/simpleapp/__init__.py', 'src/simpleapp/console.py'
+      , pure: true
```
rgommers, line 81: Please remove `pure: true` here. And the line above needs to be a list, it's missing square brackets to group the two .py files

docs/tutorials/entrypoints.rst
```diff
@@ -8,10 +8,96 @@
...
+      , subdir: 'simpleapp'
+    )
+
+The entry points are then specified in the ``pyproject.toml`` metadata specification.
+
+
+.. code-block:: toml
+   :caption: pyproject.toml
+
```
rgommers, line 90: To make this self-contained, I suggest adding:
```toml
[build-system]
build-backend = 'mesonpy'
requires = ['meson-python']
```

docs/tutorials/entrypoints.rst
```diff
@@ -8,10 +8,96 @@
...
+    name = "simpleapp"
+    description = "simple app"
+    requires-python = ">=3.6"
+    dependencies = ["docopt"]
+    version = "0.0.1"
+
+    [project.scripts]
+    simpleapp = "simpleapp.console:main"
+
+    [build-system]
```
rgommers, line 101: This table typically comes at the top, since it's the most important one and is kinda why pyproject.toml exists at all. Can you move it up? Also, the dependency specification should be:

```toml
[build-system]
build-backend = 'mesonpy'
requires = ['meson-python']
```
The other dependencies should not be there.

docs/tutorials/entrypoints.rst
```diff
@@ -8,10 +8,96 @@
...
+    [project]
+    name = "simpleapp"
+    description = "simple app"
+    requires-python = ">=3.6"
+    dependencies = ["docopt"]
+    version = "0.0.1"
+
+    [project.scripts]
+    simpleapp = "simpleapp.console:main"
+
```
rgommers, line 100: I suggest adding two `project.entry-points.NAME` entries, that will be instructive. I suggest using the actual ones from numpy (`pyinstaller40` and `array_api`), because those are real-world things. The actual implementation can be a one-liner, something like:
```python
print("You discovered the PyInstaller plugin for simpleapp")
```

rgommers (changes requested):
Thanks for writing this @daknuett. I added some suggestions. In addition, you have to uncomment the entry for this file in `doc/index.rst` so the doc build includes it.

docs/tutorials/entrypoints.rst
daknuett: This part of the documentation is way better compared to the entry point specification.

I think the entry point specification is not outdated but just confusing. Apparently the entry point group is still called `console_scripts` but the table is called `scripts`.

docs/tutorials/entrypoints.rst
daknuett: Interesting; I always thought toml would be required for reading the `pyproject.toml` (which is of course redundant because without reading the file one could not determine the requirements). And meson and ninja would be required to to build the package.

rgommers: You never want to specify transitive dependencies of your actual direct dependencies. `meson-python` will need a toml reader indeed and it depends on `meson` (and has a mechanism to install `ninja` from PyPI when it's not already on the system). So those dependencies are in `meson-python`'s pyproject.toml (ninja is elsewhere, because it's a dynamic dependency for build envs only.

PR #508, opened 2023-10-03
ENH: count installed files starting at one

dnicolodi:
Counting starting at zero is nice and nerdy, but logging that meson-python is installing zero of zero files when is actually installing something is confusing.


rgommers (approved):
LGTM. Counting starting at one is clearly the right thing to do - this was probably a small bug rather than on purpose. And the small refactor is an improvement. In it goes. Thanks @dnicolodi

PR #515, opened 2023-10-09
TST: allow for unknown platforms

### LINKED ISSUE #514: test_wheel: account for `sys.platform`s that include `uname -r` appended
https://github.com/mesonbuild/meson-python/blob/8b0733fd85c98b7e8528e68edb86c629a0958adf/tests/test_wheel.py#L36
https://github.com/mesonbuild/meson-python/blob/8b0733fd85c98b7e8528e68edb86c629a0958adf/tests/test_wheel.py#L41
With `sys.platform` in the Unix-like systems category, only Linux, AIX, and Darwin do not append a version string derived from `uname -r`. The entire test suite fails to run at all on these other Unix-like systems due to the ensuing `KeyError`. This can be hacked around on a local system using `sys.platform.strip('0123456789')`, but wouldn't work for the `win32` case generally. Perhaps there's a better way to enumerate the different `LIB_SUFFIX`es proliferated?

dnicolodi:
Assume that the library filename suffix is ".so" on all platforms not otherwise accounted for. This fixes the tests on FreeBSD, and probably other BSDs.

Fixes #514.


dnicolodi:
@rgommers Would you like to review this? The fixes are straightforward so I'm going to merge in a couple of days unless you object

rgommers (approved):
Thanks for the fix and for adding a FreeBSD CI job @dnicolodi. The latter is quite useful, because it's a fairly non-standard platform.

Apologies for the delay - deadlines + holiday didn't help.

PR #518, opened 2023-10-18
DOC: Remove extra '.' in meson args examples

QuLogic:
There is already one at the end of the command (which matches the `pypa/build` version.)


rgommers (approved):
Thanks @QuLogic!

PR #520, opened 2023-10-21
BUG: Fix wheels not being compressed

### LINKED ISSUE #519: wheel files are not compressed
I haven't looked into it (yet), but while porting from setuptools to meson-python I noticed that the whl files are about twice as large as the setuptools created ones. Zipping the content manually without compression leads to the same file size, so it looks like there is no compression at all maybe (?)

lazka:
As pointed out in
https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.writestr
when creating a ZipInfo it defaults to ZIP_STORED and when passing it
to writestr it will use the compression info from the ZipInfo instance
instead of the ZipFile. This leads to the wheel file being
uncompressed.

To fix this explicitely specify the compression type and level for
writestr() so it gets applied to the ZipInfo instance before writing.
While the compression level currently isn't used I also added it to
avoid future surprises.

Fixes #519

### DIFF
--- a/tests/test_wheelfile.py
+++ b/tests/test_wheelfile.py
@@ -4,6 +4,7 @@
 
 import contextlib
 import time
+import zipfile
 
 import wheel.wheelfile
 
@@ -41,3 +42,16 @@ def test_source_date_epoch(tmp_path, monkeypatch):
     with wheel.wheelfile.WheelFile(path, 'r') as w:
         for entry in w.infolist():
             assert entry.date_time == time.gmtime(epoch)[:6]
+
+
+def test_compression(tmp_path):
+    # create a wheel and test that everything is compressed
+    path = tmp_path / 'test-1.0-py3-any-none.whl'
+    bar = tmp_path / 'bar'
+    bar.write_bytes(b'bar')
+    with mesonpy._wheelfile.WheelFile(path, 'w') as w:
+        w.writestr('foo', b'test')
+        w.write(bar, 'bar')
+    with zipfile.ZipFile(path, 'r') as w:
+        for entry in w.infolist():
+            assert entry.compress_type == zipfile.ZIP_DEFLATED
--- a/mesonpy/_wheelfile.py
+++ b/mesonpy/_wheelfile.py
@@ -87,7 +87,10 @@ class WheelFileWriter(WheelFile):
         else:
             zinfo = zipfile.ZipInfo(zinfo_or_arcname, date_time=self.timestamp())
             zinfo.external_attr = 0o664 << 16
-        self.archive.writestr(zinfo, data)
+        self.archive.writestr(
+            zinfo, data,
+            compress_type=self.archive.compression,
+            compresslevel=self.archive.compresslevel)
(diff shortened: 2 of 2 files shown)


lazka:
In my case this reduced the file size for one wheel from ~500kb to ~130kb

tests/test_wheelfile.py
```diff
@@ -41,3 +42,16 @@ def test_source_date_epoch(tmp_path, monkeypatch):
...
+
+def test_compression(tmp_path):
+    # create a wheel and test that everything is compressed
+    path = tmp_path / 'test-1.0-py3-any-none.whl'
+    bar = tmp_path / 'bar'
+    bar.write_bytes(b'bar')
+    with mesonpy._wheelfile.WheelFile(path, 'w') as w:
+        w.writestr('foo', b'test')
+        w.write(bar, 'bar')
+    with wheel.wheelfile.WheelFile(path, 'r') as w:
```
dnicolodi, line 55: I would use `ZipFile` directly here. I think that the `RECORD` file in the zip archive can be compressed too, but you patch does not do that, see the `close()` method, however, `WheelFile.inforlist()` does not yield the `RECORD` file and the test misses it.

dnicolodi (changes requested):
Good catch. I'm surprised we didn't notice this before: `meson-python` is used for some large projects! Thanks for the patch. It looks good, but i think you missed one instance of `.writestr()` that needs handling, and the test needs to be updated to catch it. Other than that, can you please wrap the commit message at 72 characters? And, just to be extra picky, we use a convention where "Fix" in the commit message subject line is lowercase not capitalized. Thanks!

tests/test_wheelfile.py
lazka: Good point, I've changed it to Zipfile now.

But I don't understand the other parts, close() is patched and inforlist() is never overridden anywhere and shouldn't have made a difference.

dnicolodi: Yes, sorry, I misread the patch. I assumed the methods patched where `write()` and `writestr()`.

lazka:
> Good catch. I'm surprised we didn't notice this before: `meson-python` is used for some large projects!Thanks for the patch. It looks good, but i think you missed one instance of `.writestr()` that needs handling, and the test needs to be updated to catch it.

I can only see two calls to writestr() and both are handled :/

> Other than that, can you please wrap the commit message at 72 characters? And, just to be extra picky, we use a convention where "Fix" in the commit message subject line is lowercase not capitalized. Thanks!

done

dnicolodi:
> I can only see two calls to writestr() and both are handled :/

Yes, sorry, my bad. Not enough coffee this morning yet.

dnicolodi (approved):

lazka:
Thanks!

rgommers:
Thanks a lot, great catch!

> I'm surprised we didn't notice this before: `meson-python` is used for some large projects!

I think why I didn't notice is that we always run numpy/scipy wheels through `auditwheel` & co, and those tools will write out a new compressed wheel with dependencies vendored, so in those cases it didn't matter. Without that, the difference is very large indeed.

PR #521, opened 2023-10-24
DOC/REL: prepare 0.15.0 changelog and release

rgommers:
We accumulated a few important bug fixes and new features, so it seems time for a release.

@dnicolodi I set the date to today - in case it gets merged later, that date should be updated right before the merge.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,38 @@
 Changelog
 +++++++++
 
+0.15.0
+======
+
+- Enable compression for wheel files. The may result in several times
+  smaller wheels.
+- Require Meson 1.2.3 or later on Python 3.12 or later. Meson 1.2.3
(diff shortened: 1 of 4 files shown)


dnicolodi:
@rgommers Thanks for working on this. The changelog entries for other releases use the present tense, but the ones for 0.15.0 used the past tense. I took the freedom to fix them to use the present tense and to do some rewording PTAL.

rgommers:
Thanks @dnicolodi, LGTM! Let me just bump the date once more, and then I'll get this release done.

dnicolodi:
Thanks @rgommers. The changelog in the present tense reads a bit strange to me too, but this is how it was started, and I think there is value in keeping it all in the same form.

rgommers:
Yes, agreed. We may want to change it at some point, but let's consider it only if we make other changes like adding change categories (bug fix, enhancement, backwards compatibility, etc.).

Issue #522, opened 2023-10-28
readthedocs stuck to old version


dnicolodi:
The documentation at https://meson-python.readthedocs.io/en/latest/ is stuck to some old version of the documentation.

Also, https://meson-python.readthedocs.io/ redirects to the latest documentation, which is the one build from the main branch, not from the latest release. The only other version provided by readthedocs is the stable version (which currently points to version 0.13.2, thus a couple of releases ago). Other than getting the latest version built and uploaded, we should probably make the stable version the default.

Do we really need to rely on readthedocs for the documentation? I think that for the basic needs of the project, using Github Pages would be more than adequate and it would be much easier to make sure that the published documentation stays up to date as the build and upload would be integrated in the CI infrastructure.

rgommers:
Thanks - I opened gh-523 to patch it up quickly. Also, do you have a RTD account/username? If so, I can add you to the project there.

> Do we really need to rely on readthedocs for the documentation? I think that for the basic needs of the project, using Github Pages would be more than adequate and it would be much easier to make sure that the published documentation stays up to date as the build and upload would be integrated in the CI infrastructure.

Maybe easier CI wise, however deployment is also a pain. Using a `gh-pages` branch within the same repo makes a mess, and deploying to another repo is better (no spurious automated commits in the main repo) but not well-supported by GitHub Actions. And if you want to store the built docs for multiple versions, it gets worse. Hence I'd be inclined to not touch it.

The other issue here is that the notifications on build failures were not enabled for me (they are now), otherwise I would have easily caught this problem.

rgommers:
Reopening for the "shall we move to Github Pages" part.

dnicolodi:
I agree that the extra branch was a pain, but that is not needed anymore. See for example https://github.com/dnicolodi/python-siphash24/blob/master/.github/workflows/docs.yml

As for the multiple versions, we are just using "latest" pointing to the main branch (which IMO is the wrong choice) and "stable" pointing to the latest release. I don't think that having the docs for the development version published is useful (building the docs locally is trivial) and having the docs just built for the latest release would be straightforward.

dnicolodi:
Can we have Read The Docs use the "stable" version by default, instead of "latest"?

dnicolodi:
Something is still wrong with RTD: https://meson-python.readthedocs.io/en/stable/ still has the documentation for 0.13.2, not 0.15.0.

rgommers:
Yes, that's easy:

[...]

PR #523, opened 2023-10-28
DOC: fix Read the Docs builds

rgommers:
The problem was that they migrated to a new yaml version, plus that notifications on build failures in the RTD UI was not enabled for anyone.

Closes gh-522

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -4,8 +4,12 @@
 
 version: 2
 
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
 python:
-  version: 3.8
   install:
     - method: pip
       path: .


dnicolodi:
@rgommers I don't have a way to check that this does the right thing, thus please merge it when you are happy.

dnicolodi:
I think that using readthedocs does not buy us anything and having the docs built on another platform is more trouble, thus I would be happy to migrate to Github Pages for hosting the documentation.

rgommers:
> I think that using readthedocs does not buy us anything and having the docs built on another platform is more trouble, thus I would be happy to migrate to Github Pages for hosting the documentation.

Let's discuss on gh-522? Happy to consider it of course.

> I don't have a way to check that this does the right thing, thus please merge it when you are happy

We used to have a CI job for this, but it's no longer running. Not sure why, probably the webhook needs an update. I'll wait with that until we resolve gh-522.

rgommers:
This passes says the RTD interface, so I'll merge it.

Issue #525, opened 2023-11-04
Building Windows wheels with libraries fails


jpakkane:
When building a CapyPDF releases I now get the following error on Windows:

```
Traceback (most recent call last):
  File "C:\python\lib\site-packages\pyproject_hooks\_in_process\_in_process.py", line 353, in <module>
    main()
  File "C:\python\lib\site-packages\pyproject_hooks\_in_process\_in_process.py", line 335, in main
    json_out['return_val'] = hook(**hook_input['kwargs'])
  File "C:\python\lib\site-packages\pyproject_hooks\_in_process\_in_process.py", line 251, in build_wheel
    return _build_backend().build_wheel(wheel_directory, config_settings,
  File "C:\Users\Jussi Pakkanen\AppData\Local\Temp\build-env-t_mdgsd0\lib\site-packages\mesonpy\__init__.py", line 985, in wrapper
    return func(*args, **kwargs)
  File "C:\Users\Jussi Pakkanen\AppData\Local\Temp\build-env-t_mdgsd0\lib\site-packages\mesonpy\__init__.py", line 1039, in build_wheel
    return project.wheel(out).name
  File "C:\Users\Jussi Pakkanen\AppData\Local\Temp\build-env-t_mdgsd0\lib\site-packages\mesonpy\__init__.py", line 890, in wheel
    return builder.build(directory)
  File "C:\Users\Jussi Pakkanen\AppData\Local\Temp\build-env-t_mdgsd0\lib\site-packages\mesonpy\__init__.py", line 455, in build
    self._install_path(whl, src, dst)
  File "C:\Users\Jussi Pakkanen\AppData\Local\Temp\build-env-t_mdgsd0\lib\site-packages\mesonpy\__init__.py", line 414, in _install_path
    mesonpy._rpath.fix_rpath(origin, libspath)
  File "C:\Users\Jussi Pakkanen\AppData\Local\Temp\build-env-t_mdgsd0\lib\site-packages\mesonpy\_rpath.py", line 64, in fix_rpath
    raise NotImplementedError(f'Bundling libraries in wheel is not supported on {sys.platform}')
NotImplementedError: Bundling libraries in wheel is not supported on win32

ERROR Backend subprocess exited when trying to invoke build_wheel
```

Ths is on 3.12.0, but I also tried downgrading to 3.11 and 3.10.11 (which used to work  when I made the previous release AFAICR). Buildng the package on macOS 3.10.5 works without problems.

To reproduces:

[...]

PR #526, opened 2023-11-05
BUG: fix detection of platform dependent components in wheel

dnicolodi:
Correctly detect a wheel shipping shared libraries as being platform dependent also when it does not ship compiled executables or Python extension modules.


rgommers (approved):
LGTM, thanks @dnicolodi.

For the case of gh-525, with this PR the wheel name now looks correct: `capypdf-0.7.0-py3-none-macosx_14_0_arm64.whl`.

PR #527, opened 2023-11-05
BUG: use 11.0 as minimum macOS platform ABI tag on arm64

### LINKED ISSUE #524: Limit macOS platform tag by compiled deployment target?
We were recently notified that our moved-to-Meson nightly [macOS arm64 wheels do not install](https://github.com/matplotlib/matplotlib/issues/27243). This is because they have a platform tag with `10_12`, but arm64 only existed on macOS 11 and above, so those wheels are never considered. These wheels are built with cibuildwheel using x86_64 and cross-compiled arm64 and `MACOSX_DEPLOYMENT_TARGET=10.12`. So the `10_12` does make sense from the environment variable point of view.

_However_, our older wheels built with setuptools use this same environment, but get `11_0` in their platform tag.

Some digging shows that `wheel` actually [looks at all `.dylib` and `.so` in the produced wheel](https://github.com/pypa/wheel/blob/a899f1c69ffddf564b904a5de974457c213dc36a/src/wheel/macosx_libfile.py#L426-L438), and overrides the platform tag to pick a minimum version that matches the minimum deployment target of the compiled extensions. See also the pypa/wheel#314.

A [comment elsewhere](https://github.com/pypa/wheel/blob/a899f1c69ffddf564b904a5de974457c213dc36a/src/wheel/macosx_libfile.py#L310) indicates that Xcode will not produce a deployment target below 11 for arm64 builds.

So these two things work together in a setuptools-based build to produce a platform tag with `11_0` on Apple Silicon, even though the deployment target environment variable is set lower.

For Matplotlib, we can work around this by splitting the job by arch and setting `MACOSX_DEPLOYMENT_TARGET` as necessary.

But for meson-python, since it re-implements `wheel`, the question here is whether it should _also_ implement the check on compiled extensions, and then pick a platform tag that reflects what's in the wheel?

dnicolodi:
Fixes #524.

### DIFF
--- a/mesonpy/_tags.py
+++ b/mesonpy/_tags.py
@@ -129,6 +129,16 @@ def _get_macosx_platform_tag() -> str:
     # can therefore ignore the issue and generate the slightly
     # incorrect tag.
 
+    # The minimum macOS ABI version on arm64 is 11.0.  The macOS SDK
+    # on arm64 silently bumps any compatibility version specified via
+    # the MACOSX_DEPLOYMENT_TARGET environment variable to 11.0.
+    # Despite the platform ABI tag being intended to be a minimum
+    # compatibility version, pip refuses to install wheels with a
+    # platform tag specifying an ABI version lower than 11.0.  Use
+    # 11.0 as minimum ABI version on arm64.
+    if arch == 'arm64' and version < (11, 0):
+        version = (11, 0)
+
     major, minor = version
 
     if major >= 11:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -80,3 +80,9 @@ freebsd_task:
     image_family: freebsd-13-2
   install_script: pkg install -y git ninja
   << : *test
+
+macos-arm64_task:
+  macos_instance:
+    image: ghcr.io/cirruslabs/macos-ventura-xcode
+  install_script: brew install git ninja python@3.11
+  << : *test
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -5,6 +5,7 @@
 import importlib.machinery
 import os
 import pathlib
+import platform
 import sys
 import sysconfig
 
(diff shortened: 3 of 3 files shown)


mesonpy/_tags.py
```diff
@@ -129,6 +129,16 @@ def _get_macosx_platform_tag() -> str:
     # can therefore ignore the issue and generate the slightly
     # incorrect tag.
 
+    # The minimum macOS ABI version on arm64 is 11.0.  The macOS SDK
+    # or arm64 silently bumps any compatibility version specified via
```
rgommers, line 133: typo "on arm64"

mesonpy/_tags.py
```diff
@@ -129,6 +129,16 @@ def _get_macosx_platform_tag() -> str:
...
+    # the MACOSX_DEPLOYMENT_TARGET environment variable to 11.0.
+    # Depsite the platform ABI tag being intended to be a minimum
```
rgommers, line 135: typo: Despite

mesonpy/_tags.py
```diff
@@ -129,6 +129,16 @@ def _get_macosx_platform_tag() -> str:
...
+    # compatibiloty version, pip refuses to install wheels with a
```
rgommers, line 136: typo: compatibility

rgommers (changes requested):
Thanks @dnicolodi. This fix LGTM in principle, however it does make an older test fail:
```
    def test_macos_platform_tag(monkeypatch):
        for minor in range(9, 16):
            monkeypatch.setenv('MACOSX_DEPLOYMENT_TARGET', f'10.{minor}')
>           assert next(packaging.tags.mac_platforms((10, minor))) == mesonpy._tags.get_platform_tag()
E           AssertionError: assert 'macosx_10_9_arm64' == 'macosx_11_0_arm64'
E             - macosx_11_0_arm64
E             ?         ^ ^
E             + macosx_10_9_arm64
E             ?   
```

This highlights that we have a test coverage gap that we should fix (can be done later, I can test locally for now). We're under the `mesonbuild` org, so it's not possible to see our credits usage on Cirrus CI - and hence I'm not sure if we can add `arm64` on Cirrus. Meson itself doesn't use Cirrus though, so it seems like we can try without causing an issue; worst-case we'd exhaust free credits in one month.Or, we wait for GitHub Actions to finally roll out support.

Shall I fix up the tests locally and push that?

dnicolodi:
The test failure is odd. I would have thought that `packaging` would refuse to generate a tag on macOS with an ABI version lower than 11.0 on arm64. I'll see if there is a more suitable `packaging` function to use in this test.

> This highlights that we have a test coverage gap that we should fix (can be done later, I can test locally for now).

It mostly highlights that my laptop is still an Intel mac. We don't test on architectures other than amd64 for any other platform. Extending the CI to run on other architectures would be nice, but not very high priority.

dnicolodi:
> We're under the `mesonbuild` org, so it's not possible to see our credits usage on Cirrus CI - and hence I'm not sure if we can add `arm64` on Cirrus. Meson itself doesn't use Cirrus though, so it seems like we can try without causing an issue; worst-case we'd exhaust free credits in one month.

AFAIK Cirrus CI does not use a credit system for using their service within the limits allocated to public repositories. https://cirrus-ci.org/faq/#are-there-any-limits So far we don't use any macOS worker on Cirrus CI so far, so we can setup one to run on arm64 and still have all the workers run in parallel

dnicolodi:
Correction. The page I linked to actually reports that there is also a monthly usage cap. I don't think we are likely to hit the 10000 CPU minutes limit for Linux tasks, but the 500 minutes per month for macOS task is not very high. I'll setup the macOS arm64 jobs and see how we do.

dnicolodi:
There does not seem to be a better `packaging` API to use for the test. I fixed it the ugly way. The macOS arm64 CI job seems to be running fine.

rgommers (approved):
LGTM now, thanks @dnicolodi!

PR #528, opened 2023-11-05
MAINT: use placeholder email address for Thomas Li in pyproject.toml

dnicolodi:
This is the same email address that is used in the for the git commits. This should fix how the Maintainers metadata field is rendered on the PyPI package page.


rgommers (approved):
Thanks! LGTM, let's give this a go.

PR #529, opened 2023-11-07
DOC: deploy documentation using Github Pages

dnicolodi:
This is very simple and does not allow for multiple versions of the documentation to be published. I think it would not be too hard to add, but I don't think it would add much value for the meson-python documentation.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -6,6 +6,8 @@ name: docs
 
 on:
   push:
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+'
   pull_request:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -14,14 +16,37 @@ on:
...
+    needs: build
+    permissions:
+      pages: write
+      id-token: write
+    environment:
+      name: github-pages
+    runs-on: ubuntu-latest
+    if: ${{ github.event.push.tag || github.event.inputs.tag }}
+    steps:
+      - uses: actions/deploy-pages@v1
```
rgommers, line 52: any reason not to use `v2` here? Latest

rgommers (approved):
Thanks @dnicolodi, this LGTM. Only one minor question about using the latest version of the deploy action.

> does not allow for multiple versions of the documentation to be published. I think it would not be too hard to add, but I don't think it would add much value for the meson-python documentation.

I think we can cross that bridge when we get to it. For now it should be okay to have only the latest released version. Having a development docs version is the second most useful, and older releases after that. But that indeed adds a fair bit of complexity. As do doc build previews on PRs. We can leave all that out, and get this going.

Also, regarding logistics - I think we merge this, build the latest tag, ensure it shows up on https://mesonbuild.com/meson-python, and if it all looks good then we update the doc links in README and `pyproject.toml` in a follow-up PR.

I did enable Pages already in the repo settings, so this should be good to go I believe.

.github/workflows/docs.yml
dnicolodi: No reason other than I copy-pasted most of this from a project of mines and I didn't check if there has been another version released since I wrote this the first time. Let me check if v2 has any important change before moving to it.

dnicolodi: Updated the actions to v2.

rgommers:
Great, thanks!

PR #531, opened 2023-11-09
Allow duplicate -arch in ARCHFLAGS

matthew-brett:
We came across a situation where the build added duplicate archs in
ARCHFLAGS, e.g. `-arch arm64 -arch arm64`.  These usually pass through
compilers etc without problem, but mesonpy was raising in that case,
accusing the build of being multiarch.   Fix and test.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -647,7 +647,7 @@ class Project():
             archflags = os.environ.get('ARCHFLAGS', '').strip()
             if archflags:
                 arch, *other = filter(None, (x.strip() for x in archflags.split('-arch')))
-                if other:
+                if [a for a in other if a != arch]:
                     raise ConfigError(f'Multi-architecture builds are not supported but $ARCHFLAGS={archflags!r}')
                 macver, _, nativearch = platform.mac_ver()
                 if arch != nativearch:
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -10,8 +10,9 @@ import subprocess
 import sys
 import sysconfig
 import textwrap
-from pathlib import Path
+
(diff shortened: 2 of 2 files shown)


dnicolodi:
I am not convinced that we should fix this. `ARCHFLAGS` is a `setuptools` thing that we support for compatibility, the compiler or other parts of the toolchain do not know anything about it. IIRC `setuptools` allows you to pass any flag accepted by the compiler via `ARCHFLAGS`, meson-python does not allow this. I think it is much easier to fix your build scripts.

matthew-brett:
So - is there a specific reason that an error is desirable here?    I can't see any situation where the error would be anything but spurious.  And if that is so, it seems to me unhelpful not to fix it, and so to force all your users to (independently) discover the problem, debug it, and find their own workarounds.

dnicolodi:
Why would anyone set `ARCHFLAGS='-arch arm64 -arch arm64'`? It is much easier to fix the build script that generates this environment variable than to fix meson-python to accept it and wait till the next meson-python release to be able to run your builds.

As I explained, `setuptools` allows to pass anything via `ARCHFLAGS` but `meson-python` does not. Thus there are more sensible setting for the environment variable that are refused. I agree that the parsing of the environment variable and the error messages can be improved, but I don't think handling arbitrary flags to be passed via `ARCHFLAGS` is desirable.

matthew-brett:
Could I ask you again - what is the purpose of the error?    Who would be served by it?     Just a personal opinion - but I think build systems really have to be careful not to trip people up for things that won't in fact cause a problem.

dnicolodi:
Better fix in #532

matthew-brett:
Thanks - that's very helpful.

Issue #535, opened 2023-11-16
Build fails on windows when one of the bundled data files has non-ascii characters


ankith26:
This error only happens on windows, and only happens when I try to include a file that has some korean characters in its path string. Perhaps the fix to this could be to force utf-8 usage? Apologies in advance if this is not actually a meson-python issue, I'm just assuming it is from the traceback
```
Traceback (most recent call last):
 File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-a3rpdcsp\cp310-win_amd64\build\venv\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 353, in <module>
   main()
 File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-a3rpdcsp\cp310-win_amd64\build\venv\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 335, in main
   json_out['return_val'] = hook(**hook_input['kwargs'])
 File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-a3rpdcsp\cp310-win_amd64\build\venv\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 152, in prepare_metadata_for_build_wheel
   whl_basename = backend.build_wheel(metadata_directory, config_settings)
 File "C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-1c4y2iz0\overlay\Lib\site-packages\mesonpy\__init__.py", line 985, in wrapper
   return func(*args, **kwargs)
 File "C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-1c4y2iz0\overlay\Lib\site-packages\mesonpy\__init__.py", line 1039, in build_wheel
   return project.wheel(out).name
 File "C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-1c4y2iz0\overlay\Lib\site-packages\mesonpy\__init__.py", line 890, in wheel
   return builder.build(directory)
 File "C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-1c4y2iz0\overlay\Lib\site-packages\mesonpy\__init__.py", line 445, in build
   counter.update(src)
 File "C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-1c4y2iz0\overlay\Lib\site-packages\mesonpy\_util.py", line 69, in update
   print(line)
[...]

PR #536, opened 2023-11-16
BUG: use errors=replace for stdout

dnicolodi:


dnicolodi:
Unfortunately the added test catches the issue reported in #535 only when run without pytest stdout capture (namely with the `-s` command line flag).

ankith26:
I can confirm that this PR fixes my issue. Instead of an error and build fail now, I get question mark symbols (`?`) in stdout in place of the korean characters, and the build finishes correctly.

dnicolodi:
> I can confirm that this PR fixes my issue. Instead of an error and build fail now, I get question mark symbols (`?`) in stdout in place of the korean characters, and the build finishes correctly.

Thanks for testing. We can choose other alternative representations for characters that cannot be represented in the current encoding. I went for the `?` because it is the simplest. Other options that make sense are 'backslashreplace' or 'namereplace'. https://docs.python.org/3/library/codecs.html#error-handlers I don't have an opinion about which one is the best choice.

ankith26:
IMHO the `'replace'` error handling is fine

rgommers:
> I went for the `?` because it is the simplest

+1 from me, seems like a good choice.

mesonpy/_util.py
```diff
@@ -71,3 +71,20 @@ def update(self, description: str) -> None:
...
+    ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x04
+
+    kernel = windll.kernel32
+    stdout = kernel.GetStdHandle(STD_OUTPUT_HANDLE)
+    mode = DWORD()
+
+    if not kernel.GetConsoleMode(stdout, byref(mode)):
+        return False
+
+    return bool(kernel.SetConsoleMode(stdout, mode.value | ENABLE_VIRTUAL_TERMINAL_PROCESSING))
```
rgommers, line 90: The one difference with `mesonbuild/mlog.py` is that that adds `or os.environ.get('ANSICON')` (see https://github.com/adoxa/ansicon/blob/37f92009f6ca4938a34a1d7cecff41d331c7423c/readme.txt#L119 for the docs of that env var). It's not clear to me that that case is covered by the checks before calling `setup_windows_console` here - should it be added?

rgommers (approved):
Looks great, thanks @dnicolodi. This looks fine to merge as is, and quite nice that we can drop a dependency. Just one question about the one difference with the corresponding code in Meson for color handling.

dnicolodi:
I'm moving dropping the colorama dependency to another PR and merging this one.

mesonpy/_util.py
dnicolodi: I think it should be handled indeed. I had the impression that ansicon was some legacy thing, but I checked and the `ANSICON` environment variable is used by other terminal emulators on Windows. I'll add checking for the environment variable.

PR #542, opened 2023-11-18
DOC: remove stubs

dnicolodi:
The main purpose of the exercise is to remove the warnings when building the documentation.


rgommers (approved):
Thanks @dnicolodi! This is a nice cleanup - having the doc build clean of warnings and enforcing that in CI is quite useful. There are open issues for the main things that still need to be documented (thanks for opening those), so nothing is really lost by removing these empty/todo pages.

PR #544, opened 2023-11-25
Add project 'werpy' to list of projects using meson-python.

rossarmstrong:
This pull request proposes the addition of our package, werpy, to the list of working examples in the meson-python documentation.

Werpy is a fast, lightweight Python package for calculating and analyzing Word Error Rate (WER) between two sets of text.

This package demonstrates a successful use case of meson-python for building and distributing Python packages with C extensions and provides a good use case for others to view.

### DIFF
--- a/docs/projects-using-meson-python.rst
+++ b/docs/projects-using-meson-python.rst
@@ -30,6 +30,13 @@ Here's a curated list of projects using ``meson-python``.
        an example of how to use `cibuildwheel`_ to produce Python wheels for
        several platforms.
 
+   * - `werpy <https://github.com/analyticsinmotion/werpy>`_
+     - This project provides a straightforward example for utilizing
+       meson-python to package a Python project with compiled extensions. 
+       It shows a practical use case of integrating optimized native code 
+       written in Cython_ into a Python package through seamless 
+       cross-platform compilation.
+
 
 .. _CPython extension: https://docs.python.org/3/extending/extending.html
 .. _CPython extensions: https://docs.python.org/3/extending/extending.html


dnicolodi:
I had a quick look at the project and I don't see it adding anything to what is already demonstrated by the other projects listed in the examples page. Actually, it does not use the recommended way to integrate Cython in the build process.

The projects listed so far have strong involvement form the meson-python maintainers thus we can be fairly sure that they follow best practices without having to review them constantly. I don't think we can realistically open this up to list arbitrary project, unless there are very good reasons for doing so.

docs/projects-using-meson-python.rst
```diff
@@ -30,6 +30,13 @@ Here's a curated list of projects using ``meson-python``.
        an example of how to use `cibuildwheel`_ to produce Python wheels for
        several platforms.
 
+   * - `werpy <https://github.com/analyticsinmotion/werpy>`_
+     - This project provides a straightforward example for utilizing
+       meson-python to package a Python project with compiled extensions. 
+       It shows a practical use case of integrating optimized native code 
+       written in Cython_ into a Python package through seamless 
+       cross-platform compilation.
```
dnicolodi, line 38: What do you mean with "cross-platform" here? meson-python does not support cross platform compilation

rossarmstrong: Hi @dnicolodi, appreciate the quick reply. I have replied to your main points below:

- Opening up the list in your documentation - Can completely understand and agree with your points on this. Happy for this PR to be closed.

- Cross-platform comment - it relates more to the fact our package uses cibuildwheel to run across linux, macOS and windows, rather than anything from meson-python.

- Best practice implementation - we have done similar projects with setuptools previously, but this was our first with meson-python. We reviewed several of the bigger projects that use meson-python. We would obviously want to follow the recommended method for Cython integration. Could you point us in the right direction (e.g. a project, or part of documentation etc.) to assist in following the recommended way to integrate Cython in the build process? We have several other projects in the pipeline that would benefit greatly from this.
Thanks again for your comments.

dnicolodi: > * Cross-platform comment - it relates more to the fact our package uses cibuildwheel to run across linux, macOS and windows, rather than anything from meson-python.

"cross-platform compilation" is usually used to mean compiling on one platform for execution on another, for example compiling on Linux for execution on Windows via Cygwin on MinGW. 

> * Could you point us in the right direction (e.g. a project, or part of documentation etc.) to assist in following the recommended way to integrate Cython in the build process?

They are listed in the examples page in the documentation. For example: https://github.com/dnicolodi/python-siphash24/blob/eba0bf7c1babf2e4991525f8cb49148c65bb2f82/meson.build In a few words: add `cython` to the languages used by the project in the Meson `project()` definition and pass the Cython source code directly to the `py.extension_module()` definition.

rossarmstrong: Hi @dnicolodi, I appreciate the time you have taken to respond with these examples. We will definitely be reviewing the code again this week. Thanks again.

eli-schwartz: > "cross-platform compilation" is usually used to mean compiling on one platform for execution on another, for example compiling on Linux for execution on Windows via Cygwin on MinGW.

No, that's cross-compilation. Cross-platform (anything) typically means the source code is compatible with many platforms, not just one.

That being said I would usually expect to see the word "support" not "compilation" in that context. The use of "compilation" here makes it resemble the term "cross-compilation" (just elide the "platform" word).

PR #545, opened 2023-11-25
ENH: drop dependency on colorama on Windows

dnicolodi:
Simply enable ANSI escape sequences processing on the Windows console. This works only on Windows 10 and later, and colorama does the same thing when it can.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -18,6 +18,7 @@ import difflib
 import functools
 import importlib.machinery
 import io
+import itertools
 import json
 import os
 import pathlib
@@ -56,7 +57,7 @@ from mesonpy._compat import cached_property, read_binary
 if typing.TYPE_CHECKING:  # pragma: no cover
     from typing import Any, Callable, DefaultDict, Dict, List, Literal, Optional, Sequence, TextIO, Tuple, Type, TypeVar, Union
 
-    from mesonpy._compat import Collection, Iterator, Mapping, ParamSpec, Path
+    from mesonpy._compat import Collection, Iterator, Mapping, ParamSpec, Path, Self
 
     P = ParamSpec('P')
     T = TypeVar('T')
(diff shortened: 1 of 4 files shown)


rgommers:
I finally have access to a Windows machine, so I tested this. There is still something wrong here, both in this PR and on `main`. When trying on a fresh Windows 11 install in the default Powershell terminal, I see that both the terminal itself, Pip and Meson show colors. But meson-python does not, and Meson also doesn't.

I also see one instance where an ansi escape code leaks into the output, for error messages. In an env without ninja installed, I get:
```
> pip install . --no-build-isolation
Processing c:\users\ralfg\code\meson-python
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [2 lines of output]

      ('\x1b[31m',)meson-python: error: Could not find ninja version 1.8.2 or newer.
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.
```

The `pip`-written lines contain color, the meson-python error does not:

image

I don't have other terminals like Git Bash installed yet.

This is still a bit puzzling, things are inconsistent:
- `meson setup build` in a fresh terminal does show color
- `python -m build -wnx` shows color too
- `pip install . --no-build-isolation -v` does NOT show color

There is no regression with respect to `main`, so this PR could be merged as is. But it may be preferable to address the error ansi code issue in this PR too.

dnicolodi:
> When trying on a fresh Windows 11 install in the default Powershell terminal, I see that both the terminal itself, Pip and Meson show colors. But meson-python does not, and Meson also doesn't.

This is because `pip` redirects stdout to something that is not a tty. This is done because the command output is shown only in case of error. I don't think there is much we can do about this.

> I also see one instance where an ansi escape code leaks into the output, for error messages.

I'm looking into this.

dnicolodi:
> I also see one instance where an ansi escape code leaks into the output, for error messages.

Fixed.

rgommers (approved):
LGTM now, thanks @dnicolodi!

The one CI failure (homebrew 3.8) is unrelated, so in it goes.

PR #551, opened 2023-12-09
ENH: support shared libraries on Windows when explicitly enabled

dnicolodi:


docs/reference/pyproject-settings.rst
```diff
@@ -34,6 +34,17 @@ use them and examples.
...
    environment variable.
 
+.. option:: tool.meson-python.shared-libs-win32
+
+   A boolean indicating whether shared libraries should be supported on
+   Windows. ``meson-python`` installs shared libraries in a dedicated location
+   and uses RPATH or equivalent mechanisms to have Python modules and native
+   executables load them form there. Windows does not have an equivalent
+   mechanism to set the DLL load path. Supporting shared libraries on Windows
+   requires collaboration from the package. To make sure that package authors
```
QuLogic, line 44: This doesn't seem to explain (or link to something that explains) what sort of "collaboration" is needed.

mesonpy/__init__.py
```diff
@@ -430,6 +432,9 @@ def _install_path(self, wheel_file: mesonpy._wheelfile.WheelFile, origin: Path,
 
         if self._has_internal_libs:
             if _is_native(origin):
+                if sys.platform == 'win32' and not self._shared_libs_win32:
+                    raise NotImplementedError(f'Bundling libraries in wheel is not supported on {sys.platform}')
```
QuLogic, line 436: This should mention the configuration option?

mesonpy/__init__.py
```diff
@@ -757,6 +763,10 @@ def __init__(
             if not value:
                 self._limited_api = False
 
+        # Shared library support on Windows requires collaboration
+        # from the package, make sure the developpers aknowledge this.
```
QuLogic, line 767:
```suggestion
        # from the package; make sure the developers acknowledge this.
```

rgommers (commented):
This looks like a good approach, I like it. A couple of thoughts:

1. Internal shared libraries can be made to work today already in a package using `meson-python`, when explicitly installing the shared library without the package and using `delvewheel` (in an elaborate way) and `install_rpath` on other platforms, as demonstrated by https://github.com/scipy/scipy/pull/20321. I think this PR is fine as is; I did have a concern that not having the option set shouldn't start raising an exception, so let's make sure that that remains the case.
2. As @QuLogic commented, it'd be good to elaborate on the best ways for a package to do this. I think there are two:

    - Using `os.add_dll_directory`, which should be reliable on Python >= 3.10
    - Preloading the DLL with `ctypes.WinDLL` - we've done this for a long time in NumPy and SciPy already for external shared libraries: https://github.com/scipy/scipy/blob/bb92c8014e21052e7dde67a76b28214dd1dcb94a/tools/openblas_support.py#L239-L274 (it should require fewer lines of code for an internal shared library that will always be present)

dnicolodi:
With the documentation work done in #700 it is time to revisit this, I think. I think this PR is a good approach and the implementation is simple enough. However, I really don't like the option name I came up with. Do anyone have a better idea?

rgommers:
A good descriptive name is not so easy. Ideally the name should make clear it's a boolean and allows meson-python to fold _internal_ shared libraries targeted for install _outside_ site-packages into the wheel on _Windows_. That's too much info though. Although `shared-libs-win32` can probably be improved on. Perhaps something like `allow-windows-internal-sharedlib`? (I don't like `win32` much, since in wheel tags it's specific to 32-bit Python).

dnicolodi:
I've adjusted the option name and updated the documentation. Please have a look.

mesonpy/__init__.py
```diff
@@ -422,6 +424,12 @@ def _install_path(self, wheel_file: mesonpy._wheelfile.WheelFile, origin: Path,
 
         if self._has_internal_libs:
             if _is_native(origin):
+                if sys.platform == 'win32' and not self._windows_shared_libs:
+                    raise NotImplementedError(
+                        'Loading shared libraries bundled in the Python wheel on Windows requires '
+                        'setting the DLL load path or pre-loading. See the documentation for '
+                        'the "tool.meson-pyhton.allow-windows-internal-shared-libs" option.')
```
rgommers, line 431: Typo in the option name here (pyhton)

mesonpy/__init__.py
```diff
@@ -784,6 +793,10 @@ def __init__(
                 'The package targets Python\'s Limited API, which is not supported by free-threaded CPython. '
                 'The "python.allow_limited_api" Meson build option may be used to override the package default.')
 
+        # Shared library support on Windows requires collaboration
+        # from the package, make sure the developers aknowledge this.
```
rgommers, line 797: typ: "acknowledge"

mesonpy/__init__.py
```diff
@@ -309,10 +309,12 @@ def __init__(
         metadata: Metadata,
         manifest: Dict[str, List[Tuple[pathlib.Path, str]]],
         limited_api: bool,
+        windows_shared_libs: bool,
     ) -> None:
         self._metadata = metadata
         self._manifest = manifest
         self._limited_api = limited_api
+        self._windows_shared_libs = windows_shared_libs
```
rgommers, line 317: Allow `allow_` into this name may be useful? If you considered and rejected it based on the name being too long, then the current name is fine with me.

rgommers (commented):
The added docs and the option name look quite good to me, as do the code changes.

The docs CI job has a complaint still. And it seems like test coverage still needs to be added.

dnicolodi:
Thanks for the review @rgommers I didn't add tests because I didn't want to deal with compilation with MSVC... I added them and indeed I have no idea why MSVC complains during linking (there is an issue with an invalid linker argument that I'll fix, but that does not seem to be the cause of the link failure). Any help would be appreciated

dnicolodi:
All green!

docs/reference/limitations.rst
```diff
@@ -31,16 +31,13 @@ Shared libraries on Windows
...
-be found at run-time.
-
-This limitation can be overcome with static linking or using
-`delvewheel`_ to post-process the Python wheel to bundle the required
-shared libraries and include the setup code to properly set the
-library search path.
+be found at run-time without cooperation from the Python package, see
+the documentation for the
+:option:`tool.meson-python.allow-windows-internal-shared-libs` option
+and the :ref:`shared-libraries` guide.
```
rgommers, line 37: I'm wondering whether it makes sense to delete this section completely - it's not actually a limitation anymore after this PR, it works with a few lines of code, and in case someone runs into it when forgetting the opt-in step, the error message points at the thing to do.

rgommers (approved):
This looks good to go, nice work! Glad adding the test wasn't too much trouble.

One more docs suggestion inline.

One test suggestion here: the first commit and the FreeBSD `patchelf` addition show that RPATH handling works fine on FreeBSD now, so it may be nice to add `freebsd` to these lines in `test_wheel.py`:
```python
@pytest.mark.skipif(sys.platform not in {'linux', 'darwin', 'sunos5'}, reason='Not supported on this platform')
```

Both of these are optional suggestions - happy for this to be merged as is if you prefer.

docs/reference/limitations.rst
dnicolodi: Agreed. I'll remove it

dnicolodi:
> One test suggestion here: the first commit and the FreeBSD `patchelf` addition show that RPATH handling works fine on FreeBSD now, so it may be nice to add `freebsd` to these lines in `test_wheel.py`:

Indeed. I'll do

rgommers:
🎉 nice work!

Issue #559, opened 2024-01-11
Fix Windows TemporaryDirectory cleanup


thalassemia:
In scipy/scipy#19855, I was experiencing Windows CI build failures with logs like the following:
```
+ meson dist --allow-dirty --no-tests --formats gztar
WARNING: Repository has uncommitted changes that will not be included in the dist tarball
Created D:\a\scipy\scipy\.mesonpy-rct2q19y\meson-dist\SciPy-1.13.0.dev0.tar.gz
Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.11.7\x64\Lib\shutil.py", line 624, in _rmtree_unsafe
    os.rmdir(path)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: '.\\.mesonpy-rct2q19y\\meson-private\\cmake_scipy-openblas\\CMakeFiles\\CMakeScratch\\TryCompile-cewmca'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.11.7\x64\Lib\tempfile.py", line 878, in onerror
    _os.unlink(path)
PermissionError: [WinError 5] Access is denied: '.\\.mesonpy-rct2q19y\\meson-private\\cmake_scipy-openblas\\CMakeFiles\\CMakeScratch\\TryCompile-cewmca'
```

The logs then alternate between the two functions until reaching a recursion depth error. The issue appears to stem from the use of `TemporaryDirectory` [here](https://github.com/mesonbuild/meson-python/blob/9d0fa1336fe18ffcd258b31c327ff3c5a020532e/mesonpy/__init__.py#L945C15-L945C15). This is a known issue (python/cpython#79325) that should be fixed in the next releases of Python 3.11/3.12.

Once that fix is in, we can use the `ignore_cleanup_errors` flag ([docs here](https://docs.python.org/3.11/library/tempfile.html#tempfile.TemporaryDirectory)) to silence this on Windows systems, where failed file deletions are a [known issue](https://bugs.python.org/issue29982). I'll create a draft PR for that.

PR #560, opened 2024-01-11
BUG: Ignore cleanup errors on Windows

thalassemia:
Closes gh-559.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -940,9 +940,16 @@ def _project(config_settings: Optional[Dict[Any, Any]] = None) -> Iterator[Proje
     build_dir = settings.get('build-dir')
     editable_verbose = bool(settings.get('editable-verbose'))
 
+    ignore_cleanup_errors = False
+    # Files failing to delete on Windows is a known issue
+    # (https://bugs.python.org/issue29982)
+    if os.name == 'nt':
+        ignore_cleanup_errors = True
     with contextlib.ExitStack() as ctx:
         if build_dir is None:
-            build_dir = ctx.enter_context(tempfile.TemporaryDirectory(prefix='.mesonpy-', dir=source_dir))
+            build_dir = ctx.enter_context(tempfile.TemporaryDirectory(
+                prefix='.mesonpy-', dir=source_dir,
+                ignore_cleanup_errors=ignore_cleanup_errors))
         yield Project(source_dir, build_dir, meson_args, editable_verbose)
 
 
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -940,16 +940,18 @@ def _project(config_settings: Optional[Dict[Any, Any]] = None) -> Iterator[Proje
     build_dir = settings.get('build-dir')
     editable_verbose = bool(settings.get('editable-verbose'))
 
-    ignore_cleanup_errors = False
+    temp_dir_kwargs = {
+        'prefix': '.mesonpy-',
+        'dir': source_dir
+    }
     # Files failing to delete on Windows is a known issue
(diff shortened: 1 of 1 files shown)


rgommers:
Thanks!

> Wait for Python 3.11.8/3.12.2.

We do need to be able to work with all bug-fix versions that we can find out in the wild. `.0` releases with bugs may be ignored, however we should have code that works with Python `3.11.7`. So this shouldn't have to wait for a next CPython bug fix release - or did you mean waiting here for testing purposes?

thalassemia:
Yeah, I think it would be nice to have some kind of test for this. I need to think more about how to set that up.

dnicolodi:
Why hasn't this been a problem so far? What changed?

dnicolodi:
Why do we need to wait for the new Python releases? The argument was added in Python 3.10.

thalassemia:
> Why hasn't this been a problem so far? What changed?

I identified `BackgroundDownload.exe` as the process preventing `TemporaryDirectory` from deleting certain files. It does not always happen, but calling Microsoft's Handle tool during the build process would occasionally give output like the following (see attempts 1 and 2 in [this PR check](https://github.com/thalassemia/scipy/actions/runs/7499460152/job/20416292926?pr=7)):

```
------------------------------------------------------------------------------
BackgroundDownload.exe pid: 1304 fv-az1387-368\runneradmin
   40: File  (RW-)   D:\a\scipy\scipy\.mesonpy-n0tbbfzi\meson-private\cmake_scipy-openblas\CMakeFiles\CMakeScratch\TryCompile-s9yijw
  188: Section       \BaseNamedObjects\Cor_Private_IPCBlock_v4_1304
  18C: Section       \...\Cor_SxSPublic_IPCBlock
  284: File  (R--)   C:\Windows\assembly\pubpol11.dat
  288: File  (R-D)   C:\Program Files (x86)\Microsoft Visual Studio\Installer\resources\app\ServiceHub\Services\Microsoft.VisualStudio.Setup.Service\Microsoft.VisualStudio.Setup.dll
```

A new Windows Server 2019 runner image (the version used for these PR checks) was released on January 8th, which may coincide with the first reports of this issue. Interestingly, the `BackgroundDownload.exe` process disappears along with the failed cleanup errors when building with [Windows Server 2022](https://github.com/thalassemia/scipy/actions/runs/7498897745/job/20418197169?pr=6).

The weird behavior of `BackgroundDownload.exe` exposed an infinite recursion in the `TemporaryDirectory` cleanup code. That bug should be fixed in the next releases of Python 3.11/3.12. 

I've modified the PR to make it more targeted while maintaining all current behavior/compatibility. After some more thought, I've also changed my mind about waiting to test this, because the relevant bugs are all beyond the scope of this repo. We can merge whenever ready.

dnicolodi:
IIUC, `BackgroundDownload.exe` is used to download Visual Studio updates. I don't understand why it should be preventing the cleanup of the temporary build directory. Please file a bug with GitHub to fix their images and with Microsoft for fixing their tools. I don't think we should work around this, especially if it means leaving temporary directories around.

If you need to work around this, you can simply pass a build directory explicitly with the `build-dir` build option. This disables the build directory cleanup.

Also, if this is a widespread issue, `pip` would run into the same cleanup problem when building a package for an sdist for local isntallation: `pip` unpacks the sdist into a temporary directory, and meson-python creates a build directory into that directory. If meson-python cleanup of the temporary build directory fails, the cleanup of the `pip` temporarty directory also fails.

thalassemia:
Good point about the `build-dir` workaround and the downstream effects of a build dir cleanup failure (our failures were with `build` but I'm sure `pip` would fail too)! I'll close this and open an issue on the GitHub actions image.

PR #579, opened 2024-02-22
Editable install: do not show any output in  verbose mode when there is no work to do

lesteve:
### Why the current behaviour is problematic

The "ninja: no work to do" output can be confusing as noted for example by @ogrisel in https://github.com/scikit-learn/scikit-learn/pull/28040#issuecomment-1886676007. @eli-schwartz mentioned that it could potentially be changed in meson-python https://github.com/scikit-learn/scikit-learn/pull/28040#issuecomment-1888470251.

Personally, when working on scikit-learn code, I find it reassuring to see some output if the sklearn import takes a while and get the feed-back that some compilation is happening. In the case when there is nothing to do I would rather have nothing printed. One thing I realised recently is that the "no work to do" output can be printed multiple times if you use subprocesses (makes complete sense since sklearn is imported in each subprocess). This can happen easily in scikit-learn for example with this snippet.

```py
# /tmp/test.py
from sklearn.linear_model import LogisticRegression
from sklearn.datasets import make_classification
from sklearn.model_selection import cross_val_score

model = LogisticRegression()
X, y = make_classification()

print(cross_val_score(model, X, y, n_jobs=4))
```

Output is:
```
❯ python /tmp/test.py
+ /home/lesteve/micromamba/envs/scikit-learn-dev/bin/ninja
ninja: no work to do.
+ /home/lesteve/micromamba/envs/scikit-learn-dev/bin/ninja
+ /home/lesteve/micromamba/envs/scikit-learn-dev/bin/ninja
+ /home/lesteve/micromamba/envs/scikit-learn-dev/bin/ninja
+ /home/lesteve/micromamba/envs/scikit-learn-dev/bin/ninja
ninja: no work to do.
ninja: no work to do.
ninja: no work to do.
ninja: no work to do.
[1.   0.85 0.9  0.95 0.95]
```

### Current implementation

The simplest thing I could find was to first execute the `ninja` command in dry-run mode and to check if "no work to do" is in the captured stdout. This may be a bit brittle in case the ninja output changes. Maybe there is a better way to check with `ninja` that there is no work to do?

There does not seem to be any test right now for the editable verbose behavior, but I can try to add some if you think this is worth it.

### Alternative implementation

If you think the overhead of having a `ninja` dry-run command is too much overhead, one possible alternative implementation would be to capture `stdout` with `p = subprocess.Popen(..., stdout=subprocess.PIPE)` and look at the first two lines of output. In my experience, this kind of manipulation are always more complicated than you would hope, but maybe I have been doing it wrong all this time ...

The kind of complications I have in mind:
- `p.stdout.read` is blocking so in the case that there is no output for some reason, it will wait for ever. Maybe we know there will always be some output with the `ninja` command. You can also put you can use `fcntl` to put stdout in non-blocking mode.
- displaying the captured stdout live (i.e. without waiting for the command to finish) is also always more complicated than you would hope:
  + decide whether to read line by line, a few lines, a number of bytes
  + decide whether to sleep between two reads (I have seen some cases where not sleeping had a high CPU usage)
  + you also need to remember to redisplay all the remaining stdout once the command finishes otherwise you miss some stdout.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -116,7 +116,7 @@ jobs:
 
       - name: Run tests
         run: >-
-          python -m pytest --showlocals -vv --cov --cov-report=xml -s
+          python -m pytest --showlocals -vv --cov --cov-report=xml
 
       - name: Upload coverage report
         uses: codecov/codecov-action@v3
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -116,7 +116,7 @@ jobs:
 
       - name: Run tests
         run: >-
-          python -m pytest --showlocals -vv --cov --cov-report=xml
+          python -m pytest --showlocals -vv --cov --cov-report=xml -s
 
       - name: Upload coverage report
         uses: codecov/codecov-action@v3
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -309,7 +309,7 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
             dry_run_build_cmd = self._build_cmd + ['-n']
             p = subprocess.run(dry_run_build_cmd, cwd=self._build_path, env=env, capture_output=True)
             if b'ninja: no work to do.' not in p.stdout and b'samu: nothing to do' not in p.stdout:
-                print('+ ' + ' '.join(self._build_cmd))
+                print('+ ' + ' '.join(self._build_cmd), flush=True)
                 subprocess.run(self._build_cmd, cwd=self._build_path, env=env)
         else:
             subprocess.run(self._build_cmd, cwd=self._build_path, env=env, stdout=subprocess.DEVNULL)
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -316,6 +316,7 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
         env = os.environ.copy()
         env[MARKER] = os.pathsep.join((env.get(MARKER, ''), self._build_path))
 
+        sys.stderr.write(f"{env.get(VERBOSE, '')=}\n")
         if self._verbose or bool(env.get(VERBOSE, '')):
             # We do not want any output if there is no work to do. Code is adapted from:
             # https://github.com/mesonbuild/meson/blob/a35d4d368a21f4b70afa3195da4d6292a649cb4c/mesonbuild/mtest.py#L1635-L1636
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -316,8 +316,9 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
         env = os.environ.copy()
         env[MARKER] = os.pathsep.join((env.get(MARKER, ''), self._build_path))
 
-        sys.stderr.write(f"{env.get(VERBOSE, '')=}\n")
+        sys.stderr.write('VERBOSE value:' + env.get(VERBOSE, '') + '\n')
         if self._verbose or bool(env.get(VERBOSE, '')):
+            sys.stderr.write('this line gets executed contrary to what codecov says\n')
             # We do not want any output if there is no work to do. Code is adapted from:
             # https://github.com/mesonbuild/meson/blob/a35d4d368a21f4b70afa3195da4d6292a649cb4c/mesonbuild/mtest.py#L1635-L1636
             if sys.platform == 'win32':
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -304,12 +304,14 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
         env[MARKER] = os.pathsep.join((env.get(MARKER, ''), self._build_path))
 
         if self._verbose or bool(env.get(VERBOSE, '')):
-            print('+ ' + ' '.join(self._build_cmd))
-            stdout = None
+            dry_run_build_cmd = self._build_cmd + ['-n']
+            # We do not want any output if there is no work to do
+            p = subprocess.run(dry_run_build_cmd, cwd=self._build_path, env=env, capture_output=True)
+            if b'no work to do' not in p.stdout:
+                print('+ ' + ' '.join(self._build_cmd))
+                subprocess.run(self._build_cmd, cwd=self._build_path, env=env)
         else:
-            stdout = subprocess.DEVNULL
-
-        subprocess.run(self._build_cmd, cwd=self._build_path, env=env, stdout=stdout, check=True)
+            subprocess.run(self._build_cmd, cwd=self._build_path, env=env, stdout=subprocess.DEVNULL)
 
         install_plan_path = os.path.join(self._build_path, 'meson-info', 'intro-install_plan.json')
         with open(install_plan_path, 'r', encoding='utf8') as f:
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -316,9 +316,7 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
         env = os.environ.copy()
         env[MARKER] = os.pathsep.join((env.get(MARKER, ''), self._build_path))
 
-        sys.stderr.write('VERBOSE value:' + env.get(VERBOSE, '') + '\n')
         if self._verbose or bool(env.get(VERBOSE, '')):
-            sys.stderr.write('this line gets executed contrary to what codecov says\n')
             # We do not want any output if there is no work to do. Code is adapted from:
             # https://github.com/mesonbuild/meson/blob/a35d4d368a21f4b70afa3195da4d6292a649cb4c/mesonbuild/mtest.py#L1635-L1636
             if sys.platform == 'win32':
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -308,7 +308,7 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
         tree = self._rebuild()
         return find_spec(fullname, tree)
 
-    def _work_to_do(self, env):
+    def _work_to_do(self, env: dict[str, str]) -> bool:
         # Code is adapted from:
         # https://github.com/mesonbuild/meson/blob/a35d4d368a21f4b70afa3195da4d6292a649cb4c/mesonbuild/mtest.py#L1635-L1636
         if sys.platform != 'win32':
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -313,7 +313,7 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
             else:
                 dry_run_build_cmd = self._build_cmd + ['-n']
 
-            p = subprocess.run(dry_run_build_cmd, cwd=self._build_path, env=env, capture_output=True)
+            p = subprocess.run(dry_run_build_cmd, cwd=self._build_path, env=env, capture_output=True, check=True)
             if b'ninja: no work to do.' not in p.stdout and b'samu: nothing to do' not in p.stdout:
                 module_names = ' '.join(sorted(self._top_level_modules))
                 build_command = ' '.join(self._build_cmd)
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -304,10 +304,11 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
         env[MARKER] = os.pathsep.join((env.get(MARKER, ''), self._build_path))
 
         if self._verbose or bool(env.get(VERBOSE, '')):
+            # We do not want any output if there is no work to do. Code is adapted from:
+            # https://github.com/mesonbuild/meson/blob/a35d4d368a21f4b70afa3195da4d6292a649cb4c/mesonbuild/mtest.py#L1635-L1636
             dry_run_build_cmd = self._build_cmd + ['-n']
-            # We do not want any output if there is no work to do
             p = subprocess.run(dry_run_build_cmd, cwd=self._build_path, env=env, capture_output=True)
-            if b'no work to do' not in p.stdout:
+            if b'ninja: no work to do.' not in p.stdout and b'samu: nothing to do' not in p.stdout:
                 print('+ ' + ' '.join(self._build_cmd))
                 subprocess.run(self._build_cmd, cwd=self._build_path, env=env)
         else:
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -319,7 +319,7 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
         if self._verbose or bool(env.get(VERBOSE, '')):
             # We do not want any output if there is no work to do. Code is adapted from:
             # https://github.com/mesonbuild/meson/blob/a35d4d368a21f4b70afa3195da4d6292a649cb4c/mesonbuild/mtest.py#L1635-L1636
-            if self._build_cmd[0] == 'meson':
+            if sys.platform == 'win32':
                 # On Windows meson compile is used so to do a dry run you need
                 # --ninja-args
                 dry_run_build_cmd = self._build_cmd + ['--ninja-args=-n']
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -312,8 +312,7 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
                 module_names = ' '.join(sorted(self._top_level_modules))
                 build_command = ' '.join(self._build_cmd)
                 info_msg = (
-                    f'+ meson-python: building {module_names}\n'
-                    f'+ meson-python: executing {build_command}'
+                    f'meson-python: building {module_names} with {build_command}'
                 )
                 print(info_msg, flush=True)
                 subprocess.run(self._build_cmd, cwd=self._build_path, env=env)
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -309,7 +309,13 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
             dry_run_build_cmd = self._build_cmd + ['-n']
             p = subprocess.run(dry_run_build_cmd, cwd=self._build_path, env=env, capture_output=True)
             if b'ninja: no work to do.' not in p.stdout and b'samu: nothing to do' not in p.stdout:
-                print('+ ' + ' '.join(self._build_cmd), flush=True)
+                module_names = ' '.join(sorted(self._top_level_modules))
+                build_command = ' '.join(self._build_cmd)
+                info_msg = (
+                    f"+ meson-python: building {module_names}\n"
+                    f'+ meson-python build command: {build_command}'
+                )
+                print(info_msg, flush=True)
                 subprocess.run(self._build_cmd, cwd=self._build_path, env=env)
         else:
             subprocess.run(self._build_cmd, cwd=self._build_path, env=env, stdout=subprocess.DEVNULL)
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -312,7 +312,7 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
                 module_names = ' '.join(sorted(self._top_level_modules))
                 build_command = ' '.join(self._build_cmd)
                 info_msg = (
-                    f"+ meson-python: building {module_names}\n"
+                    f'+ meson-python: building {module_names}\n'
                     f'+ meson-python build command: {build_command}'
                 )
                 print(info_msg, flush=True)
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -309,7 +309,7 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
             if self._build_cmd[0] == 'meson':
                 # On Windows meson compile is used so to do a dry run you need
                 # --ninja-args
-                dry_run_build_cmd = self._build_cmd + ['--ninja-args="-n"']
+                dry_run_build_cmd = self._build_cmd + ['--ninja-args=-n']
             else:
                 dry_run_build_cmd = self._build_cmd + ['-n']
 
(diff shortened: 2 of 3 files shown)


ogrisel:
I think the lines:

```
+ /home/lesteve/micromamba/envs/scikit-learn-dev/bin/ninja
ninja: no work to do.
```

would be ok, if they were prefixed by some user friendlier message:

```
meson-python: importing sklearn installed in editable mode
+ /home/lesteve/micromamba/envs/scikit-learn-dev/bin/ninja
ninja: no work to do.
```

But I agree that the repeated message is a bit annoying when using parallelism that spawn Python workers that import `sklearn` concurrently.

ogrisel:
But I am also fine with what @lesteve suggests in this PR.

mesonpy/_editable.py
```diff
@@ -304,12 +304,14 @@ def rebuild(self) -> Node:
         env[MARKER] = os.pathsep.join((env.get(MARKER, ''), self._build_path))
 
         if self._verbose or bool(env.get(VERBOSE, '')):
-            print('+ ' + ' '.join(self._build_cmd))
-            stdout = None
+            dry_run_build_cmd = self._build_cmd + ['-n']
+            # We do not want any output if there is no work to do
+            p = subprocess.run(dry_run_build_cmd, cwd=self._build_path, env=env, capture_output=True)
+            if b'no work to do' not in p.stdout:
```
eli-schwartz, line 310: Relevant: https://github.com/mesonbuild/meson/blob/a35d4d368a21f4b70afa3195da4d6292a649cb4c/mesonbuild/mtest.py#L1636

rgommers:
Thanks! I agree that this message can be confusing. I hadn't seen it printed multiple times yet, but that makes it worse. So looks like a good idea to suppress it.

mesonpy/_editable.py
rgommers: already tested code, always nice:)

rgommers: Probably worth taking that over verbatim, and adding a code comment showing its origin so that if we need to make further tweaks, we can keep the two pieces of code in sync?

eli-schwartz: That would be my recommendation, yes.

mesonpy/_editable.py
```diff
@@ -304,12 +304,15 @@ def rebuild(self) -> Node:
...
 
         if self._verbose or bool(env.get(VERBOSE, '')):
-            print('+ ' + ' '.join(self._build_cmd))
-            stdout = None
+            # We do not want any output if there is no work to do. Code is adapted from:
+            # https://github.com/mesonbuild/meson/blob/a35d4d368a21f4b70afa3195da4d6292a649cb4c/mesonbuild/mtest.py#L1635-L1636
+            dry_run_build_cmd = self._build_cmd + ['-n']
+            p = subprocess.run(dry_run_build_cmd, cwd=self._build_path, env=env, capture_output=True)
+            if b'ninja: no work to do.' not in p.stdout and b'samu: nothing to do' not in p.stdout:
+                print('+ ' + ' '.join(self._build_cmd))
```
ogrisel, line 312: I find the build command not very informative because it does not reflect which editable install is being rebuilt, nor does it state that `meson-python`'s editable install is in charge of making this build happen.

So maybe consider:

```suggestion
                module_names = " ".join(sorted(self._top_level_modules))
                print(f"meson-python: building {module_names}")
```

Here is the resulting user experience:

```
$ python -c "import sklearn"  # nothing to rebuild => no output
$ touch sklearn/ensemble/_gradient_boosting.pyx
$ python -c "import sklearn"
meson-python: building sklearn
[3/3] Linking target sklearn/ensemble/_gradient_boosting.cpython-311-darwin.so
```

rgommers: I like it!

lesteve: Two questions:
- do we want to keep the `ninja` build command for troubleshooting or are we fine not showing it?
- do we want to keep the `+` at the beginning of the first line? I guess maybe this was a marker to make it stand out in the output?

mesonpy/_editable.py
```diff
@@ -304,12 +304,15 @@ def rebuild(self) -> Node:
         env[MARKER] = os.pathsep.join((env.get(MARKER, ''), self._build_path))
 
         if self._verbose or bool(env.get(VERBOSE, '')):
-            print('+ ' + ' '.join(self._build_cmd))
-            stdout = None
+            # We do not want any output if there is no work to do. Code is adapted from:
+            # https://github.com/mesonbuild/meson/blob/a35d4d368a21f4b70afa3195da4d6292a649cb4c/mesonbuild/mtest.py#L1635-L1636
+            dry_run_build_cmd = self._build_cmd + ['-n']
```
dnicolodi, line 309: This does not work on Windows, where the compilation command is `meson compile`, which does not accept a `-n` argument. This demonstrates that this feature needs tests.

dnicolodi (changes requested):

mesonpy/_editable.py
ogrisel: I don't have an opinion on those points as long as the first message mentions "meson-python" and the name of the package(s) being rebuilt prior to displaying build outputs.

tests/test_editable.py
```diff
@@ -189,3 +190,39 @@ def test_editble_reentrant(venv, editable_imports_itself_during_build):
...
+        # Need to filter some warning on OSX like
+        # 'ld: warning: -undefined dynamic_lookup may not work with chained fixups'
+        output_lines = [line for line in output.splitlines() if 'warning' not in line]
+        expected_pattern_list =  [
+            'meson-python: building complex',
+            'meson-python build command:.+(ninja|samu)',
+            'Compiling Cython source',
+            'Compiling C object',
+            'Linking target'
+        ]
```
dnicolodi, line 219: Why do you check for the exact content of the log messages? It seems fragile (they are not under the control of meson-python) and not really required. The output is also very different on Windows.

tests/test_editable.py
```diff
@@ -189,3 +190,39 @@ def test_editble_reentrant(venv, editable_imports_itself_during_build):
...
+    cython_content = cython_path.read_text()
+    try:
+        cython_path.write_text(cython_content + '\n')
+        output = venv.python('-c', 'import complex').strip()
+        # Need to filter some warning on OSX like
+        # 'ld: warning: -undefined dynamic_lookup may not work with chained fixups'
+        output_lines = [line for line in output.splitlines() if 'warning' not in line]
+        expected_pattern_list =  [
+            'meson-python: building complex',
+            'meson-python build command:.+(ninja|samu)',
```
dnicolodi, line 215: The compilation command on Windows is neither `ninja` nor `samu`.

mesonpy/_editable.py
```diff
@@ -304,12 +304,21 @@ def rebuild(self) -> Node:
...
+            p = subprocess.run(dry_run_build_cmd, cwd=self._build_path, env=env, capture_output=True)
+            if b'ninja: no work to do.' not in p.stdout and b'samu: nothing to do' not in p.stdout:
+                module_names = ' '.join(sorted(self._top_level_modules))
+                build_command = ' '.join(self._build_cmd)
+                info_msg = (
+                    f'+ meson-python: building {module_names}\n'
+                    f'+ meson-python build command: {build_command}'
```
dnicolodi, line 316: For consistency, I would spell this `meson-python: executing {build_command}`.

dnicolodi: Or you can make it a single line, combining the two messages: `meson-python: building {module_names}: {build_command}`. And I would loose the `+ ` prefix.

mesonpy/_editable.py
```diff
@@ -304,12 +304,21 @@ def rebuild(self) -> Node:
         env[MARKER] = os.pathsep.join((env.get(MARKER, ''), self._build_path))
 
         if self._verbose or bool(env.get(VERBOSE, '')):
-            print('+ ' + ' '.join(self._build_cmd))
-            stdout = None
+            # We do not want any output if there is no work to do. Code is adapted from:
+            # https://github.com/mesonbuild/meson/blob/a35d4d368a21f4b70afa3195da4d6292a649cb4c/mesonbuild/mtest.py#L1635-L1636
+            dry_run_build_cmd = self._build_cmd + ['-n']
+            p = subprocess.run(dry_run_build_cmd, cwd=self._build_path, env=env, capture_output=True)
```
dnicolodi, line 310: Right now, this is failing on Windows, where `meson compile -n` cannot work.

tests/test_editable.py
lesteve: I was a bit overoptimistic indeed, I am going to only check the first two lines of output since they are under meson-python control.

lesteve:
It would be great if someone with the necessary rights could enable the Github actions workflow :pray:! 

I can reproduce the Alpine failure in a Docker container and the bad news is that `ninja -n` seems to segfault (`ninja` alone works fine). The ninja version is 1.9 (released January 2019) and I have seen at least one test in meson-python that is skipped if the ninja version < 1.10, so maybe this could be an option?

My basic gdb skills only allowed me to get this information:
```
(gdb) r
Starting program: /usr/bin/ninja -C build/cp312 -n
warning: Error disabling address space randomization: Operation not permitted
ninja: entering directory 'build/cp312'

Program received signal SIGSEGV, Segmentation fault.
0x00007b65464a9458 in strlen (s=<optimized out>, 
    s@entry=0x8 <error: Cannot access memory at address 0x8>) at src/string/strlen.c:17
warning: 17	src/string/strlen.c: No such file or directory
```

Suggestions how to get more info more than welcome ...

eli-schwartz:
> I can reproduce the Alpine failure in a Docker container and the bad news is that `ninja -n` seems to segfault (`ninja` alone works fine). The ninja version is 1.9 (released January 2019) and I have seen at least one test in meson-python that is skipped if the ninja version < 1.10, so maybe this could be an option?

This is:

https://github.com/michaelforney/samurai/issues/66
https://github.com/michaelforney/samurai/issues/81
https://github.com/michaelforney/samurai/commit/fb61f22c7e690715d309c41812412c4f432ef53a

Possibly something Alpine should backport...

lesteve:
I have skipped the test on Alpine for now based on detecting musllinux.

To know whether the test pass on Windows it would be nice that someone enable Github workflows. Full disclosure: locally the Windows tests don't pass and I don't really know why yet.

rgommers:
> To know whether the test pass on Windows it would be nice that someone enable Github workflows.

Done. this can be a pain because the approve button has to be pressed after every push. If you want to submit a trivial PR to side-step this problem, I'm happy to merge that straight away (e.g, in the readme, change `to use Meson` to `using Meson`).

lesteve:
Thanks, the Windows tests fail in a similar way locally, the dry-run command fails with a exit code 1 and I can't find a reason in the captured stderr as to why, but maybe I am missing something ...

I'll be offline next week but I can try to pick this up when I come back. Any meson-python maintainer should not hesitate to push into my branch if that makes things easier.

mesonpy/_editable.py
```diff
@@ -304,12 +304,26 @@ def rebuild(self) -> Node:
...
+            else:
+                dry_run_build_cmd = self._build_cmd + ['-n']
+
+            p = subprocess.run(dry_run_build_cmd, cwd=self._build_path, env=env, capture_output=True, check=True)
+            if b'ninja: no work to do.' not in p.stdout and b'samu: nothing to do' not in p.stdout:
+                module_names = ' '.join(sorted(self._top_level_modules))
+                build_command = ' '.join(self._build_cmd)
+                info_msg = (
+                    f'meson-python: building {module_names} with {build_command}'
+                )
```
dnicolodi, line 322: Assigning the message to a variable is unnecessary.

mesonpy/_editable.py
```diff
@@ -304,12 +304,26 @@ def rebuild(self) -> Node:
         env[MARKER] = os.pathsep.join((env.get(MARKER, ''), self._build_path))
 
         if self._verbose or bool(env.get(VERBOSE, '')):
-            print('+ ' + ' '.join(self._build_cmd))
-            stdout = None
+            # We do not want any output if there is no work to do. Code is adapted from:
+            # https://github.com/mesonbuild/meson/blob/a35d4d368a21f4b70afa3195da4d6292a649cb4c/mesonbuild/mtest.py#L1635-L1636
+            if self._build_cmd[0] == 'meson':
```
dnicolodi, line 309: The build command on Windows can be something else that the literal string `meson`.

tests/test_editable.py
```diff
@@ -189,3 +190,37 @@ def test_editble_reentrant(venv, editable_imports_itself_during_build):
...
+    assert len(output_lines) > 1
+    # Only checking the first output line which meson-python controls.
+    # The rest of the output varies across platforms.
+    checked_output_lines = output_lines[:1]
+    expected_pattern_list =  [
+        'meson-python: building complex with',
+    ]
+    assert len(checked_output_lines) == len(expected_pattern_list)
+    for expected_pattern, output_line in zip(expected_pattern_list, checked_output_lines):
+        assert re.search(expected_pattern, output_line), f'{expected_pattern} was not found in {output_line}'
```
dnicolodi, line 223: The past 10 lines can simply be:
```
assert output.startswith('meson-python: building complex with')
```

tests/test_editable.py
```diff
@@ -189,3 +190,37 @@ def test_editble_reentrant(venv, editable_imports_itself_during_build):
...
+
+    # Second import should have no output since the project has already been built
+    assert venv.python('-c', 'import complex').strip() == ''
+
+    # Touch the pyx and make sure that the Compiling lines are seen
+    complex_package_dir = venv.python(
+        '-c', 'import os; import complex; print(os.path.dirname(complex.__file__))').strip()
+    cython_path = pathlib.Path(complex_package_dir).parent / 'test.pyx'
+    cython_content = cython_path.read_text()
+    cython_path.write_text(cython_content)
```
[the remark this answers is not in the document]
lesteve: As the comment above is trying to say, touching the `.pyx` ensures that recompilation happens.

mesonpy/_editable.py
```diff
@@ -304,12 +304,26 @@ def rebuild(self) -> Node:
...
+                # On Windows meson compile is used so to do a dry run you need
+                # --ninja-args
+                dry_run_build_cmd = self._build_cmd + ['--ninja-args="-n"']
```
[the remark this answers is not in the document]
lesteve: Good catch the quoting was the issue indeed, now the test passes locally on a Windows VM.

About multiple `--ninja-args`, you probably know a lot more than me about this, so any suggestions welcome!

mesonpy/_editable.py
lesteve: Do you have any advice on a more robust way to handle the Windows case?

lesteve:
> If you want to submit a trivial PR to side-step this problem, I'm happy to merge that straight away (e.g, in the readme, change `to use Meson` to `using Meson`).

@rgommers I have opened https://github.com/mesonbuild/meson-python/pull/585 with this change.

rgommers:
> @rgommers I have opened #585 with this change.

Sounds good - merged now. Should fix the issue with CI running here.

mesonpy/_editable.py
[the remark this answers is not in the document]
dnicolodi: I don't know whether `meson compile` accepts multiple `--ninja-args` options. This is why I am asking. If I knew I would not have posed this as a question. You are the proponent of this feature, thus it is on you to verify and to find a solution.

tests/test_editable.py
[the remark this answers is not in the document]
dnicolodi: The comment above is precisely the reason why I am asking. "to touch" a file is usually understood as in updating the mtime in the file metadata without altering the file content. What you are doing here is to rewrite the content of the file. Which is, per your comment, unnecessary.

dnicolodi:
> This is:
> 
> [michaelforney/samurai#66](https://github.com/michaelforney/samurai/issues/66) [michaelforney/samurai#81](https://github.com/michaelforney/samurai/issues/81) [michaelforney/samurai@fb61f22](https://github.com/michaelforney/samurai/commit/fb61f22c7e690715d309c41812412c4f432ef53a)
> 
> Possibly something Alpine should backport...

@eli-schwartz Are you implying that Alpine uses samurai but calls it ninja? I wonder if we can pull in a fixed package in our CI image...

dnicolodi:
Yes, it seems that installing ninja on Alpine Linux puts the executable in a location outside `$PATH` and that the samuari package install a symlink that makes ninja actually be samurai. samurai seems to be installed with the build-base package, otherwise I don't understand why we get it in our CI images. Anyhow, installing `ninja-is-really-ninja` should make the `ninja` command actually be what is expected.

The bug @eli-schwartz mentions was fixed in samurai back in 2021, but samurai hasn't seen a release since 2020.

lesteve:
The CI failures for the development meson version are also `main` and do not seem to be related to this PR.

tests/test_editable.py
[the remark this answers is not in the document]
lesteve: I changed this to only `touch` the file

mesonpy/_editable.py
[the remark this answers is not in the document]
dnicolodi: It is not acceptable to discard the user supplied argument and keep only the `-n`.

mesonpy/_editable.py
dnicolodi: The easiest is probably to check for `if sys.platform == 'win32'`. If that is true, the command being executed is `$MESON compile` possibly with a `--ninja-args=...` argument. Otherwise is `ninja` or something with a compatible command line interface.

mesonpy/_editable.py
```diff
@@ -317,12 +317,23 @@ def _rebuild(self) -> Node:
...
+                dry_run_build_cmd = self._build_cmd + ['--ninja-args=-n']
+            else:
+                dry_run_build_cmd = self._build_cmd + ['-n']
+
+            p = subprocess.run(dry_run_build_cmd, cwd=self._build_path, env=env, capture_output=True, check=True)
+            if b'ninja: no work to do.' not in p.stdout and b'samu: nothing to do' not in p.stdout:
+                module_names = ' '.join(sorted(self._top_level_modules))
+                build_command = ' '.join(self._build_cmd)
+                print(f'meson-python: building {module_names} with {build_command!r}', flush=True)
+                subprocess.run(self._build_cmd, cwd=self._build_path, env=env)
```
dnicolodi, line 334: Why does Codecov think that these lines are not covered by the tests?

lesteve: Hmmm not sure what is happening ...

lesteve: It seems like a but somewhere in the toolchain (maybe `coverage`, maybe `codecov`, not sure ...) that does not see the line covered although the line is executed. 

- codecov says the line is not covered [build log](https://github.com/mesonbuild/meson-python/pull/579/checks?check_run_id=22255522195) or
image

- the line is executed as you can see in this [build log](https://github.com/mesonbuild/meson-python/actions/runs/8143472943/job/22255402547?pr=579#step:9:204) for example

mesonpy/_editable.py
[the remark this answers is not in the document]
lesteve: Point taken, my current plan for this is to fine the last arguments in `self._build_cmd` that starts with `--ninja-args` and if there is one to append `,-n` to it. This does seem a bit complicated though, let me know if you have a better suggestion!

In order to test this (at least locally), do you have an example in mind on Windows where ninja arguments would be already in the build command (i.e. before adding `-n`)?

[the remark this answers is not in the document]
eli-schwartz: My assumption is this would happen when users specify `-C...` options to the build frontend.

mesonpy/_editable.py
ogrisel: Maybe sklearn's editable install `find_spec` is called prior to the initialization of the `coverage` tracer?

dnicolodi: > Maybe sklearn's editable install's

What does sklearn has to do with this?

mesonpy/_editable.py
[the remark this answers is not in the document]
dnicolodi: As @eli-schwartz says, you can pass install with `pip install -e . -Cbuild-args=-j1` for example.

As for the implementation, it is ok to check if the component of the command starts with `--ninja-args`, then you need to parse the value passed to the option and add `-n` to the list. You can take some inspiration for how to do it from the code in this test https://github.com/mesonbuild/meson-python/blob/b3373346f4f60b9f0929d7a1042fa909badc6a48/tests/test_project.py#L49-L56

mesonpy/_editable.py
ogrisel: Sorry, I typed too quickly. I wanted to mention the editable modules of the tests.

.github/workflows/tests.yml
```diff
@@ -116,7 +116,7 @@ jobs:
 
       - name: Run tests
         run: >-
-          python -m pytest --showlocals -vv --cov --cov-report=xml
+          python -m pytest --showlocals -vv --cov --cov-report=xml -s
```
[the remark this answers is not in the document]
lesteve: Yep I was about to do it :wink:, done in my last commit.

mesonpy/_editable.py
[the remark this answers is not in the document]
lesteve: I have changed the code to deal with existing and multiple `--ninja-args` in `self._build_cmd` but I haven't written a test for it yet.

mesonpy/_editable.py
lesteve: I have added a test to cover these lines in the main process, by adding a parametrized tests to `test_mesonpy_meta_finder` with `verbose=True`

> coverage (unless tricks are applied) does not cross process boundaries when collecting coverage data.

I have seen `parallel = True` in `.coveragerc` files to be able to cover lines in subprocesses, see [this](https://coverage.readthedocs.io/en/latest/cmd.html#combining-data-files-coverage-combine) for more details. Not sure if you would like to consider this option. We are doing this in scikit-learn and this seems to work fine but there may be some edge cases I am unaware of.

dnicolodi:
Please don't merge the main branch into your branch the next time: it makes rebasing the commits harder without reason.

dnicolodi:
I added the missing tests and fixed up the last issue and submitted as #594

Issue #460, opened 2023-08-25
Document environment variables used by Meson


dnicolodi:
See https://github.com/mesonbuild/meson-python/issues/459#issuecomment-1692971600

Add a note about environment variables injected by conda.

rgommers:
- the environment variables used by `meson-python` are documented at https://mesonbuild.com/meson-python/reference/environment-variables.html
-  the environment variables used by `meson` are documented at https://mesonbuild.com/Reference-tables.html#compiler-and-linker-flag-environment-variables
- the note on Conda injecting environment variables is present on https://mesonbuild.com/meson-python/how-to-guides/debug-builds.html

The main thing that can be improved still is to cross-link from our docs to the Meson docs on this topic. I'll have a go at that.

PR #583, opened 2024-02-28
Extend documentation on environment variables

rgommers:
- cross link to the tables of compiler/linker flags in the Meson docs
- mention `PKG_CONFIG_PATH`
- add a warning about Conda's habit of injecting env vars also on the environment variables reference page

Closes gh-460

### DIFF
--- a/docs/reference/environment-variables.rst
+++ b/docs/reference/environment-variables.rst
@@ -8,6 +8,36 @@
 Environment variables
 *********************
 
+Environment variables can be used to influence ``meson-python``'s behavior, as
+well as the behavior of Meson and other tools that may be used during the
+build. This page lists all the environment variables directly used by
+``meson-python``.
+
+Meson recommends using command line arguments instead of environment variables,
+but does support a number of environment variables for compatibility with other
+build systems:
+
+- `Compiler and linker flag variables <https://mesonbuild.com/Reference-tables.html#compiler-and-linker-flag-environment-variables>`__
+- `Compiler and linker selection variables <https://mesonbuild.com/Reference-tables.html#compiler-and-linker-selection-variables>`__
+
+Other environment variables may influence how other tools used during the setup
+or build steps operate. For example, ``pkg-config`` supports
+``PKG_CONFIG_PATH`` which influences the search path for ``.pc`` files
+describing the available dependencies.
+
+.. warning::
+
+    Conda sets a number of environment variables during environment activation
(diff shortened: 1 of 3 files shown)


rgommers:
CI failure is Sphinx version too old:
```
Running Sphinx v4.5.0

Sphinx version error:
The sphinxcontrib.applehelp extension used by this project needs at least Sphinx v5.0; it therefore cannot be built with this version.
```

Will add a second commit for that.

docs/reference/environment-variables.rst
```diff
@@ -8,6 +8,35 @@
...
+``meson-python``.
+
+Meson recommends using command line arguments instead of environment variables,
+but does support a number of environment variables for compatibility with other
+build systems:
+
+- `Compiler and linker flag variables <https://mesonbuild.com/Reference-tables.html#compiler-and-linker-flag-environment-variables>`__
+- `Compiler and linker selection variables <https://mesonbuild.com/Reference-tables.html#compiler-and-linker-selection-variables>`__
+
+Other tools each may support their own variables. E.g., ``pkg-config`` supports
```
dnicolodi, line 23: The beginning of this sentence is a bit convoluted. I would simply write: "Other environment variables may influence how other tools used during the setup or build steps operate. For example, ``pkg-config`` supports..."

docs/reference/environment-variables.rst
```diff
@@ -8,6 +8,35 @@
...
+``PKG_CONFIG_PATH`` which influences the search path for dependencies at build
+time.
```
dnicolodi, line 25: Technically `pkg-config` is used at setup (configure) time, not at build time. Maybe "at build time" can be simply omitted without loss of relevant information. Also, maybe, "the search path for dependencies" should be something like "the search path for ``.pc`` files describing the available dependecies".

dnicolodi (changes requested):
Thanks @rgommers. Looks nice, except a couple of small things.

docs/reference/environment-variables.rst
rgommers: thanks, adopted this phrasing

docs/reference/environment-variables.rst
rgommers: agreed, adopted

dnicolodi:
Thanks @rgommers. I was indeed wondering why sphynx was pinned to such an old release. Just one more thing: maybe the commit messages should follow the usual convention and be prepended with `DOC:` and the remaining should not be capitalized.

PR #584, opened 2024-02-28
ENH: support custom_target() installing a directory

dnicolodi:
Meson support custom_target() whose output is not a file but a directory. Support these in meson-python mapping of installation paths to wheel locations. Replaces #571.


rgommers (approved):
LGTM, thanks @dnicolodi!

PR #585, opened 2024-02-29
Minor README tweak

lesteve:
The main goal is to be able to have CI run in each push in https://github.com/mesonbuild/meson-python/pull/579. cc @rgommers


rgommers (approved):
Thanks @lesteve. I tweaked the commit message a bit, since our policy here is to rebase-merge all commits and using NumPy style DOC/BUG/etc. commit prefixes.

PR #599, opened 2024-03-12
TST: add test package depending on a cmake subproject

dnicolodi:


dnicolodi:
It depends on your definition of draft PR 🙂. The PR is complete, but there is a bug in Meson that makes merging it not productive. I keep this open just to remember ourselves of the bug: I don't know cmake at all and I don't have the spare cycles required to investigate this.

rgommers:
👍🏼 

In my book: 
- draft = "no need to review it"
- non-draft & green = "ready for review & merge"

dnicolodi:
This was supposed to be non-draft and red, thus also not ready for review, probably. But it is so old that the test results have been flushed out. Setting it as draft now.

PR #605, opened 2024-03-24
CI: run tests also with GraalPy 24.0

dnicolodi:


dnicolodi:
The tests do not pass with GraalPy but pytest is weird on this platform making debugging the problem not really pleasant and I don't really know if it is worth it. At a quick look, it seems that we generate wheel tags that are not compatible with the ones generated by `packaging`. It also looks like that GraalPy, at least in the version installed on the GitHub Action runners, patches `packaging`. Unless someone asks for it, I'm not really motivated to untangle this.

rgommers:
Agreed. I know of GraalPy, but haven't encountered it in real-world situations yet.

dnicolodi:
Yeah. This actually started with me checking that we run the tests on PyPy 3.10 after I saw mesonbuild/meson#12979 and noticing that GitHub Actions offer GraalPy too. For the records, we run the tests with PyPy 3.10.

PR #612, opened 2024-04-15
CI: run tests with pyproject-metadata 0.8.0

dnicolodi:


rgommers:
Once `pyproject-metadata` 0.8.0 is released, should we tweak CI to ensure we have at least one job with the oldest supported version, and the rest continue defaulting to the latest version?

dnicolodi:
I'm debating whether once pyproject-metadata 0.8.0 is released we should keep compatibility with older releases or require 0.8.0 with the next (intended as 0.17.0, not the upcoming 0.16.0) meson-python release and drop the workaround for the bugs fixed in pyproject-metadata (see #613 for what these are). In the past we updated the required minimum version quite aggressively. I don't know if we should be a bit more conservative this time around, given that meson-python is now a build dependency for more packages.

rgommers:
I think we can still bump fairly aggressively if there's a significant benefit (bug fix we need etc.). We want to be a bit more conservative on average though probably. The main issue occurs for distros that don't have build isolation I think. They tend to unvendor vendored deps, so if we bump the lower bound then we're bumping it for other tools that use `pyproject-metadata` as well.

`pyproject-metadata` is quite small and well-behaved, so we probably can't cause too much pain. I was in the process of writing "it's worse for our dependency on `packaging`" - and then realized that we don't actually express that dependency. I don't have a memory of us discussing that before - but that seems wrong to me. We seem to rely on `pyproject-metadata` continuing to depend on it. WDYT about adding it with a reasonable lower bound (in the 19.0-22.0 range probably)?

dnicolodi:
I didn't realize either that we don't have an explicit dependency on packaging. I think it is an oversight and not a decision. We definitely should add it. pyproject-metadata requires packaging >= 19.0. AFAIK we don't depend on anything introduced in later versions, so we should add the same constraint.

rgommers:
Okay, I'll open a separate PR now.

PR #613, opened 2024-04-15
MAINT: bump minimum required pyproject-metadata version to 0.8.0

dnicolodi:


dnicolodi:
This is ready to merge when we are ready to bump the minimum required pyproject-metadata version.

rgommers:
I think it's okay to bump to a fairly recent version. PDM and scikit-build-core now both vendor `pyproject-metadata`, so it doesn't seem likely that we'll cause problems for anyone in pretty aggressively requiring the most recent version.

It may be useful to check `pyproject_metadata.__version__` to guard against potential mismatches? It's easy for distros to miss keeping their dependency metadata for meson-python fully in sync, and if we somehow pick up a `pyproject_metadata` 0.7.1 somewhere, it could otherwise yield confusing errors.

rgommers:
It's been 6 months, so I think this can be merged now after a small update to resolve the merge conflict, and possibly add the below check:

> It may be useful to check `pyproject_metadata.__version__` to guard against potential mismatches?

dnicolodi:
I think it could be useful to cut a release (let's call it 0.16.1 or 0.17.0) before merging this. There are a couple of bug fixes on main that it would be nice to get out with minimal effort from distributors. I can prepare the release.

I'm not convinced that we should start verifying the version of all dependencies, just because someone may have done a poor job in packaging. We haven't checked that pyproject-metadata is version 0.7.1 or later so far, thus adding a check for version 0.8.0 or later should not be necessary either. We check for ninja and meson versions because these are not Python libraries and we had evidence of the wrong ones being picked up because of `$PATH` being set incorrectly.

rgommers:
Good point, +1 for doing a release first.

Okay, fair enough re version check - we can always add it later in case bug reports show a real-world need.

dnicolodi:
I've prepared the release, see #687

PR #614, opened 2024-04-16
Prepare release 0.16.0

dnicolodi:
@rgommers Unless you have anything against it I'm going to get this out of the door in the next few days. The pending PRs will be material for 0.17.0

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,36 @@
 Changelog
 +++++++++
 
+0.16.0
+======
+
+- Correctly detect a wheel shipping shared libraries as being platform
+  dependent also when it does not ship compiled executables or Python
+  extension modules.
+- Use 11.0 as minimum macOS platform ABI tag on arm64: lower versions
+  do not exist on arm64.
+- Improve parsing of the ``$ARCHFLAGS`` environment
+  variable. Correctly handle repeated ``-arch`` flags.
+- Do not break when trying to log filename containing Unicode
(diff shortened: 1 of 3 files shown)


rgommers (approved):
Thanks @dnicolodi. I think that covers all the user-visible changes (outside of docs updates).

The one thing I'd consider adding is to note that the docs moved to https://mesonbuild.com/meson-python/. Worth including?

rgommers:
Ah, and: pre-commit has a minor complaint

rgommers:
Do you want to do this release, or do you want me to take care of it? Either way you may want to add your GPG key to `docs/about.rst`?

dnicolodi:
#594 should be mentioned too, probably. I'm not sure mentioning the documentation location change in a changelog that is visible only at the new location is that useful. But I don't have anything against it.

rgommers:
> #594 should be mentioned too, probably

Ah yes, agreed. There are two editable install related entries; this one is deserving of a separate mention indeed.

dnicolodi:
If you have the time for it, I'll appreciate if you could cut the release once the last bits are committed.

rgommers:
Sure, will do. I've been utterly swamped for too long, but I should just make time. Turning off Slack now and will keep working on meson-python for most of today:)

dnicolodi:
Thanks @rgommers! I updated the changelog, PTAL. I think there isn't anything else that we want to include in 0.16.0

rgommers:
LGTM, in it goes.

Issue #540, opened 2023-11-18
Document how to use a persistent build directory and associated caveats


dnicolodi:
- Explain why you'd want to use the ``build-dir`` option
- Mention it works properly when using ``--cross-file`` (eg. https://github.com/scipy/scipy/blob/1c836efe5ff37ffa4490756269b060a464690e62/.github/workflows/wheels.yml#L180-L188)
- Explain how ``build-dir`` works
- Warn that ``meson-python`` version mismatches in the build directory can cause trouble

Issue #246, opened 2022-12-15
Improve logging on build failure by displaying file contents of `meson-log.txt`


rgommers:
The current output from `pip` when encountering a build error is very poor. gh-231 will fix one part of that, by reducing the traceback of irrelevant pip/meson-python internals. There is a second part however which we need to improve on, which is showing the actual root cause of a build failure when possible. Now what happens is the user sees something like:

```
../../scipy/meson.build:40:0: ERROR: Command "/usr/bin/python3 -c import os; os.chdir(".."); import numpy; print(numpy.get_include())" failed with status 1.

A full log can be found at /tmp/pip-install-3okqoo15/scipy_a5afe2fae16c4250875f20d9db43706f/.mesonpy-u88p7q3f/build/meson-logs/meson-log.txt
```
and is then left wondering "_why does `import numpy` fail, given `numpy` is a build dependency?"_. The answer is in `meson-log.txt`, which `pip` has unhelpfully already cleaned up. So it's impossible to diagnose the issue (latest example for scipy [here](https://github.com/scipy/scipy/issues/16784#issuecomment-1352898268)).

Two options to improve here that I can think of:
1. Don't clean up the log file
2. Print to log file contents to stdout (possibly followed by repeating the root cause from the original stdout content)

(2) is probably better than (1).

rgommers:
Follow-up: the way to not clean up the build dir (and hence the log file) is:
```
pip install . --config-settings=builddir=build
```
I'm not sure that that is guaranteed to stay working, it seems like a side-effect of the way `Project.with_temp_working_dir` is implemented. And it's actually `meson-python` that cleans things up; `pip`'s `--no-clean` works as advertised for the build env itself.

FFY00:
> And it's actually `meson-python` that cleans things up; `pip`'s `--no-clean` works as advertised for the build env itself.

We should consider this a bug, I think. There's no reason for us to clean things up on errors.

rgommers:
I think we should do both avoid cleaning up ourselves _and_ improve terminal output. Right now I think `pip` will still clean up for a regular isolated build, so the user trying to diagnose a build failure still has a problem.

rgommers:
I had a look at this, it's a little complex with multiple nested context managers and a subprocess, so let me add my thoughts here first. This is the code in question that yields a new `Project` instance in a tmpdir:

[...]

PR #616, opened 2024-04-17
DOC: add docs on using the `build-dir` config setting

rgommers:
Closes gh-540

Also touches on gh-246, which is a common enough hiccup that it's useful to explicitly show dealing with `meson-log.txt` in CI.

### DIFF
--- a/docs/how-to-guides/config-settings.rst
+++ b/docs/how-to-guides/config-settings.rst
@@ -19,7 +19,7 @@ the ``-C`` short command line option:
 
 .. tab-set::
 
-    .. tab-item:: pypa/buid
+    .. tab-item:: pypa/build
         :sync: key_pypa_build
 
 	.. code-block:: console
@@ -59,3 +59,70 @@ immediately install it, replace ``wheel`` with ``install``.  See the
 
 .. _build: https://pypa-build.readthedocs.io/en/stable/
 .. _pip: https://pip.pypa.io/
+
+
+Using a persistent build directory
+==================================
+
+By default, ``meson-python`` uses a temporary build directory which is
+deleted when the build terminates. A persistent build directory allows
+faster incremental builds and to access build logs and intermediate
+build artifacts. The ``build-dir`` :ref:`config setting
+<reference-config-settings>` instructs ``meson-python`` to use a
+user-specified build directory which will not be deleted. For example:
+
+.. tab-set::
+
+    .. tab-item:: pypa/build
+        :sync: key_pypa_build
+
+	.. code-block:: console
+
+	   $ python -m build -Cbuild-dir=build
+
+    .. tab-item:: pip
+        :sync: key_pip
+
+	.. code-block:: console
+
+	   $ python -m pip install . -Cbuild-dir=build
+
+After running this command, the ``build`` directory will contain all
+the build artifacts and support files created by ``meson``, ``ninja``
+and ``meson-python``.  The same build directory can be used by
+subsequent invocations of ``meson-python``, avoiding the need to
+rebuild the whole project when testing changes during development.
+
+Using a permanent build directory also aids in debugging a failing
+build by allowing access to build logs and intermediate build outputs,
+including the Meson introspection files and detailed log. The latter
+is stored in the ``meson-logs/meson-log.txt`` file in the build
+directory and can be useful to diagnose why a build fails at the
+project configuration stage. For example, to understand why dependency
+detection failed, it is often necessary to look at the ``pkg-config``
+or other dependency detection methods output.
+
+Access to detailed logs and intermediate build outputs is particularly
+helpful in CI setups where introspecting the build environment is
+usually more difficult than on a local system. Therefore, it can be
+useful to show the more detailed log files when the CI build step
+fails. For example, the following GitHub Actions workflow file snippet
+shows the detailed Meson setup log when the build fails:
+
+.. code-block:: yaml
+
+    - name: Build the package
+      run: python -m build --wheel -Cbuild-dir=build
+    - name: Show meson-log.txt
+      if: failure()
+      run: cat build/meson-logs/meson-log.txt
+
+Replacing ``failure()`` with ``always()`` in the code above will
+result in the Meson log file always being shown. See the GitHub
+Actions documentation__ for more details. Please be aware that the
+setup log can become very long for complex projects, and the GitHub
+Actions web interface becomes unresponsive when the running job emits
+many output lines.
+
+
+__ https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions
--- a/docs/reference/config-settings.rst
+++ b/docs/reference/config-settings.rst
@@ -23,10 +23,6 @@ them.
    exists and contains a valid Meson build directory setup, the
    project will be reconfigured using ``meson setup --reconfigure``.
 
-   The same build directory can be used by subsequent invocations of
-   ``meson-python``. This avoids having to rebuild the whole project
-   when testing changes during development.
-
    For backward compatibility reasons, the alternative ``builddir``
    spelling is also accepted.
 


docs/how-to-guides/config-settings.rst
```diff
@@ -59,3 +59,63 @@ immediately install it, replace ``wheel`` with ``install``.  See the
 
 .. _build: https://pypa-build.readthedocs.io/en/stable/
 .. _pip: https://pip.pypa.io/
+
+
+Using a persistent build directory
+==================================
+
+By default, ``meson-python`` will create a temporary build directory, which
+gets cleaned up after the sdist or wheel has been built, or if the build fails.
```
dnicolodi, line 68: I would tweak this to "By default, meson-python creates a temporary build directory, which is deleted when the build terminates." It is a bit more direct and avoid the expression "cleaned up" which is not well defined.

docs/how-to-guides/config-settings.rst
```diff
@@ -59,3 +59,63 @@ immediately install it, replace ``wheel`` with ``install``.  See the
...
+The ``build-dir`` config seting (see :ref:`reference-config-settings`) can be
```
dnicolodi, line 69: I would simply link "config settings" to the relative documentation page.

docs/how-to-guides/config-settings.rst
```diff
@@ -59,3 +59,63 @@ immediately install it, replace ``wheel`` with ``install``.  See the
...
+used to build in a user-specified build directory instead. For example:
```
dnicolodi, line 70: Should we say immediately here that the user-specified build directory is left alone? Maybe it is obvious.

docs/how-to-guides/config-settings.rst
```diff
@@ -59,3 +59,63 @@ immediately install it, replace ``wheel`` with ``install``.  See the
...
+
+    .. tab-item:: pip
+        :sync: key_pip
+
+	.. code-block:: console
+
+	   $ python -m pip install . -Cbuild-dir=build
+
+
+The ``build`` directory will now contain all the build output created by
```
dnicolodi, line 89: I would say "build artifacts and support files" instead than "build output". I think we use the latter to indicate the console logs usually.

docs/how-to-guides/config-settings.rst
```diff
@@ -59,3 +59,63 @@ immediately install it, replace ``wheel`` with ``install``.  See the
...
+- For accessing build logs and intermediate build outputs when debugging
+
+The same build directory can be used by subsequent invocations of
+``meson-python``. This avoids having to rebuild the whole project when testing
+changes during development. This will work with any config settings flags,
+including for cross compilation. The one thing to keep in mind is that changing
+the version of ``meson`` or ``meson-python`` in the build environment between
+separate builds in the same build directory is not supported (it often works,
+but may fail - and the error message cannot be guaranteed to be clear about the
+reason).
```
dnicolodi, line 103: I don't think this is the case anymore. We fixed meson-python so that `--reconfigure` is used when needed.

docs/how-to-guides/config-settings.rst
```diff
@@ -59,3 +59,63 @@ immediately install it, replace ``wheel`` with ``install``.  See the
...
+
+Using ``build-dir`` makes it possible to access the log files that Meson
+creates. When a build fails, Meson's CLI logging will display the reason for
```
dnicolodi, line 106: "...to access the Meson's log and introspection files." Shorter and more precise.

docs/how-to-guides/config-settings.rst
```diff
@@ -59,3 +59,63 @@ immediately install it, replace ``wheel`` with ``install``.  See the
...
+the failure in most cases. However, it is sometimes necessary to look at more
+detailed logs like ``meson-logs/meson-log.txt``. A common reason for this is
```
dnicolodi, line 108: I would make this more to the point and more precise. I would remove the sentence starting "When a build fails..." and I would reword the next sentence to something like: "To diagnose why a build fails at the project configuration stage, it is sometimes necessary to inspect the detailed Meson log in the ``meson-logs/meson-log.txt`` file in the build directory. The complete path the the log file is available in the command output when the build fails. For example, when dependency detection fails it is often..."

docs/reference/config-settings.rst
```diff
@@ -23,10 +23,6 @@ them.
    exists and contains a valid Meson build directory setup, the
    project will be reconfigured using ``meson setup --reconfigure``.
 
-   The same build directory can be used by subsequent invocations of
-   ``meson-python``. This avoids having to rebuild the whole project
-   when testing changes during development.
-
```
dnicolodi, line 29: Why is this paragraph removed?

dnicolodi (changes requested):
Thanks @rgommers It is a very useful addition to the docs. I have a few stylistic comment and a remark regarding a limitation that is described but I don't think is there.

docs/reference/config-settings.rst
dnicolodi: I'm ok with removing it.

docs/how-to-guides/config-settings.rst
rgommers: I'm not sure that `meson setup --reconfigure` is 100% the same as `rm -rf builddir && meson setup builddir` though. I recently ran into a very odd failure (https://github.com/scipy/scipy/issues/20433) which had to do with a pickled version of `meson` that is reloaded at build time.

The Meson docs don't say anything about this either way I believe, but my take-away from https://github.com/mesonbuild/meson/pull/12357#issuecomment-2050016590 was "if you upgrade Meson, no guarantees unless you delete the build dir". I could be wrong though - @eli-schwartz?

eli-schwartz: Reconfigure is *definitely* a section of the codebase covered by meson's explicit check for coredata version.

If meson detects that you've upgraded meson and the version of the coredata doesn't match, it essentially does a wipe and from-scratch configure, instead of the normal reconfigure logic. Alternatively, if you haven't asked to setup the project it isn't apropos to do so automatically just because the coredata is incompatible -- from memory, things like test/install will just print a "fatal error: xxx. Try running meson setup --wipe".

Reconfigure is indeed different from deleting the build directory. It will:
- keep dependency lookup caches instead of actually running pkg-config/cmake for dependency()
- keep incremental build outputs, since they will anyways be invalidated if the command line changes
- preserve current option states

docs/how-to-guides/config-settings.rst
rgommers: adopted

docs/how-to-guides/config-settings.rst
rgommers: thanks both. removed these sentences now.

docs/how-to-guides/config-settings.rst
rgommers: adopted

docs/how-to-guides/config-settings.rst
rgommers: good idea I think, added "which will not be deleted"

docs/how-to-guides/config-settings.rst
rgommers: adopted

docs/reference/config-settings.rst
rgommers: okay, resolving this then.

rgommers:
Addressed all comments, I think this is ready now.

rgommers:
Cirrus is not happy. Can be ignored for this PR, since its docs-only.

dnicolodi:
@rgommers I find fixing prose via GitHub reviews not very efficient. I edited your text to make it flow better and pushed it to your branch. No conceptual changes intended. Please have a look.

henryiii:
We could change the other places that have a pip version to use `-C` now, I think? (Not in this PR!)

rgommers:
> We could change the other places that have a pip version to use `-C` now, I think? (Not in this PR!)

Agreed.

rgommers:
The rephrasings LGTM - time to get this in. Thanks @dnicolodi and @henryiii

PR #618, opened 2024-04-17
CI: fix docs build and deploy workflow 

dnicolodi:
Did I ever complain about the GitHub Actions documentation being an horrible mess?

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -31,7 +31,7 @@ jobs:
           python-version: '3.11'
       - uses: actions/checkout@v4
         with:
-          ref: ${{ github.event.push.tag || github.event.inputs.tag }}
+          ref: ${{ github.event.inputs.tag || github.ref }}
       - run: python -m pip install .[docs]
       - run: python -m sphinx -W docs/ build/docs/
       - uses: actions/upload-pages-artifact@v3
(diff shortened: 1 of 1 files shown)


rgommers (approved):
Thanks @dnicolodi. I assume you tested locally, so go ahead and merge when happy. If it takes an extra iteration that's also fine - this stuff tends to be tricky to test.

> Did I ever complain about the GitHub Actions documentation being an horrible mess?

No, but you may:)

dnicolodi:
No, I haven't tested this locally and I run out of time for the moment. I was hoping that someone with more experience with GitHub Actions could comment on the validity of what I've done here. I'll test this in my fork. Anyhow, for some reason the GitHub Actions runners haven't pick up this PR...

rgommers:
Ah okay. Let me compare with what works in another project. I have a fair amount of experience with GHA, but this kind of thing is mostly trial-and-error rather than understand-then-do unfortunately. 

The updated docs are up on https://mesonbuild.com/meson-python/, so this isn't a burning issue I think.

agriyakhetarpal:
Hi, I just looked at the changes made here, and I think they should work. 👍 Just to note that `github.ref` is `heads/refs/<name of the tag>` and not just `<name of the tag>`, so that could be the only possible place where this can break.

dnicolodi:
> The updated docs are up

I triggered the workflow manually to publish them. I was not sure whether the issue was with the trigger or with the job definition, so doing that ruled out the latter.

dnicolodi:
I've testes this and it works as expected. It needs an adjustment to the environment protection rules in the project settings to be able to deploy the documentation from a tag. I've already done this.

Issue #499, opened 2023-09-19
Need to augment wheel tag computation code to account for nogil builds


dnicolodi:
Starting with Python 3.13 there will be a nogil variant of the interpreter that will need to be get an interpreter tag in the wheel tag string. See https://discuss.python.org/t/python-abis-and-pep-703/34018

dnicolodi:
See also https://github.com/pypa/packaging/issues/727 https://github.com/pypa/packaging/pull/728 https://github.com/pypa/packaging/pull/747

rgommers:
gh-621 addresses part of this. As of today, it's not yet very easy to test with a free-threaded CPython build and make the test suite pass. The hiccups are:

- `pip` does not yet have support. Two issues: it incorrectly installs incompatible `abi3` wheels, and it refuses to install wheels with the `cp313t`/`cp313td` ABI tag. The former doesn't affect meson-python, the latter was a little painful to work around (see below for how).
- `cython` does not yet have full support. The latest release (3.0.10) fails to build one of our test packages (the `complex` one), however the branch from https://github.com/cython/cython/pull/6137 does work.
- CPython needs to be built from source (`./configure --with-pydebug --disable-gil`)

In order to make `pip` install wheels with the `t` tag, the `_fail_if_link_is_unsupported_wheel` function in `pip/_internal/resolution/resolvelib/factory.py` needs to be patched - simply replacing `raise UnsupportedWheel(msg)` with `return` is enough.

Then the `venv` fixture in our test suite needs patching, because it installs a fresh copy of `pip` in the venv's it creates. This hacky patch does that (replace path to site-packages as needed):

```diff
diff --git a/tests/conftest.py b/tests/conftest.py
index 80303a14f..68df9e2ef 100644
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -101,7 +101,14 @@ class VEnv(EnvBuilder):
 @pytest.fixture()
 def venv(tmp_path_factory):
     path = pathlib.Path(tmp_path_factory.mktemp('mesonpy-test-venv'))
-    return VEnv(path)
+    fixture = VEnv(path)
+
+    # Apply fix to `pip` installed into venv to avoid unknown tag issue
+    sitepkg_dir = '/home/rgommers/mambaforge/envs/nogil/lib/python3.13/site-packages'
+    sitepkg_venv = str(path) + '/lib/python3.13/site-packages'
+    _file = '/pip/_internal/resolution/resolvelib/factory.py'
[...]

PR #621, opened 2024-04-17
MAINT: improve support for free-threaded CPython

rgommers:
As noted in https://github.com/pypa/packaging/issues/727, free-threaded CPython does not support the limited API. Hence raise an exception if a `limited-api` build is requested.

Also skip/update a few tests that were failing when run under a free-threaded interpreter.

Note that we are already producing the correct wheel tags, since that comes from `sysconfig.get_config_var('EXT_SUFFIX')`.

Addresses part of gh-499. There will be more to do - most importantly adding a CI job. This probably only makes sense to do once the `setup-python` action adds support (xref https://github.com/actions/setup-python/issues/771). This may happen around the CPython 3.13 beta 1 release (May 7th).

I'll add notes on how to test with a free-threaded build on gh-499 (it's not completely trivial yet).


mesonpy/__init__.py
```diff
@@ -758,6 +758,11 @@ def __init__(
             if not value:
                 self._limited_api = False
 
+            if self._limited_api:
+                if bool(sysconfig.get_config_var("Py_GIL_DISABLED")):
```
dnicolodi, line 762: We use single quotes for strings. This is why pre-commit-hook is complaining.

mesonpy/__init__.py
```diff
@@ -758,6 +758,11 @@ def __init__(
...
+                    raise BuildError('Free-threaded CPython does not support '
+                                     'the Python limited API')
```
dnicolodi, line 764: You can most likely write the error message on one line without exceeding the maximum line length we enforce.

However, I'm not sure whether this error message is clear enough for users. Let's imagine an user trying to `pip install` a package from PyPI. If the user is using a nogil Python and the package uses the limited API, this message may not be enough to understand what is going on, ie why is meson-python trying to use the limited API. What do you think?

Maybe something on the lines of "The package requires the Python limited API but it is not supported by free-threaded Python" but with a better grammar.

By the way, I find that "free threaded" is not the most user-friendly name that could have been used for this. But this boat has sailed...

tests/test_editable.py
```diff
@@ -16,7 +18,17 @@
 
 from mesonpy import _editable
 
-from .test_wheel import EXT_SUFFIX
+from .test_wheel import EXT_SUFFIX, NOGIL_BUILD
+
+
+def find_cython_version():
+    _cython_version_str = subprocess.run(['cython', '--version'], check=True,
```
dnicolodi, line 25: It does not matter, but why the leading `_`?

tests/test_wheel.py
```diff
@@ -299,7 +302,11 @@ def test_limited_api(wheel_limited_api):
 @pytest.mark.skipif(MESON_VERSION < (1, 2, 99), reason='Meson version too old')
 @pytest.mark.xfail('__pypy__' in sys.builtin_module_names, reason='PyPy does not use special modules suffix for stable ABI')
 def test_limited_api_bad(package_limited_api, tmp_path):
-    with pytest.raises(mesonpy.BuildError, match='The package declares compatibility with Python limited API but '):
+    msg_start = 'The package declares compatibility with Python limited API but '
+    if NOGIL_BUILD:
+        msg_start = 'Free-threaded CPython does not support '
+
+    with pytest.raises(mesonpy.BuildError, match=msg_start):
```
dnicolodi, line 309: I think it would be clearer if the test would be simply skipped for nogil Python, otherwise it is harder to understand the purpose of the test.

dnicolodi (changes requested):
Thanks @rgommers for looking into this. This will be important as soon as the first Python 3.13 is out and more people starts testing the nogil interpreter.

mesonpy/__init__.py
rgommers: Yeah ... always takes me one iteration, since I am not touching pre-commit locally. Fixed now.

tests/test_wheel.py
rgommers: Sure, changed to skipping it.

tests/test_editable.py
rgommers: Taken over from `_meson_ver_str` in `test_wheel.py`. Agreed the underscore isn't needed, removed now.

mesonpy/__init__.py
rgommers: I made it more verbose - along the lines of your suggestion, and also added a hint about how to get around it, which should always work.

> By the way, I find that "free threaded" is not the most user-friendly name that could have been used for this. But this boat has sailed...

Yeah agreed ... but unlikely to change at this point.

rgommers:
Looks like a typo fix in `pyproject-metadata` 0.8.0 is messing with CI here:
```
FAILED tests/test_sdist.py::test_pep621 - AssertionError: assert 'Metadata-Ver...1 metadata!\n' == 'Metadata-Ver...1 metadata!\n'
  
  Skipping 116 identical leading characters in diff, use -v to show
  Skipping 906 identical trailing characters in diff, use -v to show
  - ords: full metadata
  ?           ^
  + ords: full,metadata
  ?           ^
```

gh-613 has a draft fix already, but with more changes. Do you want to finish that PR, or should we fix this typo failure separately first?

dnicolodi:
> Looks like a typo fix in `pyproject-metadata` 0.8.0 is messing with CI here:

This should be fixed in main already. I think you just need to rebase on current main.

tests/test_editable.py
```diff
@@ -16,7 +18,17 @@
...
 
-from .test_wheel import EXT_SUFFIX
+from .test_wheel import EXT_SUFFIX, NOGIL_BUILD
+
+
+def find_cython_version():
+    cython_version_str = subprocess.run(['cython', '--version'], check=True,
+                                        stdout=subprocess.PIPE, text=True).stdout
+    version_regex = re.compile(r"(\d{1,4}\.\d{1,4}\.?\d{0,4})")
+    version_str = version_regex.search(cython_version_str).group(0)
```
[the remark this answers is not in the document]
rgommers: good point, adopted

mesonpy/__init__.py
```diff
@@ -758,6 +758,12 @@ def __init__(
             if not value:
                 self._limited_api = False
 
+            if self._limited_api:
+                if bool(sysconfig.get_config_var('Py_GIL_DISABLED')):
+                    raise BuildError('Package targets Python\'s Limited API, which is not supported by free-threaded '
+                                     'CPython. The `python.allow_limited_api` Meson build option may be used to '
+                                     'override the package default.')
+
```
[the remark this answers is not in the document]
rgommers: Added (two indents actually, with a simplification), and fixed the grammar higher up. Which in turn required a regex in `test_project.py` to be updated too.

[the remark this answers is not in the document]
dnicolodi: I think half of the dedent got lost. We use double quotes to delimit literals everywhere else in the error messages. Maybe we should do the same here. Sorry for not noticing this earlier. I can fixup these things when merging.

dnicolodi (approved):

mesonpy/__init__.py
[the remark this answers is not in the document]
dnicolodi: I fixed the tiny issues and pushed to your branch. Merging when the CI jobs finish.

Issue #624, opened 2024-05-01
Document setting limited-api via build option better


lpsinger:
The [`limited-api` option](https://meson-python.readthedocs.io/en/latest/reference/pyproject-settings.html#cmdoption-arg-tool.meson-python.limited-api) can only be set through pyproject.toml, not through the [build config settings](https://meson-python.readthedocs.io/en/latest/reference/config-settings.html) that you can provide on the command line to `pip` or `python -m build`.

Would you please consider supporting `limited-api` as a build config setting? I am using cibuildwheel to build a package that uses the limited API, but my build fails when it tries to install the build package and run my tests under pypy with the following error:

```
  ERROR: hpx-0.0.1-pp39-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl is not a supported wheel on this platform.
```

I would like to be able to customize the build so that it sets `limited-api` to false for PyPy, but I can't do this because `limited-api` is not a build config setting. Here's what I would like to do:

```toml
[tool.cibuildwheel]
config-settings.limited-api = true

# PyPy does not support the Python limited API.
[[tool.cibuildwheel.overrides]]
select = "pp*"
config-settings.limited-api = false
```

eli-schwartz:
build config settings, as in the kind that are compatible with `pip install --config-settings="...."`, take the form:

```
--config-settings=setup-args="-Dpython.allow_limited_api=true"
```

as documented by https://mesonbuild.com/Builtin-options.html#module-options

rgommers:
Yes, we should be handling `python.allow_limited_api`: https://github.com/mesonbuild/meson-python/blob/6c728bd047372ea9ad7427851d5fc414159bb4a6/mesonpy/__init__.py#L758

And we just added a more descriptive error a few lines below that (not in the 0.16.0 release yet), but only for free-threaded CPython and not PyPy.

Guess we should improve the error message and the docs a bit more.

lpsinger:
Yes, this worked, although somehow I found it hard to find this in the documentation.

rgommers:
Great, thanks for confirming. Changed the issue title to reflect that we need to improve the docs here.

trim21:
What's the best way to do this currently?

Enable limited-api pyproject.toml and disable it with `setup-args`?

rgommers:
[...]

PR #626, opened 2024-05-03
Disable abi3 wheel tag for PyPy

lpsinger:
Pip does not support installing PyPy wheels with the abi3 tag. If you build a wheel for PyPy and set the ABI tag to `abi3`, pip will not be able to install it.

```
$ docker run --rm -it quay.io/pypa/manylinux2014_x86_64
[root@cd7b2d465170 /]# /opt/python/pp39-pypy39_pp73/bin/python
Python 3.9.19 (a2113ea87262, Apr 21 2024, 05:40:24)
[PyPy 7.3.16 with GCC 10.2.1 20210130 (Red Hat 10.2.1-11)] on linux
>>>> from pip._internal.utils.compatibility_tags import get_supported
>>>> for tag in get_supported():
....     print(tag)
....
pp39-pypy39_pp73-manylinux_2_17_x86_64
pp39-pypy39_pp73-manylinux2014_x86_64
```

Full output:

<details>

```
>>>> from pip._internal.utils.compatibility_tags import get_supported
>>>> for tag in get_supported():
....     print(tag)
....
pp39-pypy39_pp73-manylinux_2_17_x86_64
pp39-pypy39_pp73-manylinux2014_x86_64
pp39-pypy39_pp73-manylinux_2_16_x86_64
pp39-pypy39_pp73-manylinux_2_15_x86_64
pp39-pypy39_pp73-manylinux_2_14_x86_64
pp39-pypy39_pp73-manylinux_2_13_x86_64
pp39-pypy39_pp73-manylinux_2_12_x86_64
pp39-pypy39_pp73-manylinux2010_x86_64
pp39-pypy39_pp73-manylinux_2_11_x86_64
pp39-pypy39_pp73-manylinux_2_10_x86_64
pp39-pypy39_pp73-manylinux_2_9_x86_64
pp39-pypy39_pp73-manylinux_2_8_x86_64
pp39-pypy39_pp73-manylinux_2_7_x86_64
pp39-pypy39_pp73-manylinux_2_6_x86_64
pp39-pypy39_pp73-manylinux_2_5_x86_64
pp39-pypy39_pp73-manylinux1_x86_64
pp39-pypy39_pp73-linux_x86_64
pp39-none-manylinux_2_17_x86_64
pp39-none-manylinux2014_x86_64
pp39-none-manylinux_2_16_x86_64
pp39-none-manylinux_2_15_x86_64
pp39-none-manylinux_2_14_x86_64
pp39-none-manylinux_2_13_x86_64
pp39-none-manylinux_2_12_x86_64
pp39-none-manylinux2010_x86_64
pp39-none-manylinux_2_11_x86_64
pp39-none-manylinux_2_10_x86_64
pp39-none-manylinux_2_9_x86_64
pp39-none-manylinux_2_8_x86_64
pp39-none-manylinux_2_7_x86_64
pp39-none-manylinux_2_6_x86_64
pp39-none-manylinux_2_5_x86_64
pp39-none-manylinux1_x86_64
pp39-none-linux_x86_64
py39-none-manylinux_2_17_x86_64
py39-none-manylinux2014_x86_64
py39-none-manylinux_2_16_x86_64
py39-none-manylinux_2_15_x86_64
py39-none-manylinux_2_14_x86_64
py39-none-manylinux_2_13_x86_64
py39-none-manylinux_2_12_x86_64
py39-none-manylinux2010_x86_64
py39-none-manylinux_2_11_x86_64
py39-none-manylinux_2_10_x86_64
py39-none-manylinux_2_9_x86_64
py39-none-manylinux_2_8_x86_64
py39-none-manylinux_2_7_x86_64
py39-none-manylinux_2_6_x86_64
py39-none-manylinux_2_5_x86_64
py39-none-manylinux1_x86_64
py39-none-linux_x86_64
py3-none-manylinux_2_17_x86_64
py3-none-manylinux2014_x86_64
py3-none-manylinux_2_16_x86_64
py3-none-manylinux_2_15_x86_64
py3-none-manylinux_2_14_x86_64
py3-none-manylinux_2_13_x86_64
py3-none-manylinux_2_12_x86_64
py3-none-manylinux2010_x86_64
py3-none-manylinux_2_11_x86_64
py3-none-manylinux_2_10_x86_64
py3-none-manylinux_2_9_x86_64
py3-none-manylinux_2_8_x86_64
py3-none-manylinux_2_7_x86_64
py3-none-manylinux_2_6_x86_64
py3-none-manylinux_2_5_x86_64
py3-none-manylinux1_x86_64
py3-none-linux_x86_64
py38-none-manylinux_2_17_x86_64
py38-none-manylinux2014_x86_64
py38-none-manylinux_2_16_x86_64
py38-none-manylinux_2_15_x86_64
py38-none-manylinux_2_14_x86_64
py38-none-manylinux_2_13_x86_64
py38-none-manylinux_2_12_x86_64
py38-none-manylinux2010_x86_64
py38-none-manylinux_2_11_x86_64
py38-none-manylinux_2_10_x86_64
py38-none-manylinux_2_9_x86_64
py38-none-manylinux_2_8_x86_64
py38-none-manylinux_2_7_x86_64
py38-none-manylinux_2_6_x86_64
py38-none-manylinux_2_5_x86_64
py38-none-manylinux1_x86_64
py38-none-linux_x86_64
py37-none-manylinux_2_17_x86_64
py37-none-manylinux2014_x86_64
py37-none-manylinux_2_16_x86_64
py37-none-manylinux_2_15_x86_64
py37-none-manylinux_2_14_x86_64
py37-none-manylinux_2_13_x86_64
py37-none-manylinux_2_12_x86_64
py37-none-manylinux2010_x86_64
py37-none-manylinux_2_11_x86_64
py37-none-manylinux_2_10_x86_64
py37-none-manylinux_2_9_x86_64
py37-none-manylinux_2_8_x86_64
py37-none-manylinux_2_7_x86_64
py37-none-manylinux_2_6_x86_64
py37-none-manylinux_2_5_x86_64
py37-none-manylinux1_x86_64
py37-none-linux_x86_64
py36-none-manylinux_2_17_x86_64
py36-none-manylinux2014_x86_64
py36-none-manylinux_2_16_x86_64
py36-none-manylinux_2_15_x86_64
py36-none-manylinux_2_14_x86_64
py36-none-manylinux_2_13_x86_64
py36-none-manylinux_2_12_x86_64
py36-none-manylinux2010_x86_64
py36-none-manylinux_2_11_x86_64
py36-none-manylinux_2_10_x86_64
py36-none-manylinux_2_9_x86_64
py36-none-manylinux_2_8_x86_64
py36-none-manylinux_2_7_x86_64
py36-none-manylinux_2_6_x86_64
py36-none-manylinux_2_5_x86_64
py36-none-manylinux1_x86_64
py36-none-linux_x86_64
py35-none-manylinux_2_17_x86_64
py35-none-manylinux2014_x86_64
py35-none-manylinux_2_16_x86_64
py35-none-manylinux_2_15_x86_64
py35-none-manylinux_2_14_x86_64
py35-none-manylinux_2_13_x86_64
py35-none-manylinux_2_12_x86_64
py35-none-manylinux2010_x86_64
py35-none-manylinux_2_11_x86_64
py35-none-manylinux_2_10_x86_64
py35-none-manylinux_2_9_x86_64
py35-none-manylinux_2_8_x86_64
py35-none-manylinux_2_7_x86_64
py35-none-manylinux_2_6_x86_64
py35-none-manylinux_2_5_x86_64
py35-none-manylinux1_x86_64
py35-none-linux_x86_64
py34-none-manylinux_2_17_x86_64
py34-none-manylinux2014_x86_64
py34-none-manylinux_2_16_x86_64
py34-none-manylinux_2_15_x86_64
py34-none-manylinux_2_14_x86_64
py34-none-manylinux_2_13_x86_64
py34-none-manylinux_2_12_x86_64
py34-none-manylinux2010_x86_64
py34-none-manylinux_2_11_x86_64
py34-none-manylinux_2_10_x86_64
py34-none-manylinux_2_9_x86_64
py34-none-manylinux_2_8_x86_64
py34-none-manylinux_2_7_x86_64
py34-none-manylinux_2_6_x86_64
py34-none-manylinux_2_5_x86_64
py34-none-manylinux1_x86_64
py34-none-linux_x86_64
py33-none-manylinux_2_17_x86_64
py33-none-manylinux2014_x86_64
py33-none-manylinux_2_16_x86_64
py33-none-manylinux_2_15_x86_64
py33-none-manylinux_2_14_x86_64
py33-none-manylinux_2_13_x86_64
py33-none-manylinux_2_12_x86_64
py33-none-manylinux2010_x86_64
py33-none-manylinux_2_11_x86_64
py33-none-manylinux_2_10_x86_64
py33-none-manylinux_2_9_x86_64
py33-none-manylinux_2_8_x86_64
py33-none-manylinux_2_7_x86_64
py33-none-manylinux_2_6_x86_64
py33-none-manylinux_2_5_x86_64
py33-none-manylinux1_x86_64
py33-none-linux_x86_64
py32-none-manylinux_2_17_x86_64
py32-none-manylinux2014_x86_64
py32-none-manylinux_2_16_x86_64
py32-none-manylinux_2_15_x86_64
py32-none-manylinux_2_14_x86_64
py32-none-manylinux_2_13_x86_64
py32-none-manylinux_2_12_x86_64
py32-none-manylinux2010_x86_64
py32-none-manylinux_2_11_x86_64
py32-none-manylinux_2_10_x86_64
py32-none-manylinux_2_9_x86_64
py32-none-manylinux_2_8_x86_64
py32-none-manylinux_2_7_x86_64
py32-none-manylinux_2_6_x86_64
py32-none-manylinux_2_5_x86_64
py32-none-manylinux1_x86_64
py32-none-linux_x86_64
py31-none-manylinux_2_17_x86_64
py31-none-manylinux2014_x86_64
py31-none-manylinux_2_16_x86_64
py31-none-manylinux_2_15_x86_64
py31-none-manylinux_2_14_x86_64
py31-none-manylinux_2_13_x86_64
py31-none-manylinux_2_12_x86_64
py31-none-manylinux2010_x86_64
py31-none-manylinux_2_11_x86_64
py31-none-manylinux_2_10_x86_64
py31-none-manylinux_2_9_x86_64
py31-none-manylinux_2_8_x86_64
py31-none-manylinux_2_7_x86_64
py31-none-manylinux_2_6_x86_64
py31-none-manylinux_2_5_x86_64
py31-none-manylinux1_x86_64
py31-none-linux_x86_64
py30-none-manylinux_2_17_x86_64
py30-none-manylinux2014_x86_64
py30-none-manylinux_2_16_x86_64
py30-none-manylinux_2_15_x86_64
py30-none-manylinux_2_14_x86_64
py30-none-manylinux_2_13_x86_64
py30-none-manylinux_2_12_x86_64
py30-none-manylinux2010_x86_64
py30-none-manylinux_2_11_x86_64
py30-none-manylinux_2_10_x86_64
py30-none-manylinux_2_9_x86_64
py30-none-manylinux_2_8_x86_64
py30-none-manylinux_2_7_x86_64
py30-none-manylinux_2_6_x86_64
py30-none-manylinux_2_5_x86_64
py30-none-manylinux1_x86_64
py30-none-linux_x86_64
pp39-none-any
py39-none-any
py3-none-any
py38-none-any
py37-none-any
py36-none-any
py35-none-any
py34-none-any
py33-none-any
py32-none-any
py31-none-any
py30-none-any
```

</details>

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -406,22 +406,23 @@ class _WheelBuilder():
 
     @cached_property
     def _stable_abi(self) -> Optional[str]:
-        if self._limited_api:
-            # PyPy does not use a special extension module filename
-            # suffix for modules targeting the stable API.
-            if '__pypy__' not in sys.builtin_module_names:
-                # Verify stable ABI compatibility: examine files installed
-                # in {platlib} that look like extension modules, and raise
-                # an exception if any of them has a Python version
-                # specific extension filename suffix ABI tag.
-                for path, _ in self._manifest['platlib']:
-                    match = _EXTENSION_SUFFIX_REGEX.match(path.name)
-                    if match:
-                        abi = match.group('abi')
-                        if abi is not None and abi != 'abi3':
-                            raise BuildError(
-                                f'The package declares compatibility with Python limited API but extension '
-                                f'module {os.fspath(path)!r} is tagged for a specific Python version.')
+        # PyPy supports the limited API but does not provide a stable
+        # ABI, therefore extension modules using the limited API do
+        # not use the stable ABI filename suffix and wheels should not
+        # be tagged with the abi3 tag.
+        if self._limited_api and '__pypy__' not in sys.builtin_module_names:
+            # Verify stable ABI compatibility: examine files installed
+            # in {platlib} that look like extension modules, and raise
+            # an exception if any of them has a Python version
+            # specific extension filename suffix ABI tag.
+            for path, _ in self._manifest['platlib']:
+                match = _EXTENSION_SUFFIX_REGEX.match(path.name)
+                if match:
+                    abi = match.group('abi')
+                    if abi is not None and abi != 'abi3':
+                        raise BuildError(
+                            f'The package declares compatibility with Python limited API but extension '
+                            f'module {os.fspath(path)!r} is tagged for a specific Python version.')
             return 'abi3'
         return None
 
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -103,11 +103,13 @@ def test_tag_stable_abi():
     builder = wheel_builder_test_factory({
         'platlib': [f'extension{ABI3SUFFIX}'],
     }, limited_api=True)
-    assert str(builder.tag) == f'{INTERPRETER}-abi3-{PLATFORM}'
+    # PyPy does not support the stable ABI.
+    abi = 'abi3' if '__pypy__' not in sys.builtin_module_names else ABI
+    assert str(builder.tag) == f'{INTERPRETER}-{abi}-{PLATFORM}'
 
 
 @pytest.mark.xfail(sys.version_info < (3, 8) and sys.platform == 'win32', reason='Extension modules suffix without ABI tags')
-@pytest.mark.xfail('__pypy__' in sys.builtin_module_names, reason='PyPy does not use special modules suffix for stable ABI')
+@pytest.mark.xfail('__pypy__' in sys.builtin_module_names, reason='PyPy does not support the stable ABI')
 def test_tag_mixed_abi():
     builder = wheel_builder_test_factory({
         'platlib': [f'extension{ABI3SUFFIX}', f'another{SUFFIX}'],
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -290,6 +290,7 @@ def test_skip_subprojects(package_subproject, tmp_path, arg):
 # Requires Meson 1.3.0, see https://github.com/mesonbuild/meson/pull/11745.
 @pytest.mark.skipif(MESON_VERSION < (1, 2, 99), reason='Meson version too old')
 @pytest.mark.skipif(NOGIL_BUILD, reason='Free-threaded CPython does not support the limited API')
+@pytest.mark.xfail('__pypy__' in sys.builtin_module_names, reason='PyPy does not support the abi3 platform tag for wheels')
 def test_limited_api(wheel_limited_api):
     artifact = wheel.wheelfile.WheelFile(wheel_limited_api)
     name = artifact.parsed_filename


rgommers:
Thanks for working on this @lpsinger. The current behavior certainly doesn't seem quite right - we should never be producing something with a tag that `pip` won't understand.

Related to gh-624. If I understand correctly, with the build option discussed in gh-624 you can disable asking for limited API usage, which will then produce a tag like `pp310-pypy310_pp73-linux_x86_64` that can be installed. This PR is a little different in that you want to avoid having to think about using that build option at all, still use the limited C API, but produce a tag like `pp310-pypy310_pp73-linux_x86_64`. Is that right? If so, that seems reasonable to me, since it reduced friction and the end result should be the same as erroring out and forcing the user to specify `-C-Dpython.allow_limited_api=false`.

CI is a little unhappy, the tests need updating to match:

```
FAILED tests/test_tags.py::test_tag_stable_abi - AssertionError: assert 'pp310-pypy310_pp73-linux_x86_64' == 'pp310-abi3-linux_x86_64'
  
  - pp310-abi3-linux_x86_64
  + pp310-pypy310_pp73-linux_x86_64
FAILED tests/test_wheel.py::test_limited_api - AssertionError: assert 'pypy310_pp73' == 'abi3'
```

lpsinger:
> This PR is a little different in that you want to avoid having to think about using that build option at all, still use the limited C API, but produce a tag like `pp310-pypy310_pp73-linux_x86_64`. Is that right?

That's correct. I'm looking to avoid having to add this line to projects that use the limited API and cibuildwheel: https://github.com/nasa-gcn/hpx/blob/main/pyproject.toml#L81-L84

lpsinger:
> CI is a little unhappy, the tests need updating to match:

I pushed a change for that; let's see if it fixes the CI.

dnicolodi:
My understanding is that PyPy supports the limited API but does not use a dedicated filename suffix for extension modules compiled for the limited API. The Python support in Meson implies the same understanding https://github.com/mesonbuild/meson/blob/cfd57180eef9036c7167c5682b9f3055a540fccc/mesonbuild/scripts/python_info.py#L103-L107 

What I am not sure about is whether, when using the limited API, PyPy implements a stable ABI, and thus the resulting wheels should be taggeg `abi3` or not. If this is the case, the problem being solved here is a bug in `pip` and not something we need to work-around in meson-python.

@mattip Can you confirm that my understanding is correct?

I added the `tool.meson-python.limited-api` setting based on the assumption that PyPy supports the limited API without a dedicated module filename suffix and thus it is not sufficient to look at the filenames of the installed extension modules to determine whether a package uses the limited API only. If the assumption is wrong and PyPy does not support the limited API or supports it without exporting a stable ABI, the setting could in principle be removed.

mattip:
PyPy does not implement a stable ABI. Wheels for PyPy should not be tagged `abi3`. The output of `pip`'s `get_supported` is correct.

dnicolodi:
Thanks for the clarification @mattip. My reasoning for implementing the `tool.meson-python.limited-api` setting is wrong. I think we should remove this setting and rely on the filename suffix of the installed extension modules to decide the ABI tag for the created wheel. We would still accept the parameter for backward compatibility but emit a deprecation warning when we encounter it. @rgommers, what do you think?

dnicolodi:
One of the reasons to remove the `tool.meson-python.limited-api` setting is that it is misnamed. When I introduced it, I thought that using the limited API always results in a stable ABI, and thus in wheels to be tagged with the `abi3` tag. However, this is not the case on PyPy. Therefore, mixing the two concepts is not a good idea. To my defense, the CPython documentation does not make this distinction at all.

rgommers:
My impression is that the current build option is fine, since it works as designed for CPython. For PyPy, proceeding with the regular minor-version-specific tag as done in this PR would seem reasonable to me. The question I have is whether we should be emitting a warning when doing that.

> However, this is not the case on PyPy. Therefore, mixing the two concepts is not a good idea. 

PyPy could implement it in the future in principle (it's just a missing feature now), so I wouldn't attach too much value to this since conceptually it does make sense.

Moreover, PyPy is probably going to fade away pretty quickly over the next 1-2 years - there is unlikely to be a 3.11 release for PyPy, so next year when Python 3.10 starts being dropped that's the end of it. Conda-forge is also starting to drop PyPy builds because of maintenance issues and no real future. Hence deprecating because of PyPy isn't needed imho.

henryiii:
The way selecting `cp311` works is that it's only applicable on CPython with GIL 3.11+. Any older version or other interpreter (including PyPy) ignores this choice. (It wasn't quite right for PyPy until the most recent release, when I was working on free-threading I cleaned it up a little). A user is expected to skip requesting it CMake like this:

```cmake
if(NOT "${SKBUILD_SABI_COMPONENT}" STREQUAL "")
  python_add_library(some_ext MODULE WITH_SOABI USE_SABI 3.11 ...)
else()
  python_add_library(some_ext MODULE WITH_SOABI ...)
endif()
```

(USE_SABI also sets the Limited API define) Though I haven't checked to see exactly how CMake's FindPython module handles it, it might do the right thing (IMO which is to ignore this setting if it doesn't apply).

Not setting the define just means you are not limited to the Limited API, so shouldn't hurt user code.

lpsinger:
> Moreover, PyPy is probably going to fade away pretty quickly over the next 1-2 years - there is unlikely to be a 3.11 release for PyPy, so next year when Python 3.10 starts being dropped that's the end of it. Conda-forge is also starting to drop PyPy builds because of maintenance issues and no real future. Hence deprecating because of PyPy isn't needed imho.

Does that mean that this is a "won't fix?"

rgommers:
> Does that mean that this is a "won't fix?"

I don't think so. This is a minor change that to me seems strictly an improvement over the status quo. So I'd be in favor of merging it as is.

The larger change @dnicolodi was thinking about ([this comment above](https://github.com/mesonbuild/meson-python/pull/626#issuecomment-2109795364)) is orthogonal, and may or may not happen. I'd suggest that it shouldn't be blocking for this PR.

PR #627, opened 2024-05-13
CI: test free threading Python 3.13

dnicolodi:
The manylinux docker images added support for Python 3.13 build with free threading support. Trade python3.11 for python3.13t in the manylinux jobs.


henryiii:
FYI, `python -m venv ...` still installs pip 24.0. Maybe what you are seeing?

dnicolodi:
> FYI, `python -m venv ...` still installs pip 24.0. Maybe what you are seeing?

Yes, I believe this is the problem. I'm working on a fix.

PR #631, opened 2024-05-17
MAINT: improve error message when meson isn't found

rgommers:
This is a follow-up to gh-495.

There are several ways that meson may go missing:
- a non-isolated build and meson isn't installed
- meson executable is specified in pyproject.toml and git submodule is containing it is not initialized (reported as resulting in a confusing error in https://github.com/numpy/numpy/issues/26397)
- the  `MESON` environmnent variable is used by the user and is misspelled

The git submodule case was ending with:
```
meson-python: error: Could not find meson version 0.63.3 or newer, found .
```
and after this ends with:
```
meson-python: error: Could not find specified meson: vendored-meson/meson/meson.py
```

If the executable is missing, the build ended with a very long traceback. Changing `FileNotFoundError` to `ConfigError` elides the traceback and clearly reports that the executable wasn't found. Easy to verify with any package with:
```
$ MESON=nonsense python -m build -wnx
* Building wheel...

meson-python: error: meson executable `nonsense` not found

```

Closes https://github.com/numpy/numpy/issues/26397

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -988,6 +988,8 @@ def _get_meson_command(
     # directly. For packages that vendor a forked Meson, the `meson.py` in the
     # root of the Meson repo can be used this way.
     if meson.endswith('.py'):
+        if not os.path.exists(meson):
+            raise ConfigError(f'Could not find the specified meson: "{meson}"')
         cmd = [sys.executable, meson]
     else:
         cmd = [meson]
@@ -997,7 +999,11 @@ def _get_meson_command(
     # but the corresponding meson command is not available in $PATH. Implement
     # a runtime check to verify that the build environment is setup correcly.
     required_version = _parse_version_string(version)
-    meson_version = subprocess.run(cmd + ['--version'], check=False, text=True, capture_output=True).stdout
+    try:
+        meson_version = subprocess.run(cmd + ['--version'], check=False, text=True, capture_output=True).stdout
+    except FileNotFoundError as err:
+        raise ConfigError(f'meson executable "{meson}" not found') from err
+
     if _parse_version_string(meson_version) < required_version:
         raise ConfigError(f'Could not find meson version {version} or newer, found {meson_version}.')
 


mesonpy/__init__.py
```diff
@@ -988,6 +988,8 @@ def _get_meson_command(
     # directly. For packages that vendor a forked Meson, the `meson.py` in the
     # root of the Meson repo can be used this way.
     if meson.endswith('.py'):
+        if not os.path.exists(meson):
+            raise ConfigError(f"Could not find specified meson: {meson}")
```
dnicolodi, line 992: `Could not find the specified meson: "{meson}"` maybe, adding the article and quotes around the path. I think having the quotes helps when there is white space in the filename.

mesonpy/__init__.py
```diff
@@ -997,7 +999,11 @@ def _get_meson_command(
     # but the corresponding meson command is not available in $PATH. Implement
     # a runtime check to verify that the build environment is setup correcly.
     required_version = _parse_version_string(version)
-    meson_version = subprocess.run(cmd + ['--version'], check=False, text=True, capture_output=True).stdout
+    try:
+        meson_version = subprocess.run(cmd + ['--version'], check=False, text=True, capture_output=True).stdout
+    except FileNotFoundError as err:
+        raise ConfigError(f'meson executable `{meson}` not found') from err
```
dnicolodi, line 1005: We generally use double quotes `"` around literals in the error messages.

dnicolodi (changes requested):
Thanks @rgommers. It looks good, except for some nitpicking about the spelling of the error messages. Curiously I have similar patch queued locally. In that I also check the return code of `meson --version` and raise an error with what was printed on the standard error if the return code is not zero.

rgommers:
Thanks @dnicolodi. Adopted your suggestions for error messages.

> In that I also check the return code of `meson --version` and raise an error with what was printed on the standard error if the return code is not zero.

If you have that already, do you want to put that on top of this PR (either push a commit or do in a follow-up PR)?

dnicolodi:
Sure, I will do that later today.

dnicolodi (approved):

PR #652, opened 2024-07-31
DOC: make the documentation reproducible when rebuilt

smcv:
* From: @lamby

    Make the documentation reproducible when rebuilt
    
    The Reproducible Builds project aims to arrange for rebuilding the same
    source code in a sufficiently similar environment to produce the same
    installable packages every time, as a way to discourage supply-chain
    attacks by making it possible to verify that a particular installable
    package was built from the claimed source code.
    
    Previously, if meson-python was built twice, at least a year apart,
    then its documentation would contain different copyright dates. The
    SOURCE_DATE_EPOCH environment variable is used here to avoid that
    difference: the intention is that environments that want to produce
    reproducible packages will set SOURCE_DATE_EPOCH to some suitable fixed
    date (perhaps the date of the most recent git commit) which is held
    constant across rebuilds.
    
    See the specification for SOURCE_DATE_EPOCH for more details:
    https://reproducible-builds.org/docs/source-date-epoch/
    
    [smcv: Added commit message]
    
    Co-authored-by: @smcv

---

The actual change here is a patch originally sent to <https://bugs.debian.org/1076806>.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -16,15 +16,21 @@
 
 import datetime
 import os
+import time
 import sys
 sys.path.insert(0, os.path.abspath('..'))
 import mesonpy
+
+build_date = datetime.datetime.fromtimestamp(
+    int(os.environ.get('SOURCE_DATE_EPOCH', time.time())),
+    tz=datetime.timezone.utc,
+)
 
 
 # -- Project information -----------------------------------------------------
 
 project = 'meson-python'
-copyright = f'2021\N{EN DASH}{datetime.date.today().year} The meson-python developers'
+copyright = f'2021\N{EN DASH}{build_date.year} The meson-python developers'
 
 # The short X.Y version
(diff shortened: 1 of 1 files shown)


lamby:
Thanks for forwarding this upstream, @smcv !

smcv:
I hope you're sufficiently happy with the long commit message that I added. I didn't want to leave it at just what you said in the original patch, because as an upstream developer (in other projects, not this one), if I received a similar patch without having the context of knowing about the reproducible builds goal within Debian, I would have wanted to see more details in the commit message before accepting it.

lamby:
Oh, very happy. In fact, I may use this as a template to improve my own PRs. Thanks :)

rgommers (approved):
LGTM, thank you @smcv and @lamby. Build reproducibility improvements are always very welcome!

PR #657, opened 2024-08-02
CI: use a token for the Codecov action to upload

rgommers:
This is required now, it was a breaking change in `codecov-action@v4`. Follows up on https://github.com/mesonbuild/meson-python/pull/479#issuecomment-2265134409.

The token didn't exist yet; I just added it as a repository secret.

PR #658, opened 2024-08-02
CI: point action to codecov.xml as generated by pytest-cov

rgommers:
Addresses the issue in https://github.com/mesonbuild/meson-python/pull/657#issuecomment-2265481244

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -138,8 +138,9 @@ jobs:
       - name: Upload coverage report
         uses: codecov/codecov-action@v4
         if: ${{ always() }}
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          files: ./coverage.xml
 
   msvc:
     runs-on: windows-latest


rgommers:
Okay, that isn't it, no change. This broke something in the last 4 weeks says the dashboard for PRs:

image

It worked 4 weeks ago and doesn't work today, nothing changed on our side. There is zero useful logging in CI (it claims success) or in the dashboard. `codecov-action` didn't change in that time, and neither did `pytest-cov`. One `coverage release (7.6.0) but no relevant open issues. I'll park this for now.

dnicolodi:
Does anyone know hot to get codecov unstuck?

@rgommers Should we switch to something else? coveralls.io seems to be another popular option.

rgommers:
Maybe it fixed itself in the meantime and there's nothing left to do? The percentages are no longer missing and https://app.codecov.io/gh/mesonbuild/meson-python/commits looks like:

image

It seems like the code browser is also up to date again (https://app.codecov.io/gh/mesonbuild/meson-python/blob/main/mesonpy%2F__init__.py). 

> Should we switch to something else? 

I rarely look at codecov, so I'd be fine with anything.

dnicolodi:
I just look at the reports that pop up along the CI tests results. I haven't seen these working in a very long time. Therefore I thought codecov was still broken for us. Do you know how to get these back?

rgommers:
Ah, in https://github.com/mesonbuild/meson-python/settings/installations I now see that the GitHub App is suspended, so that's probably why the notifications don't show up:

image

I don't know what happened though; it can only be re-enabled by an org owner. @jpakkane would you mind hitting the _Configure_ button on that link and then the _Update Access_ one on the next screen?

jpakkane:
We disabled codecov on the main repo because it was too spammy and not really helpful. Should be fixed now.

rgommers:
I think everything is fixed; in the most recent push for gh-681 I see:

image

I'll close this PR.

PR #676, opened 2024-10-06
BUG: allow install_subdir() of missing directories

dnicolodi:
Remove the assert that made sure the source path for install_subdir() is an existing directory. Meson warns in this case but allows it. Empty directories for missing source directories are not created in the Python wheel as these are hardly meaningful.


dnicolodi:
Replaces #675

rgommers (approved):
Seems reasonable to do nothing"instead of erroring out, given the discussion in gh-675. 

Meson does display a deprecation warning in the build log like:
```
WARNING: Deprecated features used:
 * 0.60.0: {'install_subdir with empty directory'}
 ```

Issue #270, opened 2023-01-23
Allow `license` to be listed in the dynamic section


tristan957:
Meson already has `project(license:)`. Would it be possible to autodetect it dynamically? For instace, I have `'Apache-2.0 OR MIT'` as the value in `project()`. I would like this to populate the `license = { text = "XXX" }` metadata like this project already does with `version`.

FFY00:
Thanks for the suggestion! However, I think the semantics are a little bit different, `XXX` in your example would be the actual license text, so "Apache-2.0 OR MIT" wouldn't be the correct value. There's no way to specify that currently, but hopefully there will be with [PEP 639](https://peps.python.org/pep-0639/), which adds a `License-Expression` core metadata field and assigns the value of `project.license` in pyproject.toml to the expression. So, this is another issue where we are blocked until that PEP, or an alternative, lands.

tristan957:
It's unfortunate this wasn't thought of earlier when the pyproject.toml was designed. Seems so fundamental, but hopefully the PEP is accepted.

For now, I'll just leave it as is since I see no better way to express dual licensed software.

FFY00:
My recommendation for that case would be to have a single license file with both licenses, and specifying it in `pyproject.tml`. But I agree, it's unfortunate, hopefully the PEP will land though and fix this.

tristan957:
Unfortunately that wouldn't be REUSE compliant from what I understand.

https://reuse.software/

FFY00:
I didn't know about REUSE, thank you for pointing it out. I've only skimmed it a bit, but I can't tell how it wouldn't be compliant? You can still have the SPDX headers, you'd just have an extra license file for the Python tooling to read. In fact, using the REUSE spec you should even be able to automatically generate this file! Am I missing something here? :rofl:

tristan957:
Hmm `touch LICENSE` in the root doesn't seem to cause the tool to fail. Oh well. Just throw REUSE on the pile of reasons that PEP should be advanced.

FFY00:
Agreed. I have pinged the discourse thread. Let's see if we can get it to move forward.

(cc @CAM-Gerlach)

tristan957:
Thanks for your efforts.

CAM-Gerlach:
[...]

PR #681, opened 2024-10-12
ENH: add support for PEP 639

dnicolodi:
Closes gh-270


mesonpy/__init__.py
```diff
@@ -53,6 +53,17 @@
 from mesonpy._compat import cached_property, read_binary
 
 
+try:
+    from packagng.licenses import canonicalize_license_expression
```
henryiii, line 57:
```suggestion
    from packaging.licenses import canonicalize_license_expression
```

docs/reference/meson-compatibility.rst
```diff
@@ -39,6 +39,14 @@ versions.
    Meson 1.3.0 or later is required for compiling extension modules
    targeting the Python limited API.
 
+.. option:: 1.6.0
+
+   Meson 1.6.0 or later is required to support ``license`` and
```
henryiii, line 44: Do you have a link to a PR or something for the 1.6 change? curious as to what needed updates in meson, was expecting it to be just meson-python.

henryiii (commented):
Looks pretty good from a quick review. I'll keep this in mind when I implement it (very soon) in scikit-build-core.

docs/reference/meson-compatibility.rst
eli-schwartz: Probably due to depending on https://github.com/mesonbuild/meson/pull/13783

mesonpy/__init__.py
dnicolodi: Ups. Thanks.

docs/reference/meson-compatibility.rst
dnicolodi: Indeed, as @eli-schwartz says. Meson support is required only to get the license and license files declared in `meson.build` when they are declared dynamic or when there is no `project` section in `pyproject.toml`.

dnicolodi:
> Looks pretty good from a quick review. I'll keep this in mind when I implement it (very soon) in scikit-build-core.

It is a bit uglier than what I would like because I wanted to keep compatibility with older pyproject-metadata. meson-python supports anything back to version 0.7.1.

QuLogic:
I gave this a shot in https://github.com/matplotlib/matplotlib/pull/28982, and `License-Expression` is correctly pulled from `meson.build`, as are `License-File` from the main project, but what I'd be missing there is `License-File` from the subprojects.

dnicolodi:
I'm not very well versed in licensing issues, but I don't think the way PEP 639 supports declaring package licenses works well for this use case. The main problem is that it expects the license of the package's sdist and wheel to be exactly the same. For packages like matplotlib that link statically with some libraries, this is not the case: the license of the source code is the matplotlib license, the license of the wheel is the union of the matplotlib license with the license of all the libraries linked to it.

Concretely, the issue is that in the most common case the subprojects directory does not contain the source code of the subprojects but only references to them (in the matplotlib case in the form of meson wrap files). However, the `License-File` fields need to point to actual files distributed in the sdist and in the wheel. The sdist in these cases does not contain the license files, thus the `License-File` fields for the subprojects cannot be added. The metadata information for the sdist and the wheel need to be the same, thus the fields cannot be added to the wheel metadata either.

I don't know how the problem could be solved. The most formally right solution would to allow source and binary distributions to have different licenses. But this is a quite significant change in behavior for the package metadata system.

The pragmatic solution is to add the subprojects' license files to the source distribution and accept that the sdist as a whole might have a more restrictive license than what it could have.

eli-schwartz:
> However, the `License-File` fields need to point to actual files distributed in the sdist and in the wheel. The sdist in these cases does not contain the license files, thus the `License-File` fields for the subprojects cannot be added. The metadata information for the sdist and the wheel need to be the same, thus the fields cannot be added to the wheel metadata either.

This is not true. Metadata can be dynamic and vary between the sdist and the wheel. It's already dynamic as a result of being calculated from meson introspection, so you just need to mark that in the sdist as well.

The decision to make sdist metadata static and forbidden to change is... definitely a decision of all time. But the good news is it comes with an escape hatch. This escape hatch is needed since the spec also says that it's illegal per the spec to apply open source patches to an sdist metadata if the fields are marked as static. :shrug:

rgommers:
> Metadata can be dynamic and vary between the sdist and the wheel.

I think this is technically true but also it's ignored by most tools (e.g., `pip`), who assume they can read metadata from PyPIs JSON API once per version of a package. So in practice different metadata between sdists and wheels doesn't really work.

henryiii:
METADATA 2.2 added the ability to specify if the wheels will have different metadata from the SDist. (Technically, it was implemented the other way around - earlier versions have no guarantees, while 2.2+ guarantees that anything not listed in a `Dynamic` header is fixed between SDist and wheel. Anything with this header might be different per wheel. This seems like a very valid use case for that feature.

eli-schwartz:
> I think this is technically true but also it's ignored by most tools (e.g., `pip`), who assume they can read metadata from PyPIs JSON API once per version of a package. So in practice different metadata between sdists and wheels doesn't really work.

Also in practice, pip doesn't care what the license of a package is via the JSON api as pip considers that to be informational-only. :)

But I was under the impression that the major reason for PEP 658 is precisely because pip and similar tools do indeed need to directly consider the metadata from a built wheel rather than making any assumptions about the JSON api.

dnicolodi:
> This is not true. Metadata can be dynamic and vary between the sdist and the wheel. 

I was convinced that `License-Expression` and `License-File` were excluded from being dynamic as `Name` and `Version` are. However, scanning through PEP 639, this does not seem to be the case.

> It's already dynamic as a result of being calculated from meson introspection, so you just need to mark that in the sdist as well.

It is not dynamic in the sense of PEP 658. It is calculated at sdist build time and not declared as `Dynamic` in the metadata.

Declaring these fields as dynamic is an interesting solution for the problem at hand, thanks @eli-schwartz for pointing it out. However, how to expose this to the users of meson-python is not really clear to me. I'm not sure that the mere presence of subprojects should be a strong enough indicator that `License-Files` should be treated as a PEP 658 `Dynamic` field. This would also raise the question of what to do with `License-Expression`. Should it also marked `Dynamic` and have it set to a `(root project license) AND (subproject 1 license) AND (subproject 2 license) AND ...` equivalent license expression?

dnicolodi:
> Though there's a statement or two about license-files in the PEP that might be incorrect for this sort of usage.

Which ones?

henryiii:
> If a License-File is listed in a [Source Distribution](https://packaging.python.org/en/latest/glossary/#term-Source-Distribution-or-sdist) or [Built Distribution](https://packaging.python.org/en/latest/glossary/#term-Built-Distribution)’s Core Metadata:
>
> * That file MUST be included in the [distribution archive](https://packaging.python.org/en/latest/glossary/#term-Distribution-Archive) at the specified path relative to the root license directory.

henryiii:
I plan to work on the dynamic-metadata project soon and that would provide a way to do this from a Python hook. Not sure if this would help with integration with Meson, though.

eli-schwartz:
> Declaring these fields as dynamic is an interesting solution for the problem at hand, thanks @eli-schwartz for pointing it out. However, how to expose this to the users of meson-python is not really clear to me. I'm not sure that the mere presence of subprojects should be a strong enough indicator that `License-Files` should be treated as a PEP 658 `Dynamic` field.

It's less an "interesting solution for the problem" and more just... the correct way to define it?

It is dynamic because the final value depends on which subprojects end up configured when building a wheel. You can (and should) always mark it as PEP 658 `dynamic` because meson-python reserves the right to recalculate it later on.

> This would also raise the question of what to do with `License-Expression`. Should it also marked `Dynamic` and have it set to a `(root project license) AND (subproject 1 license) AND (subproject 2 license) AND ...` equivalent license expression?

Yes? It's the exact same topic. A subproject can define both the license SPDX expression and the license files. If the wheel build has configured that subproject, it is part of the resulting metadata for the binary distribution.

dnicolodi:
I think that this reasoning is correct if the subprojects are linked statically. However, if they are linked dynamically, the binary distribution in itself is not bound by the license of the dynamically linked libraries. Is it? What about subprojects that are only used as build tools? Or any other use of subprojects that does not end up in static linking to the distributed binaries?

Similarly, what if subprojects are only fallbacks for dependencies that are not found locally? If the binary distribution links statically to a locally installed library, it is still bound by the terms of the license of said library, but the subproject would not be build.

> Yes? It's the exact same topic. A subproject can define both the license SPDX expression and the license files. If the wheel build has configured that subproject, it is part of the resulting metadata for the binary distribution.

The question was more whether concatenating the licenses with a logic `AND` is always the appropriate thing to do. I think it is, but IANAL.

eli-schwartz:
> I think that this reasoning is correct if the subprojects are linked statically. However, if they are linked dynamically, the binary distribution in itself is not bound by the license of the dynamically linked libraries. Is it? What about subprojects that are only used as build tools? Or any other use of subprojects that does not end up in static linking to the distributed binaries?

If you distribute a dynamically linked library that has been auditwheel'd into your wheel, you most certainly are bound by its license.

dynamically linked libraries have different license terms than statically linked ones, some of the time, such as for example being able to use the LGPL instead of the GPL. Still bound by the license.

> What about subprojects that are only used as build tools?

I'm not sure what to say there besides for "difficult to detect... :("

> Similarly, what if subprojects are only fallbacks for dependencies that are not found locally? If the binary distribution links statically to a locally installed library, it is still bound by the terms of the license of said library, but the subproject would not be build.

Then meson can't help you report this, but it is still true that those are the terms you must follow. However typical wheel building processes tend to happen on systems where dependencies, if available, are dynamically available -- which means either you distribute it yourself via e.g. auditwheel or require the system to provide one in which case it's technically not your responsibility to list the terms.



> The question was more whether concatenating the licenses with a logic `AND` is always the appropriate thing to do. I think it is, but IANAL.

"AND", because you have to fulfill both terms. The lawyers are the people you ask to find out what "fulfill both sets of requirements" means. :)

dnicolodi:
This is why I think that the mere presence of subprojects is not a very good indicator for how the license of the resulting wheel needs to be "composed".

dnicolodi:
> IMO automatic determination based on subproject usage is the _only_ reason to care about PEP639's dynamism

What do you mean? PEP 639 does not define any dynamism. It only puts some order on how license information is recorded in Python package metadata.

What I am trying to say is that basing the definition of the license of the package on the presence of configured subprojects has the potential to produce incorrect results. I believe that the implementing a feature that has the potential of producing incorrect results in some cases and save typing a few hundred characters in the best case is good tradeoff. Therefore, I will not work on it.

However, if someone comes up with a solution that implements dynamic license determination based on the configured subprojects, that can optionally be enabled, and that does not break any of the features already implemented, I will be happy to review and eventually merge it.

henryiii:
Clear for pyproject-metadata release?

dnicolodi:
Yes, green light 🙂

rgommers (commented):
Thanks for this, great to see PEP 639 support coming along!

I tried this out on PyWavelets, and all looks good there: https://github.com/PyWavelets/pywt/pull/772

One change I'd suggest in this PR is to unconditionally require `pyproject-metadata>=0.9.0`, since we don't want projects to have to manage a transitive dependency like that when starting to use PEP 639 metadata.

I'll have a look at NumPy and SciPy as well, but from the PEP I can already tell that it's going to be a little painful, since apparently anything that uses `auditwheel` and co is "relatively niche" (sigh): https://peps.python.org/pep-0639/appendix-rejected-ideas/#different-licenses-for-source-and-binary-distributions. I think the statement there that per-file licenses don't have a mechanism is no longer true I think, that predates [PEP 658](https://peps.python.org/pep-0658/). So hopefully after this gets rolled out, there'll be an update to enable per-file metadata.

henryiii:
Can't this already be done using dynamic-metadata? The point of dynamic-metadata is to allow the wheels to have different metadata than the source distribution. And the new entries are valid listed in dynamic-metadata. It might not be shown effectively by PyPI, etc, but it should at least work. And METADATA 2.4 > 2.2, so it's always available to use.

henryiii:
Are you just making sure the license is already normalized? Without a packaging release, we don't have license-expression normalization yet. The PR is merged but there hasn't been a release yet.

rgommers:
> Can't this already be done using dynamic-metadata? The point of dynamic-metadata is to allow the wheels to have different metadata than the source distribution.

Perhaps yes, although it's quite hard to come up with a reasonable way of using that, since meson-python cannot know whether or not `auditwheel` will be used after `meson-python` itself is done. More importantly, declaring metadata is dynamic in `pyproject.toml` is kinda awful, because (a) it removes useful data that static analysis tools would otherwise be able to read, and (b) it makes `pip` & co unable to use any static metadata from `pyproject.toml` directly (in practice today, not in principle), thereby potentially having to do much more work in some cases.

The better way would be to actually make it a first class use case with something like:
```toml
[project]
license = "MIT"

[pypi-wheels]
# platform specifiers need some more thought - maybe reuse the cibuildwheel syntax?
license = {
    win32 = "MIT AND Other",
    linux = "MIT AND LGPL-2.1-and-later",
}
# Add `license-files` in a similar way
```
This could perhaps be done in a tool-specific way first, with `[tool.meson-python.pypi-wheels]`. And if we'd come up with something that works for both `meson-python` and `scikit-build-core`, that would probably be a good basis for future standardization?

What we do right now in NumPy is quite a bit more hacky:
https://github.com/numpy/numpy/blob/88a6a0070d9d50f034c3f31f5e1d9ceea5f1065e/tools/wheels/cibw_before_build.sh#L9-L20

eli-schwartz:
> More importantly, declaring metadata is dynamic in `pyproject.toml` is kinda awful, because (a) it removes useful data that static analysis tools would otherwise be able to read, and (b) it makes `pip` & co unable to use any static metadata from `pyproject.toml` directly (in practice today, not in principle), thereby potentially having to do much more work in some cases.

Both of these reasons are invalid.

- static analysis tools per definition cannot know whether you are running auditwheel, what would they "analyze" about this anyway? Furthermore, static analysis tools that care about *licenses* do not depend on clumsy functionality such as "pypa metadata", which in nearly all cases isn't useful because it isn't even a python software being analyzed. e.g. https://www.fossology.org/ and friends. There are anyways existing standards that unlike "pypa metadata" actually allow you to declare your licenses in a machine-readable fashion, including breaking down licenses file by file and recording the licenses of vendored software, git submodules, and much much more
- `pip` & co already have to deal with being unable to use dynamic metadata from pyproject.toml directly -- static metadata works quite fine -- and this doesn't matter in practice today OR in principle anytime given that most metadata fields are simply ignored by pip altogether due to not participating in install-time dependency resolution


Also none of this matters if you expect people to use wheels

Also also none of this matters if you expect people to use PyPI-based sdists, since you can emplace the license metadata into the sdist and mark it as "dynamic" at the same time -- it is trivially parseable and simply isn't accurate for anything beyond the sdist itself due to the fact that the sdist can be configured and built and download new (checksummed) dependencies over the internet and change its effective license in, depending on the project, maybe even dozens of different configurations that are only machine-knowable by asking the resulting configured build to announce itself via introspection.

rgommers:
> Also none of this matters if you expect people to use wheels [...] Also also none of this matters if you expect people to use PyPI-based sdists,

Not quite, my problem is the opposite. I want metadata in `pyproject.toml` to be generically correct to the extent possible (PyPI is a special case with weird rules/behavior). The problem with declaring anything as dynamic is that you have to delete the static info completely. Removing `license = ...' under `[project]` completely because you have to do `dynamic = ['license']` just makes the metadata much less discoverable. 

Maybe I misunderstood what @henryiii meant by "dynamic-metadata". If it's not `dynamic = ['license']` but only marked dynamic in `PKG-INFO` then I retract my point. There's no way I think to actually use that though without talking about UX in `pyproject.toml`.

eli-schwartz:
If you just want people to use pyproject.toml and nothing other than pyproject.toml then what is the purpose of PKG-INFO in a sdist?

That file contains possibly still-dynamic versions of the metadata that was discovered during sdist building -- it is machine readable and there's no real reason not to use it as a statically deducible alternative to `dynamic = ` values for scenarios where you don't want to run the project metadata generation hook.

rgommers:
> If you just want people to use pyproject.toml and nothing other than pyproject.toml then what is the purpose of PKG-INFO in a sdist?

`PKG-INFO` is a _generated_ file, so that's not something that package authors can use - there must be some UX either inside `pyproject.toml` or in some other way (e.g., a CLI flag to the build backend) to be able to put anything inside `PKG-INFO`.

`PKG-INFO` also isn't version-controlled, and I'm interested in tools that analyze a repo rather than an sdist.

eli-schwartz:
> `PKG-INFO` also isn't version-controlled, and I'm interested in tools that analyze a repo rather than an sdist.

Ok so maybe what you want in order to analyze a repo for license information is actually https://www.fossology.org/ ?

eli-schwartz:
Anyway I think people should have the right to use dynamic metadata if they want to, and use pyproject.toml metadata if they want to.

With that in mind, the ability to have different metadata for sdist vs. wheel by making use of "Dynamic" entries in current versions of the Core Metadata, is a powerful escape hatch for those who choose to do so.

rgommers:
> Anyway I think people should have the right to use dynamic metadata if they want to, and use pyproject.toml metadata if they want to.

+1 completely agreed. My only point is that _as a Python package author_ I don't wanted to be _forced_ to use dynamic metadata via `dynamic =` under `[project]` when I statically know what the licenses of the project, the sdist and the PyPI wheels I publish are. 

> With that in mind, the ability to have different metadata for sdist vs. wheel by making use of "Dynamic" entries in current versions of the Core Metadata, is a powerful escape hatch for those who choose to do so.

For build backend authors/features, sure. Not directly relevant to package authors though - Core Metadata is a thing between packaging tools and index servers, not something one can directly access or really ever thinks about at all as a package author. Hence my proposal for a UX that package authors can use to put in their statically known sdist and PyPI wheels metadata, which the build backend can then translate to a Dynamic entry in Core Metadata.

henryiii:
FYI, having something listed in the project table and also dynamic allowing the metadata to be added to is still something I want to write a PEP for.

dnicolodi:
> One change I'd suggest in this PR is to unconditionally require `pyproject-metadata>=0.9.0`, since we don't want projects to have to manage a transitive dependency like that when starting to use PEP 639 metadata.

I added a commit on top that bumps the required pyproject-metadata to 0.9.0.

@rgommers Is there anything else to be done before we can merge this?

rgommers:
> @rgommers Is there anything else to be done before we can merge this?

It's looking pretty good, but let me try to do a slightly more careful line-by-line review - so far I just quickly browsed the diff and tested with PyWavelets.

> I have no idea of what the timeline to get support for PEP 639 in PyPI is and I don't know if waiting to land this till that is available is a good idea or not.

That's a good point. https://github.com/pypi/warehouse/pull/16949 is moving fairly quickly it looks like, but it may be wise to ensure that this PR lines up with what lands in PyPI somehow before merging it. Not 100% sure about that, but without that it seems to me that there's a risk of some users trying out the new feature and ending up with wheels with license-related issues or that cannot be uploaded.

rgommers:
Something I noticed that is correct but mildly surprised me is that the `.dist-info` directory gets quite nested. E.g., for NumPy:
```
% tree -a numpy-2.2.0.dev0.dist-info/licenses/
numpy-2.2.0.dev0.dist-info/licenses/
├── .spin
│   └── LICENSE
├── LICENSE.txt
├── numpy
│   ├── _build_utils
│   │   └── tempita
│   │       └── LICENSE.txt
│   ├── _core
│   │   └── include
│   │       └── numpy
│   │           └── libdivide
│   │               └── LICENSE.txt
│   └── linalg
│       └── lapack_lite
│           └── LICENSE.txt
└── vendored-meson
    └── meson
        └── COPYING
```
Not a problem, and matches the PEP (as usual the PEP is short of examples though, so one needs to read carefully), so just noting here as "indeed looks as it should".

I noticed that `canonicalize_license_expression` relies on the not-yet-released `packaging` 24.2. I tested locally that it works, but that's another thing that could potentially change still. And rolling out PyPI support also depends on `packaging` 24.2. So probably best to merge this post `packaging` and PyPI feature availability.

rgommers (approved):
I now went over this more carefully, and it all LGTM.

The one thing that came to mind was that including Apache/LGPL license files is probably going to trigger some license scanner tooling and give some distro maintainers work to silence the false positives, but not much that can be done about that.

Ready to merge I think - but I'd hold off until `packaging` and PyPI have support.

rgommers:
> There is no technical reason why these files need to be named like that or include the actual licenses text. I just copied them over from the source code of a package I had at hand. They can be replaced with placeholders.

Perhaps worth doing? Using `MIT AND BSD-3-Clause` might save some minor trouble here. Only if you feel like it though, I'm fine either way.

dnicolodi:
> Also, noticed there's no quick and easy link to the source from readthedocs

I'm not sure about what you mean with this. meson-python documentation is not on readthedocs and we have a "Source Code" link in the left sidebar (and typing this made me realize that the capitalization of the title of the link is inconsistent with the other titles).

dnicolodi:
> packaging 24.2 is out!

Thanks. I'll add tests making sure the validation of the license fields works as expected.

dnicolodi:
> Anyway I think people should have the right to use dynamic metadata if they want to, and use pyproject.toml metadata if they want to.
> 
> With that in mind, the ability to have different metadata for sdist vs. wheel by making use of "Dynamic" entries in current versions of the Core Metadata, is a powerful escape hatch for those who choose to do so.

Dynamic metadata is seen as such a nice feature that there is a discussion about eliminating it:
https://discuss.python.org/t/brainstorming-elliminating-dynamic-metadata/71405/1

I don't have the time to read these long threads on Python discourse (even more so because discourse is horrible) let alone contribute any opinions. But I thought that some here may be interested...

eli-schwartz:
> Dynamic metadata is seen as such a nice feature that there is a discussion about eliminating it: https://discuss.python.org/t/brainstorming-elliminating-dynamic-metadata/71405/1
> 
> I don't have the time to read these long threads on Python discourse (even more so because discourse is horrible) let alone contribute any opinions. But I thought that some here may be interested...

Agreed on both counts about reading the Python discourse.

But from a quick look at the initial post before the site became impossible to read, it appears to make a few basic arguments about why dynamic metadata is "bad".

- users of editable installs have inaccurate versions from `git describe`
- poetry and uv rerun metadata generation constantly and would rather not have to execute a build backend hook to find out things like the version or dependencies, as it is "slow"

I find both arguments weak enough that I suspect the dislike of dynamic metadata came first and the supporting arguments were created to fit the narrative. None of this matters for sdists or wheels because both contain static metadata explicitly intended to be a cache, and editable installs are an extremely specific use case that users have already chosen to make some sacrifices in order to use -- plainly they are not worried about the version number or if they care they are using `myprog --version` which isn't actually covered by pypa standards about dynamic vs static metadata.

Lockfiles can just avoid self-inflicted wounds by generating them when asked, but it doesn't actually matter either way since lockfiles only have to do ANY of this on bare source trees, not dependencies obtained via sdist/ wheel, which means that users who use poetry / uv can choose to avoid dynamic metadata and users who prefer dynamic metadata can generate lockfiles less frequently.

This is yet another example of people who think the point of a standard is to hard-line forbid people from doing perfectly reasonable things that are optional, because the people doing the forbidding don't *personally* value that thing, and cannot abide the idea that anyone else would value it.

This contrasts with the fact that the standard currently allows all those perfectly reasonable things precisely because they are perfectly reasonable, and forbidding them would upset the people who are perfectly justified in doing perfectly reasonable things.


The whole argument is undoubtedly just perl code. They'll get tired of it after a while and nothing will change, not even opinions.

dnicolodi:
> Ready to merge I think - but I'd hold off until `packaging` and PyPI have support.

`packaging` releases a version with the license validation and normalization. PyPI is ironing out the kinks in their support for PEP 639. I've tried to upload to PyPI a package using PRP 639 fields, but neither `twine` nor `uv publish` support Metadata 2.4 at the moment. I don't know of anyone working on adding support to `twine`. I wonder what is the recommended way to upload to PyPI nowadays...

dnicolodi:
I've looked into what it would take to add support for metadata 2.4 to `twine`, and the answer is worst than what I expected: it would take a miracle to have it any time soon. `twine` uses `pkginfo` to parse the package metadata. So adding support for metadata 2.4 requires a new release of `pkginfo`. However, `twine` is not compatible with the latest `pkginfo` release. There is a pull request to make `twine` compatible with the latest `pkginfo` but it is open since June and it stalled.

I don't know why `twine` uses `pkginfo` and not `packaging` to parse the metadata. `warehouse` uses `packaging` thus I would have expected `twine` to use it too. `pkginfo` looks to be a one man show maintained on Launchpad in a Bazar repository https://code.launchpad.net/pkginfo

agriyakhetarpal:
In the meantime, it is possible to install `pkginfo>=1.11` after you isntall `twine` (i.e., by breaking the dependency tree) and `twine check --strict` should then work with wheels with the metadata version as 2.4. PyPI now accepts metadata-2.4 wheels, too – [`zig-pypi`](https://github.com/ziglang/zig-pypi) (https://pypi.org/project/ziglang/0.13.0.post1/#files) is an example.

agriyakhetarpal:
In the meantime, it is possible to install `pkginfo>=1.11` after you isntall `twine` (i.e., by breaking the dependency tree) and `twine check --strict` should then work with wheels with the Metadata-Version set as 2.4. PyPI now accepts those wheels, too, if uploaded manually – [`zig-pypi`](https://github.com/ziglang/zig-pypi) (https://pypi.org/project/ziglang/0.13.0.post1/#files) is an example.

agriyakhetarpal:
In the meantime, it is possible to install `pkginfo>=1.11` after you install `twine` (i.e., by breaking the dependency tree) and `twine check --strict` should then work with wheels with the Metadata-Version set as 2.4. PyPI now accepts those wheels, too, if uploaded manually – [`zig-pypi`](https://github.com/ziglang/zig-pypi) (https://pypi.org/project/ziglang/0.13.0.post1/#files) is an example.

dnicolodi:
@agriyakhetarpal The dependecy problem is not a problem in itself. It is only a problem because implementing metadatya 2.4 support in twine requires adding support to pkginfo and make twine consume the added metadata fields. If there is no maintainer interest in supporting a version of pkginfo release a long while back, I don't see how there could be interest in merging patches to work with a future release. It is my understanding that, without support in the upload tool, the license information recorded in the `License-Expression` metadata field is not picked up by PyPI. For example, the ziglang release you link to does not show any license information on the PyPI page. Getting accurate information shown on PyPI is the purpose of this whole exercise, thus I don't think that what you show is a successful example.

dnicolodi:
For what is worth, `uv publish` can upload metadata 2.4 distributions, but the lack of explicit metadata 2.4 support means that the license is not recorded in the PyPI index and not shown in the package page on PyPI.

agriyakhetarpal:
> For what is worth, `uv publish` can upload metadata 2.4 distributions, but the lack of explicit metadata 2.4 support means that the license is not recorded in the PyPI index and not shown in the package page on PyPI.

I understand the state of things better now, thanks for the context!

frenzymadness:
Do you have a plan to move this forward? I see the PR for twine has already been merged.

rgommers:
Very close though, which is great to see!

frenzymadness:
Twine 6.1.0 is out for more than a month now. Would it make sense to release meson-python as well?

henryiii:
Scikit-build-core 0.11.0 includes PEP 639 support. Hatchling uses METADATA 2.4 now. So FWIW, I think it would be a good idea. ;)

(PS: there's a bug if your LICENSE file contains form feeds and you don't use PEP 639 and you are using Python older than 3.12.8, will be releasing a new pyproject-metadata with a fix soon. But I don't think it affects this at all.)

dnicolodi:
There is a lengthy discussion on the Python Discourse regarding issues similar to the one discussed here regarding how to express the license of distribution that embed multiple projects either via vendoring, subprojects, or embedding shared libraries into wheels: https://discuss.python.org/t/pep-639-round-3-improving-license-clarity-with-better-package-metadata/53020/196

I haven't read all the posts (usual caveats about the structure of exchanges on Discourse and the dynamics of the Python Discourse apply) but the discussion does not seem to be going anywhere. However, who has interest in the topic (@rgommers for NumPy and SciPy, @QuLogic for Matplotlib, maybe @henryiii for his work on scikit-core) may want to have a look.

henryiii:
~~Is license-files supported?~~ I'm trying out cookie and meson-python is failing because the LICENSE file, specified as `license-files = ["LICENSE"]`, is not in the SDist anywhere.

Ah, this isn't in a release, got it.

On the topic above, I think the partially dynamic metadata PEP will help, and there's further ideas now on the thread above.

rgommers:
> However, who has interest in the topic (@rgommers for NumPy and SciPy, @QuLogic for Matplotlib, maybe @henryiii for his work on scikit-core) may want to have a look.

Oh yes, thanks for pointing that out. I somehow didn't get notifications. This is a problem for complex projects. I have only read the first ~10 posts, but we've been dealing with this problem in NumPy and SciPy for a long time in hacky ways like dynamically amending license files when `cibuildwheel` runs. PEP 639 is an improvement, but doesn't completely solve this issue due to ambiguities between "project license" and "project + vendored stuff in various build scenarios".

It shouldn't require a change in `meson-python` or block us from releasing support.

dnicolodi:
> I have only read the first ~10 posts

I stopped reading the thread after the first 100 posts or so went in circles trying to define whether the intent of the PEP was for the metadata to capture the source code license or the distribution artifact license...

> It shouldn't require a change in `meson-python` or block us from releasing support.

Indeed. Nothing to do yet on our side.

PR #687, opened 2024-10-19
Prepare release 0.17.0

dnicolodi:


rgommers (approved):
Thanks @dnicolodi! The content added LGTM. I double checked the PRs that went in for 0.17.0 and have one question: did you leave doc PRs gh-616, gh-652 out of the release notes on purpose?

I should have time to do the release tomorrow, unless you want to do it.

dnicolodi:
Yes. We haven't been including documentation updates in the changelog so far and I didn't think #616 is any more important than past updates. I was unsure about #652. I decided that, because the built documentation is not distributed with the package, the issue fixed is not really relevant for users, thus listing it in the changelog is not necessary.

I like to keep the changelog as short as possible because it makes it easier for users to see what user facing changes have actually been implemented. We can add entries for this changes, if you think it they should be mentioned.

rgommers:
That rationale for excluding the doc PRs sounds good to me - I just wanted to check. In it goes!

PR #690, opened 2024-10-21
MAINT: set version to 0.18.0.dev0

rgommers:
The 0.17.0 release is up.

In the `0.17.0` tag we did unfortunately forget to bump the version string in the docs, so the deployed docs say `0.17.0.dev0`. Not a major issue, but a slight annoyance - needs a manual docs deployment to fix.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: MIT
 
-project('meson-python', version: '0.17.0')
+project('meson-python', version: '0.18.0.dev0')
 
 py = import('python').find_installation()
 
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -65,7 +65,7 @@ if typing.TYPE_CHECKING:  # pragma: no cover
     MesonArgs = Mapping[MesonArgsKeys, List[str]]
 
 
-__version__ = '0.17.0.dev0'
+__version__ = '0.18.0.dev0'
 
 
 _NINJA_REQUIRED_VERSION = '1.8.2'


rgommers:
Ugh no, remembered that I need to make a branch in this repo, fix the version number to exactly 0.17.0 there, and then manually build. Not ideal.

Either way I was thinking about always deploying the latest docs, since most doc updates are either narrative docs or bug fixes, and new features specific to a version are more rare. And the latter would benefit from using `.. versionadded::` anyway. @dnicolodi WDYT?

dnicolodi:
We can definitely deploy the docs at every push to main. The reason for not doing it so far is that the Sphinx configuration takes the version number from `mesonpy.__version__` and if we push the documentation from the main branch, that is always a `.devX` version of the next release, which would be confusing. We can set the version number explicitly in the Sphinx configuration, but then we need to update it in three places, instead of the two places where we need to update it now, which is already annoying. I would keep things as they are, regarding the documentation deployment.

More than the documentation, which we can fix, in the published release, `mesonpy.__version__` and the metadata version do not agree. I don't know how bad this is, given that no one should be using `mesonpy` as a library... But if it does not matter, maybe we can remove `mesonpy.__version__`, and move the version number from `meson.build` to `pyproject.toml`, have the Sphinx documentation pick it up from there, and be left with only one version number to update.

rgommers:
> But if it does not matter, maybe we can remove `mesonpy.__version__`, and move the version number from `meson.build` to `pyproject.toml`, have the Sphinx documentation pick it up from there, and be left with only one version number to update.

That sounds like an improvement, yes - we don't need `mesonpy.__version__` I think. 



> if we push the documentation from the main branch, that is always a `.devX` version of the next release, which would be confusing.

We could just strip the `.devX`? Having new docs unavailable is pretty annoying - usually they're written to address some issue or some gap, and it shouldn't take a new release to improve the rendered docs.

Perhaps even better: we can drop the version number in the sidebar completely. Since we no longer have versioned docs, why display a version number at all? Meson itself also doesn't do that, and it looks better.

dnicolodi:
I think the version number have been kept so far mostly to indicate that the documentation on readthedocs is outdated. Have we found a way to get rid of the documentation on readthedocs? If redirecting to the proper location is too difficult, deleting it would be the next best thing.

rgommers:
I'll work on that redirect.

PR #692, opened 2024-10-21
CI: version updates

dnicolodi:


rgommers (approved):
That all seems to work as advertised, thanks!

Issue #539, opened 2023-11-18
Document how shared libraries distributed with the wheel are handled


dnicolodi:
- Explain why bundle internal dependencies
- Explain how we bundle internal dependencies
- Explain how we patch the ``RPATH``
- Mention our limitation regarding executables with internal dependencies
- Link to the relevant *Limitations* section
- Explain how to work around it (using a static library)
- Mention how the ``pyproject.toml`` settings can be used to only enable the workaround on meson-python builds (by adding a custom Meson option and enabling it on meson-python builds)

PR #700, opened 2024-10-27
DOC: add documentation about using shared libraries

rgommers:
~_Opening this PR now to get feedback on topics and structure, not ready for detailed review yet._~

We get a lot of questions about shared libraries, and this is a tricky thing to get right. So try to document how to use internal libraries as well as link to external shared libraries as well as possible. Subproject-related questions also come up more and more, and there are some extra gotchas here, so treat those as a third "source" of shared (or static) libraries.

- Closes gh-539
- xref gh-525 and gh-551 for internal shared libraries on Windows
- xref gh-410 and gh-556 for optionally bundled a shared library
- xref gh-629 for shared library in subproject use cases

### DIFF
--- a/docs/how-to-guides/shared-libraries.rst
+++ b/docs/how-to-guides/shared-libraries.rst
@@ -1,0 +1,251 @@
+.. SPDX-FileCopyrightText: 2024 The meson-python developers
+..
+.. SPDX-License-Identifier: MIT
+
+.. _shared-libraries:
+
+**********************
+Using shared libraries
+**********************
+
+Python projects may build shared libraries as part of their project, or link
+with shared libraries from a dependency. This tends to be a common source of
+issues, hence this page aims to explain how to include shared libraries in
+wheels, any limitations and gotchas, and how support is implemented in
+``meson-python`` under the hood.
+
+We distinguish between *internal* shared libraries that are built as part of
+the project, and *external* shared libraries that are provided by project
+dependencies and that are linked with the project build artifacts.
+For internal shared libraries, we also distinguish whether the shared library
+is being installed to its default system location (typically
+``/usr/local/lib`` on Unix-like systems, and ``C:\\lib`` on Windows - we call
+this ``libdir`` in this guide) or to a location in ``site-packages`` within the
+Python package install tree. All these scenarios are (or will be) supported,
+with some caveats:
+
++-----------------------+------------------+---------+-------+-------+
+| shared library source | install location | Windows | macOS | Linux |
++=======================+==================+=========+=======+=======+
+| internal              | libdir           | no (1)  | ✓     | ✓     |
++-----------------------+------------------+---------+-------+-------+
+| internal              | site-packages    | ✓       | ✓     | ✓     |
++-----------------------+------------------+---------+-------+-------+
+| external              | n/a              | ✓ (2)   | ✓     | ✓     |
++-----------------------+------------------+---------+-------+-------+
+
+.. TODO: add subproject as a source
+
+1. Internal shared libraries on Windows cannot be automatically handled
+   correctly, and currently ``meson-python`` therefore raises an error for them.
+   `PR meson-python#551 <https://github.com/mesonbuild/meson-python/pull/551>`__
+   may improve that situation in the near future.
+
+2. External shared libraries require ``delvewheel`` usage on Windows (or some
+   equivalent way, like amending the DLL search path to include the directory
+   in which the external shared library is located). Due to the lack of
+   `RPATH <https://en.wikipedia.org/wiki/Rpath>`__ support on Windows, there is
+   no good way around this.
+
+.. _internal-shared-libraries:
+
+Internal shared libraries
+=========================
+
+A shared library produced by ``library()`` or ``shared_library()`` built like this
+
+.. code-block:: meson
+
+    example_lib = shared_library(
+        'example',
+        'examplelib.c',
+        install: true,
+    )
+
+is installed to ``libdir`` by default. If the only reason the shared library exists
+is to be used inside the Python package being built, then it is best to modify
+the install location to be within the Python package itself:
+
+.. code-block:: python
+
+   install_path: py.get_install_dir() / 'mypkg/subdir'
+
+Then an extension module in the same install directory can link against the
+shared library in a portable manner by using ``install_rpath``:
+
+.. code-block:: meson
+
+    py3.extension_module('_extmodule',
+        '_extmodule.c',
+        link_with: example_lib,
+        install: true,
+        subdir: 'mypkg/subdir',
+        install_rpath: '$ORIGIN'
+    )
+
+The above method will work as advertised on macOS and Linux; ``meson-python`` does
+nothing special for this case. Windows needs some special handling though, due to
+the lack of RPATH support:
+
+.. literalinclude:: ../../tests/packages/sharedlib-in-package/mypkg/__init__.py
+   :start-after: start-literalinclude
+   :end-before: end-literalinclude
+
+If an internal shared library is not only used as part of a Python package, but
+for example also as a regular shared library in a C/C++ project or as a
+standalone library, then the method shown above won't work. The library is
+then marked for installation into the system default ``libdir`` location.
+Actually installing into ``libdir`` isn't possible with wheels, hence
+``meson-python`` will instead do the following *on platforms other than
+Windows*:
+
+1. Install the shared library to ``<project-name>.mesonpy.libs`` (i.e., a
+   top-level directory in the wheel, which on install will end up in
+   ``site-packages``).
+2. Rewrite RPATH entries for install targets that depend on the shared library
+   to point to that new install location instead.
+
+This will make the shared library work automatically, with no other action needed
+from the package author. *However*, currently an error is raised for this situation
+on Windows. This is documented also in :ref:`reference-limitations`.
+
+
+External shared libraries
+=========================
+
+External shared libraries are installed somewhere on the build machine, and
+usually detected by a ``dependency()`` or ``compiler.find_library()`` call in a
+``meson.build`` file. When a Python extension module or executable uses the
+dependency, the shared library will be linked against at build time.
+
+If the shared library is located in a directory on the loader search path,
+the wheel created by ``meson-python`` will work locally when installed.
+If it's in a non-standard location however, the shared library will go
+missing at runtime. The Python extension module linked against it needs an
+RPATH entry - and Meson will not automatically manage RPATH entries for you.
+Hence you'll need to add the needed RPATH yourself, for example by adding
+``-Wl,rpath=/path/to/dir/sharedlib/is/in`` to ``LDFLAGS`` before starting
+the build. In case you run into this problem after a wheel is built and
+installed, adding that same path to the ``LD_LIBRARY_PATH`` environment variable is a quick way of
+checking if that is indeed the problem.
+
+On Windows, the solution is similar - the shared library can either be
+preloaded, or the directory that the library is located in added to the DLL
+search path with ``os.add_dll_directory``, or vendored into the wheel with
+``delvewheel`` in order to make the built Python package usable locally.
+
+Publishing wheels which depend on external shared libraries
+-----------------------------------------------------------
+
+On all platforms, wheels which depend on external shared libraries usually need
+post-processing to make them usable on machines other than the one on which
+they were built. This is because the RPATH entry for an external shared library
+contains a path specific to the build machine. This post-processing is done by
+tools like ``auditwheel`` (Linux), ``delvewheel`` (Windows), ``delocate``
+(macOS) or ``repair-wheel`` (any platform, wraps the other tools).
+
+Running any of those tools on a wheel produced by ``meson-python`` will vendor
+the external shared library into the wheel and rewrite the RPATH entries (it
+may also do some other things, like symbol mangling).
+
+On Windows, the package author may also have to add the preloading like shown
+above with ``_append_to_sharedlib_load_path()`` to the main ``__init__.py`` of
+the package, ``delvewheel`` may or may not take care of this (please check its
+documentation if your shared library goes missing at runtime).
+
+Note that we don't cover using shared libraries contained in another wheel
+and depending on such a wheel at runtime in this guide. This is inherently
+complex and not recommended (you need to be in control of both packages, or
+upgrades may be impossible/breaking).
+
+
+Using libraries from a Meson subproject
+=======================================
+
+It can often be useful to build a shared library in a
+`Meson subproject <https://mesonbuild.com/Subprojects.html>`__, for example as
+a fallback in case an external dependency isn't detected. There are two main
+strategies for folding a library built in a subproject into a wheel built with
+``meson-python``:
+
+1. Build the library as a static library instead of a shared library, and
+   link it into a Python extension module that needs it.
+2. Build the library as a shared library, and either change its install path
+   to be within the Python package's tree, or rely on ``meson-python`` to fold
+   it into the wheel when it'd otherwise be installed to ``libdir``.
+
+Option (1) tends to be easier, so unless the library of interest cannot be
+built as a static library or it would inflate the wheel size too much because
+it's needed by multiple Python extension modules, we recommend trying option
+(1) first.
+
+A typical C or C++ project providing a library to link against tends to provide
+(a) one or more ``library()`` targets, which can be built as shared, static, or both,
+and (b) headers, pkg-config files, tests and perhaps other development targets
+that are needed to use the ``library()`` target(s). One of the challenges to use
+such projects as a subproject is that the headers and other installable targets
+are targeting system locations (e.g., ``<prefix>/include/``) which isn't supported
+by wheels and hence ``meson-python`` errors out when it encounters such an install
+target. This is perhaps the main issue one encounters with subproject usage,
+and the following two sections discuss how options (1) and (2) can work around
+that.
+
+Static library from subproject
+------------------------------
+
+The major advantage of building a library target as static and folding it directly
+into an extension module is that no targets from the subproject need to be installed.
+To configure the subproject for this use case, add the following to the
+``pyproject.toml`` file of your package:
+
+.. code-block:: toml
+
+    [tool.meson-python.args]
+    setup = ['--default-library=static']
+    install = ['--skip-subprojects']
+
+This ensures that ``library`` targets are built as static, and nothing gets installed.
+
+To then link against the static library in the subproject, say for a subproject
+named ``bar`` with the main library target contained in a ``bar_dep`` dependency,
+add this to your ``meson.build`` file:
+
+.. code-block:: meson
+
+    bar_proj = subproject('bar')
+    bar_dep = bar_proj.get_variable('bar_dep')
+
+    py.extension_module(
+        '_example',
+        '_examplemod.c',
+        dependencies: bar_dep,
+        install: true,
+    )
+
+That is all!
+
+Shared library from subproject
+------------------------------
+
+If we can't use the static library approach from the section above and we need
+a shared library, then we must have ``install: true`` for that shared library
+target. This can only work if we can pass some build option to the subproject
+that tells it to *only* install the shared library and not headers or other
+targets that we don't need. Install tags don't work per subproject, so
+this will look something like:
+
+.. code-block:: meson
+
+    foo_subproj = subproject('foo',
+        default_options: {
+            # This is a custom option - if it doesn't exist, can you add it
+            # upstream or in WrapDB?
+            'only_install_main_lib': true,
+        })
+    foo_dep = foo_subproj.get_variable('foo_dep')
+
+Now we can use ``foo_dep`` like a normal dependency, ``meson-python`` will
+include it into the wheel in ``<project-name>.mesonpy.libs`` just like an
+internal shared library that targets ``libdir`` (see
+:ref:`internal-shared-libraries`).
+*Remember: this method doesn't support Windows (yet)!*
--- a/tests/packages/link-library-in-subproject/meson.build
+++ b/tests/packages/link-library-in-subproject/meson.build
@@ -1,0 +1,17 @@
+# SPDX-FileCopyrightText: 2022 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+project(
+    'link-library-in-subproject',
+    'c',
+    version: '1.0.0',
+    meson_version: '>=1.2.0',
+)
+
+py = import('python').find_installation(pure: false)
+
+bar_proj = subproject('bar')
+bar_dep = bar_proj.get_variable('bar_dep')
+
+subdir('foo')
--- a/tests/packages/link-library-in-subproject/pyproject.toml
+++ b/tests/packages/link-library-in-subproject/pyproject.toml
@@ -1,0 +1,15 @@
+# SPDX-FileCopyrightText: 2022 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+[build-system]
+build-backend = 'mesonpy'
+requires = ['meson-python']
+
+[project]
+name = 'link-library-in-subproject'
+version = '1.2.3'
+
+[tool.meson-python.args]
+setup = ['--default-library=static']
+install = ['--skip-subprojects']
--- a/tests/packages/link-library-in-subproject/subprojects/bar/examplelib.c
+++ b/tests/packages/link-library-in-subproject/subprojects/bar/examplelib.c
@@ -1,0 +1,9 @@
+// SPDX-FileCopyrightText: 2022 The meson-python developers
+//
+// SPDX-License-Identifier: MIT
+
+#include "bar_dll.h"
+
+BAR_DLL int sum(int a, int b) {
+    return a + b;
+}
--- a/tests/packages/link-library-in-subproject/subprojects/bar/meson.build
+++ b/tests/packages/link-library-in-subproject/subprojects/bar/meson.build
@@ -1,0 +1,36 @@
+# SPDX-FileCopyrightText: 2022 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+project('bar', 'c', version: '1.2.3', meson_version: '>= 1.3.0')
+
+if get_option('default_library') == 'shared' and meson.get_compiler('c').get_id() in ['msvc', 'clang-cl', 'intel-cl']
+    export_dll_args = ['-DBAR_DLL_EXPORTS']
+    import_dll_args = ['-DBAR_DLL_IMPORTS']
+else
+    export_dll_args = []
+    import_dll_args = []
+endif
+
+example_lib = library(
+    'examplelib',
+    'examplelib.c',
+    c_shared_args: export_dll_args,
+    install: true,
+)
+
+# A second library that we don't link from `foo`. If we install the subproject,
+# this second library also ends up in the wheel. To prevent that, we need to
+# skip installing this `bar` subproject, and statically link `example_lib`.
+unwanted_lib = library(
+    'unwantedlib',
+    'examplelib.c',
+    c_shared_args: export_dll_args,
+    install: true,
+)
+
+bar_dep = declare_dependency(
+    compile_args: import_dll_args,
+    link_with: example_lib,
+    include_directories: '.',
+)
--- a/tests/packages/sharedlib-in-package/mypkg/__init__.py
+++ b/tests/packages/sharedlib-in-package/mypkg/__init__.py
@@ -1,0 +1,52 @@
+# SPDX-FileCopyrightText: 2024 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+import os
+import sys
+
+
+# start-literalinclude
+def _append_to_sharedlib_load_path():
+    """
+    Ensure the shared libraries in this package can be loaded on Windows.
+
+    Windows lacks a concept equivalent to RPATH: Python extension modules
+    cannot find DLLs installed outside the DLL search path. This function
+    ensures that the location of the shared libraries distributed inside this
+    Python package is in the DLL search path of the process.
+
+    The Windows DLL search path includes the object depending on it is located:
+    the DLL search path needs to be augmented only when the Python extension
+    modules and the DLLs they require are installed in separate directories.
+    Cygwin does not have the same default library search path: all locations
+    where the shared libraries are installed need to be added to the search
+    path.
+
+    This function is very similar to the snippet inserted into the main
+    ``__init__.py`` of a package by ``delvewheel`` when it vendors external
+    shared libraries.
+
+    .. note::
+
+        `os.add_dll_directory` is only available for Python 3.8 and later, and
+        in the Conda ``python`` packages it works as advertised only for
+        version 3.10 and later. For older Python versions, pre-loading the DLLs
+        with `ctypes.WinDLL` may be preferred.
+    """
+    basedir = os.path.dirname(__file__)
+    subdir = os.path.join(basedir, 'sub')
+    if os.name == 'nt':
+        os.add_dll_directory(subdir)
+    elif sys.platform == 'cygwin':
+        os.environ['PATH'] = os.pathsep.join((os.environ['PATH'], basedir, subdir))
+
+
+_append_to_sharedlib_load_path()
+# end-literalinclude
+
+
+from ._example import example_prod, example_sum  #noqa: E402
+
+
+__all__ = ['example_prod', 'example_sum']
--- a/tests/packages/sharedlib-in-package/mypkg/examplelib.c
+++ b/tests/packages/sharedlib-in-package/mypkg/examplelib.c
@@ -1,0 +1,9 @@
+// SPDX-FileCopyrightText: 2022 The meson-python developers
+//
+// SPDX-License-Identifier: MIT
+
+#include "sub/mypkg_dll.h"
+
+MYPKG_DLL int sum(int a, int b) {
+    return a + b;
+}
--- a/tests/packages/sharedlib-in-package/mypkg/meson.build
+++ b/tests/packages/sharedlib-in-package/mypkg/meson.build
@@ -1,0 +1,41 @@
+# SPDX-FileCopyrightText: 2022 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+if meson.get_compiler('c').get_id() in ['msvc', 'clang-cl', 'intel-cl']
+    export_dll_args = ['-DMYPKG_DLL_EXPORTS']
+    import_dll_args = ['-DMYPKG_DLL_IMPORTS']
+else
+    export_dll_args = []
+    import_dll_args = []
+endif
+
+example_lib = shared_library(
+    'examplelib',
+    'examplelib.c',
+    c_args: export_dll_args,
+    install: true,
+    install_dir: py.get_install_dir() / 'mypkg',
+)
+
+example_lib_dep = declare_dependency(
+    compile_args: import_dll_args,
+    link_with: example_lib,
+)
+
+subdir('sub')
+
+py.extension_module(
+    '_example',
+    '_examplemod.c',
+    dependencies: [example_lib_dep, example_lib2_dep],
+    include_directories: 'sub',
+    install: true,
+    subdir: 'mypkg',
+    install_rpath: '$ORIGIN',
+)
+
+py.install_sources(
+    ['__init__.py'],
+    subdir: 'mypkg',
+)
--- a/tests/packages/sharedlib-in-package/mypkg/sub/examplelib2.c
+++ b/tests/packages/sharedlib-in-package/mypkg/sub/examplelib2.c
@@ -1,0 +1,9 @@
+// SPDX-FileCopyrightText: 2022 The meson-python developers
+//
+// SPDX-License-Identifier: MIT
+
+#include "mypkg_dll.h"
+
+MYPKG_DLL int prod(int a, int b) {
+    return a * b;
+}
--- a/tests/packages/sharedlib-in-package/mypkg/sub/meson.build
+++ b/tests/packages/sharedlib-in-package/mypkg/sub/meson.build
@@ -1,0 +1,16 @@
+# SPDX-FileCopyrightText: 2022 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+example_lib2 = shared_library(
+    'examplelib2',
+    'examplelib2.c',
+    c_args: export_dll_args,
+    install: true,
+    install_dir: py.get_install_dir() / 'mypkg/sub',
+)
+
+example_lib2_dep = declare_dependency(
+    compile_args: import_dll_args,
+    link_with: example_lib2,
+)
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -187,6 +187,13 @@ def test_sharedlib_in_package(venv, wheel_sharedlib_in_package):
     assert int(output) == 42
 
 
+@pytest.mark.skipif(MESON_VERSION < (1, 3, 0), reason='Meson version too old')
+def test_link_library_in_subproject(venv, wheel_link_library_in_subproject):
+    venv.pip('install', wheel_link_library_in_subproject)
+    output = venv.python('-c', 'import foo; print(foo.example_sum(3, 6))')
+    assert int(output) == 9
+
+
 @pytest.mark.skipif(sys.platform not in {'linux', 'darwin', 'sunos5'}, reason='Not supported on this platform')
 def test_rpath(wheel_link_against_local_lib, tmp_path):
     artifact = wheel.wheelfile.WheelFile(wheel_link_against_local_lib)
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -179,6 +179,14 @@ def test_local_lib(venv, wheel_link_against_local_lib):
     assert int(output) == 3
 
 
+def test_sharedlib_in_package(venv, wheel_sharedlib_in_package):
+    venv.pip('install', wheel_sharedlib_in_package)
+    output = venv.python('-c', 'import mypkg; print(mypkg.example_sum(2, 5))')
+    assert int(output) == 7
+    output = venv.python('-c', 'import mypkg; print(mypkg.example_prod(6, 7))')
+    assert int(output) == 42
+
+
 @pytest.mark.skipif(sys.platform not in {'linux', 'darwin', 'sunos5'}, reason='Not supported on this platform')
 def test_rpath(wheel_link_against_local_lib, tmp_path):
     artifact = wheel.wheelfile.WheelFile(wheel_link_against_local_lib)
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -82,6 +82,7 @@ the use of ``meson-python`` and Meson for Python packaging.
    how-to-guides/config-settings
    how-to-guides/meson-args
    how-to-guides/debug-builds
+   how-to-guides/shared-libraries
    reference/limitations
    projects-using-meson-python
 
--- a/tests/packages/link-library-in-subproject/foo/__init__.py
+++ b/tests/packages/link-library-in-subproject/foo/__init__.py
@@ -1,0 +1,8 @@
+# SPDX-FileCopyrightText: 2024 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+from ._example import example_sum
+
+
+__all__ = ['example_sum']
--- a/tests/packages/link-library-in-subproject/foo/_examplemod.c
+++ b/tests/packages/link-library-in-subproject/foo/_examplemod.c
@@ -1,0 +1,37 @@
+// SPDX-FileCopyrightText: 2022 The meson-python developers
+//
+// SPDX-License-Identifier: MIT
+
+#include <Python.h>
+
+#include "examplelib.h"
+
+static PyObject* example_sum(PyObject* self, PyObject *args)
+{
+    int a, b;
+    if (!PyArg_ParseTuple(args, "ii", &a, &b)) {
+        return NULL;
+    }
+
+    long result = sum(a, b);
+
+    return PyLong_FromLong(result);
+}
+
+static PyMethodDef methods[] = {
+    {"example_sum", (PyCFunction)example_sum, METH_VARARGS, NULL},
+    {NULL, NULL, 0, NULL},
+};
+
+static struct PyModuleDef module = {
+    PyModuleDef_HEAD_INIT,
+    "_example",
+    NULL,
+    -1,
+    methods,
+};
+
+PyMODINIT_FUNC PyInit__example(void)
+{
+    return PyModule_Create(&module);
+}
--- a/tests/packages/link-library-in-subproject/foo/meson.build
+++ b/tests/packages/link-library-in-subproject/foo/meson.build
@@ -1,0 +1,16 @@
+# SPDX-FileCopyrightText: 2022 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+py.extension_module(
+    '_example',
+    '_examplemod.c',
+    dependencies: bar_dep,
+    install: true,
+    subdir: 'foo',
+)
+
+py.install_sources(
+    ['__init__.py'],
+    subdir: 'foo',
+)
--- a/tests/packages/link-library-in-subproject/subprojects/bar/bar_dll.h
+++ b/tests/packages/link-library-in-subproject/subprojects/bar/bar_dll.h
@@ -1,0 +1,19 @@
+// SPDX-FileCopyrightText: 2024 The meson-python developers
+//
+// SPDX-License-Identifier: MIT
+
+#pragma once
+
+// When building the `examplelib` DLL, this macro expands to `__declspec(dllexport)`
+// so we can annotate symbols appropriately as being exported. When used in
+// headers consuming a DLL, this macro expands to `__declspec(dllimport)` so
+// that consumers know the symbol is defined inside the DLL. In all other cases,
+// the macro expands to nothing.
+// Note: BAR_DLL_{EX,IM}PORTS are set in meson.build
+#if defined(BAR_DLL_EXPORTS)
+    #define BAR_DLL __declspec(dllexport)
+#elif defined(BAR_DLL_IMPORTS)
+    #define BAR_DLL __declspec(dllimport)
+#else
+    #define BAR_DLL
+#endif
--- a/tests/packages/link-library-in-subproject/subprojects/bar/examplelib.h
+++ b/tests/packages/link-library-in-subproject/subprojects/bar/examplelib.h
@@ -1,0 +1,7 @@
+// SPDX-FileCopyrightText: 2022 The meson-python developers
+//
+// SPDX-License-Identifier: MIT
+
+#include "bar_dll.h"
+
+BAR_DLL int sum(int a, int b);
--- a/tests/packages/sharedlib-in-package/meson.build
+++ b/tests/packages/sharedlib-in-package/meson.build
@@ -1,0 +1,9 @@
+# SPDX-FileCopyrightText: 2022 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+project('sharedlib-in-package', 'c', version: '1.0.0')
+
+py = import('python').find_installation(pure: false)
+
+subdir('mypkg')
--- a/tests/packages/sharedlib-in-package/mypkg/_examplemod.c
+++ b/tests/packages/sharedlib-in-package/mypkg/_examplemod.c
@@ -1,0 +1,51 @@
+// SPDX-FileCopyrightText: 2022 The meson-python developers
+//
+// SPDX-License-Identifier: MIT
+
+#include <Python.h>
+
+#include "examplelib.h"
+#include "examplelib2.h"
+
+static PyObject* example_sum(PyObject* self, PyObject *args)
+{
+    int a, b;
+    if (!PyArg_ParseTuple(args, "ii", &a, &b)) {
+        return NULL;
+    }
+
+    long result = sum(a, b);
+
+    return PyLong_FromLong(result);
+}
+
+static PyObject* example_prod(PyObject* self, PyObject *args)
+{
+    int a, b;
+    if (!PyArg_ParseTuple(args, "ii", &a, &b)) {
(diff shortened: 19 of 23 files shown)


tests/packages/sharedlib-in-package/mypkg/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+    within wheels. So pre-load it.
+    """
+    if os.name == "nt":
+        from ctypes import WinDLL
+        basedir = os.path.dirname(__file__)
+        dll_path = os.path.join(basedir, "example_lib.dll")
+        WinDLL(dll_path)
+
+
+_load_sharedlib()
```
dnicolodi, line 22: Isn't `os.add_dll_directory()` the recommended way to deal with this? https://docs.python.org/3/library/os.html#os.add_dll_directory

rgommers: That's true now, yes. The pain point is that it can only be relied on for Python >=3.10. It wasn't useful until the documented changes in 3.8, and those took 2 more minor versions to propagate to at least Conda's `python` builds.

We could use `os.add_dll_directory` in the docs here, and note that if a package still supports <3.10, they should use `ctypes.WinDLL` preloading instead?

rgommers: We're also still testing with Python 3.7, which is EOL. Maybe time to drop that for the next release?

dnicolodi: I'm not sure I'm following. The documentation states that this function was added in Python 3.8. Are you saying that Conda's Python carried some patches that broke this and it was fixed only in Conda's Python 3.10?

The `ctypes.WinDLL` does not work either, judging from the results of CI 😄

dnicolodi: Oh, but the CI issue is just a typo in the name of the library that is being loaded.

rgommers: Yes `WinDLL` should work, since we're using it in SciPy.

And re Conda: I remember it was a long story, but the tl;dr I took from it was that CPython upstream's changes were quite disruptive (not surprising, since Conda is the only real distro on Windows, and had already needed to implement its own DLL loading rules/precedence) and it took two years for Conda to adapt.

dnicolodi: > We're also still testing with Python 3.7, which is EOL. Maybe time to drop that for the next release?

So far it would have required more effort to drop support for Python 3.7 than to keep it, so it hasn't happened. If there is any reason for dropping it (including support from the CI infrastructure) I don't see any reason why we should keep it.

dnicolodi: I think mentioning that the `os.add_dll_directory()` thing does not work on Conda's Python before 3.10. Not knowing better, I would have used it and debugging while it does not work on Conda would have not been fun.

rgommers: Yeah it wasn't at the time, we had a bad time with wheels vendoring shared libs not working with Conda's Python. 

I'll finish debugging this new test on my own fork or find a Windows machine, enough abusing CI here.

.cirrus.yml
```diff
@@ -55,12 +55,12 @@ manylinux-python3.13t_task:
...
   << : *test
 
-manylinux-python3.7_task:
+manylinux-python3.8_task:
   container:
     dockerfile: ci/manylinux.docker
     cpu: 1
   env:
-    PATH: "/opt/python/cp37-cp37m/bin/:${PATH}"
+    PATH: "/opt/python/cp38-cp38m/bin/:${PATH}"
```
dnicolodi, line 63: The `m` should not be there for CPython 3.8

rgommers:
Okay, changed to `os.add_dll_directory` with some compat notes added. The new test case passes on all Windows configs now except for Cygwin. I don't know what's going on there - I don't think we do anything special in SciPy for Cygwin with the same kind of shared-library-inside-python-package setup. The build completes but at runtime the shared library goes missing:

<details>

```
[1/5] Compiling C object mypkg/cygexamplelib.dll.p/examplelib.c.o
[2/5] Compiling C object mypkg/_example.cpython-39-x86_64-cygwin.dll.p/_examplemod.c.o
[3/5] Linking target mypkg/cygexamplelib.dll
[4/5] Generating symbol file mypkg/cygexamplelib.dll.p/cygexamplelib.dll.symbols
[5/5] Linking target mypkg/_example.cpython-39-x86_64-cygwin.dll
[1/5] /cygdrive/d/a/meson-python/meson-python/tests/packages/sharedlib-in-package/.mesonpy-8c69nja6/mypkg/cygexamplelib.dll
[2/5] /cygdrive/d/a/meson-python/meson-python/tests/packages/sharedlib-in-package/.mesonpy-8c69nja6/mypkg/libexamplelib.dll.a
[3/5] /cygdrive/d/a/meson-python/meson-python/tests/packages/sharedlib-in-package/.mesonpy-8c69nja6/mypkg/_example.cpython-39-x86_64-cygwin.dll
[4/5] /cygdrive/d/a/meson-python/meson-python/tests/packages/sharedlib-in-package/.mesonpy-8c69nja6/mypkg/_example.cpython-39-x86_64-cygwin.dll.a
[5/5] /cygdrive/d/a/meson-python/meson-python/tests/packages/sharedlib-in-package/mypkg/__init__.py
----------------------------- Captured stderr call -----------------------------
  File "/tmp/pytest-of-runneradmin/pytest-0/mesonpy-test-venv4/lib/python3.9/site-packages/mypkg/__init__.py", line 32, in <module>
    from ._example import example_sum
ImportError: No such file or directory
```

</details>

@DWesl could you perhaps have a look at this? I'm not even sure it's supposed to be working, or it requires support within Cygwin somehow.

dnicolodi:
> `[1/5] /cygdrive/d/a/meson-python/meson-python/tests/packages/sharedlib-in-package/.mesonpy-8c69nja6/mypkg/cygexamplelib.dl`

Why does the shared library have this funny name?

tests/test_wheel.py
```diff
@@ -161,6 +161,13 @@ def test_local_lib(venv, wheel_link_against_local_lib):
     assert int(output) == 3
 
 
+@pytest.mark.skipif(MESON_VERSION < (0, 64, 0), reason='Meson version too old')
```
dnicolodi, line 164: What does break with older Meson versions?

rgommers:
> Why does the shared library have this funny name?

Seems like the fun of building on Windows - with MinGW we get `libexamplelib.dll`, with MSVC we get `examplelib.dll` and with Cygwin we get `cygexamplelib.dll`.

tests/test_wheel.py
rgommers:
```
../meson.build:7:0: ERROR: python.find_installation got unknown keyword arguments "pure"
```

dnicolodi: Oh, this. We work around it in other test packages. I think we should either bump the minimum required Meson version or work around it here too passing the `pure` argument to the individual functions in the `python` module. Meson 0.64 was released November 6th 2022, thus almost 2 years ago. Maybe we can update the minimum required version to 0.64 without feeling bad 😃

rgommers: Yes, I'd prefer to bump to 0.64 rather than continue to use less idiomatic code. Especially since I plan to link to this test package from the docs as a complete and working example.

tests/packages/sharedlib-in-package/mypkg/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+        If you require support for older versions, pre-loading the DLL
+        with `ctypes.WinDLL` may be preferred (the SciPy code base has an
+        example of this).
+    """
+    if os.name == "nt":
+        basedir = os.path.dirname(__file__)
+        os.add_dll_directory(basedir)
+
+
+_enable_sharedlib_loading()
```
dnicolodi, line 29: Does this really need to define a function and immediately call it? Given that the setup code is reduced to three lines (that can be condensed to two) I think it is clearer to simply have the code directly at module level. Also, I think that it would be nice to have the code actually shown in the documentation page with a Sphinx code include directive. Then the function doc string can be included in the documentation page (I don't think it is easy to have it rendered nicely in the documentation page otherwise).

dnicolodi:
Would it be worth to run `pyupgrade --py38-plus` as part of dropping support for Python 3.7? I've just tried and it upgrades almost all typing annotations (and introduces some bugs about encoding...) thus the noise may be not worth the simplifications.

tests/packages/link-library-in-subproject/meson.build
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: 2022 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+project(
+    'link-library-in-subproject',
+    'c',
+    version: '1.0.0',
+    meson_version: '>=1.2.3',
```
dnicolodi, line 9: The requirement for this Meson version should probably be documented here https://mesonbuild.com/meson-python/reference/meson-compatibility.html#meson-compatibility I completely forgot about it

tests/packages/link-library-in-subproject/pyproject.toml
```diff
@@ -0,0 +1,17 @@
...
+# SPDX-License-Identifier: MIT
+
+[build-system]
+build-backend = 'mesonpy'
+requires = ['meson-python']
+
+[project]
+name = 'link-library-in-subproject'
+version = '1.2.3'
+license = 'MIT'
```
dnicolodi, line 12: This makes the test dependent on pyproject-metadata 0.9.0 (the test fails on the CI job that tests with pyproject-metadata 0.8.0, which is currently the minimum required version)

rgommers:
> Would it be worth to run pyupgrade --py38-plus as part of dropping support for Python 3.7? I've just tried and it upgrades almost all typing annotations (and introduces some bugs about encoding...) thus the noise may be not worth the simplifications.

I'll have a look, but if that's useful enough then I'd prefer to do it in a separate PR I think.

tests/packages/link-library-in-subproject/pyproject.toml
rgommers: Yes, seeing the failure - I'll drop this line then.

DWesl:
> Okay, changed to `os.add_dll_directory` with some compat notes added. The new test case passes on all Windows configs now except for Cygwin. I don't know what's going on there - I don't think we do anything special in SciPy for Cygwin with the same kind of shared-library-inside-python-package setup. The build completes but at runtime the shared library goes missing:
> 
> ```
> [1/5] Compiling C object mypkg/cygexamplelib.dll.p/examplelib.c.o
> [2/5] Compiling C object mypkg/_example.cpython-39-x86_64-cygwin.dll.p/_examplemod.c.o
> [3/5] Linking target mypkg/cygexamplelib.dll
> [4/5] Generating symbol file mypkg/cygexamplelib.dll.p/cygexamplelib.dll.symbols
> [5/5] Linking target mypkg/_example.cpython-39-x86_64-cygwin.dll
> [1/5] /cygdrive/d/a/meson-python/meson-python/tests/packages/sharedlib-in-package/.mesonpy-8c69nja6/mypkg/cygexamplelib.dll
> [2/5] /cygdrive/d/a/meson-python/meson-python/tests/packages/sharedlib-in-package/.mesonpy-8c69nja6/mypkg/libexamplelib.dll.a
> [3/5] /cygdrive/d/a/meson-python/meson-python/tests/packages/sharedlib-in-package/.mesonpy-8c69nja6/mypkg/_example.cpython-39-x86_64-cygwin.dll
> [4/5] /cygdrive/d/a/meson-python/meson-python/tests/packages/sharedlib-in-package/.mesonpy-8c69nja6/mypkg/_example.cpython-39-x86_64-cygwin.dll.a
> [5/5] /cygdrive/d/a/meson-python/meson-python/tests/packages/sharedlib-in-package/mypkg/__init__.py
> ----------------------------- Captured stderr call -----------------------------
> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
>   File "/tmp/pytest-of-runneradmin/pytest-0/mesonpy-test-venv4/lib/python3.9/site-packages/mypkg/__init__.py", line 32, in <module>
>     from ._example import example_sum
> ImportError: No such file or directory
> ```
> 
> @DWesl could you perhaps have a look at this? I'm not even sure it's supposed to be working, or it requires support within Cygwin somehow.

Most of the packages I've worked with have only depended on system DLLs/shared libraries, but the reference BLAS and Lapack libraries are in a non-standard location so I have some experience with this from NumPy.  The easy things to check are permissions (both shared libraries need execute permissions to work properly) and whether the `PATH` is set properly to find dependent shared libraries actually linked into the extension module (`ldd` is the tool you're probably familiar with, though I tend to use `cygcheck`).

From what I remember of the last time I compiled SciPy on Cygwin a few years back, SciPy doesn't link to itself; dependencies are Python-level or header-only, so this problem wouldn't come up.  I haven't run into `os.add_dll_directory` (in the CPython repo it looks Windows-only and would not be available on Cygwin), I've only ever directly modified PATH outside Python.  I wonder whether something like:
```python
if sys.platform == "cygwin":
    def add_dll_directory(path: str):
        os.environ["PATH"] = f"{os.environ['PATH']:s}:{path:s}"
    os.add_dll_directory = add_dll_directory
```
might help.

tests/packages/sharedlib-in-package/mypkg/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+        `os.add_dll_directory` is only available for Python >=3.8, and with
+        the Conda `python` packages only works as advertised for >=3.10.
+        If you require support for older versions, pre-loading the DLL
+        with `ctypes.WinDLL` may be preferred (the SciPy code base has an
+        example of this).
+    """
+    if os.name == "nt":
+        basedir = os.path.dirname(__file__)
+        os.add_dll_directory(basedir)
+
```
DWesl, line 27: Given there's an explicit function for this, the better route may be:
```suggestion
    elif sys.platform == "cygwin":
        basedir = os.path.dirname(__file__)
        os.environ["PATH"] = f"{os.environ['PATH']:s}:{basedir:s}"

```

tests/packages/link-library-in-subproject/meson.build
dnicolodi: Our tests say that the minimum Meson version for `--skip-subprojects` to work is 1.2.0

tests/packages/link-library-in-subproject/pyproject.toml
dnicolodi: In #681 we are updating the minimum required version to 0.9.0. I would kip this.

tests/packages/link-library-in-subproject/meson.build
dnicolodi: Documentation note added in #700

rgommers:
Thanks for the input @DWesl!

> From what I remember of the last time I compiled SciPy on Cygwin a few years back, SciPy doesn't link to itself; dependencies are Python-level or header-only, so this problem wouldn't come up.

That changed recently, the SciPy 1.14.0 release has a shared library:
https://github.com/scipy/scipy/blob/ea916c6f7f487bd53e98de082649d542cc6106ed/scipy/special/meson.build#L37

> I've only ever directly modified PATH outside Python. I wonder whether something like: [...] might help

I'll try but I doubt it, since modifying `os.environ` shouldn't affect the existing process IIRC.

> but the reference BLAS and Lapack libraries are in a non-standard location so I have some experience with this from NumPy.

Is there a patch/repo somewhere for how the `numpy` package in Cygwin is built?

tests/packages/sharedlib-in-package/mypkg/__init__.py
rgommers: Makes sense, if it stays at ~2 lines I'll drop the function. `.. literalinclude::` is useful too, will do.

For completeness, `delvewheel` inserts this for external shared libraries that it vendors in:
```python
# start delvewheel patch
def _delvewheel_patch_1_8_2():
    import os
    libs_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir, 'numpy.libs'))
    if os.path.isdir(libs_dir):
        os.add_dll_directory(libs_dir)


_delvewheel_patch_1_8_2()
del _delvewheel_patch_1_8_2
# end delvewheel patch
```
The `del` is a nice touch if one wants to keep the namespace very clean.

DWesl:
> Thanks for the input @DWesl!
> 
> > but the reference BLAS and Lapack libraries are in a non-standard location so I have some experience with this from NumPy.
> 
> Is there a patch/repo somewhere for how the `numpy` package in Cygwin is built?

For NumPy:
https://github.com/numpy/numpy/blob/main/.github/workflows/cygwin.yml

For SciPy, most recent I have:
https://github.com/DWesl/scipy/pull/7

> That changed recently, the SciPy 1.14.0 release has a shared library

It might still work, depending on what Windows thinks is the current directory for a `dlopen` call and which extensions load it.  I can't test at the moment, since Cygwin doesn't have Python>=3.10 (numpy/numpy#26247)

tests/packages/sharedlib-in-package/mypkg/__init__.py
rgommers: This does indeed work 🎉

rgommers:
The Cygwin tests pass now, thanks for the pointers @DWesl. Amending `os.environ['PATH']` worked (`'LD_LIBRARY_PATH'` did not), and is necessary when the shared library is in another directory _and also_ when it is right next to a Python extension module that needs it. For regular Windows, curdir is searched by default so `os.add_dll_directory` is only necessary when the shared library is elsewhere.

DWesl:
> The Cygwin tests pass now, thanks for the pointers @DWesl. Amending `os.environ['PATH']` worked (`'LD_LIBRARY_PATH'` did not),

I think `LD_LIBRARY_PATH` gets used by `dlopen` on Cygwin, but Python already has its own logic for that (`sys.path`), so that's not relevant here.

> and is necessary when the shared library is in another directory _and also_ when it is right next to a Python extension module that needs it. For regular Windows, curdir is searched by default so `os.add_dll_directory` is only necessary when the shared library is elsewhere.

It sounds like Windows does some helpful `chdir` things/`PATH` manipulation around its version of `dlopen` then (`LoadDynamicLibraryEx` maybe?), that Cygwin (and the lower-level Windows function it passes things off to) does not.  
Interesting to know about in case it comes up again.

docs/how-to-guides/shared-libraries.rst
```diff
@@ -0,0 +1,200 @@
...
+   and `meson#13046 <https://github.com/mesonbuild/meson/issues/13046>`__ for
+   issues this can cause.
+
+   TODO: describe workarounds, e.g. via ``-Wl,-rpath`` or setting ``LD_LIBRARY_PATH``.
+
+On Windows, the shared library can either be preloaded, or vendored with
+``delvewheel`` in order to make the built Python package usable locally.
+
+
+Publishing wheels which depend on external shared libraries
```
virtuald, line 158: There's no mention here of the topic of linking to libraries contained in other wheels. Even if it's not explained in the final draft, it should probably be mentioned and TODO added.

rgommers: Thanks for the suggestion @virtuald. I'm not quite sure I want to add that topic though. Just to clarify, did you mean (a) an actual shared library contained in a wheel (a la `scipy-openblas32`), or (b) a Python package that exports a C API (a la NumPy's C API)?

I think (b) doesn't really fit topic-wise, and (a) isn't recommended, since it requires a bunch of manual workarounds to get to work at all, and then there's no good way to deal with versioning and upgrades.

virtuald: I mean (a), as I think when you start discussing 'external' libraries eventually that thought comes up, so it should be given at least some mention. As I said, even if it's just a footnote that says "you probably shouldn't do this" or something else to that effect, I think that would be appropriate enough for now.

rgommers: Makes sense, thanks - I'll add a note like that.

HaoZeke (commented):
This seems to be shaping up very nicely, and is super useful (drive by spelling nits).

tests/test_wheel.py
rgommers: Upgrade to 0.64.0 minimum version was done, so I dropped this line - resolving.

docs/how-to-guides/shared-libraries.rst
rgommers: I added this note.

docs/how-to-guides/shared-libraries.rst
```diff
@@ -0,0 +1,200 @@
...
+| internal              | libdir           | no (1)  | ✓     | ✓     |
++-----------------------+------------------+---------+-------+-------+
+| internal              | site-packages    | ✓       | ✓     | ✓     |
++-----------------------+------------------+---------+-------+-------+
+| external              | n/a              | ✓ (2)   | ✓     | ✓     |
++-----------------------+------------------+---------+-------+-------+
+
+.. TODO: add subproject as a source
+
+1: Internal shared libraries on Windows cannot be automaticall handled
```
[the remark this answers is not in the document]
rgommers: thanks for catching these spelling errors - incorporated now.

rgommers:
> Would it be worth to run `pyupgrade --py38-plus` as part of dropping support for Python 3.7? I've just tried and it upgrades almost all typing annotations (and introduces some bugs about encoding...) thus the noise may be not worth the simplifications.

I tried now. A very large diff indeed with things that mostly seemed pretty uninteresting, so I prefer to skip this exercise.

eli-schwartz:
FWIW, nearly all changes that pyupgrade does, it does for `--py37-plus` as well. There's no granularity whatsoever with regard to what sort of changes you're interested in making.

Consider using `ruff` instead, as you can enable/disable your choice of 42 distinct rules in the UP range: https://docs.astral.sh/ruff/rules/#pyupgrade-up

tests/packages/link-library-in-subproject/pyproject.toml
rgommers: Removed this in the end; it served no real purpose and almost all other test packages don't have a `license =` entry.

rgommers:
This should be good to go now.

tests/packages/link-library-in-subproject/subprojects/bar/examplelib.c
```diff
@@ -0,0 +1,9 @@
+// SPDX-FileCopyrightText: 2022 The meson-python developers
+//
+// SPDX-License-Identifier: MIT
+
+#include "bar_dll.h"
+
+BAR_DLL int sum(int a, int b) {
+	return a + b;
```
dnicolodi, line 8: It does not matter much, but we use 4 spaces indentation for `.c` and `.cpp` files in all other files, keeping consistency is always nice 🙂

tests/packages/link-library-in-subproject/subprojects/bar/meson.build
```diff
@@ -0,0 +1,36 @@
...
+
+project('bar', 'c', version: '1.2.3', meson_version: '>= 1.3.0')
+
+if get_option('default_library') == 'shared' and meson.get_compiler('c').get_id() in ['msvc', 'clang-cl', 'intel-cl']
+  export_dll_args = ['-DBAR_DLL_EXPORTS']
+  import_dll_args = ['-DBAR_DLL_IMPORTS']
+else
+  export_dll_args = []
+  import_dll_args = []
+endif
```
dnicolodi, line 13: I think we decided to use 4 spaces indentation for all `meson.build` files.

tests/packages/link-library-in-subproject/subprojects/bar/meson.build
```diff
@@ -0,0 +1,36 @@
...
+    'examplelib.c',
+    c_shared_args: export_dll_args,
+    install: true,
+)
+
+bar_dep = declare_dependency(
+  compile_args: import_dll_args,
+  link_with: example_lib,
+  include_directories: '.',
+)
```
dnicolodi, line 36: Also here.

tests/packages/sharedlib-in-package/mypkg/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+# SPDX-License-Identifier: MIT
+
+import os
+import sys
+
+
+# Note: the below function is used in the docs via a literalinclude directive.
+#       When making any changes to this file, please ensure the right set of
+#       lines is pulled into the docs, and adjust the :lines: attribute of
+#       the literalinclude if needed.
```
dnicolodi, line 12: I think that Sphynx can include Python code by object name, so the issue describe here can be avoided using
```rest
.. literlainclude:: .../__init__.py
   pyobejct: _enable_sharedlib_loading
```
or something like this, maybe?

tests/packages/sharedlib-in-package/mypkg/examplelib.c
```diff
@@ -0,0 +1,9 @@
+// SPDX-FileCopyrightText: 2022 The meson-python developers
+//
+// SPDX-License-Identifier: MIT
+
+#include "sub/mypkg_dll.h"
+
+MYPKG_DLL int sum(int a, int b) {
+	return a + b;
```
dnicolodi, line 8: Indentation.

tests/packages/sharedlib-in-package/mypkg/meson.build
```diff
@@ -0,0 +1,41 @@
...
+#
+# SPDX-License-Identifier: MIT
+
+if meson.get_compiler('c').get_id() in ['msvc', 'clang-cl', 'intel-cl']
+  export_dll_args = ['-DMYPKG_DLL_EXPORTS']
+  import_dll_args = ['-DMYPKG_DLL_IMPORTS']
+else
+  export_dll_args = []
+  import_dll_args = []
+endif
```
dnicolodi, line 11: Indentation

tests/packages/sharedlib-in-package/mypkg/sub/examplelib2.c
```diff
@@ -0,0 +1,9 @@
+// SPDX-FileCopyrightText: 2022 The meson-python developers
+//
+// SPDX-License-Identifier: MIT
+
+#include "mypkg_dll.h"
+
+MYPKG_DLL int prod(int a, int b) {
+	return a * b;
```
dnicolodi, line 8: Indentation

tests/packages/sharedlib-in-package/mypkg/sub/meson.build
```diff
@@ -0,0 +1,16 @@
...
+    'examplelib2',
+    'examplelib2.c',
+    c_args: export_dll_args,
+    install: true,
+    install_dir: py.get_install_dir() / 'mypkg/sub',
+)
+
+example_lib2_dep = declare_dependency(
+  compile_args: import_dll_args,
+  link_with: example_lib2,
```
dnicolodi, line 15: Indentation

tests/packages/sharedlib-in-package/mypkg/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def _enable_sharedlib_loading():
+    """
+    Ensure the shared libraries in this dir and the `sub` subdir can be loaded
+    on Windows.
+
+    One shared library is installed alongside this ``__init__.py`` file.
```
dnicolodi, line 20: There is an inconsistent use of single and double backticks used to delimit literals in this docstring. It does not matter much, but as it is a code example maybe we should be a bit more careful than usual.

tests/packages/sharedlib-in-package/mypkg/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    Windows can load it because it searches for DLLs in the directory a ``.pyd``
```
dnicolodi, line 21: I'm not a native English speaker, but I think this needs to be "it searches for DLLs in the directory where a ``.pyd``
(Python extension module) is located."

tests/packages/sharedlib-in-package/mypkg/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    (Python extension module) is located in. Cygwin does not though. For a
+    shared library in another directory inside the package, Windows also needs
+    a hint.
+
+    This function is Windows-specific due to lack of RPATH support on Windows.
+    It cannot find shared libraries installed within wheels unless we either
+    amend the DLL search path or pre-load the DLL.
```
dnicolodi, line 28: I find this docstring a bit repetitive and contrived in places. I think it can be made much more concise. I can try to come up with a better formulation, if you like.

tests/packages/sharedlib-in-package/mypkg/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+
+    Note that `delvewheel` inserts a similar snippet into the main
+    `__init__.py` of a package when it vendors external shared libraries.
```
dnicolodi, line 31: Is this really relevant for users? The topic is already complex and I would like to avoid to overwhelm the readers with information.

docs/how-to-guides/shared-libraries.rst
```diff
@@ -0,0 +1,253 @@
...
+Python projects may build shared libraries as part of their project, or link
+with shared libraries from a dependency. This tends to be a common source of
+issues, hence this page aims to explain how to include shared libraries in
+wheels, any limitations and gotchas, and how support is implemented in
+``meson-python`` under the hood.
+
+We distinguish between *internal* shared libraries, i.e. they're built as part
+of the build executed by ``meson-python``, and *external* shared libraries that
+are only linked against from targets (usually Python extension modules) built
+by ``meson-python``. For internal shared libraries, we also distinguish whether
```
dnicolodi, line 20: I find the distinction between internal and external shared libraries a bit hard to read. What about something like:

We distinguish between *internal* shared libraries that are built as part of the project, and *external* shared libraries provided by project dependencies and that are linked with the project build artifacts.

or something like this.

docs/how-to-guides/shared-libraries.rst
```diff
@@ -0,0 +1,253 @@
...
+the shared library is being installed to its default location (i.e. ``libdir``,
```
dnicolodi, line 21: Maybe "default system location" to indicate that the install location is outside the Python environment?

docs/how-to-guides/shared-libraries.rst
```diff
@@ -0,0 +1,253 @@
...
+usually something like ``<prefix>/lib/``) or to a location in ``site-packages``
```
dnicolodi, line 22: I would substitute the parenthetical statement with "(typically `/usr/local/lib` on Unix-like systems, and `?` on Windows)". I don't know what the default prefix is on Windows and I'm not sure the default prefix is `/usr/local`.

dnicolodi (changes requested):
Thanks for working on this @rgommers! I have a first set of comments based on a first quick read through the additions. I'll come back with more.

basnijholt:
@rgommers, just wanted to thank you for writing these docs! 

It has been key in getting the Windows builds in https://github.com/basnijholt/pfapack/pull/21 to work. 

If only I found this PR 5 hours ago 😅

tests/packages/sharedlib-in-package/mypkg/__init__.py
dnicolodi: The `start-after` and `end-before` arguments for `literalinclude` are also an option that does not require to hard-code line numbers https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-option-literalinclude-start-after

tests/packages/link-library-in-subproject/subprojects/bar/meson.build
rgommers: Sorry about all those - syncing from SciPy with its 2-space indent made those sneak in.

docs/how-to-guides/shared-libraries.rst
rgommers: Yes, `/usr/local/lib` and `C:\\lib`, see https://mesonbuild.com/Builtin-options.html#universal-options.

docs/how-to-guides/shared-libraries.rst
rgommers: thanks, adopted.

docs/how-to-guides/shared-libraries.rst
rgommers: That sounds better, adopted.

tests/packages/sharedlib-in-package/mypkg/__init__.py
rgommers: Good point. I did them more carefully, only keeping the stdlib module/object that will get linked almost certainly correctly by intersphinx as single backtick (link), and the rest as double backtick (code). That way it works well when copying into some other project.

tests/packages/sharedlib-in-package/mypkg/__init__.py
rgommers: I'd keep it - there is no other documentation that explains any of this well, so I'd err on the side of completeness. Users reading this are probably fairly technical (but this is still new info to many of them), so I'm not too worried about explaining too much.

tests/packages/sharedlib-in-package/mypkg/__init__.py
rgommers: Yes please, that could make it better.

tests/packages/sharedlib-in-package/mypkg/__init__.py
rgommers: Thanks, fixed now - much easier to maintain this way.

rgommers:
Thanks for the review @dnicolodi, and for the feedback @basnijholt!

docs/how-to-guides/shared-libraries.rst
```diff
@@ -0,0 +1,254 @@
...
+.. literalinclude:: ../../tests/packages/sharedlib-in-package/mypkg/__init__.py
+   :start-after: start-literalinclude
+   :end-before: end-literalinclude
+
+If an internal shared library is not only used as part of a Python package, but
+for example also as a regular shared library in a C/C++ project or as a
+standalone library, then the method shown above won't work - the library has to
+be installed to the default ``libdir`` location. In that case, ``meson-python``
+will detect that the library is going to be installed to ``libdir`` - which is
+not a recommended install location for wheels, and not supported by
```
dnicolodi, line 97: I think "recommended install location" is a bit weak: there is no way for a Python wheel to install in `libdir`. I'm pointing it out because I foresee someone reading this and interpreting it as meson-python enforcing a recommendation as a strict rule and asking for this to be supported (we already had cases of this interpretation and of this request in the issue tracker). Maybe a stronger wording is justified here.

docs/how-to-guides/shared-libraries.rst
```diff
@@ -0,0 +1,254 @@
...
+   An RPATH entry alone is not always enough - if the directory that the shared
+   library is located in is not on the loader search path, then the shared
+   library may go missing at runtime. See, e.g., `meson#2121
+   <https://github.com/mesonbuild/meson/issues/2121>`__ and
+   `meson#13046 <https://github.com/mesonbuild/meson/issues/13046>`__ for
+   issues this can cause.
+
+   If this happens, workarounds include adding the directory that the shared
+   library is located in to the search path by amending the ``LD_LIBRARY_PATH``
+   environment variable, or by using a compile flag ``-Wl,-rpath,/path/to/sharedlib/dir``.
```
dnicolodi, line 134: I don't understand this note. Adding an RPATH is equivalent to setting LD_LIBRARY_PATH and the compiler flag is just a way to add to RPATH. Without studying them in detail, what is going on in the linked issues is that RPATH is absolute/relative but the user would like it to be relative/absolute, respectively.

The only way for the library to go missing at load time is when the RPATH is relative and the library install location is changed or the library is not at the location it was when the artifact linking to it was built.

Am I missing something?

dnicolodi (changes requested):
Thanks @rgommers, especially for bearing with me for the tedious work of keeping indentation consistent. I've spotted a couple more things that may warrant some attention. Otherwise I think we can just merge this and iterate on it in follow up patches, if needed. 

There is another tedious thing I have noticed: the copyright years in some files are wrong. I don't care and I remember suggesting quite strongly that having the copyright year in the SPDX headers is just a waste of time, but I lost that battle 😃 and if we have the years we should have them accurate.

rgommers:
> There is another tedious thing I have noticed: the copyright years in some files are wrong. I don't care and I remember suggesting quite strongly that having the copyright year in the SPDX headers is just a waste of time, but I lost that battle 😃 and if we have the years we should have them accurate.

IIRC what I did is for files that I copied from another test package to keep the year unchanged, and for files that I wrote from scratch I added 2024. Happy to do it some other way (2024 for everything?), but I did try to follow a consistent rule at least. No preference from my side - please let me know what you prefer here.

docs/how-to-guides/shared-libraries.rst
rgommers: There is a way: use the `data` directory. Unpack an `mkl` or `pybind11-global` wheel to verify - they install into `prefix/lib` and `prefix/include`. It is indeed a recommendation that we enforce as of today.

docs/how-to-guides/shared-libraries.rst
rgommers: What is going on is that C/C++/Fortran libraries like `zlib`, `openblas`, etc. expect to be installed into a path that's on the loader search path. As a result, the link line of their `.pc` file doesn't include any RPATH. It will include something like `-lmylib` or `-L/path/to/mylib -lmylib`. This then works fine when the library is indeed in `/usr/lib/` or another standard location. 

However, if you build such a library yourself and put it elsewhere, an RPATH is needed, but not present at all. Meson isn't aware of what directories are or aren't in the loader search path, and hence it's not able to add (or avoid stripping) of RPATHs to such directories at install time. This is particularly confusing because Autotools does know how to do this (xref https://github.com/mesonbuild/meson/issues/13046#issuecomment-2089364514).

It's actually not uncommon for this to bite you in practice if your Python package depends on a shared library, because you'll have it in some custom directory as soon as you want to do any development on it (or put it into its own wheel, like we did with `scipy-openblas32`).

docs/how-to-guides/shared-libraries.rst
dnicolodi: I think we may be referring to two different things. `libdir` as defined earlier in this document is a system wide installation location and the location where meson would install a shared library. What these wheels do is to install in the `lib` directory of the current Python environment, which is not the same thing.

rgommers: Ah okay. Yeah not the same, although it does end up being the same if you're not using a virtualenv on top of your system Python.

Let me think about how to rephrase this.

docs/how-to-guides/shared-libraries.rst
dnicolodi: I understand the problem you are describing in your latest comment, but I am not able to make the connection from what this PR adds to the documentation and the problem you describe. The documentation added here, almost literally says "An RPATH entry alone is not always enough... if this is the case, add an RPATH" because setting LD_LIBRARY_PATH or the compiler flag just do that.

I now understand what you have in mind, but I think the explanation in the documentation is a bit too terse: I think I have an above average understanding of these issues and still I could not make the connection. I'll try to come up with a concrete suggestion on how to explain this.

docs/how-to-guides/shared-libraries.rst
eli-schwartz: It is definitely not at all the same thing. The issue isn't that wheels cannot install outside of site-packages. They absolutely can do that by brute force, as long as the location is still in sys.prefix. The issue is that even if you assume the garden path of sys.prefix being /use and wanting to install to a location within /usr...

... on some systems this is /usr/lib, and on other systems it is /usr/lib64, and on yet more systems it is /usr/lib/x86_64-linux-gnu

There are more configurations too, this is just the most common ones for Linux specifically.

If you have system-specific wheels then this all is entirely fine. It's not physically possible to design a wheel that is simultaneously installable on Arch Linux and on Debian and on Gentoo, without adding new features to the wheel standard (described in a PEP that was rejected years ago).

docs/how-to-guides/shared-libraries.rst
eli-schwartz: It's always enough to add your own RPATH by adding -Wl,-rpath=... to your personal LDFLAGS with the same value that you'd add to LD_LIBRARY_PATH, I guess.

The problem is not with rpath, it's with assuming that rpath will as of today be configured out of the box for you.

eli-schwartz: > hence it's not able to add (or avoid stripping) of RPATHs to such directories at install time

Once upon a long time ago meson stripped rpaths unless they were listed in the install_rpath in meson.

These days it fixed that, and now only strips rpaths that *are* listed exclusively in build_rpath (or point to locations inside the build tree).

So your note is not true for versions of meson that meson-python supports.

rgommers: > The problem is not with rpath, it's with assuming that rpath will as of today be configured out of the box for you.

Ah, that's a helpful way of phrasing it I think, thanks.

dnicolodi:
> IIRC what I did is for files that I copied from another test package to keep the year unchanged, and for files that I wrote from scratch I added 2024. Happy to do it some other way (2024 for everything?), but I did try to follow a consistent rule at least. No preference from my side - please let me know what you prefer here.

It makes sense. I don't have a strong opinion. I would just remove the copyright years, but IANAL. Things are good as they are. Thanks @rgommers

docs/how-to-guides/shared-libraries.rst
rgommers: I rephrased this section and make it a bit shorter as well - please take another look.

Also, I defined what `libdir` means right above its first usage, in the table.

docs/how-to-guides/shared-libraries.rst
rgommers: I rewrote this section - shorter now, and hopefully more correct and easier to understand.

tobiasdiez:
Thanks a lot for these very helpful and easy-to-understand instructions!

rgommers:
@dnicolodi the reviews comments were addressed a while back. WDYT about hitting the green button on this?

dnicolodi:
Sorry for being slow on this @rgommers. There are still a couple of todo items assigned to myself in the comments and I wanted to address them before merging, mostly because otherwise I'll forget to address them. I'll try to work on this today or over the weekend. If I'll not come to it I'll merge this as it is and we will iterate on it later.

docs/how-to-guides/shared-libraries.rst
```diff
@@ -0,0 +1,250 @@
...
+is being installed to its default system location (typically
+``/usr/local/lib`` on Unix-like systems, and ``C:\\lib`` on Windows - we call
+this ``libdir`` in this guide) or to a location in ``site-packages`` within the
+Python package install tree. All these scenarios are (or will be) supported,
+with some caveats:
+
++-----------------------+------------------+---------+-------+-------+
+| shared library source | install location | Windows | macOS | Linux |
++=======================+==================+=========+=======+=======+
+| internal              | libdir           | no (1)  | ✓     | ✓     |
```
dnicolodi, line 30: I think there is a very serious lack of use of emojis in our documentation 😄. I was thinking to use emojis for the check mark columns of this table ✔️ and 🚧 maybe

docs/how-to-guides/shared-libraries.rst
```diff
@@ -0,0 +1,250 @@
...
+nothing special for this case. Windows needs some special handling though, due to
+the lack of RPATH support:
+
+.. literalinclude:: ../../tests/packages/sharedlib-in-package/mypkg/__init__.py
+   :start-after: start-literalinclude
+   :end-before: end-literalinclude
+
+If an internal shared library is not only used as part of a Python package, but
+for example also as a regular shared library in a C/C++ project or as a
+standalone library, then the method shown above won't work. The library is`
```
dnicolodi, line 95:
```suggestion
standalone library, then the method shown above won't work. The library is
```

docs/how-to-guides/shared-libraries.rst
```diff
@@ -0,0 +1,250 @@
...
+
+1: Internal shared libraries on Windows cannot be automatically handled
+correctly, and currently ``meson-python`` therefore raises an error for them.
+`PR meson-python#551 <https://github.com/mesonbuild/meson-python/pull/551>`__
+may improve that situation in the near future.
+
+2: External shared libraries require ``delvewheel`` usage on Windows (or
+some equivalent way, like amending the DLL search path to include the directory
+in which the external shared library is located). Due to the lack of RPATH
+support on Windows, there is no good way around this.
```
dnicolodi, line 47: Should this be formatted as an enumerated list?

docs/how-to-guides/shared-libraries.rst
```diff
@@ -0,0 +1,250 @@
...
+.. TODO: add subproject as a source
+
+1: Internal shared libraries on Windows cannot be automatically handled
+correctly, and currently ``meson-python`` therefore raises an error for them.
+`PR meson-python#551 <https://github.com/mesonbuild/meson-python/pull/551>`__
+may improve that situation in the near future.
+
+2: External shared libraries require ``delvewheel`` usage on Windows (or
+some equivalent way, like amending the DLL search path to include the directory
+in which the external shared library is located). Due to the lack of RPATH
```
dnicolodi, line 46: Should we assume that readers are familiar with the concept of RPATH? It may be something unknown to users working predominantly on Windows.

tests/packages/sharedlib-in-package/mypkg/__init__.py
```diff
@@ -0,0 +1,51 @@
+# SPDX-FileCopyrightText: 2024 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+import os
+import sys
+
+
+# start-literalinclude
+def _enable_sharedlib_loading():
```
dnicolodi, line 10: I find the function name a bit misleading. What the function does is to add a path to the list of paths where shared libraries are searched. The capability of loading shared library loading is unaffected. I would name this `_append_to_sharedlib_load_path()` or something like this.

tests/packages/sharedlib-in-package/mypkg/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+    ``__init__.py`` of a package when it vendors external shared libraries.
+
+    .. note::
+
+        `os.add_dll_directory` is only available for Python >=3.8, and with
+        the Conda ``python`` packages only works as advertised for >=3.10.
+        If you require support for older versions, pre-loading the DLL
+        with `ctypes.WinDLL` may be preferred (the SciPy code base has an
+        example of this).
+    """
```
dnicolodi, line 35:
```suggestion
    """
    Ensure the shared libraries in this package loaded on Windows.
    
    Windows lacks a concept equivalent to RPATH: Python extension modules cannot 
    find DLLs installed outside the DLL search path. This function ensured that the 
    location of the shared libraries distributed with this Python wheel is in the DLL
    search path of the process.
    
    The Windows DLL search path includes the object depending on it is located:
    the DLL search path needs to be augmented only when the Python extension
    modules and the DLLs they require are installed in separate directories.
    Cygwin does not have the same default library search path: all locations where
    the shared libraries are installed need to be added to the search path.

    This function is very similar to the snippet inserted into the main ``__init__.py`` of a
    package by ``delvewheel`` when it vendors external shared libraries.

    .. note::

        `os.add_dll_directory` is only available for Python 3.8 and later, and in the 
        Conda ``python`` packages it works as advertised only  for version 3.10 and later.
        For older Python versions, pre-loading the DLLs with `ctypes.WinDLL` may be preferred.
    """
```

I find this a bit clearer. The line breaks are to be adjusted.

dnicolodi:
I've added a few suggestions in the places that I thought needed some little work. @rgommers please have a look.

docs/how-to-guides/shared-libraries.rst
rgommers: They may not be. I don't really feel like explaining in our docs, but maybe a link to Wikipedia or another source is in order?

docs/how-to-guides/shared-libraries.rst
```diff
@@ -0,0 +1,250 @@
...
+RPATH entry - and Meson will not automatically manage RPATH entries for you.
+Hence you'll need to add the needed RPATH yourself, for example by adding
+``-Wl,rpath=/path/to/dir/sharedlib/is/in`` to ``LDFLAGS`` before starting
+the build. In case you run into this problem after a wheel is built and
+installed, adding that same path to ``LD_LIBRARY_PATH`` is a quick way of
+checking if that is indeed the problem.
+
+On Windows, the solution is similar - the shared library can either be
+preloaded, or the directory that the library is in added to ``PATH`` or with
+``os.add_dll_directory``, or vendored into the wheel with ``delvewheel`` in
```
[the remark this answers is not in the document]
rgommers: Good catch. It used to be true, but no longer is I believe. I'll verify and fix.

tests/packages/sharedlib-in-package/mypkg/__init__.py
rgommers: SGTM, will change to that

rgommers: Adopted this function name, resolving.

docs/how-to-guides/shared-libraries.rst
rgommers: Reformatted, resolving.

docs/how-to-guides/shared-libraries.rst
rgommers: I gave that a try, looks reasonable. The ✔️ didn't work because it doesn't play well with the dark theme, but 🗹 did.

Kept it in a separate commit so it's easily reversible. Messing with table formatting is a bit fiddly.

dnicolodi: It looks fine with the GitHub dark mode. Is it an issue with the Sphynx theme? This was just a silly suggestion, the table looks fine as it was without the emojis and we can keep it like that if it makes editing it cumbersome.

rgommers:
A good descriptive name is not so easy. Ideally the name should make clear it's a boolean and allows meson-python to fold _internal_ shared libraries targeted for install _outside_ site-packages into the wheel on _Windows_. That's too much info though. Although `shared-libs-win32` can probably be improved on. Perhaps something like `allow-windows-internal-sharedlib`? (I don't like `win32` much, since in wheel tags it's specific to 32-bit Python).

docs/how-to-guides/shared-libraries.rst
rgommers: Probably yes. Just that emoji didn't seem to change color, while the other one did.

I'm happy to keep the commit, it's slightly nicer I'd say. Now that I've manually realigned the table it should be fine.

docs/how-to-guides/shared-libraries.rst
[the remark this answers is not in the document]
rgommers: Rephrased now. This is good to go I think.

docs/how-to-guides/shared-libraries.rst
dnicolodi: I build the docs locally and on macOS 12.7 the unicode character you choose is not shown correctly. It looks like the glyph is missing in the font used by the Sphinx theme. Or something like this. I'm sure we can fix it, somehow, but I'm not sure it is worth the effort. Maybe we should go back to the simpler symbols in the table, at least until someone does not figure out what's wrong with emojis and the Furo theme.

tests/packages/sharedlib-in-package/mypkg/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+        the Conda ``python`` packages it works as advertised only for version
+        3.10 and later. For older Python versions, pre-loading the DLLs with
+        `ctypes.WinDLL` may be preferred.
+    """
+    basedir = os.path.dirname(__file__)
+    subdir = os.path.join(basedir, 'sub')
+    if os.name == 'nt':
+        os.add_dll_directory(subdir)
+    elif sys.platform == 'cygwin':
+        os.environ['PATH'] = f'os.environ["PATH"]:{basedir}:{subdir}'
```
dnicolodi, line 40: I think this should be
```suggestion
        os.environ['PATH'] = f'{os.environ["PATH"]}:{basedir}:{subdir}'
```

docs/how-to-guides/shared-libraries.rst
rgommers: > Maybe we should go back to the simpler symbols in the table,

Agreed, let me drop the last commit.

tests/packages/sharedlib-in-package/mypkg/__init__.py
dnicolodi, line 40: or maybe
```suggestion
        os.environ['PATH'] = ':'.join((os.environ['PATH'], basedir, subdir))
```

dnicolodi: Assuming `os.pathsep` is `:` and not `;` on cygwin.

tests/packages/sharedlib-in-package/mypkg/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+    This function is very similar to the snippet inserted into the main ``__init__.py``
+    of a package by ``delvewheel`` when it vendors external shared libraries.
+
+    .. note::
+
+        `os.add_dll_directory` is only available for Python 3.8 and later, and in
+        the Conda ``python`` packages it works as advertised only for version
+        3.10 and later. For older Python versions, pre-loading the DLLs with
+        `ctypes.WinDLL` may be preferred.
+    """
```
dnicolodi, line 34: If you fix the small issue below, maybe you can also reflow the docstring. It looks untidy now because of the lines with wildly different length. Line wrapping at 76 characters works well visually, but anything between 72 and 80 is fine.

tests/packages/sharedlib-in-package/mypkg/__init__.py
rgommers: Thanks. I went with this one - CI will tell us if Cygwin is still happy.

tests/packages/sharedlib-in-package/mypkg/__init__.py
rgommers: done at 80, which is my standard editor setting

dnicolodi:
I've rebased to kick cirrus CI into behaving

dnicolodi (approved):

PR #701, opened 2024-10-28
MAINT: bump minimum required Meson version to 0.64.0

dnicolodi:
Depending on 0.64.0 or later allows to use the pure argument when invoking the import('python').find_installation() Meson function, which makes for some nicer test and example code. Meson 0.64.0 was released on November 6th, 2022, just over a month after the current minimum supported version.

### DIFF
--- a/docs/reference/meson-compatibility.rst
+++ b/docs/reference/meson-compatibility.rst
@@ -26,6 +26,12 @@ versions.
    and similar installation functions. On older Meson versions, these
    arguments have no effect.
 
+.. option:: 1.2.0
+
+   Meson 1.2.0 or later is required to support the ``--skip-subprojects``
+   option that can be passed to ``meson install`` to not include files
+   installed by some or all subprojects in the Python wheel.
+
 .. option:: 1.2.3
 
    Meson 1.2.3 or later is required to support Python 3.12 or later.
--- a/docs/reference/meson-compatibility.rst
+++ b/docs/reference/meson-compatibility.rst
@@ -15,9 +15,9 @@ maintain compatibility with as old as possible Meson releases.
 However, some functionality is available only with more recent Meson
 versions.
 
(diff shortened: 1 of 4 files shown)


docs/reference/meson-compatibility.rst
```diff
@@ -26,6 +26,13 @@ versions.
    and similar installation functions. On older Meson versions, these
    arguments have no effect.
 
+.. option:: 1.2.0
+
+   Meson 1.2.0 or later is required to support the
+   ``--skip-subprojects`` option that can be passed to ``meson
+   install`` to do not include in the Python wheel files installed by
+   some or all subprojects.
```
rgommers, line 34: These last two lines have a small grammar issue ("to do not" -> "to not") and I'd put "in the Python wheel" at the end of the sentence.

rgommers (commented):
Thanks for the PR, LGTM!

There's one mention of 0.64 in the code base, maybe you also want to update this (either way is fine with me)?
https://github.com/mesonbuild/meson-python/blob/9faca831e794879040c87cff529e0fecd2a98b01/tests/packages/subdirs/meson.build#L9-L10

This change is nice because it enables us to use `python.find_installation(pure: false)` as discussed in gh-700.

dnicolodi:
> There's one mention of 0.64 in the code base, maybe you also want to update this (either way is fine with me)?

Thanks, I didn't find this reference. I'll update this test too.

docs/reference/meson-compatibility.rst
dnicolodi: > These last two lines have a small grammar issue ("to do not" -> "to not")

This is something that somehow got stuck wired in the wrong in my brain. I write it wrong all the time...

rgommers (approved):
LGTM now, in it goes.

PR #719, opened 2025-01-27
MAINT: put back mesonpy.__version__

dnicolodi:
The module constant has been used to check meson-python version in other projects, see discussion on #691.  Add a test to ensure that the version recorded in `pyproject.toml` and `mesonpy/__init__.py` are kept in sync.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -41,9 +41,9 @@ debian-unstable_task:
     cpu: 1
   << : *test
 
-fedora-37_task:
+fedora-41_task:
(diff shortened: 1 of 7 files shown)


dnicolodi:
I would really like to know why CirrusCI gets stuck sometimes until the docker images are refreshed

rgommers (approved):
Thanks for taking care of this @dnicolodi!

ianthomas23:
Thanks for putting this back in.

Issue #711, opened 2024-12-06
RPATH goes missing when using both `install_rpath` and an internal shared library dependency


rgommers:
This is a case that probably hasn't been exercised before: a Python extension module that links against both a shared library that's part of the package _and_ a shared library built in a subproject that's folded into the wheel by `meson-python.

I ran into this in SciPy when adding a subproject. This worked as advertised, except for [this `scipy.special._ufuncs`](
https://github.com/scipy/scipy/blob/dd08de8c9b0e2726f5688a8b76a699c1f28e6514/scipy/special/meson.build#L144-L162) extension module, which depends on [this `libsf_error` shared library](https://github.com/scipy/scipy/blob/dd08de8c9b0e2726f5688a8b76a699c1f28e6514/scipy/special/meson.build#L36-L42) as well as on the library in the subproject.
What one sees then is that the installed wheel can no longer find `libsf_error.so`, because the RPATH entry for `install_rpath: '$ORIGIN'` goes missing.

With a default SciPy build (no subproject), the `Library rpath|runpath` entries on the `_ufuncs` target in the build and install directories looks like this (has `$ORIGIN` as it should):
```bash
$ readelf -d build/scipy/special/_ufuncs.cpython-312-x86_64-linux-gnu.so | rg "Library r"
 0x000000000000000f (RPATH)              Library rpath: [/home/rgommers/mambaforge/envs/scipy-dev-py312/lib:$ORIGIN/]

$ readelf -d build-install/lib/python3.12/site-packages/scipy/special/_ufuncs.cpython-312-x86_64-linux-gnu.so | rg "Library r"
 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN:/home/rgommers/mambaforge/envs/scipy-dev-py312/lib]
```

On the branch with the subproject, we see this instead:
```bash
$ readelf -d build-whl/scipy/special/_ufuncs.cpython-312-x86_64-linux-gnu.so | rg Library
 0x000000000000001d (RUNPATH)            Library runpath: [/home/rgommers/code/pixi-dev-scipystack/scipy/.pixi/envs/openblas-src/lib:$ORIGIN/../../.scipy.mesonpy.libs:$ORIGIN/../../.scipy.mesonpy.libs:$ORIGIN/../../.scipy.mesonpy.libs:$ORIGIN/../../.scipy.mesonpy.libs]

[...]

PR #724, opened 2025-02-13
Fix support for `install_rpath`

dnicolodi:


tests/packages/sharedlib-in-package/mypkg/meson.build
```diff
@@ -32,7 +32,7 @@ py.extension_module(
     include_directories: 'sub',
     install: true,
     subdir: 'mypkg',
-    install_rpath: '$ORIGIN',
+    install_rpath: build_machine.system() == 'darwin' ? '@loader_path' : '$ORIGIN',
```
dnicolodi, line 35: So... This test was working before just because of a bug: we did no strip the RPATH pointing to the build directory added by meson when the package does not contain libraries relocated to `.<package-name>.mesonpy.libs`. Now we do and the test fails. The fix above is required because AFAICT meson does not substitures `@loader_path` for `$ORIGIN` on macOS. However, it is only half the fix.

The test installs one library in `site-packages/mypkg/` and one in `site-packages/mypkg/sub/`. An RPATH to both of these directories would need to be added. However, meson does not provide a mechanism to set `install_rpath` to a list of strings.

I don't know how to fix this other than extending meson to accept a list for `install_rpath` or to simply change the test.

dnicolodi: @rgommers You added the test. Which use case did you have in mind?

rgommers: > The test installs one library in `site-packages/mypkg/` and one in `site-packages/mypkg/sub/`. An RPATH to both of these directories would need to be added. However, meson does not provide a mechanism to set `install_rpath` to a list of strings.

This was helpful to smoke out the details of Windows vs. Cygwin, namely this sentence in the docs:
```
The Windows DLL search path includes the path to the object attempting
to load the DLL: it needs to be augmented only when the Python extension
modules and the DLLs they require are installed in separate directories.
```

Now that it's done though, the test case isn't essential. I suggest that we remove the shared library right next to the extension, and keep the one in `sub`.

rgommers: > The fix above is required because AFAICT meson does not substitures `@loader_path` for `$ORIGIN` on macOS.

Hmm, something seems wrong there, `$ORIGIN` really should be made to work without this hack. At least SciPy releases depend on this already, and it's conceptually right. I don't fully understand this yet, but it seems if there's something to fix, it should be done in Meson in a backwards-compatible way.

dnicolodi: When building a wheel, `install_rpath` is handled by meson-python, thus we can easily replace `$ORIGIN` with `@loader_path`. However, I think it is better to keep handling of `install_rpath` consistent between meson and meson-python. I haven't tested it, but the documentation does not mention it, and reading the code I didn't find anything in meson that does the translation, therefore I didn't implement it for meson-python either.

What do you mean when you say that SciPy depend on this? `install_rpath` did not do anything when building a wheel before this PR.

dnicolodi: > I suggest that we remove the shared library right next to the extension, and keep the one in `sub`.

It is a bit more convoluted to do, but one think that could be done to keep the test coverage on Windows is to make one library depend on the other and set `install_rpath` on the libraries accordingly. I'll look into how complex this gets.

rgommers: > What do you mean when you say that SciPy depend on this? `install_rpath` did not do anything when building a wheel before this PR.

https://github.com/scipy/scipy/blob/e70ed9c96bdbc75aaf0fb5060b3cdaa8b0f6abdc/scipy/special/meson.build#L87

It certainly was necessary to use `install_rpath: '$ORIGIN'`. See also https://labs.quansight.org/blog/libsf-error-state (search for `$ORIGIN`).

You're saying that `$ORIGIN` doesn't work on macOS I think - how sure are you of that? It does not sound correct to me, IIRC `@loader_path` and `@rpath` are different, and the latter should be understanding `$ORIGIN`.

> However, I think it is better to keep handling of `install_rpath` consistent between meson and meson-python.

Yes, agreed.

dnicolodi: I was just looking at that commit. I don't understand how that may work: `install_rpath` is handled by meson when `meson install` is called and meson-python does not do that...

Thinking about it, it works for the same reason that the test in meson-python worked before and it needs to be adjusted now: meson-python did not remove the rpath entries pointing to the build directory added by meson unless the package contains at least one library relocated to `.<package-name>.mesonpy.libs`. Therefore the SciPy modules never had their RPATH fixed and contained the build RPATH entry pointing to `$ORIGIN` or `@loader_path` depending on the platform. This PR corrects this and exposes the problem.

> You're saying that `$ORIGIN` doesn't work on macOS I think - how sure are you of that? It does not sound correct to me, IIRC `@loader_path` and `@rpath` are different, and the latter should be understanding `$ORIGIN`.

I don't understand. `@rpath` is not recognized as a special root handle by the dynamic linker. It can be used in the identification name of a dynamic shared library. It is not another field in the search path where `$ORIGIN` can be substituted.

dnicolodi: I think the best way to deal with `$ORIGIN` vs `@loader_path` is to do the substitution in meson-python and emit a warning about it. This would avoid breaking SciPy and possibly other packages while bringing users to do the right thing.

dnicolodi:
This should fix all identified issue and it is ready for review. The failed tests are due to https://github.com/mesonbuild/meson/issues/14254

rgommers:
@dnicolodi sorry I'm struggling to find the time to think about this more thoroughly right now. This is fixing a pre-existing bug that's been there for a long time, right? And nothing is urgent (I can work around gh-711 for the time being). If so, wouldn't it be better to defer this PR until after 0.18.0?

dnicolodi:
I don't see a problem with postponing this. On the other hand, I don't see any reason why we need to release 0.18.0 now. We can merge this for the next release or we can release 0.18.0 once this is merged. It is up to you.

dnicolodi:
> what is the status of this PR?

It is not merged yet. What else would you like to know?

> Is `install_rpath` handled properly now?

This PR is not merged yet, thus, no, it is not handled in the latest release of meson-python.

dnicolodi:
@rgommers I'm inclined to merge this soon. Do you still want to have a better look at it before?

rgommers:
> doing some final testing now - but I expect to find nothing

Okay, that was a bit too optimistic.

Started testing with `scipy` 1.15.1, as the last released version to use an internal shared library. In all cases, the `intro-install_plan.json` entry contains `"install_rpath": "$ORIGIN"`.

Linux, `meson` 1.7.1, `meson-python` main branch:
```
$ python -m build -wnx -Cbuild-dir=build
$ cd dist
$ unzip scipy-1.15.1-cp313-cp313-linux_x86_64.whl
$ readelf -d scipy/special/_ufuncs.cpython-313-x86_64-linux-gnu.so | rg "rpath|runpath"
 0x000000000000000f (RPATH)              Library rpath: [/home/rgommers/mambaforge/envs/scipy-dev/lib:$ORIGIN/]
```

Linux, `meson` 1.7.1, `meson-python` from this PR:
```
$ python -m build -wnx -Cbuild-dir=build
$ cd dist
$ unzip scipy-1.15.1-cp313-cp313-linux_x86_64.whl
$ readelf -d scipy/special/_ufuncs.cpython-313-x86_64-linux-gnu.so | rg "rpath|runpath"
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN]
```

`LDFLAGS` contains ` -Wl,-rpath`:
```
$ echo $LDFLAGS   # from conda-forge compiler activation
-Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,--allow-shlib-undefined -Wl,-rpath,/home/rgommers/mambaforge/envs/scipy-dev/lib -Wl,-rpath-link,/home/rgommers/mambaforge/envs/scipy-dev/lib -L/home/rgommers/mambaforge/envs/scipy-dev/lib
```
So the behavior on `main` is correct and this PR breaks usage of `-Wl,-rpath`. As can be seen further from `ldd` output; no libraries are found from the conda-forge environment's libdir:
```
$ ldd scipy/special/_ufuncs.cpython-313-x86_64-linux-gnu.so 
        linux-vdso.so.1 (0x00007d47fdcea000)
        libsf_error_state.so => /home/rgommers/code/bldscipy/dist/scipy/special/libsf_error_state.so (0x00007d47fdbf0000)
        libopenblas.so.0 => not found
        libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x00007d47fd800000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007d47fdad0000)
        libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007d47fdaa2000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007d47fd60e000)
        /usr/lib64/ld-linux-x86-64.so.2 (0x00007d47fdcec000)
```

We do have a test case for `-Wl,-rpath`, but you removed it in this PR. That has to be incorrect; Conda, Nix, Spack at least rely on setting RPATHs explicitly, and those have to be preserved.

tests/packages/link-against-local-lib/meson.build
```diff
@@ -9,7 +9,7 @@ if meson.get_compiler('c').get_id() in ['msvc', 'clang-cl', 'intel-cl']
     link_args = ['-DEXAMPLE_DLL_IMPORTS']
 else
     lib_compile_args = []
-    link_args = ['-Wl,-rpath,custom-rpath']
```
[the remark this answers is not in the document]
rgommers: I see a comment in the commit message:

> Drop test checking correct handling of RPATH set via linker arguments: modern Meson highly discourages doing this.

Do you have a reference for this? I think that if this is really the case right now, that is misinformed - this is pretty essential for some distros.

tests/packages/sharedlib-in-package/mypkg/meson.build
rgommers: Could we land these `sharedlib-in-package` test package changes separately? This PR is very tricky, and making major changes to both tests and implementation at the same time makes it quite easy for regressions to slip in. None of the tests are actually wrong, but if cleaning up this test a bit helps then sure let's do it.

tests/packages/link-against-local-lib/meson.build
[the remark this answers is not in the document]
rgommers: I see the failures in gh-768, looking into them now.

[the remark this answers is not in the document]
rgommers: If we do that in a separate PR, it will unblock gh-768 it looks like. Want me to do this?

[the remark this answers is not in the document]
dnicolodi: > This test case needs to stay, it's important that this works.

This is working. I'll drop this commit from the PR.

eli-schwartz:
IIRC meson treats install_rpath as additive, not a replacement. The other angle is that it removes build_rpath -- and tracks various internally added rpaths needed for `meson test` and `meson devenv` as build rpaths.

So maybe the solution here is to treat the introspection files as incomplete and extend it with even more info?

eli-schwartz:
> If there is no other way to know what the build-rpath(s) is/are, yes maybe. I'm not yet fully convinced that we cannot know - we know in `meson-python` what the build directory is after all, so we may be able to compute this somehow from that build dir + `install_rpath`? Even if there are multiple, there is no way `-Wl,-rpath` should be pointing to within the build dir.

That might work, sure. Though FWIW internally meson does have `target.rpath_dirs_to_remove` inside the backend build data, and that's what `meson install` uses.

dnicolodi:
> What I pointed out is that this PR seems to break adding an RPATH with `LDFLAGS=-Wl,-rpath,/some/path`, which will break conda-forge et al.

This is true only of the added RATH is anchored to `$ORIGIN`.

tests/packages/sharedlib-in-package/mypkg/meson.build
dnicolodi: The tests are actually wrong and they work just because meson-python fails to clean up the RPATH entries added by meson pointing to the build directory. This PR fixes this and the tests start to fail.

dnicolodi: > It just occurred to me after writing this comment that this is verifiable, by building the test package as a wheel, then installing it and manually checking that the imports work (the build dir will no longer exist).

The build RPATH entries added by Meson are relative to `$ORIGIN` and do not contain any absolute reference to the build directory. I think I've stated this multiple times now, and there is a long comment in one of the commits explaining this. Therefore, the imports fail only if the layout in the build directory is different from the layout in the in the install location. This is not the case for the current tests.

dnicolodi: > The tests are for case (1). You can't argue that tests for (1) are wrong because you're trying to fix case (2).

This PR touches different test packages. Some check for (1) and others check for (1) and (2), but succeed by accident. Some use `install_rpath` as they should (althooug, prior to this PR it has no effect). Some other rely on the bogus meson-python behavior. The tests that use `install_rpath` are anyhow relying on the bogus meson-python behavior because they use RPATH entries relative to `$ORIGIN`, which has no meaning on macOS and prior to this patch there is not provision in meson-python to translate `$ORIGIN` to `@loader_path` on macOS.

As I explained, this PR assumes that all RPATH entries already present before applying `install_rpath` that are relative to `$ORIGIN` (or `@loader_path` on macOS) are Meson build RPATH entries and must be removed. This is the reason why RPATH entries added with linker arguments (either via `$LDFLAGS` or via linker arguemts in `meson.build` and that are relative to `$ORIGIN` (or `@loader_path` on macOS) are removed. This is what needs fixing. I have a patch ready. The correct fix depends on https://github.com/mesonbuild/meson/pull/14819

> I know `install_rpath: '$ORIGIN'` was necessary and worked for SciPy

I don't know why you feel it is necessary to state this. I am nowhere questioning that adding RPATH entries is ncessary. However, before this PR, there is nothing in meson-python that applies `install_rpath: '$ORIGIN'` thus I cannot believe it makes a difference, and indeed, I don't find any `install_rpath` in the whole SciPy source code.

dnicolodi:
Superseded by #788. Closing.

Issue #673, opened 2024-09-30
Subproject with static libraries fails the build in 0.13.0 onward (but not in 0.12.0)


blochl:
When building with a subproject which provides both static and dynamic libraries, I get the following error with 0.13.0 onward:
```
meson-python: error: Could not map installation path to an equivalent wheel directory: '{libdir_static}/mylibrary.a'
```
With meson-python 0.12.0 everything works fine.

Looking at #35, I understand that there is no location in wheels for static libraries, and they aren't needed at runtime, but for me `--skip-subprojects` install option is not a solution, as I do need the dynamic libraries which this subproject installs. What can I do (besides just using version 0.12.0, where it works :smile: )?

rgommers:
Shared libraries also have nowhere to go in a wheel, so you'll run into that same problem. I think what you want is only use the dependency from the subproject, e.g. `libsimple_dep` from the examples in https://mesonbuild.com/Subprojects.html.

I'll note that a static library is a lot easier to deal with in a wheel. If you want to use a shared library, there may be issues (I actually don't know how to do it, it kinda requires running `auditwheel` & co before returning control to the build frontend).

blochl:
@rgommers , thanks for your reply!
1. But in the subproject the libraries' dependency is specified with `install : true`, and I can't change that (not my code)...
2. The shared libraries have to go in a wheel - it's a project which uses Python C extensions. And it works great with 0.12.0 - the wheel contains the shared libs, and installs them just fine. How do you suggest to overcome this?

blochl:
@rgommers ,
Just checked: the shared libs go into the wheel and are installed just fine in all versions, including the latest (0.16.0).

rgommers:
> Just checked: the shared libs go into the wheel and are installed just fine in all versions,

I think they go under the data directory right, and then get installed to `<prefix>/lib/`? That's not a good default for redistributable packages (since it risks clobbering things in other virtualenv or in `/usr/lib/`), but I do agree that shared and static libraries should be treated the same way here.

Is your package public perhaps? We're lacking a test package for this config right now unfortunately - we should add that, with more docs, since this isn't the first question about subprojects with shared/static libraries. 

> 1. But in the subproject the libraries' dependency is specified with `install : true`, and I can't change that (not my code)...

[...]

PR #732, opened 2025-04-22
Add libdir static

edtanous:
Some packages with static libraries report an error.

meson-python: error: Could not map installation path to an equivalent
wheel directory: '{libdir_static}/libjson-c.a'

Possibly related to https://github.com/mesonbuild/meson/issues/12846

Example of this error occurring is here:
https://github.com/edtanous/libcper/tree/pypi

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -106,6 +106,7 @@ _INSTALLATION_PATH_MAP = {
     # custom location
     '{libdir}': 'mesonpy-libs',
     '{libdir_shared}': 'mesonpy-libs',
+    '{libdir_static}': 'mesonpy-libs',
 }
 
 


edtanous:
also possibly related to #673

dnicolodi:
Not mapping `{libdir_static}` to a location in the Python wheel is a design decision. We can revisit this, however, we would need a good reason for doing so. What is the reason to include a static library into a Python wheel? In all cases I've seen so far, the proper solution is to do not include the static library in the wheel, either disabling installation of the static library or not selecting the static library component for installation via arguments to `meson install`.

rgommers:
> In all cases I've seen so far, the proper solution is to do not include the static library in the wheel, either disabling installation of the static library or not selecting the static library component for installation via arguments to `meson install`.

NumPy ships a static library (two actually). If that's the intent for any Python package, the solution is to explicitly pick an install location without the package structure:

https://github.com/numpy/numpy/blob/9ead596ce4f8df0189f9ba3d54937e22e2785a5e/numpy/_core/meson.build#L577-L584

Which is also what the last comment on the linked issue says: https://github.com/mesonbuild/meson/issues/12846#issuecomment-2276648655. That way it is a fixed location that the package authors can actually advertise to end users, so there is then a point in shipping the wheel (putting it in an auto-generated location like this PR does won't achieve that).

@edtanous is that the right fix for you? If not, why not?

dnicolodi:
> NumPy ships a static library (two actually). If that's the intent for any Python package

I expect the number of packages where shipping a static library makes sense to be counted on one ~~hand~~ finger 🙂

edtanous:
Then maybe there's another bug/feature.  The intent in my case is not to _ship_ a static library, but this option seemed to be required regardless of whether the subproject was installed or not.  Because of that at least in my testing I needed a mapping very similar to #673 to get anything to compile.

I can try the numpy example, but it would be really nice if this "just worked" without having to manually modify the meson file with wheel install specifics. In my case, I do not get *.a in the final wheel, but this mapping was required to make the build run at all.

edtanous:
```
install = ['--skip-subprojects']
```

Seems to have done the trick here.  It's not clear why it wasn't working before, but it seems to work now.  Would be nice if there was a better warning message that conveyed "meson-python doesn't support implicitly installing static libraries." rather than an error around a path lookup.  If I changed this patch to add a better error, would that be reasonable?

dnicolodi:
The current error message is very descriptive of what is going on:
```
meson-python: error: Could not map installation path to an equivalent wheel directory: '{libdir_static}/libjson-c.a'
```
Maybe it requires knowing how meson-python operates to grok it fully, but I don't know how to rephrase it in different terms that are at the same time technically correct, concise, and informative enough to allow the user to fix the problem.

meson-python maps installations paths for the build artifacts of the Meson project to wheel install locations. Some Meson installation paths can be directly mapped to wheel install locations, some require some extra works (like `{libdir_shared}`), other (like `{libdir_static}`) do not make sense as there is not an equivalent concept in Python wheels. However, there is no guarantee that what is installed in `{libdir_static}` are static libraries: anything can be installed at that prefix (like there is no guarantee that what is in `{libdir_shared}` is a shared librsry, or that what is in `{py_purelib}` is a Python module). Therefore, the only correct message that meson-python can provide is "I don not know what to do of what would be installed in {libdir_shared}". Such an error message is however too vague for the user to be able to fix the problem. What is necessary to solve the problem is identifying what exactly would be installed in `{libdir_shared}`. The best way to do that is to list the exact installation path. And this is what the current error message does.

dnicolodi:
> ```
> install = ['--skip-subprojects']
> ```
> Seems to have done the trick here.

It seems that the use case that this PR was aiming to support already works as expected. Closing.

edtanous:
> Maybe it requires knowing how meson-python operates to grok it fully

I think the difference is that the message phrased it in terms of the internal data structures and specifics around how mapping was done, not that the operation that was being done (adding a static library) was something that was an explicit design decision to not support.  I had made a suggestion earlier for a log message, but it sounds like what's there now is preferred, so that's fine, I'll let it go.

I appreciate the discussion, and thank you for supporting a project that was useful to me.

rgommers:
> Maybe it requires knowing how meson-python operates to grok it fully, but I don't know how to rephrase it in different terms that are at the same time technically correct, concise, and informative enough to allow the user to fix the problem.

Possibly we could keep what is there now, and add something more below, e.g., a blank line and then mention `--skip-subprojects` as a possible solution, or whatever else would have helped @edtanous?

edtanous:
skip-subprojects wasn't fully what I needed, it also needed a build that didn't install things like the packageconfig (which this also more rightly barfs on).  I can try to bake up a better error message.

eli-schwartz:
Possibly relevant is my port of this C library + python bindings: https://github.com/dgibson/dtc/commit/ecb21febfdd32fa9e0cbd9517f9e92cc9c703ffb#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R24

rgommers:
Better error messages are always very welcome!

> Possibly relevant is my port of this C library + python bindings

Nice example. I've needed a single `wheel-only` flag as well (because it's kind of a special case), IIRC in one of the SciPy subproject build configs.

PR #734, opened 2025-04-24
MAINT: remove nox configuration

dnicolodi:
nox is not used by any of the current maintainers and its configuration is becoming outdated. Instead of trying to keep it updated without any effective testing, remove it.


rgommers:
thanks!

PR #735, opened 2025-04-24
MAINT: bump minimum packaging version to 23.2

### LINKED ISSUE #733: Incorrect lower bound on `packaging` versions supported
I found this because a SciPy CI run on a runner image that already had `packaging` 19.1 installed was failing with the latest `meson-python` (0.17.1), and manually upgrading `packaging` fixed it. xref https://github.com/scipy/scipy/pull/22840#discussion_r2057110457

In CI we tests with old versions of `meson`, but not old versions of other dependencies. We do an unconditional `import packaging.metadata` in `conftest.py`, which required `packaging` 23.1 to work.

The SciPy job failed with:
```
TypeError: canonicalize_name() got an unexpected keyword argument 'validate' ;
```
which is caused by this except clause not working with older `packaging`:

https://github.com/pypa/pyproject-metadata/blob/92752c624bf2c3be3cf934525abe67e0650c7f3e/pyproject_metadata/__init__.py#L514-L516

That was fixed already in https://github.com/pypa/pyproject-metadata/pull/239, but isn't part of the latest release (`pyproject-metadata` 0.9.1).

I think what we need here is:
1. A new `pyproject-metadata` release
2. Bump our minimum required `packaging` version to 23.2 (from October 1, 2023)
3. Add a CI job that tests the minimum versions of all our dependencies.

@dnicolodi do you agree that that's the way to go here?

EDIT @dnicolodi: fixed `pypackaging-metadata` to `pyproject-metadata`

dnicolodi:
pyproject-metadata 0.9.1 requires packaging 23.2 or later, but does not record this dependency, see https://github.com/pypa/pyproject-metadata/pull/239

This becomes a problem when meson-python is installed on a platform where packaging 19.0 (the current minimum required version) is installed. This does not trigger an update of packaging to the latest available version and results in run time errors, see https://github.com/scipy/scipy/pull/22840#issuecomment-2817494211

This could be fixed by a new release of pyproject-metadata, but we do not want meson-python to depend on very recent releases. meson-python supports pyproject-metadata 0.9.0 or later and still carries some work-arounds for issues in that release.

The least invasive and most effective solution is to bump the packaging requirement for meson-python to 23.2. This also fixes the (intentional) use of API available only with packaging 23.2 or later in the meson-python tests.

Fixes #733.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ backend-path = ['.']
 requires = [
   'meson >= 0.64.0; python_version < "3.12"',
   'meson >= 1.2.3; python_version >= "3.12"',
-  'packaging >= 19.0',
+  'packaging >= 23.2',
   'pyproject-metadata >= 0.9.0',
   'tomli >= 1.0.0; python_version < "3.11"',
 ]
@@ -36,7 +36,7 @@ classifiers = [
 dependencies = [
   'meson >= 0.64.0; python_version < "3.12"',
   'meson >= 1.2.3; python_version >= "3.12"',
-  'packaging >= 19.0',
+  'packaging >= 23.2',
   'pyproject-metadata >= 0.9.0',
   'tomli >= 1.0.0; python_version < "3.11"',
 ]
@@ -44,7 +44,6 @@ dependencies = [
 [project.optional-dependencies]
 test = [
   'build',
-  'packaging >= 23.1',
   'pytest >= 6.0',
   'pytest-cov[toml]',
   'pytest-mock',


rgommers (approved):
LGTM, thanks!

rgommers:
Oh actually, there's a test dependency (`'packaging >= 23.1'`) further down in `pyproject.toml` that can be removed.

dnicolodi:
I knew the use of the newer packaging API was intentional in the tests, but I forgot that we actually recorded that in the test dependencies. I'll remove it. Maybe it is also time to use a dependency group for the test dependencies instead of extra dependencies. Both `pip` and `uv` should have support for dependency groups nowadays.

dnicolodi:
Turns out that a release of pip that supports dependency groups is scheduled for tomorrow, so we are not there yet. I don't think that switching to uv just for this is worth the trouble. The conversion to dependency groups will have to wait a tiny bit more.

rgommers:
> Turns out that a release of pip that supports dependency groups is scheduled for tomorrow, so we are not there yet. I don't think that switching to uv just for this is worth the trouble. The conversion to dependency groups will have to wait a tiny bit more.

Agreed, best to not be an early adopter for that, good to wait a while until versions of `pip` et al. that support it have been available for 6-12 months so devs are likely to have them installed.

PR #736, opened 2025-04-24
CI: drop Pyston from test matrix

dnicolodi:
Pyston is unmaintained. The most recent Pyston packages are only available for for Ubuntu 20.04. GitHub Actions is dropping support for this distribution. With this it is becoming more demanding to test this interpreter. The latest Pyston release support Python 3.8, which is EOL for CPython. meson-python does not have any Pyston specific code. Maintaining test coverage for this interpreter does not seem worth the effort that it would require.


rgommers (approved):
Thanks @dnicolodi. Makes sense, last commit on the Pyston repo was 3 years ago, it's really dead.

PR #737, opened 2025-04-25
MAINT: switch to PEP639 license declaration for meson-python itself

dnicolodi:
meson-python uses itself as a build backend, thus we can switch to PEP639 license declaration now that meson-python supports it.


rgommers (approved):
Thanks @dnicolodi, good idea. License-related content in `PKG-INFO` looks better, and dogfooding is nice. In it goes.

PR #745, opened 2025-05-03
BUG: fix a problem with symlink support

rgommers:
Modifies the existing `symlinks` test package so it exercises the bug. The situation without a `version` keyword in `meson.build` is more difficult because the dist tarball generated by Meson will be named `<project-name>-undefined.tar.gz`, which meson-python then changes to `<project-name-version.tar.gz`.

_Note: I didn't include the dist script part discussed in https://github.com/mesonbuild/meson-python/pull/728#issuecomment-2848691677, because I'm out of time right now. I think it should be done, but it's not blocking for 0.18.0 I think._

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -966,6 +966,9 @@ class Project():
                         warnings.warn(
                             f'symbolic link pointing to a directory ignored: {name}', stacklevel=1)
 
+                # Copy `member` before starting to modify it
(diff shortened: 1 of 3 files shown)


rgommers:
I'll leave the dist script for a follow-up and merge this, since @dnicolodi told me he is unavailable for a bit and I want to finish the 0.18.0 release.

PR #746, opened 2025-05-05
MAINT: set version to 0.19.0.dev0

rgommers:


rgommers:
Hmm, it looks like Cirrus is getting more strict - 11 jobs failing on _Not enough compute credits to prioritize tasks!_

This may be because I opened this PR while CI on gh-731 was also still running. But it looks like we'll have to deal with this problem.

PR #747, opened 2025-05-05
MAINT: avoid warnings because of symlinks pointing outside repo

rgommers:
Follows up on https://github.com/mesonbuild/meson-python/pull/728#issuecomment-2848691677 and gh-745.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,8 @@
+# SPDX-FileCopyrightText: 2025 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+# These test files contain symlinks to outside the repo (hence invalid - on
+# purpose) so exclude them to prevent warnings. See gh-728 for context.
+tests/packages/symlinks/baz.py export-ignore
+tests/packages/symlinks/qux.py export-ignore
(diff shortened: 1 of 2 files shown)


dnicolodi (changes requested):
Another solution would be to create the symlinks in the test code. This is simpler, but should we add a note in the test somewhere stating that these test files are not present in the sdist?

rgommers:
Added the note in the test case itself, that is indeed useful.

rgommers:
CI failure is unrelated, looks like a change in Meson master:
```
FAILED tests/test_editable.py::test_mesonpy_meta_finder - AssertionError: assert 'D:/a/meson-python/meson-python/tests/packages/complex/complex\\__init__.py' == 'D:\\a\\meson-python\\meson-python\\tests\\packages\\complex\\complex\\__init__.py'
  
  - D:\a\meson-python\meson-python\tests\packages\complex\complex\__init__.py
  ?   ^ ^            ^            ^     ^        ^       ^
  + D:/a/meson-python/meson-python/tests/packages/complex/complex\__init__.py
  ?   ^ ^            ^            ^     ^        ^       ^
FAILED tests/test_editable.py::test_importlib_resources - AssertionError: assert 'D:/a/meson-python/meson-python/tests/packages/simple/__init__.py' == 'D:\\a\\meson-python\\meson-python\\tests\\packages\\simple\\__init__.py'
```

PR #752, opened 2025-05-27
DOCS: Add short contributing guide

tobiasdiez:
Just to help people get started. Not sure if there are additional steps / other important things that should be mentioned.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,18 @@
+# Contributing
+
+Thank you for your interest in contributing to `meson-python`!
+
+## How to Contribute
+
+1. **Fork the repository** and create your branch from `main`.
+2. **Setup development environment** using `pip install -e .[test]`.
+   This will install as well all necessary dependencies for development and testing.
+   
+   It is recommended to use a virtual environment to avoid conflicts with system packages.
+3. **Make your changes** and add tests if applicable.
+4. **Run the test suite** to ensure all tests pass: `pytest` from the project root.
+5. **Submit a pull request** with a clear description of your changes.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,19 @@
+# Contributing
```
dnicolodi, line 1: All other meson-python documentation is in reST format. This should be too.

dnicolodi, line 1: This is missing a license header (this is the reason why the CI fails).

dnicolodi (changes requested):
I'm not really convinced this adds anything really useful to the documentation: the steps are pretty much standard for any Python project, except installing ninja, which becomes obvious once someone tries to run the tests.

In any case, if others think that this is useful, this file needs to be in reST format as the other parts of the documentation, and needs to have a license header.

CONTRIBUTING.md
tobiasdiez: Done.

dnicolodi: The only thing that the GitHub native support for `CONTRIBUTING.md` does is to show the content when a new pull request is opened. I argue that the proposed content is only relevant to help contributors in the stage of development that come before the point of preparing a PR, thus I see the GitHub native support for `CONTRIBUTING.md` only showing the added documentation in the wrong context. I much prefer having all the documentation in a uniform format and, eventually, being able to link it to the rendered HTML documentation.

dnicolodi: Did you really write this in 2021 and waited till now to submit the PR? 🙂

PR #753, opened 2025-05-27
Add VSCode settings for pytest configuration

tobiasdiez:
Include settings to enable pytest in VSCode.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+    "python.testing.pytestArgs": [
+        "."
+    ],
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true
+}


dnicolodi:
Does VSCode really need to add this to all projects? I don't think we should add a configuration file for every editor or IDE someone may use to work on the project.

tobiasdiez:
> Does VSCode really need to add this to all projects? I don't think we should add a configuration file for every editor or IDE someone may use to work on the project.

Yes, it's needed since otherwise vscode will not recognize the existing pytests and show them in its test inferface which provides a  quite convenient way to re-run only a few tests and other helpful tooling.

> Piggybacking off of meson's own well-defined test command might help though.

My impression is that vscode directly calls pytest to discover tests etc.

eli-schwartz:
> My impression is that vscode directly calls pytest to discover tests etc.

In theory it could use meson via vscode-meson! :)

rgommers:
I had a look at https://github.com/mesonbuild/vscode-meson, but not being a VSCode user, I'm not sure from the README how that would work exactly for hooking up a pytest invocation. It seems very reasonable for working on `meson-python` to suggest `vscode-meson` to VSCode users though.

eli-schwartz:
> but not being a VSCode user, I'm not sure from the README how that would work exactly for hooking up a pytest invocation.

Nor me, but I assume vscode-meson allows you to configure a build with `meson setup` and test it with `meson test`.

tobiasdiez:
Is there anything missing here? I would like to work on the other PR, but for this it's more convenient to have this PR merged first.

PR #754, opened 2025-05-27
MAINT: add config for ruff formatter

tobiasdiez:
Add configuration for the ruff formatter to enforce single quotes for strings, otherwise it constantly reformats the strings to double-quotes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,6 +99,9 @@ known-first-party = [
   'mesonpy',
 ]
 
+[tool.ruff.format]
+quote-style = 'single'
+
 
 [tool.coverage.run]
 disable_warnings = [


pyproject.toml
```diff
@@ -98,6 +98,9 @@ known-first-party = [
   'mesonpy',
 ]
 
+[tool.ruff.format]
+# Use single quotes for strings
+quote-style = "single"
```
dnicolodi, line 103: We use single quotes in `pyrpoject.toml` too.

pyproject.toml
```diff
@@ -98,6 +98,9 @@ known-first-party = [
   'mesonpy',
 ]
 
+[tool.ruff.format]
+# Use single quotes for strings
```
dnicolodi, line 102: This comment is redundant: it is pretty clear what the purpose of the option is. Please remove it.

pyproject.toml
```diff
@@ -98,6 +98,9 @@ known-first-party = [
...
+quote-style = "single"
```
dnicolodi, line 104: Please leave a double empty line between the configuration of different tools.

eli-schwartz:
It may independently be worth configuring the linter to check the quoting style. Interestingly, ruff doesn't have a consistent configuration schema, so you *will* have to configure single quotes separately for the linter and the autoformatter or they will produce different results (`format` will produce code that `check` errors out on and autofixes to something that `format` then tries to revert again).

rgommers:
> So, it does depend on whether the meson-python project devs view ruff as "eh, personally I wouldn't bother" or "please revert the style choices the autoformatter just made"

What I thought this would be doing is align style choices for code already written by the author of a PR with what `pre-commit` enforces in CI. I don't touch `pre-commit` locally, so `ruff` formatting may cause less conflicts with style choices made in CI.

Either way, yeah the PR diffs should be minimal, that's always important - a formatting should never increase those.

dnicolodi:
> > > It may independently be worth configuring the linter to check the quoting style.
> > 
> > That's already check if I'm not mistaken.
> 
> I don't see it in `lint.select`, are you sure?

It's  enforced by pre-commit-hook, I think here https://github.com/mesonbuild/meson-python/blob/9d46f41ffa4f95439b6e56409802eb69ed220ec8/.pre-commit-config.yaml#L21 Having it enforced by ruff too would not be a bad idea, just to make it easier to check for the ones of us that do not buy into the pre-commit-hook thing.

dnicolodi:
See #757

dnicolodi (approved):

PR #758, opened 2025-05-31
MAINT: fix ruff configuration in pyproject.toml

rgommers:
This change fixes something that may be invalid TOML, aligns better with the docs at https://docs.astral.sh/ruff/configuration, and avoids Pixi choking on parsing `pyproject.toml` with:

```
  79 │ lint.extend-ignore = [
     · ──┬─
     ·   ╰── first defined here
  80 │   'B019',
  81 │ ]
...
  98 │
  99 │ [tool.ruff.lint.flake8-quotes]
     ·            ──┬─
     ·              ╰── duplicate defined here
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,10 +76,15 @@ strict = true
 
 [tool.ruff]
 line-length = 127
-lint.extend-ignore = [
+
+[tool.ruff.format]
+quote-style = 'single'
+
+[tool.ruff.lint]
+extend-ignore = [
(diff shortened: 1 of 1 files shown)


dnicolodi:
I'm quite sure this is a bug in pixi because already before the recent changes that added the `tool.ruff.lint.flake8-quotes` section we had a `tool.ruff.lint.isort` section and AFAIK it didn't cause any issue.

PR #761, opened 2025-06-04
TST: harden against environment variables affecting meson-python

### LINKED ISSUE #760: tests/test_tags.py:43: AssertionError on Darwin
Links:

- https://hydra.nixos.org/build/298947878 (python312Packages.meson-python.aarch64-darwin)
- https://hydra.nixos.org/build/298947879 (python312Packages.meson-python.x86_64-darwin)
- https://hydra.nixos.org/build/298967891 (python313Packages.meson-python.aarch64-darwin)
- https://hydra.nixos.org/build/298967893 (python313Packages.meson-python.x86_64-darwin)

```python
________________________________ test_wheel_tag ________________________________

    def test_wheel_tag():
>       assert str(mesonpy._tags.Tag()) == f'{INTERPRETER}-{ABI}-{PLATFORM}'
E       AssertionError: assert 'cp313-cp313-...sx_11_0_arm64' == 'cp313-cp313-...sx_15_0_arm64'
E         
E         - cp313-cp313-macosx_15_0_arm64
E         ?                     ^
E         + cp313-cp313-macosx_11_0_arm64
E         ?                     ^

tests/test_tags.py:43: AssertionError
____________________________ test_tag_platlib_wheel ____________________________

    def test_tag_platlib_wheel():
        builder = wheel_builder_test_factory({
            'platlib': [f'extension{SUFFIX}'],
        })
>       assert str(builder.tag) == f'{INTERPRETER}-{ABI}-{PLATFORM}'
E       AssertionError: assert 'cp313-cp313-...sx_11_0_arm64' == 'cp313-cp313-...sx_15_0_arm64'
E         
E         - cp313-cp313-macosx_15_0_arm64
E         ?                     ^
E         + cp313-cp313-macosx_11_0_arm64
E         ?                     ^

tests/test_tags.py:99: AssertionError
_____________________________ test_tag_stable_abi ______________________________

    def test_tag_stable_abi():
        builder = wheel_builder_test_factory({
            'platlib': [f'extension{ABI3SUFFIX}'],
        }, limited_api=True)
        # PyPy does not support the stable ABI.
        abi = 'abi3' if '__pypy__' not in sys.builtin_module_names else ABI
>       assert str(builder.tag) == f'{INTERPRETER}-{abi}-{PLATFORM}'
E       AssertionError: assert 'cp313-abi3-macosx_11_0_arm64' == 'cp313-abi3-macosx_15_0_arm64'
E         
E         - cp313-abi3-macosx_15_0_arm64
E         ?                    ^
E         + cp313-abi3-macosx_11_0_arm64
E         ?                    ^

tests/test_tags.py:108: AssertionError
```

dnicolodi:
Fixes #760.

### DIFF
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -40,7 +40,13 @@ SUFFIX = sysconfig.get_config_var('EXT_SUFFIX')
 ABI3SUFFIX = get_abi3_suffix()
 
 
-def test_wheel_tag():
+@pytest.fixture()
+def cleanenv(monkeypatch):
+    # $MACOSX_DEPLOYMENT_TARGET affects the computation of the platform tag on macOS.
+    monkeypatch.delenv('MACOSX_DEPLOYMENT_TARGET', raising=False)
+
+
+def test_wheel_tag(cleanenv):
     assert str(mesonpy._tags.Tag()) == f'{INTERPRETER}-{ABI}-{PLATFORM}'
     assert str(mesonpy._tags.Tag(abi='abi3')) == f'{INTERPRETER}-abi3-{PLATFORM}'
 
@@ -112,14 +118,14 @@ def test_tag_purelib_wheel():
     assert str(builder.tag) == 'py3-none-any'
 
 
-def test_tag_platlib_wheel():
+def test_tag_platlib_wheel(cleanenv):
     builder = wheel_builder_test_factory({
         'platlib': [f'extension{SUFFIX}'],
     })
     assert str(builder.tag) == f'{INTERPRETER}-{ABI}-{PLATFORM}'
 
 
-def test_tag_stable_abi():
+def test_tag_stable_abi(cleanenv):
     builder = wheel_builder_test_factory({
         'platlib': [f'extension{ABI3SUFFIX}'],
     }, limited_api=True)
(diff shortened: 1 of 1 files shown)


dnicolodi:
> It's not entirely clear to me why tests in `test_wheel.py` aren't also failing for the same reason

Indeed they should. I wonder if they did not show up in the Nix CI as failing because they were disabled already.

dnicolodi:
The tests fail locally if I set `$MACOSX_DEPLOYMENT_TARGET`. I'll extend the fix to these tests too.

I've also found an error in the documentation. https://mesonbuild.com/meson-python/reference/environment-variables.html#envvar-MACOSX_DEPLOYMENT_TARGET says that `11` should be a valid value for this environment variable, but it is not, it should be `11.0`.

rgommers:
We have (or had?) some compat code to ensure that both `11` and `11.0` are valid. `packaging` made a mess in the `10.x` -> `>=11` transition, and we had users using both. `11.0` is canonical indeed, but that's also weird given that packaging disallows `11.x` with `x != 0`.

dnicolodi:
> We have (or had?) some compat code to ensure that both `11` and `11.0` are valid.

I don't remember having it and I don't find trace of it in the git history. There are a lot of details in the comments in the implementation https://github.com/mesonbuild/meson-python/blob/1e69e7a23f2b24d688dc4220e93de6f0e2bcf9d2/mesonpy/_tags.py#L94-L160

The issue is that here https://github.com/mesonbuild/meson-python/blob/1e69e7a23f2b24d688dc4220e93de6f0e2bcf9d2/mesonpy/_tags.py#L143 we expect the version to have two components, and `11` has only one.

I don't know whether it is better to fix the implementation or the docs. I don't know how to check which values are accepted by clang.

dnicolodi:
> I don't know how to check which values are accepted by clang.

That was actually easy: clang errors out when it does not understand the env var format. `11`, `11.0`, `11.0.0` are all valid.

I'll fix the implementation to accept the `11` variant too.

vcunat:
So when we applied this patch, we now see those test_wheel failures instead:
https://hydra.nixos.org/build/300061448/nixlog/1/tail

rgommers:
@vcunat that should be taken care of by the follow-up PR to this one that was merged earlier today: gh-762.

PR #766, opened 2025-06-14
ENH: add tool.meson-python.wheel.exclude setting

dnicolodi:
I am not entirely sure we want to include this facility: it may potentially be abused when better solutions exist. However, I think it is better than the solution we currently recommend https://mesonbuild.com/meson-python/how-to-guides/shared-libraries.html#shared-library-from-subproject of adding specific options to subprojects to exclude this or that parts from the installation.

### DIFF
--- a/docs/reference/pyproject-settings.rst
+++ b/docs/reference/pyproject-settings.rst
@@ -29,9 +29,9 @@ use them and examples.
 .. option:: tool.meson-python.limited-api
 
    A boolean indicating whether the extension modules contained in the
-   Python package target the `Python limited API`__.  Extension
+   Python package target the `Python limited API`_.  Extension
    modules can be compiled for the Python limited API specifying the
-   ``limited_api`` argument to the |extension_module()|__ function
+   ``limited_api`` argument to the |extension_module()|_ function
    in the Meson Python module.  When this setting is set to true, the
    value ``abi3`` is used for the Python wheel filename ABI tag.
 
@@ -63,8 +63,36 @@ use them and examples.
 
    Extra arguments to be passed to the ``meson install`` command.
 
+.. option:: tool.meson-python.wheel.exclude
 
-__ https://docs.python.org/3/c-api/stable.html?highlight=limited%20api#stable-application-binary-interface
-__ https://mesonbuild.com/Python-module.html#extension_module
+   List of glob patterns matching paths of files that must be excluded from
+   the Python wheel. The accepted glob patterns are the ones implemented by
+   the Python :mod:`fnmatch` with case sensitive matching. The paths to be
+   matched are as they appear in the Meson introspection data, namely they are
+   rooted in one of the Meson install locations: ``{bindir}``, ``{datadir}``,
+   ``{includedir}``, ``{libdir_shared}``, ``{libdir_static}``, et cetera.
+
+   Inspecting the `Meson introspection data`_ may be useful to craft the exclude
+   patterns. It is accessible as the ``meson-info/intro-install_plan.json`` JSON
+   document in the build directory.
+
+   This configuration setting is measure of last resort to exclude installed
+   files from a Python wheel. It is to be used when the project includes
+   subprojects that do not allow fine control on the installed files. Better
+   solutions include the use of Meson install tags and excluding subprojects
+   to be installed via :option:`tool.meson-python.args.install`.
+
+.. option:: tool.meson-python.wheel.include
+
+   List of glob patterns matching paths of files that must not be excluded
+   from the Python wheel. All files recorded for installation in the Meson
+   project are included in the Python wheel unless matching an exclude glob
+   pattern specified in :option:`tool.meson-python.wheel.exclude`. An include
+   glob pattern is useful exclusively to limit the effect of an exclude
+   pattern that matches too many files.
+
+.. _python limited api: https://docs.python.org/3/c-api/stable.html?highlight=limited%20api#stable-application-binary-interface
+.. _extension_module(): `https://mesonbuild.com/Python-module.html#extension_module
+.. _meson introspection data: https://mesonbuild.com/IDE-integration.html#install-plan
 
 .. |extension_module()| replace:: ``extension_module()``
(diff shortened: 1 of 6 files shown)


dnicolodi:
Another idea I was toying with is to extend the filtering by meson install tags to allow to specify different tags per subproject. meson-python does not rely on the meson implementation of the filtering anyway, thus extending the facility is possible.

tests/packages/subproject/pyproject.toml
```diff
@@ -5,3 +5,8 @@
 [build-system]
 build-backend = 'mesonpy'
 requires = ['meson-python']
+
+[tool.meson-python.wheel]
+# Meson versions before 1.3.0 install data files in {datadir}/{project name}/,
+# later versions install in the more correct {datadir}/{dubproject name}/
```
rgommers, line 11: typo dubproject

tests/packages/subproject/pyproject.toml
```diff
@@ -5,3 +5,8 @@
...
+exclude = ['{datadir}/*/data.txt']
```
rgommers, line 12: This moves install location markers into the public API, which is something that may constrain us in the future, and would need to be documented. I think it'd be better to match on paths inside the wheel, so for this it'd be `subproject-*.data/dep/data.txt'`. That pattern should not be variable, but is given by the [wheel spec](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#file-contents)

tests/packages/subproject/pyproject.toml
```diff
@@ -5,3 +5,8 @@
 [build-system]
 build-backend = 'mesonpy'
 requires = ['meson-python']
+
+[tool.meson-python.wheel]
+# Meson versions before 1.3.0 install data files in {datadir}/{project name}/,
```
rgommers, line 10: It'd be fine to support this for meson>=1.3.0 only I think.

tests/packages/subproject/pyproject.toml
dnicolodi: Another issue with using wheel locations is that it makes globbing to exclude whole subdirectories much less useful.

rgommers: Ah yes, of course. I think the `{ }` introspection data is stable, so it'll probably be fine - I wanted to raise it to be certain. 

Let me resolve this thread.

rgommers:
> We could also consider implementing both `wheel.exclude` and `wheel.include` so that the first could be used to exclude a whole subtree and the latter to include selected files in the same subtree.

I can see the appeal of that, yes. Especially on projects with lots of install targets, that seems nice. Although the cases where you can't make things concise with excluding install tags + `wheel.exclude` may be a small minority.

This could be done immediately, or later if there's demand. Given this PR is very nice and clean, may be okay to add it immediately?

tests/packages/subproject/pyproject.toml
dnicolodi: The difference is only in the glob pattern used in the test. It is more work to make the test conditional to the Meson version than to use a glob here.

rgommers:
> Working on this it came to my mind that we could also have a `tool.meson-python.wheel.map` that allows to map files installed in one location to another. I'm just worried that it may be misused.

Yeah potentially useful, but also potentially easy to misuse. This thread from yesterday is a good example: https://discuss.python.org/t/include-header-above-project-root/102221. Could potentially be a nicer way to include a C library that's one folder up in the repo, instead of symlinking it in - we've had examples of this, like `nanoarrow`. 

On the other hand, this seems quite bad: https://hatch.pypa.io/latest/config/build/#forced-inclusion. It gives an explicit example of pulling in a header file from outside the repo, which seems like a good way to create non-reproducible builds/issues.

dnicolodi:
> Yeah potentially useful, but also potentially easy to misuse. This thread from yesterday is a good example: https://discuss.python.org/t/include-header-above-project-root/102221. Could potentially be a nicer way to include a C library that's one folder up in the repo, instead of symlinking it in - we've had examples of this, like `nanoarrow`.

What I had in mind would not solve this issue. What I had in mind is a facility that maps a Meson installation path, for example to another, for example `{py_platlib}/package/foolib.so` to `{py_platlib}/package/foolib.so`. However, just while writing this I realized that a simple mapping like this is not going to be easy to use as the filename to be mapped is platform dependent.

rgommers:
> Not really sure I understand the problem. If you're building a C library one folder up in the repo, I'm guessing this is a C library and its optional bindings. You really want the entire project to be a single build including both C and python bits. Autotools / Meson / CMake can all do that already,

The point is that you can't do it through a standard build frontend interface if the C library code isn't inside the part of the repo containing `pyproject.toml`. gh-674 is the example I was thinking about, there have been others. The problem to solve is how to invoke `python -m build --wheel` or similar and pull in that C library. 

> What I had in mind would not solve this issue.. [...] I realized that a simple mapping like this is not going to be easy to use as the filename to be mapped is platform dependent.

Good point. Okay, leaving that problem for another day then:)

The added `include` part also looks good on a first read. Let me see if I can test and hit the green button on this one after dinner.

rgommers:
> However, I think it is better than the solution we currently recommend https://mesonbuild.com/meson-python/how-to-guides/shared-libraries.html#shared-library-from-subproject of adding specific options to subprojects to exclude this or that parts from the installation.

I assume you want to leave an update to those docs out of this PR?

dnicolodi:
I've updated the documentation.

docs/reference/pyproject-settings.rst
```diff
@@ -63,6 +63,29 @@ use them and examples.
 
    Extra arguments to be passed to the ``meson install`` command.
 
+.. option:: tool.meson-python.wheel.exclude
+
+   List of glob patterns matching paths of files that must be excluded from
+   the Python wheel. The accepted glob patterns are the ones implemented by
+   the Python :mod:`fnmatch` with case sensitive matching. The paths to be
+   matched are as they appear in the Meson introspection data, namely they are
```
rgommers: It will be useful to name `intro-install_plan.json` explicitly, and perhaps link to https://mesonbuild.com/IDE-integration.html#install-plan. The average reader will be unaware of the existence of this file. Maybe also explain where it can be found, or use `<build-dir>/meson-info/intro-install_plan.json` as the name.

rgommers (approved):
The docs LGTM, thanks. Two small suggestions only. Also there's a merge conflict to resolve. When that is resolved, this looks ready to go in.

PR #768, opened 2025-06-14
TST: enable --fatal-meson-warnings for all tests

dnicolodi:
The test packages also work as examples. I think it is better to make sure that they are as correct as possible.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -41,12 +41,10 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
-          # actions/setup-python do not support Python 3.7 on macOS on
-          # arm64. The latest macOS version available is macOS 14 but
-          # it is arm64 only. Stick to macos-13 as long as we want to
-          # test Python 3.7 on macOS. macOS 14 on arm64 is covered by
-          # the Cirrus CI jobs.
-          - macos-13
+          # macOS 14 on arm64 is covered by the Cirrus CI jobs, stick to macOS
+          # on x86-64 here as long as we can: macOS 15 is currently planned as
+          # the last supported image based on Intel CPU on GitHub Actions.
+          - macos-15-intel
           - windows-latest
         python:
           - '3.9'


mesonpy/__init__.py
```diff
@@ -731,11 +733,12 @@ def __init__(
 
             cross_file_data = textwrap.dedent(f'''
                 [binaries]
+                ar = '{arch}-apple-{subsystem}-ar'
                 c = '{arch}-apple-{subsystem}-clang'
                 cpp = '{arch}-apple-{subsystem}-clang++'
                 objc = '{arch}-apple-{subsystem}-clang'
                 objcpp = '{arch}-apple-{subsystem}-clang++'
-                ar = '{arch}-apple-{subsystem}-ar'
+                strip = ['strip', '-arch', {arch!r}]
```
dnicolodi, line 741: @freakboy3742 Can you please verify that this is correct? I have no experience with building for iOS. Thanks!

dnicolodi:
There are still test failures. These are the tests where we set the RPATH via explicit linker arguments. We stop doing that in #724 thus I don't think it is necessary to find a way to allow these warnings to do not make the tests fail.

mesonpy/__init__.py
dnicolodi: I think the correct way to get to the `strip` utility is to run something like `xcrun --sdk iphoneos -f strip`. The system `strip` may not work for iOS binaries.

freakboy3742: You're correct that `xcrun` is generally the mechanism you use to access compiler binaries on macOS; `xcrun --sdk iphoneos strip  <args for strip>` would be a valid invocation (`-f` prints the path to the tool instead of invoking it).

So `["xcrun", "--sdk", {sdk!r}, "strip", "--arch", {arch!r}]` would be the more reliable invocation; `sdk` would need to be set to `iphoneos` or `iphonesimulator` (which you can get from `sys.platform._multiarch.split("-")[-1]`, or with an `if` analogous to how `subsystem` is constructed.

I've done some testing using the `purelib-and-platlib` project that was generating an error in the test suite, including adding a test suite that exercises both the pure and plat libraries - and it passes that test on the iOS simulator with stripping enabled. I've also tested with my build of numpy; it also works with stripping enabled.

FWIW - the straight `strip -arch...` version seems to work as well - but I can't rule out that there might be problems that aren't being exposed by a simple example.

dnicolodi: I was thinking to use `xcrun` to get the path to the `strip` binary and then writing that into the cross file. It seems a bit more efficient than using `xcrun` for every invocation of `strip`. However, there are probably not that many invocations of it, if any, thus it does not make much of a difference.

Now that I learn about `xcrun`, I am wondering: should we use it also to invoke or to find the compilers, or should we rely on the compilers always being on the `$PATH`?

dnicolodi: It's weird that Apple decided that having the compiler binaries have the architecture triple in their name was a good idea but it was not for `strip`.

freakboy3742: They didn't. Those binaries are a shim provided as part of a Python distribution for iOS, specifically to avoid issues with having definitions for variables like `CC` that have a space in them. In a lot of older build tools, there's an implied assumption that you can split a compile command at the first space... which isn't true if you're using `xcrun`. Those shims also do a bunch of handling around `IPHONEOS_DEPLOYMENT_TARGET` handling, etc.

As for pre-calling `xcrun` and caching the result - that could work. The value returned will change over time, but not over the lifespan of a single build, so caching on a per-build basis should be fine. `xcrun` *should* always be on the path (as long as the user has Xcode installed), so that part should be reliable enough.

freakboy3742: It couldn't hurt to add a shim for `strip`. The complication is that I'm not sure if we'll be able to backport that shim to Python 3.13, since it would be a new feature... but I *can* manually add it to the support package build (as those aren't currently maintained by CPython itself), so in practice, it would be available.

dnicolodi: It does not really matter if it is not available even if we say its should be in the cross file, as long as no one is using it 🙂 Adding it to the support package build would be my preferred solution, unless you see an issue with this.

eli-schwartz: > WARNING: Cross file does not specify strip binary, result will not be stripped.

As noted by the warning, the consequences of it not being available are not earth-shattering.

Quite likely, an appstore upload will do server-side stripping anyway.

dnicolodi: @freakboy3742 I was looking for where the compiler wrappers. I see them mentioned in the documentation of that project, but I don't see the source code for them anywhere, although I admit that I got a bit lost in the Makefile.

freakboy3742: Ah - they're all defined [here](https://github.com/python/cpython/tree/main/iOS/Resources/bin).

dnicolodi: Thanks! Looking at them, now I wonder whether we should go in the other direction instead and have the content of these shims "unwrapped" into the cross file. I'm probably happy either way.

rgommers: It looks to me like the performance difference is going to be very small compared to the overall time it takes to build a wheel. So I'd go with the simplest solution, which is invoking the shims directly. Fixing a bug that may show up in a version of CPython that doesn't get bug fix releases anymore is years away, and if such a bug fix is needed, the unwrapping can then still be done.

rgommers:
Enabling `--fatal-meson-warnings` by default does seem like a good idea.

mesonpy/__init__.py
rgommers: We can, the CPython shims just look a lot more complex.

Anyway, I am fine with that solution too if you prefer, any choice is valid here. Just my $2c that I'd make the minimal change only for `strip` here (maybe I'm just lazy:)).

dnicolodi:
@rgommers I think we will keep testing adding RPATH entries with linker arguments, thus I added a way to disable fatal meson warnings for selected packages. This is ready for another review.

rgommers:
Ah that is not good indeed. If a test fails, the usual strategy I use is to build it locally to reproduce the issue. Plus they can be thought of as small example packages. So it'd be better to find another strategy to skip this check. A couple of thoughts:

- Keep the monkeypatching and `pyproject.toml` content, but add a new `_private-args` or `_testing-args` build config setting rather than using `setup-args`
- Put a skip list with package names inside the patched-in `_get_meson_args` rather than putting it in the package config.

rgommers:
> The skip list is probably the easiest and cleanest option.

I agree

PR #771, opened 2025-06-22
DOC: shared-libraries: fix and clarify install location example

### LINKED ISSUE #769: typo in internal shared libraries docs? ERROR: Got unknown keyword arguments "install_path"
I am struggling with a seemingly common issue of bundling a shared library (for a pure c library I build via a subproject) in my python wheel (#765, #673, #410) .

While trying to follow the example at, https://mesonbuild.com/meson-python/how-to-guides/shared-libraries.html I get the following error `ERROR: Got unknown keyword arguments "install_path"` from meson when attempting to set that variable for the shared_library. 

According to the meson docs, shared_library only accepts install_dir (https://mesonbuild.com/Reference-manual_functions.html#shared_library_install_dir) 

Likely this line will need to be updated:
https://github.com/mesonbuild/meson-python/blob/7efa4176758106374e7e6ef2b5d4c1968a38c570/docs/how-to-guides/shared-libraries.rst?plain=1#L71

I want to confirm that my interpretation of a typo in the docs was correct or if `install_path` is meant to be used elsewhere and not with the shared_library.

AndrewAnnex:
Updates install_path to install_dir. 
Updates code block to be more explicit with shared_library usage.

Fixes #769

### DIFF
--- a/docs/how-to-guides/shared-libraries.rst
+++ b/docs/how-to-guides/shared-libraries.rst
@@ -64,18 +64,28 @@ A shared library produced by ``library()`` or ``shared_library()`` built like th
 
 is installed to ``libdir`` by default. If the only reason the shared library exists
 is to be used inside the Python package being built, then it is best to modify
-the install location to be within the Python package itself:
+the install location, via the ``install_dir`` argument, to be within
+the Python package itself:
 
-.. code-block:: python
+.. TODO update the text block below to 'meson' when
+.. meson lexer is updated to fix
+.. https://github.com/pygments/pygments/issues/2918
 
-   install_path: py.get_install_dir() / 'mypkg/subdir'
+.. code-block:: text
+
+    example_lib = shared_library(
+        'example',
+        'examplelib.c',
+        install: true,
+        install_dir: py.get_install_dir() / 'mypkg/subdir',
+    )
 
 Then an extension module in the same install directory can link against the
 shared library in a portable manner by using ``install_rpath``:
 
 .. code-block:: meson
 
-    py3.extension_module('_extmodule',
+    py.extension_module('_extmodule',
         '_extmodule.c',
         link_with: example_lib,
         install: true,


docs/how-to-guides/shared-libraries.rst
```diff
@@ -64,11 +64,16 @@ A shared library produced by ``library()`` or ``shared_library()`` built like th
 
 is installed to ``libdir`` by default. If the only reason the shared library exists
 is to be used inside the Python package being built, then it is best to modify
-the install location to be within the Python package itself:
+the install location, via the `install_dir` argument, to be within the Python package itself:
```
dnicolodi, line 67: reST requires double backticks to format something as code. Can you also line wrap this at 78-80 characters?

docs/how-to-guides/shared-libraries.rst
```diff
@@ -64,11 +64,16 @@ A shared library produced by ``library()`` or ``shared_library()`` built like th
...
 
-.. code-block:: python
+.. code-block:: meson
 
-   install_path: py.get_install_dir() / 'mypkg/subdir'
+    example_lib = shared_library(
+        'example',
+        'examplelib.c',
+        install: true,
+        install_dir: py.get_install_dir() / 'mypkg/subdir',
```
dnicolodi, line 75: I've just realized that this example and other in this file uses the Meson's `python` module imported as `py`. However, the example just below uses `py3`. While at this, can you please fix this inconsistency too (in a separate commit maybe)?

dnicolodi (changes requested):

dnicolodi (changes requested):
Can you also please fix the commit message to comply with the format we use for this project? Namely, it should have a summary line of less than 78 characters, starting with an identifier in all caps, `DOC` in this case, followed by colons and the summary of the changes. The details go in the body. Something like
```
DOC: shared-libraries: fix and clarify install location example

Fixes #769
```

AndrewAnnex:
@dnicolodi happy to fix all the above, although for the commit message wouldn't that be easier to fix when squashing the commit on merge? I guess I can fix all the above, squash the commits, and force push to the branch too but that is a bit more involved

docs/how-to-guides/shared-libraries.rst
```diff
@@ -64,15 +64,16 @@ A shared library produced by ``library()`` or ``shared_library()`` built like th
...
+the Python package itself:
 
 .. code-block:: meson
 
     example_lib = shared_library(
         'example',
         'examplelib.c',
         install: true,
-        install_dir: py.get_install_dir() / 'mypkg/subdir',
+        install_dir: py3.get_install_dir() / 'mypkg/subdir',
```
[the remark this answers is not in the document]
AndrewAnnex: oh heh oops will do

dnicolodi:
> I guess I can fix all the above, squash the commits, and force push to the branch too but that is a bit more involved

You should do just that. The goal is to teach how to contribute to this project while freeing the maintainers from menial task.

dnicolodi (changes requested):
CI is still not happy.

dnicolodi:
I would use `text` for the code block language and leave a comment that points to the pygments issue.

AndrewAnnex:
okay cleaned up the commits, updated the code-block to text, added some comments about the pygments issue, and ran precommit locally to hopefully avoid further ci complaints

dnicolodi:
Thanks.

PR #772, opened 2025-06-30
ENH: define strip binary in synthesized cross files

dnicolodi:
Extracted from #768. `strip` has been added to the CPython distributed compiler wrappers, use that.


rgommers (approved):
Nice, thanks @dnicolodi

PR #779, opened 2025-07-16
WIP: ENH: implement support for build-details.json (PEP 739) and cross compilation

### LINKED ISSUE #321: Support cross compiling

rgommers:
*This is very much not ready yet, but I'm opening a WIP PR now to get early feedback, since there's more than one way this can be implemented.*

Depends on https://github.com/mesonbuild/meson/pull/14657 (Cc @mgorny)
Closes gh-321

Example of a cross compilation CI job for SciPy, which successfully uses a cross file and a `build-details.json` file to cross compile from x86-64 to aarch64 and then in a second job installs and tests the cross-compiled wheel:
- https://github.com/scipy/scipy/compare/main...rgommers:scipy:ci-cross
- https://github.com/rgommers/scipy/actions/runs/16294476137

The `meson-python` parts of cross compilation that need fixing are:
1. The wheel filename (interpreter, abi and platform tags)
2. The platform tag in the `WHEEL` metadata file

The changes here do both of those things.

A few things I'd like to get input on:
- I chose to use both `is_cross` and `build_details` in `_WheelBuilder`, for clarity and so that cross compilation without use of a  `build-details.json` file can also be supported (parsing the cross file can get the platform right at least, but not the interpreter). We could also decide to not worry about that case and only support supplying both files for a cross build.
- Test strategy:
    - Cross compile a single package, or multiple/all?
    - Do we do just Linux x86-64 -> aarch64 and vice versa as tests where we can validate at runtime because easy with GHA runners and toolchain setup, or anything else?
    - Conda-forge is easiest to set up (for me at least), but we can also do another distro.
    - Use minimum supported Python and Python 3.14? For the former we need to generate a `build-details.json`, for the latter it's shipped as part of CPython. 
- (minor) Static typing: `build_details` is hard to type. Use `build_details: Optional[Dict[str, str | Dict[str, Any]]]` or simply `Dict[str, Any]`?

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -312,11 +312,15 @@ class _WheelBuilder():
         manifest: Dict[str, List[Tuple[pathlib.Path, str]]],
         limited_api: bool,
         allow_windows_shared_libs: bool,
+        is_cross: bool,
+        build_details: Optional[Dict[str, str | Dict[str, Any]]] = None,
     ) -> None:
         self._metadata = metadata
         self._manifest = manifest
         self._limited_api = limited_api
         self._allow_windows_shared_libs = allow_windows_shared_libs
+        self._is_cross = is_cross
+        self._build_details = build_details
 
     @property
     def _has_internal_libs(self) -> bool:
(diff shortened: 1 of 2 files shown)


huhuang03:
Maybe tag read from cross-file is a good way.

rgommers:
I'll close this in favor of gh-829, which is further along.

PR #786, opened 2025-07-31
DOC: editable: clarify file guarantees with non-Python files

lucascolley:
closes https://github.com/numpy/numpy/issues/29490 @rgommers

### DIFF
--- a/docs/how-to-guides/editable-installs.rst
+++ b/docs/how-to-guides/editable-installs.rst
@@ -49,10 +49,11 @@ installation step to become effective.
 
 __ https://packaging.python.org/en/latest/specifications/entry-points/
 
-An editable install exposes at least all the files that would be
+An editable install exposes at least all of the Python files that would be
 available in a regular installation. However, depending on the file
 and directory organization in your project, it might also expose files
-that would not be normally available.
+that would not be normally available. Furthermore, some files written in
+other languages to Python may not be available.
 
 
 Build dependencies


docs/how-to-guides/editable-installs.rst
```diff
@@ -49,10 +49,11 @@ installation step to become effective.
 
 __ https://packaging.python.org/en/latest/specifications/entry-points/
 
-An editable install exposes at least all the files that would be
+An editable install exposes at least all of the Python files that would be
 available in a regular installation. However, depending on the file
 and directory organization in your project, it might also expose files
-that would not be normally available.
+that would not be normally available. Furthermore, some files written in
+other languages to Python may not be available.
```
dnicolodi, line 56: Uh? This is actually false. Where did you get the impression that only Python files are exposed?

dnicolodi (changes requested):

docs/how-to-guides/editable-installs.rst
lucascolley: > Where did you get the impression that only Python files are exposed?

I didn't get the impression that only Python files are exposed — that isn't what this says!

I did get the impression that some files in other languages may not be available: see https://github.com/numpy/numpy/issues/29490.

dnicolodi: All installed files are available, if you look them up with the proper Python API. 

I really don't understand what you are trying to communicate with these edits. Most likely I don't understand what you mean with "Python file". After all, meson-python is mostly used to build packages that contain Python modules written in other languages that Python, therefore, if these sometimes were to go missing, it would be a major issue.

I guess that you are trying to access installed files by their (expected) filesystem location rather than through the Python API. A few lines above the text you are amending we have this sentence:

> This will install a stub in the Python site packages directory that loads the package content from the sources and build directory.

I thought that this should be indication enough that the installed files do not exist in the filesystem at the location where they would be in an normal isntallation. But I have the impression that this is not picked up by most readers. What we really need is a "Accessing data files" in the editable install documentation that spells this out.

eli-schwartz: @dnicolodi

(In the linked numpy issue, the specific topic at hand was numpy's C headers.)

dnicolodi: > What's a "proper python API"?

The proper API is indeed `importlib.resources`. There are hooks that allow to expose arbitrary data through it. Editable install support in `meson-python` uses it.

> and if you need a file tree suitable for e.g. the C compiler's `-I -L` arguments you try to map that to a directory

This is where the impedance matching fails. `importlib.resources` exposes a `pathlib.Path`-like interface but there is no guarantee that the underlying data is accessible as files and directories in the filesystem.

> The question is how an editable loader is expected to arrange for `importlib.resources` to give a correct answer.

See how it is done here https://github.com/mesonbuild/meson-python/blob/1c520a8e181ea7d9deff75bc43739b3ef4bb9d68/mesonpy/_editable.py It is a bit convoluted but it works.

> People expect specific directory structures, among other things, which may not make any sense in an editable install e.g. files that at build time are in different directories and at install time are in a single directory.

The directory structure looks as it should when navigated via the `importlib.resources` interfaces. However, it does may not exist in the filesystem, indeed. Exposing a directory structure assembled at install time to something that does not use `importlib.resources` to access it is an issue that AFAIK cannot be solved with the editable install approach chosen by `meson-python`. It should be noted that editable installs are only one way to bump into this issue. ZIP imports are not a very common thing anymore, but they would result in the same issue. I believe that WASI has more or less the same issue: there is no real filesystem.

eli-schwartz: > It should be noted that editable installs are only one way to bump into this issue. ZIP imports are not a very common thing anymore, but they would result in the same issue. I believe that WASI has more or less the same issue: there is no real filesystem.

True. But also ZIP imports have much bigger issues (certainly for numpy): they don't support `import foo._c_ext`.

dnicolodi: I still do not understand what "some files written in other languages to Python" means (and I think it should be "other languages than Python", but I don't think this is where my understanding fails).

dnicolodi:
I'm preparing a PR that adds to the editable installs documentation a new section about data files that expands a bit what is proposed here.

docs/how-to-guides/editable-installs.rst
eli-schwartz: > > True. But also ZIP imports have much bigger issues (certainly for numpy): they don't support `import foo._c_ext`.
> 
> Is this to mean that ZIP imports do not support extension modules? I think they do.

https://docs.python.org/3/library/zipimport.html

> This module adds the ability to import Python modules (*.py, *.pyc) and packages from ZIP-format archives. It is usually not needed to use the [zipimport](https://docs.python.org/3/library/zipimport.html#module-zipimport) module explicitly; it is automatically used by the built-in [import](https://docs.python.org/3/reference/simple_stmts.html#import) mechanism for [sys.path](https://docs.python.org/3/library/sys.html#sys.path) items that are paths to ZIP archives.

> Any files may be present in the ZIP archive, but importers are only invoked for .py and .pyc files. ZIP import of dynamic modules (.pyd, .so) is disallowed.

dnicolodi: > In the linked numpy issue, the specific topic at hand was numpy's C headers.

Thinking about it, there may be a way out. If numpy were to return a list of include directories instead of a single include directory, it could list all the directories where the headers are physically located in the filesystem. This requires to keeping a list of all installed header files somewhere in a the code that returns the include directories, and assumes that the numpy headers do not use relative path includes. Then something like
```python
def get_include_paths():
    return ':'.join(set(str(importlib.resources(numpy).joinpath(header).parent) for header in HEADERS)))
```
on the numpy side, and something like:
```meson
npinc = include_directories(run_command([py, '-c', 'include numpy; print(numpy.get_include_paths())'], capture: true).split(':'))
```
should work.

PR #783, opened 2025-07-29
TST: rework the sharedlib-in-package test package

dnicolodi:
This reorganizes the test package to a flatter layout that helps visualizing all the parts involved in the test and introduces an asymmetry between the source layout and the installation layout that demonstrates the bugs in the RPATH handling.

PR #788, opened 2025-08-10
Fixes relative to RPATH handling

### LINKED ISSUE #711: RPATH goes missing when using both `install_rpath` and an internal shared library dependency

### LINKED ISSUE #813: Duplicate entries in RPATH
When building with meson python on a recent version of macOS I can no longer import my module.

```python
import gridfire
```

```
ImportError: dlopen(/opt/homebrew/anaconda3/envs/fourdstar/lib/python3.13/site-packages/gridfire.cpython-313-darwin.so, 0x0002): tried: '/opt/homebrew/anaconda3/envs/fourdstar/lib/python3.13/site-packages/gridfire.cpython-313-darwin.so' (duplicate LC_RPATH '@loader_path/.gridfire.mesonpy.libs'), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/anaconda3/envs/fourdstar/lib/python3.13/site-packages/gridfire.cpython-313-darwin.so' (no such file), '/opt/homebrew/anaconda3/envs/fourdstar/lib/python3.13/site-packages/gridfire.cpython-313-darwin.so' (duplicate LC_RPATH '@loader_path/.gridfire.mesonpy.libs')
```

When I inspect the shared object file with otool I see that it contains many duplicate LC_RPATH entries

```bash
otool -l gridfire.cpython-313-darwin.so | grep RPATH -A2
```

```
          cmd LC_RPATH
      cmdsize 48
         path @loader_path/.gridfire.mesonpy.libs (offset 12)
--
          cmd LC_RPATH
      cmdsize 48
         path @loader_path/.gridfire.mesonpy.libs (offset 12)
--
          cmd LC_RPATH
      cmdsize 48
         path @loader_path/.gridfire.mesonpy.libs (offset 12)
--
          cmd LC_RPATH
      cmdsize 48
         path @loader_path/.gridfire.mesonpy.libs (offset 12)
--
          cmd LC_RPATH
      cmdsize 48
         path @loader_path/.gridfire.mesonpy.libs (offset 12)
--
          cmd LC_RPATH
      cmdsize 48
         path @loader_path/.gridfire.mesonpy.libs (offset 12)
--
          cmd LC_RPATH
      cmdsize 48
         path @loader_path/.gridfire.mesonpy.libs (offset 12)
--
          cmd LC_RPATH
      cmdsize 48
         path @loader_path/.gridfire.mesonpy.libs (offset 12)
--
          cmd LC_RPATH
      cmdsize 48
         path @loader_path/.gridfire.mesonpy.libs (offset 12)
--
          cmd LC_RPATH
      cmdsize 56
         path /opt/homebrew/Cellar/openssl@3/3.5.2/lib (offset 12)
--
          cmd LC_RPATH
      cmdsize 48
         path /opt/homebrew/Cellar/xz/5.8.1/lib (offset 12)
--
          cmd LC_RPATH
      cmdsize 40
         path /opt/homebrew/opt/zstd/lib (offset 12)
```

If I manually remove all but one of the rpath instances using install_name_tool

```bash
install_name_tool -delete_rpath @loader_path/.gridfire.mesonpy.libs
```

Then I can import and use my module just fine. Presumably then I could write some somewhat convoluted but not too complex tool to use otool to count the number of duplicate paths and remove them; however, I'm sure this strikes me as a one-way street to a bad time (lack of portability, and just a generally convoluted build step).

> Other discussions (https://github.com/facebook/zstd/issues/4369, https://fenicsproject.discourse.group/t/duplicate-lc-rpath-errors-on-macos-15-4-sequoia/17364) have pinpointed this change in behavior with recent versions of apple's XCode command line tools. Specifically that duplicate lc_rpaths are now treated as an error when loading shared libraries. 

My question then is **how do I prevent meson-python from adding all these duplicate paths to the shared object file?**

## Details
- macOS Version 26.0.1
- CLITools version : 26.0.0.0.1.1757719676
- meson-python version: 0.18.0
- meson version: 1.9.1
- clang version: Apple clang 17.0.0 (clang-1700.3.19.1)

## Other details
- I have tried building with `pip install . --no-cache-dir` and `pip install . --no-cache-dir --isolated` and neither of these make a difference. 
- I have confirmed the same behavior in my other mesonpy library, fourdst. I see the same behavior there where if I remove all but one of the rpath's using install_name_tool then I can import and use the module just fine
- When I inspect the shared library created by meson (which I belive is generated before meson-python steps in, though I may be mistaken, there are no duplicate lc_rpaths). 

```bash
otool -l libgridfire.dylib | grep RPATH -A2
```

```
          cmd LC_RPATH
      cmdsize 72
         path @loader_path/../subprojects/libconstants/src/constants (offset 12)
--
          cmd LC_RPATH
      cmdsize 64
         path @loader_path/../subprojects/libconfig/src/config (offset 12)
--
          cmd LC_RPATH
      cmdsize 56
         path @loader_path/../subprojects/yaml-cpp (offset 12)
--
          cmd LC_RPATH
      cmdsize 72
         path @loader_path/../subprojects/libcomposition/src/composition (offset 12)
--
          cmd LC_RPATH
      cmdsize 64
         path @loader_path/../subprojects/liblogging/src/logging (offset 12)
--
          cmd LC_RPATH
      cmdsize 56
         path @loader_path/../subprojects/libplugin/src (offset 12)
--
          cmd LC_RPATH
      cmdsize 64
         path @loader_path/../subprojects/minizip-ng-4.0.10 (offset 12)
--
          cmd LC_RPATH
      cmdsize 56
         path @loader_path/../subprojects/cvode-7.3.0 (offset 12)
--
          cmd LC_RPATH
      cmdsize 56
         path /opt/homebrew/Cellar/openssl@3/3.5.2/lib (offset 12)
--
          cmd LC_RPATH
      cmdsize 48
         path /opt/homebrew/Cellar/xz/5.8.1/lib (offset 12)
--
          cmd LC_RPATH
      cmdsize 40
         path /opt/homebrew/opt/zstd/lib (offset 12)
```

Originally reported by @tboudreaux in https://github.com/mesonbuild/meson-python/discussions/812

dnicolodi:
Builds on top of #783 and replaces #724

Fixes #711
Fixes #813

### DIFF
--- a/docs/reference/meson-compatibility.rst
+++ b/docs/reference/meson-compatibility.rst
@@ -61,6 +61,12 @@ versions.
 
    Meson 1.9.0 or later is required to support building for iOS.
 
+   Meson 1.9.0 or later is also required to remove RPATH entries that
+   are added by Meson to allow to execute parts of the project from
+   the build directory and that are normally removed during ``meson
+   install``. On older Meson versions, these entries are not removed,
+   but this should not have any adverse effect.
+
 Build front-ends by default build packages in an isolated Python
 environment where build dependencies are installed. Most often, unless
 a package or its build dependencies declare explicitly a version
--- a/docs/reference/meson-compatibility.rst
+++ b/docs/reference/meson-compatibility.rst
@@ -52,6 +52,11 @@ versions.
    populate the package license and license files from the ones
    declared via the ``project()`` call in ``meson.build``.
 
+   Meson 1.6.0 or later is also required to support the
+   ``install_rpath`` argument to Meson functions that accept it, such
+   as ``library()`` and ``extension_module()`` from the ``python``
+   Meson module. On older Meson versions, this argument has no effect.
+
 .. option:: 1.9.0
 
    Meson 1.9.0 or later is required to support building for iOS.
--- a/mesonpy/_rpath.py
+++ b/mesonpy/_rpath.py
@@ -11,18 +11,57 @@ import typing
 
 
 if typing.TYPE_CHECKING:
-    from typing import Iterable, List, Union
+    from typing import List, TypeVar, Union
     Path = Union[str, os.PathLike[str]]
+    T = TypeVar('T')
 
 
-if sys.platform == 'win32' or sys.platform == 'cygwin':
+def unique(values: List[T]) -> List[T]:
+    r = []
+    for value in values:
+        if value not in r:
+            r.append(value)
+    return r
 
-    def fix_rpath(filepath: Path, libs_relative_path: str) -> None:
+
+class RPATH:
+
+    origin = '$ORIGIN'
+
+    @staticmethod
+    def get_rpath(filepath: Path) -> List[str]:
+        raise NotImplementedError
+
+    @staticmethod
+    def set_rpath(filepath: Path, old: List[str], rpath: List[str]) -> None:
+        raise NotImplementedError
+
+    @classmethod
+    def fix_rpath(cls, filepath: Path, libs_relative_path: str) -> None:
+        old_rpath = cls.get_rpath(filepath)
+        new_rpath = []
+        for path in old_rpath:
+            if path.startswith(cls.origin):
+                path = os.path.join(cls.origin, libs_relative_path)
+            new_rpath.append(path)
+        new_rpath = unique(new_rpath)
+        if new_rpath != old_rpath:
+            cls.set_rpath(filepath, old_rpath, new_rpath)
+
+
+class _Windows(RPATH):
+
+    @classmethod
+    def fix_rpath(cls, filepath: Path, libs_relative_path: str) -> None:
         pass
 
-elif sys.platform == 'darwin':
 
-    def _get_rpath(filepath: Path) -> List[str]:
+class _MacOS(RPATH):
+
+    origin = '@loader_path'
+
+    @staticmethod
+    def get_rpath(filepath: Path) -> List[str]:
         rpath = []
         r = subprocess.run(['otool', '-l', os.fspath(filepath)], capture_output=True, text=True)
         rpath_tag = False
@@ -34,17 +73,24 @@ elif sys.platform == 'darwin':
                 rpath_tag = False
         return rpath
 
-    def _replace_rpath(filepath: Path, old: str, new: str) -> None:
-        subprocess.run(['install_name_tool', '-rpath', old, new, os.fspath(filepath)], check=True)
+    @staticmethod
+    def set_rpath(filepath: Path, old: List[str], rpath: List[str]) -> None:
+        # This implementation does not preserve the ordering of RPATH
+        # entries. Meson does the same, thus it should not be a problem.
+        args: List[str] = []
+        for path in rpath:
+            if path not in old:
+                args += ['-add_rpath', path]
+        for path in old:
+            if path not in rpath:
+                args += ['-delete_rpath', path]
+        subprocess.run(['install_name_tool', *args, os.fspath(filepath)], check=True)
 
-    def fix_rpath(filepath: Path, libs_relative_path: str) -> None:
-        for path in _get_rpath(filepath):
-            if path.startswith('@loader_path/'):
-                _replace_rpath(filepath, path, '@loader_path/' + libs_relative_path)
 
-elif sys.platform == 'sunos5':
+class _SunOS5(RPATH):
 
-    def _get_rpath(filepath: Path) -> List[str]:
+    @staticmethod
+    def get_rpath(filepath: Path) -> List[str]:
         rpath = []
         r = subprocess.run(['/usr/bin/elfedit', '-r', '-e', 'dyn:rpath', os.fspath(filepath)],
             capture_output=True, check=True, text=True)
@@ -55,35 +101,32 @@ elif sys.platform == 'sunos5':
                         rpath.append(path)
         return rpath
 
-    def _set_rpath(filepath: Path, rpath: Iterable[str]) -> None:
+    @staticmethod
+    def set_rpath(filepath: Path, old: List[str], rpath: List[str]) -> None:
         subprocess.run(['/usr/bin/elfedit', '-e', 'dyn:rpath ' + ':'.join(rpath), os.fspath(filepath)], check=True)
 
-    def fix_rpath(filepath: Path, libs_relative_path: str) -> None:
-        old_rpath = _get_rpath(filepath)
-        new_rpath = []
-        for path in old_rpath:
-            if path.startswith('$ORIGIN/'):
-                path = '$ORIGIN/' + libs_relative_path
-            new_rpath.append(path)
-        if new_rpath != old_rpath:
-            _set_rpath(filepath, new_rpath)
 
-else:
-    # Assume that any other platform uses ELF binaries.
+class _ELF(RPATH):
 
-    def _get_rpath(filepath: Path) -> List[str]:
+    @staticmethod
+    def get_rpath(filepath: Path) -> List[str]:
         r = subprocess.run(['patchelf', '--print-rpath', os.fspath(filepath)], capture_output=True, text=True)
         return [x for x in r.stdout.strip().split(':') if x]
 
-    def _set_rpath(filepath: Path, rpath: Iterable[str]) -> None:
+    @staticmethod
+    def set_rpath(filepath: Path, old: List[str], rpath: List[str]) -> None:
         subprocess.run(['patchelf','--set-rpath', ':'.join(rpath), os.fspath(filepath)], check=True)
 
-    def fix_rpath(filepath: Path, libs_relative_path: str) -> None:
-        old_rpath = _get_rpath(filepath)
-        new_rpath = []
-        for path in old_rpath:
-            if path.startswith('$ORIGIN/'):
-                path = '$ORIGIN/' + libs_relative_path
-            new_rpath.append(path)
-        if new_rpath != old_rpath:
-            _set_rpath(filepath, new_rpath)
+
+if sys.platform == 'win32' or sys.platform == 'cygwin':
+    _cls = _Windows
+elif sys.platform == 'darwin':
+    _cls = _MacOS
+elif sys.platform == 'sunos5':
+    _cls = _SunOS5
+else:
+    _cls = _ELF
+
+get_rpath = _cls.get_rpath
+set_rpath = _cls.set_rpath
+fix_rpath = _cls.fix_rpath
--- a/mesonpy/_rpath.py
+++ b/mesonpy/_rpath.py
@@ -73,7 +73,7 @@ else:
 
     def _get_rpath(filepath: Path) -> List[str]:
         r = subprocess.run(['patchelf', '--print-rpath', os.fspath(filepath)], capture_output=True, text=True)
-        return r.stdout.strip().split(':')
+        return [x for x in r.stdout.strip().split(':') if x]
 
     def _set_rpath(filepath: Path, rpath: Iterable[str]) -> None:
         subprocess.run(['patchelf','--set-rpath', ':'.join(rpath), os.fspath(filepath)], check=True)
--- a/mesonpy/_rpath.py
+++ b/mesonpy/_rpath.py
@@ -39,11 +39,21 @@ class RPATH:
     @classmethod
     def fix_rpath(cls, filepath: Path, libs_relative_path: str) -> None:
         old_rpath = cls.get_rpath(filepath)
-        new_rpath = []
-        for path in old_rpath:
-            if path.startswith(cls.origin):
-                path = os.path.join(cls.origin, libs_relative_path)
-            new_rpath.append(path)
+        new_rpath = old_rpath[:]
+
+        # When an executable, libray, or Python extension module is
+        # dynamically linked to a library built as part of the project, Meson
+        # adds a build RPATH pointing to the build directory, in the form of a
+        # relative RPATH entry. We can use the presence of any RPATH entries
+        # relative to ``$ORIGIN`` as an indicator that the installed object
+        # depends on shared libraries internal to the project. In this case we
+        # need to add an RPATH entry pointing to the meson-python shared
+        # library install location. This heuristic is not perfect: RPATH
+        # entries relative to ``$ORIGIN`` can exist for other reasons.
+        # However, this only results in harmless additional RPATH entries.
+        if any(path.startswith(cls.origin) for path in old_rpath):
+            new_rpath.append(os.path.join(cls.origin, libs_relative_path))
+
         new_rpath = unique(new_rpath)
         if new_rpath != old_rpath:
             cls.set_rpath(filepath, old_rpath, new_rpath)
--- a/mesonpy/_rpath.py
+++ b/mesonpy/_rpath.py
@@ -11,7 +11,7 @@ import typing
 
 
 if typing.TYPE_CHECKING:
-    from typing import List, TypeVar, Union
+    from typing import List, Optional, TypeVar, Union
     Path = Union[str, os.PathLike[str]]
     T = TypeVar('T')
 
@@ -37,7 +37,7 @@ class RPATH:
         raise NotImplementedError
 
     @classmethod
-    def fix_rpath(cls, filepath: Path, libs_relative_path: str) -> None:
+    def fix_rpath(cls, filepath: Path, install_rpath: List[str], libs_relative_path: Optional[str]) -> None:
         old_rpath = cls.get_rpath(filepath)
         new_rpath = old_rpath[:]
 
@@ -51,8 +51,11 @@ class RPATH:
         # library install location. This heuristic is not perfect: RPATH
         # entries relative to ``$ORIGIN`` can exist for other reasons.
         # However, this only results in harmless additional RPATH entries.
-        if any(path.startswith(cls.origin) for path in old_rpath):
+        if libs_relative_path and any(path.startswith(cls.origin) for path in old_rpath):
             new_rpath.append(os.path.join(cls.origin, libs_relative_path))
+
+        # Add install_rpath.
+        new_rpath += install_rpath
 
         new_rpath = unique(new_rpath)
         if new_rpath != old_rpath:
@@ -62,7 +65,7 @@ class RPATH:
 class _Windows(RPATH):
 
     @classmethod
-    def fix_rpath(cls, filepath: Path, libs_relative_path: str) -> None:
+    def fix_rpath(cls, filepath: Path, install_rpath: List[str], libs_relative_path: str) -> None:
         pass
 
 
--- a/mesonpy/_rpath.py
+++ b/mesonpy/_rpath.py
@@ -37,9 +37,14 @@ class RPATH:
         raise NotImplementedError
 
     @classmethod
-    def fix_rpath(cls, filepath: Path, install_rpath: List[str], libs_relative_path: Optional[str]) -> None:
+    def fix_rpath(cls, filepath: Path, add: List[str], remove: List[str], libs_relative_path: Optional[str]) -> None:
         old_rpath = cls.get_rpath(filepath)
-        new_rpath = old_rpath[:]
+
+        # Meson adds a padding entry to RPATH composed of enough `X`
+        # characters to reserve enough space in the ELF header to hold
+        # the final installation RPATH. Remove this entry and other
+        # entries to be removed.
+        new_rpath = [path for path in old_rpath if path.strip('X') and path not in remove]
 
         # When an executable, libray, or Python extension module is
         # dynamically linked to a library built as part of the project, Meson
@@ -55,7 +60,7 @@ class RPATH:
             new_rpath.append(os.path.join(cls.origin, libs_relative_path))
 
         # Add install_rpath.
-        new_rpath += install_rpath
+        new_rpath += add
 
         new_rpath = unique(new_rpath)
         if new_rpath != old_rpath:
@@ -65,7 +70,7 @@ class RPATH:
 class _Windows(RPATH):
 
     @classmethod
-    def fix_rpath(cls, filepath: Path, install_rpath: List[str], libs_relative_path: str) -> None:
+    def fix_rpath(cls, filepath: Path, add: List[str], remove: List[str], libs_relative_path: Optional[str]) -> None:
         pass
 
 
--- a/tests/packages/sharedlib-in-package/src/lib.c
+++ b/tests/packages/sharedlib-in-package/src/lib.c
@@ -1,0 +1,10 @@
+// SPDX-FileCopyrightText: 2022 The meson-python developers
+//
+// SPDX-License-Identifier: MIT
+
+#include "lib.h"
+#include "sublib.h"
+
+int prodsum(int a, int b, int x) {
+    return prod(a,  x) + b;
+}
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -182,7 +182,7 @@ def test_link_library_in_subproject(venv, wheel_link_library_in_subproject):
 
 
 @pytest.mark.skipif(sys.platform in {'win32', 'cygwin'}, reason='requires RPATH support')
-def test_rpath(wheel_link_against_local_lib, tmp_path):
+def test_link_against_local_lib_rpath(wheel_link_against_local_lib, tmp_path):
     artifact = wheel.wheelfile.WheelFile(wheel_link_against_local_lib)
     artifact.extractall(tmp_path)
 
@@ -190,9 +190,7 @@ def test_rpath(wheel_link_against_local_lib, tmp_path):
     expected = {f'{origin}/../.link_against_local_lib.mesonpy.libs', 'custom-rpath',}
 
     rpath = set(mesonpy._rpath._get_rpath(tmp_path / 'example' / f'_example{EXT_SUFFIX}'))
-    # Verify that rpath is a superset of the expected one: linking to
-    # the Python runtime may require additional rpath entries.
-    assert rpath >= expected
+    assert rpath == expected
 
 
 @pytest.mark.skipif(sys.platform in {'win32', 'cygwin'}, reason='requires RPATH support')
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -168,6 +168,25 @@ def test_local_lib(venv, wheel_link_against_local_lib):
     assert int(output) == 3
 
 
+@pytest.mark.skipif(sys.platform in {'win32', 'cygwin'}, reason='requires RPATH support')
+def test_sharedlib_in_package_rpath(wheel_sharedlib_in_package, tmp_path):
+    artifact = wheel.wheelfile.WheelFile(wheel_sharedlib_in_package)
+    artifact.extractall(tmp_path)
+
+    origin = '@loader_path' if sys.platform == 'darwin' else '$ORIGIN'
+
+    rpath = set(mesonpy._rpath._get_rpath(tmp_path / 'mypkg' / f'_example{EXT_SUFFIX}'))
+    # FIXME: RPATH entries added by Meson to point to the build directory are not removed.
+    assert rpath >= {origin}
+
+    rpath = set(mesonpy._rpath._get_rpath(tmp_path / 'mypkg' / f'liblib{LIB_SUFFIX}'))
+    # FIXME: RPATH entries added by Meson to point to the build directory are not removed.
+    assert rpath >= {f'{origin}/sub'}
+
+    rpath = set(mesonpy._rpath._get_rpath(tmp_path / 'mypkg' / 'sub' / f'libsublib{LIB_SUFFIX}'))
+    assert rpath == set()
+
+
 def test_sharedlib_in_package(venv, wheel_sharedlib_in_package):
     venv.pip('install', wheel_sharedlib_in_package)
     output = venv.python('-c', 'import mypkg; print(mypkg.prodsum(2, 3, 4))')
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -187,6 +187,22 @@ def test_sharedlib_in_package_rpath(wheel_sharedlib_in_package, tmp_path):
     assert rpath == set()
 
 
+@pytest.mark.skipif(sys.platform in {'win32', 'cygwin'}, reason='requires RPATH support')
+def test_sharedlib_in_package_rpath_ldflags(package_sharedlib_in_package, tmp_path, monkeypatch):
+    origin = '@loader_path' if sys.platform == 'darwin' else '$ORIGIN'
+    extra_rpath = {f'{origin}/test-ldflags', '/usr/lib/test-ldflags'}
+    ldflags = ' '.join(f'-Wl,-rpath,{p}' for p in extra_rpath)
+    monkeypatch.setenv('LDFLAGS', ldflags)
+
+    filename = mesonpy.build_wheel(tmp_path)
+    artifact = wheel.wheelfile.WheelFile(tmp_path / filename)
+    artifact.extractall(tmp_path)
+
+    for path in f'_example{EXT_SUFFIX}', f'liblib{LIB_SUFFIX}', f'sub/libsublib{LIB_SUFFIX}':
+        rpath = set(mesonpy._rpath._get_rpath(tmp_path / 'mypkg' / path))
+        assert extra_rpath <= rpath
+
+
 def test_sharedlib_in_package(venv, wheel_sharedlib_in_package):
     venv.pip('install', wheel_sharedlib_in_package)
     output = venv.python('-c', 'import mypkg; print(mypkg.prodsum(2, 3, 4))')
@@ -210,6 +226,21 @@ def test_link_against_local_lib_rpath(wheel_link_against_local_lib, tmp_path):
 
     rpath = set(mesonpy._rpath._get_rpath(tmp_path / 'example' / f'_example{EXT_SUFFIX}'))
     assert rpath == expected
+
+
+@pytest.mark.skipif(sys.platform in {'win32', 'cygwin'}, reason='requires RPATH support')
+def test_link_against_local_lib_rpath_ldflags(package_link_against_local_lib, tmp_path, monkeypatch):
+    origin = '@loader_path' if sys.platform == 'darwin' else '$ORIGIN'
+    extra_rpath = {f'{origin}/test-ldflags', '/usr/lib/test-ldflags'}
+    ldflags = ' '.join(f'-Wl,-rpath,{p}' for p in extra_rpath)
+    monkeypatch.setenv('LDFLAGS', ldflags)
+
+    filename = mesonpy.build_wheel(tmp_path)
+    artifact = wheel.wheelfile.WheelFile(tmp_path / filename)
+    artifact.extractall(tmp_path)
+
+    rpath = set(mesonpy._rpath._get_rpath(tmp_path / 'example' / f'_example{EXT_SUFFIX}'))
+    assert extra_rpath <= rpath
 
 
 @pytest.mark.skipif(sys.platform in {'win32', 'cygwin'}, reason='requires RPATH support')
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -175,15 +175,15 @@ def test_sharedlib_in_package_rpath(wheel_sharedlib_in_package, tmp_path):
 
     origin = '@loader_path' if sys.platform == 'darwin' else '$ORIGIN'
 
-    rpath = set(mesonpy._rpath._get_rpath(tmp_path / 'mypkg' / f'_example{EXT_SUFFIX}'))
+    rpath = set(mesonpy._rpath.get_rpath(tmp_path / 'mypkg' / f'_example{EXT_SUFFIX}'))
     # FIXME: RPATH entries added by Meson to point to the build directory are not removed.
     assert rpath >= {origin}
 
-    rpath = set(mesonpy._rpath._get_rpath(tmp_path / 'mypkg' / f'liblib{LIB_SUFFIX}'))
+    rpath = set(mesonpy._rpath.get_rpath(tmp_path / 'mypkg' / f'liblib{LIB_SUFFIX}'))
     # FIXME: RPATH entries added by Meson to point to the build directory are not removed.
     assert rpath >= {f'{origin}/sub'}
 
-    rpath = set(mesonpy._rpath._get_rpath(tmp_path / 'mypkg' / 'sub' / f'libsublib{LIB_SUFFIX}'))
+    rpath = set(mesonpy._rpath.get_rpath(tmp_path / 'mypkg' / 'sub' / f'libsublib{LIB_SUFFIX}'))
     assert rpath == set()
 
 
@@ -199,7 +199,7 @@ def test_sharedlib_in_package_rpath_ldflags(package_sharedlib_in_package, tmp_pa
     artifact.extractall(tmp_path)
 
     for path in f'_example{EXT_SUFFIX}', f'liblib{LIB_SUFFIX}', f'sub/libsublib{LIB_SUFFIX}':
-        rpath = set(mesonpy._rpath._get_rpath(tmp_path / 'mypkg' / path))
+        rpath = set(mesonpy._rpath.get_rpath(tmp_path / 'mypkg' / path))
         assert extra_rpath <= rpath
 
 
@@ -224,7 +224,7 @@ def test_link_against_local_lib_rpath(wheel_link_against_local_lib, tmp_path):
     origin = '@loader_path' if sys.platform == 'darwin' else '$ORIGIN'
     expected = {f'{origin}/../.link_against_local_lib.mesonpy.libs', 'custom-rpath',}
 
-    rpath = set(mesonpy._rpath._get_rpath(tmp_path / 'example' / f'_example{EXT_SUFFIX}'))
+    rpath = set(mesonpy._rpath.get_rpath(tmp_path / 'example' / f'_example{EXT_SUFFIX}'))
     assert rpath == expected
 
 
@@ -239,7 +239,7 @@ def test_link_against_local_lib_rpath_ldflags(package_link_against_local_lib, tm
     artifact = wheel.wheelfile.WheelFile(tmp_path / filename)
     artifact.extractall(tmp_path)
 
-    rpath = set(mesonpy._rpath._get_rpath(tmp_path / 'example' / f'_example{EXT_SUFFIX}'))
+    rpath = set(mesonpy._rpath.get_rpath(tmp_path / 'example' / f'_example{EXT_SUFFIX}'))
     assert extra_rpath <= rpath
 
 
@@ -249,7 +249,7 @@ def test_uneeded_rpath(wheel_purelib_and_platlib, tmp_path):
     artifact.extractall(tmp_path)
 
     origin = '@loader_path' if sys.platform == 'darwin' else '$ORIGIN'
-    rpath = mesonpy._rpath._get_rpath(tmp_path / f'plat{EXT_SUFFIX}')
+    rpath = mesonpy._rpath.get_rpath(tmp_path / f'plat{EXT_SUFFIX}')
     for path in rpath:
         assert origin not in path
 
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -170,10 +170,8 @@ def test_local_lib(venv, wheel_link_against_local_lib):
 
 def test_sharedlib_in_package(venv, wheel_sharedlib_in_package):
     venv.pip('install', wheel_sharedlib_in_package)
-    output = venv.python('-c', 'import mypkg; print(mypkg.example_sum(2, 5))')
-    assert int(output) == 7
-    output = venv.python('-c', 'import mypkg; print(mypkg.example_prod(6, 7))')
-    assert int(output) == 42
+    output = venv.python('-c', 'import mypkg; print(mypkg.prodsum(2, 3, 4))')
+    assert int(output) == 11
 
 
 @pytest.mark.skipif(MESON_VERSION < (1, 3, 0), reason='meson too old')
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -33,6 +33,9 @@ ABI = tag.abi
 INTERPRETER = tag.interpreter
 PLATFORM = adjust_packaging_platform_tag(tag.platform)
 
+# Support for removing build RPATH entries requires Meson 1.9 or later.
+BUILD_RPATH_SUPPORT = MESON_VERSION >= (1, 9)
+
 
 def wheel_contents(artifact):
     # Sometimes directories have entries, sometimes not, so we filter them out.
@@ -176,12 +179,14 @@ def test_sharedlib_in_package_rpath(wheel_sharedlib_in_package, tmp_path):
     origin = '@loader_path' if sys.platform == 'darwin' else '$ORIGIN'
 
     rpath = set(mesonpy._rpath.get_rpath(tmp_path / 'mypkg' / f'_example{EXT_SUFFIX}'))
-    # FIXME: RPATH entries added by Meson to point to the build directory are not removed.
     assert rpath >= {origin}
+    if BUILD_RPATH_SUPPORT:
+        assert rpath == {origin}
 
     rpath = set(mesonpy._rpath.get_rpath(tmp_path / 'mypkg' / f'liblib{LIB_SUFFIX}'))
-    # FIXME: RPATH entries added by Meson to point to the build directory are not removed.
     assert rpath >= {f'{origin}/sub'}
+    if BUILD_RPATH_SUPPORT:
+        assert rpath == {f'{origin}/sub'}
 
     rpath = set(mesonpy._rpath.get_rpath(tmp_path / 'mypkg' / 'sub' / f'libsublib{LIB_SUFFIX}'))
     assert rpath == set()
@@ -225,8 +230,9 @@ def test_link_against_local_lib_rpath(wheel_link_against_local_lib, tmp_path):
     expected = {f'{origin}/../.link_against_local_lib.mesonpy.libs', 'custom-rpath',}
 
     rpath = set(mesonpy._rpath.get_rpath(tmp_path / 'example' / f'_example{EXT_SUFFIX}'))
-    # FIXME: RPATH entries added by Meson to point to the build directory are not removed.
     assert rpath >= expected
+    if BUILD_RPATH_SUPPORT:
+        assert rpath == expected
 
 
 @pytest.mark.skipif(sys.platform in {'win32', 'cygwin'}, reason='requires RPATH support')
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -225,7 +225,8 @@ def test_link_against_local_lib_rpath(wheel_link_against_local_lib, tmp_path):
     expected = {f'{origin}/../.link_against_local_lib.mesonpy.libs', 'custom-rpath',}
 
     rpath = set(mesonpy._rpath.get_rpath(tmp_path / 'example' / f'_example{EXT_SUFFIX}'))
-    assert rpath == expected
+    # FIXME: RPATH entries added by Meson to point to the build directory are not removed.
+    assert rpath >= expected
 
 
 @pytest.mark.skipif(sys.platform in {'win32', 'cygwin'}, reason='requires RPATH support')
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -127,6 +127,10 @@ def _compile_patterns(patterns: List[str]) -> Callable[[str], bool]:
 class _Entry(typing.NamedTuple):
     dst: pathlib.Path
     src: str
+    # Meson support only one install_rpath entry per target. Use a
+    # list to store install RPATH to be able to add append more
+    # entries when needed.
+    install_rpath: List[str] = []
 
 
 def _map_to_wheel(
@@ -185,7 +189,8 @@ def _map_to_wheel(
                         filedst = dst / relpath
                         wheel_files[path].append(_Entry(filedst, filesrc))
             else:
-                wheel_files[path].append(_Entry(dst, src))
+                install_rpath = target.get('install_rpath')
+                wheel_files[path].append(_Entry(dst, src, [install_rpath] if install_rpath else []))
 
     return wheel_files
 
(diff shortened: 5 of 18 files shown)


dnicolodi:
This should fix RPATH handling for good.

The only thing not included is automatic translation of the `$ORIGIN` anchor in `install_rpath` arguments to `@loader_path` on macOS. It would not be hard to implement, however, it would be a deviation from what Meson implements. I am not sure it is a good idea to implement it.

There is one case in which this may break projects that work now: when libraries or modules require setting an RPATH to dynamically link to a library installed in the Python install path (with something like `install_dir: py.get_install_dir() / 'package'`) and that work now for how because meson-python does not remove build RPATH entries added by Meson and the source layout matches the install layout. This was the case for example for SciPy 1.15, see https://github.com/mesonbuild/meson-python/pull/724#discussion_r2274385684 and previous discussion. The comments there indicated that, for what SciPy is concerned, it should be fine to break this now. Note that SciPy 1.15 used `install_rpath` arguments that are implemented here, thus it would be fine, but it uses the `$ORIGIN` anchor on macOS too, and that does not work. See above. The behavior changes only with Meson 1.9.0 or later (older Meson versions did not export the required metadata), thus projects that pin the Meson version are fine.

@rgommers I think I added test cases for all scenarios we discussed. It would be great if you could test with packages that may be affected and that do not pin the meson-python version to any released version.

dnicolodi:
I think it is about time that we straighten out our RPATH story, thus tentatively marking this for the next release. However, it still needs a decision regarding the handling of the anchors in `install_rpath`, see comments above. I'll work on testing downstream projects.

dnicolodi:
@rgommers I want to add extended tests in CI that test building SciPy (and maybe some other projects later). I had a quick look at how to do it and I didn't find what the preferred way of building SciPy is these days. Should I use pixi, spin, or something else? Or is downloading the source distribution and installing with pip or uv the way to go? The latter seems the easiest but pip does not have a way to install the build dependencies for a package and disabling build isolation is by far the simplest to test a local version of meson-python.

rgommers:
Hmm, are you sure - that's a really heavy build (10-15 min), and it's nontrivial to set up because of the BLAS dependency. Or are you thinking as a one-off or some sort of extended test suite?

We do test `meson-python`'s main branch in SciPy CI, but only on Linux I believe: https://github.com/scipy/scipy/blob/c18ebd1c4871ff1a04c397247815d859eec54de5/.github/workflows/linux.yml#L183

And that doesn't test all RPATH rewriting issues. I think what we probably need for more coverage is a macOS SciPy job against main? 

Re building:
- If we'd want it in CI here, then I'd say that a regular Ubuntu job where the system dependencies are installed with `apt` is probably best. And then a regular `pip install scipy --no-binary scipy`. There no job that can be copied 1:1, but something like [this job](https://github.com/scipy/scipy/blob/c18ebd1c4871ff1a04c397247815d859eec54de5/.github/workflows/linux.yml#L583-L586)
- `pixi` is by far the easiest for local development I think, if you clone the scipy repo and replace the `meson-python` dependency in `pixi.toml` with a path dependency to the local clone of `meson-python`.


***

I did plan on getting back to this PR after the release. It'd be great to get this merged soon.

dnicolodi:
I was thinking to add an extended test suite gated on all the existing CI jobs to complete successfully. However, I didn't expect SciPy to take 15 min to compile, it is a long time since I had to do it. Do you think it is worth having something like this in our CI?

eli-schwartz:
Does ccache not help here? I know that scipy takes a while because it has 1200 compile edges, with quite a bit of C++. On my laptop (znver5, -j16) it takes

```
$ time FEATURES=ccache ebuild scipy-1.17.1.ebuild clean install
real	4m50.092s
user	37m28.643s
sys		2m10.715s
```

But on a second run, it completes significantly faster:
```
real	0m37.002s
user	3m35.543s
sys		0m23.999s
```

eli-schwartz:
(Needing to compile all of highs despite a system package existing for it, surely is not helping compile times. Compiling that package from source with `emerge -1 =sci-mathematics/highs-1.14.0-r3` takes 48 seconds all on its own.)

rgommers:
> I was thinking to add an extended test suite gated on all the existing CI jobs to complete successfully. However, I didn't expect SciPy to take 15 min to compile, it is a long time since I had to do it. Do you think it is worth having something like this in our CI?

I'm inclined to say no here. I think there's a lot of value in our CI being simple and very fast. We'd make the completion time way longer, and the setup quite a bit more maintenance-intensive. I'd much rather add more CI jobs in SciPy's CI that use meson-python's main branch on different platforms.

I do like the idea of better integration testing though. I think we can (a) push more of our large users to test our `main` branch, and (b) have the kind of CI machinery you have in mind here in a separate repo, so it's easier to test at scale. I think https://github.com/rgommers/external-deps-build/ does a lot of what you have in mind here. We can clone that, tweak it a bit so it can be used to test PRs/branches, and populate it with a lot of packages that use `meson-python`. WDYT?

> Does ccache not help here? 

Yes it does, and we use it in SciPy - but that comes with its own extra work; two job steps per build, cache management, having to run CI jobs on `main` to populate the cache for PRs.

rgommers:
> The only thing not included is automatic translation of the `$ORIGIN` anchor in `install_rpath` arguments to `@loader_path` on macOS. It would not be hard to implement, however, it would be a deviation from what Meson implements. I am not sure it is a good idea to implement it.

I did a bit of searching, and it's recommended in our docs, and SciPy isn't the only project that used it. 

- SciPy: [code in scipy 1.15.x](https://github.com/scipy/scipy/blob/b9926e3a2adece75730156cf31df89abef3caaef/scipy/special/meson.build#L87) (meson-python<0.20.0, so won't break), and a [blog post](https://labs.quansight.org/blog/libsf-error-state) talking about that feature
- VapourSynth, current master branch (meson-python unpinned): https://github.com/vapoursynth/vapoursynth/blob/5b2d5562726a91d9a75441cc4728a90e6c9f4f27/meson.build#L308
- dwave-optimization, current main branch (`meson-python~=0.20.0`): https://github.com/dwavesystems/dwave-optimization/blob/0962fc5e3ac4adfec1d2b592ef7a0b3c016f39cd/meson.build#L162

If I can find three significant projects using the feature quickly, and we've recommended it in our docs for a while, there will be more users. We should add a backwards compat shim. That doesn't cost much; the Meson behavior of retaining a literal `$ORIGIN` isn't all that useful after all. We can document it as legacy, and in a few releases start warning about it so package authors can update to the new recommendation.

rgommers:
My review plan here is more a test plan - since I can't just read through the code and reason about changed test packages for completeness, I'll plan to add some new ones with more layouts, and run that in CI on my fork on Linux and macOS. And then add NumPy (same as SciPy, just builds faster), VapourSynth and dwave-optimization builds, and check that they all work without patching and don't contain duplicate RPATHs.

rgommers:
This PR should also update the [docs on shared libraries](https://mesonbuild.com/meson-python/how-to-guides/shared-libraries.html), so we recommend the new preferred form. Which I think is `@loader_path` on macOS, and `$ORIGIN` elsewhere.

dnicolodi:
Translating `$ORIGIN` into `@loader_path` never worked. `install_rpath` never worked, actually. Thus, all packages that ever set an `RPATH` worked by accident because of other bugs that left RPATH entries in place when they should have been removed.

This PR implements what IMO is correct handling of RPATH entries, including removal of build RPATH entries. This will break packages that have been relying (without realizing it, I assume) on build RPATH entries. The anchor translation will help the project that have (no-op till now) `install_rpath` directives with the `$ORIGIN` anchor. I suspect there may be others out there that will simply break.

rgommers:
> Translating `$ORIGIN` into `@loader_path` never worked. `install_rpath` never worked, actually. Thus, all packages that ever set an `RPATH` worked by accident because of other bugs that left RPATH entries in place when they should have been removed.

Sure, I understand that. However, it had _the effect_ of working in practice - one bug cancelling out another made that so. Let's make it easy: I'll see if I can find a simple-enough way of adding in backwards compat handling here. If I don't succeed, there's nothing to discuss. If it does work, it's probably easy to decide to add it in, so we can hard affected people rather than hard-break them in the next release.

dnicolodi:
> I'll see if I can find a simple-enough way of adding in backwards compat handling here.

If you mean adding the RPATH anchor translation, that should be easy. Knowing the code, I can do it quickly, I think.

rgommers:
Yes, that's what I meant. I was actually planning to leave it alone right now, focus on testing, see something fail (the two projects in the list above), and then get the respective CI jobs green again with a bit of backwards compat handling.

rgommers (commented):
Okay, I finally found enough time to dive deep enough into this PR/topic. The main result of the whole exercise is gh-903, which aims to map out all relevant scenarios. Quite a few of those are broken both on `main` and with the changes in this PR. 

I implemented two sets of tests in one-off CI jobs on my fork:

1. Small test packages for the scenarios listed in gh-903
2. Testing on four real-world projects: NumPy, Gridfire, VapourSynth, and dwave-optimization

For this PR to be mergeable, we'd want no regressions on both (1) and (2), and a significant improvement in the number of cases that are working correctly for (1) EDIT: which will also fix the real-world issues with at least Gridfire.

***

I did a significant amount of testing in CI (generated with Codex/Astra, not suitable for review without cleanup) and let the AI model generate ~20 test packages, because writing those all by hand and hooking up CI for them would be way too time-consuming. The branch is at https://github.com/rgommers/meson-python/tree/rpath-regression-tests; I'm not suggesting to look at that unless you're curious. What we should do is review gh-903, decide what test packages to add, write those test packages in a clean way, and them make them pass here (or `skip`/`xfail` the ones we still have problems with).

rgommers:
Oh, to be clear: I am happy to contribute of course, writing test packages from scratch, once we decide which ones we want to include.

dnicolodi:
@rgommers Thanks for trying to verify that this does the right thing. However, I find the outcome very demoralising.

This code has been sitting here for over a year (and in a form or another in other PRs for over a year) without any meaningful comment. After all this time you admit that you do not want to try to understand what the code does but will test that it works. The approach is fine for me, however, the outcome of this testing is a generic "this breaks in some AI generated test cases running in an AI generated test harness" without any indication of which these cases are and what has been tested, and an explicit suggestion to do not look into the single failure. All this seems a very elaborated way to say "this does not work" and nothing more.

Furthermore, I had a quick look at the test results and I have doubts that the tests are working properly: I see several test reporting error for failing to open a file. This PR does not change in any way how the packages are compiled or installed, thus I find this very suspicious. Other tests seem to fail because of padding entries in the `RPATH`s. These are added by Meson and so far it has never been a goal to remove them from binaries that we do not otherwise touch. Thus it has never been a goal of this PR to fix this.

rgommers:
> After all this time you admit that you do not want to try to understand what the code does

That's not at all what I said. I _will_ understand it before merging, and am up for getting this PR done very soon. However, I _also knew_ that I had to dig in deep for this one, making several full days worth of time for it. Partly because I already knew that this in its current form was a breaking change in principle (because of SciPy's previous usage, and as confirmed by https://github.com/mesonbuild/meson-python/pull/788#issuecomment-5670515132 for other packages I had to find). Hence I knew I had to start with thorough testing, rather than just reasoning about the code (which is hard!).

I don't think there's a reason to be discouraged. If we restore the original test packages - unchanged except for maybe renaming them with `-legacy` appended - and then add the new test packages (or separate extension modules added to the current test packages?) that show that the bug reports we have open get fixed, then this PR goes from "regression on some real-world packages" to "maintains backwards compat and fixes reported bugs". Which makes it way easier to merge. That's the point that I'd want to understand every line of code in detail. 

The other problem here is that I know we don't have good test coverage, and there's a lot of ways this could be breaking. Which is why I went all the way to try and summarize/test all situations I could think of. We don't have to fix them: more niche scenario's not working with the current state of `main` don't have to be fixed with this PR either, it can be deferred.

> These are added by Meson and so far it has never been a goal to remove them from binaries that we do not otherwise touch. Thus it has never been a goal of this PR to fix this.

Yes, that's perfectly fine, I agree - doesn't have to be this PR. My listing in gh-903 is "this is how ideally everything should work". We may never get to 100% on that, but I'm sure we can get close.

> This code has been sitting here for over a year (and in a form or another in other PRs for over a year) 

This thread contains the key bit for why this PR is very hard to settle in for: https://github.com/mesonbuild/meson-python/pull/724#discussion_r2217941489. PRs which only _add_ to the test suite are so much easier to review and merge. However, it was only this week you agreed on adding this back (_"If you mean adding the RPATH anchor translation, that should be easy."_ above).

***

I think we can get this in relatively quickly. If you restore backwards compat, I can convince myself of "no other regressions". And if I do find any, I will write a clean new test package by hand, to include in this PR, with a fix as well if needed.

rgommers:
Aside from the backwards compat issue though, it is looking quite good. I believe there are no other regressions except for possibly a failure for [rpath-elf-install-precedence](https://github.com/rgommers/meson-python/tree/rpath-regression-tests/tests/packages/rpath-elf-install-precedence), which I'm investigating more closely right now (it could well be an error in the test package itself, at which point all there is to do is to fix the `-legacy` thing, fix the docs issue I pointed out above, and this should be mergeable.

dnicolodi:
> This thread contains the key bit for why this PR is very hard to settle in for: [#724 (comment)](https://github.com/mesonbuild/meson-python/pull/724#discussion_r2217941489). PRs which only _add_ to the test suite are so much easier to review and merge. However, it was only this week you agreed on adding this back (_"If you mean adding the RPATH anchor translation, that should be easy."_ above).

I explained this several times, but it seems that I am not able to get the core of the issue across: the current RPATH handling is incomplete and has bugs. The test that is modified in this PR passes because the incomplete handling cancels the bugs. However, this happens only for the specific source layout used in the test. The test is modified to demonstrate exactly this: with the modified source layout things break. Being more specific, the existing test passes only because it so happens that the RPATH entries that would need to be removed are the same than the ones that would need to be added.

Because of this, the test works on macOS despite the `install_rpath` RPATH entries using the `$ORIGIN` anchor, while they would need to use the `@loader_path` anchor. However, it would have worked even if `install_rpath` would have not be used, or if it contained a wrong path.

In this sense, translating `$ORIGIN` to `@loader_path` on macOS is not really backward compatibility in the sense that it preserves a functionality that was present before. It keeps packages that use the same source layout as the test in main (of which there are a few known cases) working, as long as they also have been using an `install_rpath` using the correct path with the Linux-specific anchor. However, because `install_rpath` is effectively a no-op before this PR, it is hard to be sure that existing packages use it when it would be required and that the value they use is correct.

This brings another question: what is that we should preserve in a new test of the test modified here? The source layout or the use of `$ORIGIN` in place of `@loader_path` on macOS?

I have been asking about a decision on whether the RPATH anchor translation needs to be added for a very very long time and I don't remember any decision in a way or another. The most recent time here https://github.com/mesonbuild/meson-python/pull/788#issuecomment-4916386613

rgommers:
> I explained this several times, but it seems that I am not able to get the core of the issue across

I get how it works, I really do - I promise. 

> It keeps packages that use the same source layout as the test in main (of which there are a few known cases) working

Exactly, that is what I want. I don’t think we should release this compatibility break without warning. And given that I don't want to break them outright without warning, I had to do a lot more homework.

> In this sense, translating `$ORIGIN` to `@loader_path` on macOS is not really backward compatibility in the sense that it preserves a functionality that was present before. 

"keep packages working" is exactly what backwards compatibility means though. We do some extra work to not break those packages, that's it. The underlying mechanism or whose fault it is to have written buggy code or any other such consideration is not all that relevant. 

> This brings another question: what is that we should preserve in a new test of the test modified here? The source layout or the use of `$ORIGIN` in place of `@loader_path` on macOS?

Since the existing test package combines both, we preserve both in the regression test. If you undo the changes to the existing test package, and it keeps working, then the known real-world packages in question will also keep working (unless I missed something - I'll retest before merging). That's enough to satisfy my concern.

> I have been asking about a decision on whether the RPATH anchor translation needs to be added for a very very long time and I don't remember any decision in a way or another.

I think we're just talking past each other a little bit. That question didn't seem easy to answer quickly, and then I lost track again (I'm sorry about that, I wish I had more bandwidth, but I didn't, especially around the June/July time frame). If you had asked me "what's the best way to satisfy your concern or unblock this" I could have simply answered "keep the test package unchanged and the test for it passing", without having to think it through harder or test it. 

Now, thinking about the backwards compat fix, yes I do think translating `$ORIGIN` to `@loader_path` on macOS if we see it seems like the cleanest way to fix it (and if the explicit `install_rpath` using `$ORIGIN` correctly describes the installed layout on Linux, this translation will be enough; I don't think we have to discuss corner cases that assume `$ORIGIN` not pointing to the right place on Linux). And then whether we want to emit a warning and clean it up later, or leave it like that forever, I don't have a strong preference on.

***

Looking back, you proposed translation plus a warning in February, and you did split out the test changes into gh-783 as I requested. I didn’t follow through promptly enough, and my statement that you only agreed this week was not accurate. 

However, gh-783 still replaced the original package layout and its unconditional `$ORIGIN` usage, so it didn’t resolve my concern about preserving existing working configurations. 

Let’s make the decision explicit now: include anchor translation and retain the original test package unchanged alongside the new cases.

dnicolodi:
I added `$ORIGIN` to `@loader_path` translation for `install_rpath` arguments. And I retained a copy of the original test case. However, it fails. And it fails for a reason that I described among the reasons to restructure it (and that I forgot).

The test package builds and `_example` extension module that links with two libraries, one installed alongside the extension module, and another installed in a `sub` subdirectory. However, the original test case adds an `install_rpath` of `$ORIGIN`. For it to work it would need to add both `$ORIGIN` and `$ORIGIN/sub` to the RPATH.

I thought that this could not be fixed because I interpreted `instsall_rpath` to takes one path to be added to the RPATH. However, Meson interprets it as a `:` list of paths. This requires a small modification to the handling of `install_rpath` added here, and a fix to the test package.

dnicolodi:
There is a failure in the pixi test job:
```
FAILED tests/test_wheel.py::test_sharedlib_in_package_rpath - AssertionError: assert {'$ORIGIN', '.../default/lib'} == {'$ORIGIN'}
  
  Extra items in the left set:
  '/home/runner/work/meson-python/meson-python/.github/workflows/.pixi/envs/default/lib'
  
  Full diff:
    {
        '$ORIGIN',
  +     '/home/runner/work/meson-python/meson-python/.github/workflows/.pixi/envs/default/lib',
    }
```
I don't know where the extra RPATH entry comes from. I suspect that it is something that the build environment adds but I don't know why or where, nor what would be a good way to control for it in the test.

dnicolodi:
> I don't know where the extra RPATH entry comes from.

It comes from the `LDFLAGS` added to the Python configuration. Meson gets them from pkgconfig but it can be obtained from `sysconfig` too. I amended the tests to take it into consideration. It is not beautiful but it works.

dnicolodi:
> This PR should also update the [docs on shared libraries](https://mesonbuild.com/meson-python/how-to-guides/shared-libraries.html), so we recommend the new preferred form. Which I think is `@loader_path` on macOS, and `$ORIGIN` elsewhere.

Done.

rgommers:
Thanks for the updates!

> It comes from the `$LDFLAGS` environment variable.

Yes indeed. For a very long time, conda-forge set `CFLAGS`/`CXXFLAGS`/`LDFLAGS` in its compiler activation scripts in order to both set conda-forge defaults (e.g., `-O2`) and achieve working relocatable environments. This meant including `-Wl,-rpath=$PREFIX/lib` into `LDFLAGS`, with `$PREFIX` the root of the active conda environment.

Setting those env vars led to a lot of unwanted side effects (e.g., the `-O2` made it hard to get plain debug builds), so over the last year effort was put into "minimally activated compilers", which can be installed with `compilers>2` or `gcc>=15` (Clang version I can't remember off the top of my head). To keep relocation working while no longer population `LDFLAGS` et al., the needed compiler flags were moved directly into gcc/clang spec files. So with the most recent compilers, you still get `-rpath $PREFIX/lib` in the produced extension modules and shared libraries, but that rpath is no longer visible in compiler flags or environment variables.

I believe Homebrew uses spec files too. As do Linux distros in general, but there you typically don't notice because the system loader already has the right search paths, so you don't need extra RPATHs set in spec files. Nix/Spack use compiler wrapper scripts instead; all achieve the same end result of adding an RPATH when invoking a compiler.

In all those cases: you can get extra RPATH entries from a compiler invocation, and that may or may not be observable from either `LDFLAGS` or `build.ninja` contents.

I think the takeaway is that we can test:
- that the expected rpaths added by the test package are present
- that build/destdir directory rpaths are not present
and that extra paths may be present (we cannot reliably know if they were real, tests have to assume they're valid).

tests/test_wheel.py
```diff
@@ -45,6 +46,16 @@ def wheel_contents(artifact):
     }
 
 
+def rpath_from_sysconfig_ldflags():
+    # Account for extra RPATH entries added by compilation flags in
+    # the Python configuration.  This is required for conda/pixi.
```
rgommers, line 51: Flags from `sysconfig` never propagate, that's a `setuptools`-only behavior, so this might be misleading. For conda/pixi, it's (used to be) the environment variable `LDFLAGS` , which only happens to match `sysconfig.get_config_var('LDFLAGS')`, but that isn't guaranteed and will no longer be true if `python` gets rebuilt with "minimally activated compilers".

More importantly, we can't reliably know if extra RPATHs are valid or not, we only know when they're invalid (a build/destdir path or something we added ourselves in a test package).

rgommers: In our own CI we could assert an exact set, because we have control over that. So if we want something like that, like `assert rpaths == set(expected, extra_from_conda_env_or_compiler)`, we can - but it should then be gated on something like `MESONPY_CI_ONLY`.

rgommers:
Some of the most relevant docs/discussion/code:

- conda-forge minimally activated compilers tracking issue: [conda-forge.github.io#2595](https://github.com/conda-forge/conda-forge.github.io/issues/2595) (rationale and design discussion for moving compiler policy out of activation-time environment variables.)
- The [conda-gcc-specs](https://github.com/conda-forge/ctng-compilers-feedstock#about-conda-gcc-specs) README documents how conda-forge's GCC injects include/library paths and RPATH via GCC specs.
- And here is Homebrew doing the same `-rpath`-via-spec-file: [Homebrew/install_steps/formula_actions.rb#L57](https://github.com/Homebrew/brew/blob/2f1c682db046d37c4b6c09aa43837be6ff270c39/Library/Homebrew/install_steps/formula_actions.rb#L57)

dnicolodi:
> * And here is Homebrew doing the same `-rpath`-via-spec-file: [Homebrew/install_steps/formula_actions.rb#L57](https://github.com/Homebrew/brew/blob/2f1c682db046d37c4b6c09aa43837be6ff270c39/Library/Homebrew/install_steps/formula_actions.rb#L57)

Uhm. I wonder how the RPATH tests work without the `sysconfig` hack on Homebrew. Maybe there is something wrong with our Homebrew CI setup...

tests/test_wheel.py
dnicolodi: > Flags from `sysconfig` never propagate

I know. This is just an hack to get to what the pkgconfig contains hoping that it stays in sync with what is in `sysconfig`. I don't like this either, but I wanted to keep the tests as strict as possible and I was unable to come up with a nicer solution.

dnicolodi:
> I think the takeaway is that we can test:
> 
> that the expected rpaths added by the test package are present
> 
> that build/destdir directory rpaths are not present and that extra paths may be present (we cannot reliably know if they were real, tests have to assume they're valid).

Checking that the build paths are not present requires computing the build rpaths, which is not always straightforward. Furthermore, sometimes the build paths are the same as the expected rpaths, making the things even more complex.

I think we can assume that any rpath added by the build environment is an absolute path. Thus we can simply filter these out and keep testing that only the expected rpaths are there. What do you think?

rgommers:
> Maybe there is something wrong with our Homebrew CI setup...

That code I linked is Homebrew on Linux, in a function named `run_configure_gcc_runtime`. There's nothing wrong in CI which is macOS; we could test Homebrew on Linux as well, but I'm not sure how much it adds - I've never seen it used. If we're thinking about expanding, I'd personally be more interested in Nix. But I don't think either of them are urgent, our platform coverage is pretty good.

> I think we can assume that any rpath added by the build environment is an absolute path. Thus we can simply filter these out and keep testing that only the expected rpaths are there. What do you think?

That seems like a good approach for the test suite. 

I'd make one addition: before filtering out absolute paths, reject any that point into the source or build directory. We should always be able to know those: the build directory is either explicitly supplied by `-Cbuild-dir` or created inside the source tree. That catches absolute build path leaking in without having to do anything complicated. We can then ignore other absolute paths from the build environment and check the remaining relative paths against the expected ones. 

Of course there may be a test that explicitly adds `-Wl,-rpath=path-to-some-linked-lib`; that test package is then expected to have an absolute path. I.e., the "filter out all abspaths" may need an exception for a specific test.

I'll note that I was just polishing a test package for path ordering and (lack of) duplicates, so we can't only compare sets in general. I hope to finish that later today.

tests/test_wheel.py
rgommers: I don't think keeping the test strict is essential, but if you do want that, then I think it's correct to read `os.environ['LDFLAGS']`. The current `sysconfig.get_config_var('LDFLAGS')` happens to match today, but I expect it to start diverging as soon as conda-forge's `python` is rebuilt with the minimally activated compilers.

dnicolodi: `os.environ['LDFLAGS']` is the first thing I tried, but it is not set.

dnicolodi:
> That code I linked is Homebrew on Linux

Right. I didn't pay attention.

> I'd make one addition: before filtering out absolute paths, reject any that point into the source or build directory. We should always be able to know those: the build directory is either explicitly supplied by `-Cbuild-dir` or created inside the source tree. That catches absolute build path leaking in without having to do anything complicated. We can then ignore other absolute paths from the build environment and check the remaining relative paths against the expected ones.

This seems like a good approach. This can be extended to check for duplicates at the same time. I'll see how complex it is too do this.

> Of course there may be a test that explicitly adds `-Wl,-rpath=path-to-some-linked-lib`; that test package is then expected to have an absolute path. I.e., the "filter out all abspaths" may need an exception for a specific test.

There are just two tests that strictly check all rpath entries. I don't think it is necessary to extend it to all tests. The other tests are builds of the same package which add some rpaths via env vars or command line flags.

rgommers:
> There are just two tests that strictly check all rpath entries. I don't think it is necessary to extend it to all tests. 

Agreed

docs/how-to-guides/shared-libraries.rst
```diff
@@ -80,8 +76,9 @@ the Python package itself:
         install_dir: py.get_install_dir() / 'mypkg/subdir',
     )
 
-Then an extension module in the same install directory can link against the
-shared library in a portable manner by using ``install_rpath``:
+On Linux, macOS, and in general on all systems except Windows, an extension
+module can dynamically link against the shared library installed in the same
+directory by using ``install_rpath``:
```
rgommers: The "same directory" is slightly too specific, since `$ORIGIN/subdir` will also work and be common. So perhaps use something like: "in the same package ... by using `install_rpath` as a relative path with a local anchor (`$ORIGIN` or `@loader_path`)."

docs/reference/meson-compatibility.rst
```diff
@@ -52,10 +52,21 @@ versions.
    populate the package license and license files from the ones
    declared via the ``project()`` call in ``meson.build``.
 
+   Meson 1.6.0 or later is also required to support the
+   ``install_rpath`` argument to Meson functions that accept it, such
+   as ``library()`` and ``extension_module()`` from the ``python``
```
rgommers: nit: I think you can leave off "from the `python` Meson module" (we use it in other places in the docs as well without this).

mesonpy/_rpath.py
```diff
@@ -34,17 +91,24 @@ def _get_rpath(filepath: Path) -> List[str]:
                 rpath_tag = False
         return rpath
 
-    def _replace_rpath(filepath: Path, old: str, new: str) -> None:
-        subprocess.run(['install_name_tool', '-rpath', old, new, os.fspath(filepath)], check=True)
+    @staticmethod
+    def set_rpath(filepath: Path, old: List[str], rpath: List[str]) -> None:
+        # This implementation does not preserve the ordering of RPATH
```
rgommers: Ah, one of the test packages I wrote verifies this, and hence is failing. It kinda matters if an internal shared library has a name that matches something in an external directory that's also an rpath.

mesonpy/_rpath.py
```diff
@@ -34,17 +91,24 @@ def _get_rpath(filepath: Path) -> List[str]:
...
+        # entries. Meson does the same, thus it should not be a problem.
+        args: List[str] = []
+        for path in rpath:
+            if path not in old:
+                args += ['-add_rpath', path]
+        for path in old:
+            if path not in rpath:
+                args += ['-delete_rpath', path]
+        subprocess.run(['install_name_tool', *args, os.fspath(filepath)], check=True)
```
rgommers: If you guard this with `if args:`, then `install_name_tool` won't be invoked when there's nothing to do (it will fail on that).

mesonpy/_rpath.py
```diff
@@ -11,18 +11,75 @@
...
+    def fix_rpath(cls, filepath: Path, add: List[str], remove: List[str], libs_relative_path: Optional[str]) -> None:
+        old_rpath = cls.get_rpath(filepath)
+
+        # Meson adds a padding entry to RPATH composed of enough `X`
+        # characters to reserve enough space in the ELF header to hold
+        # the final installation RPATH. Remove this entry and other
+        # entries to be removed.
+        new_rpath = [path for path in old_rpath if path.strip('X') and path not in remove]
+
+        # When an executable, libray, or Python extension module is
```
rgommers: typo: `libray` -> `library`

tests/packages/sharedlib-in-package/src/lib.c
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: 2022 The meson-python developers
+//
+// SPDX-License-Identifier: MIT
+
+#include "lib.h"
+#include "sublib.h"
+
+int prodsum(int a, int b, int x) {
+    return prod(a,  x) + b;
```
rgommers: nit: there are two spaces before `x`

docs/reference/meson-compatibility.rst
```diff
@@ -52,10 +52,21 @@ versions.
...
+   Meson module. On older Meson versions, this argument has no effect.
+
 .. option:: 1.9.0
 
    Meson 1.9.0 or later is required to support building for iOS.
 
+   Meson 1.9.0 or later is also required to remove RPATH entries that
+   are added by Meson to allow to execute parts of the project from
```
rgommers: minor textual: "to execute" -> "executing"

docs/reference/meson-compatibility.rst
```diff
@@ -52,10 +52,21 @@ versions.
...
+   the build directory and that are normally removed during ``meson
+   install``. On older Meson versions, these entries are not removed,
+   but this should not have any adverse effect.
```
rgommers: I'd consider removing "but this should not have any adverse effect", because technically that's not true: it can affect which library gets loaded, and distro tooling that checks RPATHs may complain about it.

rgommers (commented):
This looks quite good overall. Aside from the `pixi` test issue which we're discussing, I added some inline comments, all but one are minor (the more important one is the `install_name_tool` one, don't invoke without arguments). After extensive testing, I found that with the anchor translation in place, the current PR passes the checks for all the known real-world packages I had concerns about, and there are only some smaller issues left. I'll list them here, in what I think is order of importance:

1. Whitespace in paths on macOS
2. Installation path precedence (commented on inline, in the code comment that already touched on this)
3. Repeated wheel builds failing in some cases when `install_rpath` contains multiple paths and the order after editing RPATHs changes (the fix is a one-liner, pointed out inline).
4. Transitive RPATH lookup (extmodule -> middle -> leaf): the test also fails on main, so this is not a demonstrated regression for this package. The PR converts `RPATH` to `RUNPATH`, which breaks indirect dependency lookup on glibc. This could affect real packages, and we’ve had issues with the distinction before, but I’m happy for this to be left for a follow-up if it turns out to not be a simple fix.
5. Duplicate build-only `LC_RPATH` removal: if there's a library like that as build input, it produces repeated identical `-delete_rpath` arguments in one invocation, which `install_name_tool` rejects. 

I think 1-3 should be fixed; 4 and 5 may be deferred if it takes more work; I suspect 4 would require the largest change.

I prepared 3 test packages that exercise these issues (that seemed like the minimum number of packages that made sense, and I liked it better than 5 packages), see https://github.com/rgommers/meson-python/tree/rpath-test-pkgs. Details:

- `special-library-paths`: exercises the three macOS-specific issues at once (items 1, 3, 5 above). The test for (5) requires most work, so if we defer that case, we can omit the duplicate-preparation step.
- `same-name-sharedlibs`: exercises issue 2 above
- `sharedlib-chain`: exercises issue 4 above

tests/test_wheel.py
rgommers: Ah yes, we're already using the minimally activated compilers. 

<details>

```
$ # locally with GCC 14.3.0
$ pixi ls gcc
Installed for: linux-aarch64
Name                        Version  Build               Size  Kind   Source
conda-gcc-specs             14.3.0   hadff5d6_18    30.74 KiB  conda  https://prefix.dev/conda-forge
gcc                         14.3.0   h2e72a27_18    28.75 KiB  conda  https://prefix.dev/conda-forge
gcc_impl_linux-aarch64      14.3.0   h533bfc8_18    65.95 MiB  conda  https://prefix.dev/conda-forge
gcc_linux-aarch64           14.3.0   h140ef2e_24    28.16 KiB  conda  https://prefix.dev/conda-forge
libgcc                      15.2.0   h8acb6b2_18   608.30 KiB  conda  https://prefix.dev/conda-forge
libgcc-devel_linux-aarch64  14.3.0   h25ba3ff_118    2.23 MiB  conda  https://prefix.dev/conda-forge
libgcc-ng                   15.2.0   he9431aa_18    26.92 KiB  conda  https://prefix.dev/conda-forge


$ pixi run python -c "import os; print(os.environ['LDFLAGS'])"
-Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--allow-shlib-undefined -Wl,-rpath,/workspace/.pixi/envs/default/lib -Wl,-rpath-link,/workspace/.pixi/envs/default/lib -L/workspace/.pixi/envs/default/lib


$ # change to the lock file we use in CI
$ cd .github/workflows/

$ pixi ls gcc
Installed for: linux-aarch64
Name                        Version  Build               Size  Kind   Source
conda-gcc-specs             15.3.0   hbdd0822_4     31.66 KiB  conda  https://prefix.dev/conda-forge
gcc                         15.3.0   hfdd745d_4     28.78 KiB  conda  https://prefix.dev/conda-forge
gcc_impl_linux-aarch64      15.3.0   hfe63468_4     69.58 MiB  conda  https://prefix.dev/conda-forge
libgcc                      16.2.0   h205dda4_4    613.10 KiB  conda  https://prefix.dev/conda-forge
libgcc-devel_linux-aarch64  15.3.0   h6e7e4e0_104    2.24 MiB  conda  https://prefix.dev/conda-forge


$ pixi run python -c "import os; print(os.environ['LDFLAGS'])"
    import os; print(os.environ['LDFLAGS'])
                     ~~~~~~~~~~^^^^^^^^^^^
  File "<frozen os>", line 709, in __getitem__
KeyError: 'LDFLAGS'
```

</details>

The `os.environ` contents are way cleaner. In CI it's:

```
$ pixi run python -c "import os; print(os.environ)"
environ({'CCACHE_DIR': '/ccache', 'CONDA_DEFAULT_ENV': 'meson-python', 'CONDA_PREFIX': '/workspace/.github/workflows/.pixi/envs/default', 'CONDA_SHLVL': '1', 'DEBIAN_FRONTEND': 'noninteractive', 'DISABLE_AUTOUPDATER': '1', 'DISABLE_ERROR_REPORTING': '1', 'DISABLE_NON_ESSENTIAL_MODEL_CALLS': '1', 'DISABLE_TELEMETRY': '1', 'HOME': '/root', 'INIT_CWD': '/workspace/.github/workflows', 'PATH': '/workspace/.github/workflows/.pixi/envs/default/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin', 'PIXI_CACHE_DIR': '/workspace/.pixi', 'PIXI_COLOR': 'always', 'PIXI_ENVIRONMENT_NAME': 'default', 'PIXI_ENVIRONMENT_PLATFORMS': 'linux-aarch64', 'PIXI_EXE': '/usr/local/bin/pixi', 'PIXI_IN_SHELL': '1', 'PIXI_PROJECT_MANIFEST': '/workspace/.github/workflows/pixi.toml', 'PIXI_PROJECT_NAME': 'meson-python', 'PIXI_PROJECT_ROOT': '/workspace/.github/workflows', 'PIXI_PROJECT_VERSION': 'NO_VERSION_SPECIFIED', 'PIXI_PROMPT': '(meson-python) ', 'PWD': '/workspace/.github/workflows', 'SHLVL': '0', 'TERM': 'xterm', '_': '/usr/local/bin/pixi', 'LC_CTYPE': 'C.UTF-8'})
```

While with GCC 14.3.0 it looks like:
```
pixi run -e default python -c "import os; print(os.environ)"
environ({'ADDR2LINE': 'aarch64-conda-linux-gnu-addr2line', 'AR': 'aarch64-conda-linux-gnu-ar', 'AS': 'aarch64-conda-linux-gnu-as', 'BUILD': 'aarch64-conda-linux-gnu', 'CC': '/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-cc', 'CCACHE_DIR': '/ccache', 'CC_FOR_BUILD': '/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-cc', 'CFLAGS': '-ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O3 -pipe -isystem /workspace/.pixi/envs/default/include', 'CMAKE_ARGS': '-DCMAKE_AR=/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-ar -DCMAKE_CXX_COMPILER_AR=/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-gcc-ar -DCMAKE_C_COMPILER_AR=/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-gcc-ar -DCMAKE_RANLIB=/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-ranlib -DCMAKE_CXX_COMPILER_RANLIB=/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-gcc-ranlib -DCMAKE_C_COMPILER_RANLIB=/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-gcc-ranlib -DCMAKE_LINKER=/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-ld -DCMAKE_STRIP=/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-strip -DCMAKE_BUILD_TYPE=Release', 'CMAKE_PREFIX_PATH': '/workspace/.pixi/envs/default:/workspace/.pixi/envs/default/aarch64-conda-linux-gnu/sysroot/usr', 'CONDA_BUILD_CROSS_COMPILATION': '', 'CONDA_BUILD_SYSROOT': '/workspace/.pixi/envs/default/aarch64-conda-linux-gnu/sysroot', 'CONDA_DEFAULT_ENV': 'meson-python', 'CONDA_PREFIX': '/workspace/.pixi/envs/default', 'CONDA_SHLVL': '1', 'CONDA_TOOLCHAIN_BUILD': 'aarch64-conda-linux-gnu', 'CONDA_TOOLCHAIN_HOST': 'aarch64-conda-linux-gnu', 'CPP': '/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-cpp', 'CPPFLAGS': '-DNDEBUG -D_FORTIFY_SOURCE=2 -O2 -isystem /workspace/.pixi/envs/default/include', 'CPP_FOR_BUILD': '/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-cpp', 'CXXFILT': 'aarch64-conda-linux-gnu-c++filt', 'DEBIAN_FRONTEND': 'noninteractive', 'DEBUG_CFLAGS': '-ftree-vectorize -fPIC -fstack-protector-all -fno-plt -Og -g -Wall -Wextra -fvar-tracking-assignments -pipe -isystem /workspace/.pixi/envs/default/include', 'DEBUG_CPPFLAGS': '-D_DEBUG -D_FORTIFY_SOURCE=2 -Og -isystem /workspace/.pixi/envs/default/include', 'DISABLE_AUTOUPDATER': '1', 'DISABLE_ERROR_REPORTING': '1', 'DISABLE_NON_ESSENTIAL_MODEL_CALLS': '1', 'DISABLE_TELEMETRY': '1', 'ELFEDIT': 'aarch64-conda-linux-gnu-elfedit', 'GCC': '/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-gcc', 'GCC_AR': '/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-gcc-ar', 'GCC_NM': '/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-gcc-nm', 'GCC_RANLIB': '/workspace/.pixi/envs/default/bin/aarch64-conda-linux-gnu-gcc-ranlib', 'GPROF': 'aarch64-conda-linux-gnu-gprof', 'HOME': '/root', 'HOST': 'aarch64-conda-linux-gnu', 'INIT_CWD': '/workspace/docs', 'LD': 'aarch64-conda-linux-gnu-ld', 'LDFLAGS': '-Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--allow-shlib-undefined -Wl,-rpath,/workspace/.pixi/envs/default/lib -Wl,-rpath-link,/workspace/.pixi/envs/default/lib -L/workspace/.pixi/envs/default/lib', 'LDFLAGS_LD': '-O2 --sort-common --as-needed -z relro -z now --allow-shlib-undefined -rpath /workspace/.pixi/envs/default/lib -rpath-link /workspace/.pixi/envs/default/lib -L/workspace/.pixi/envs/default/lib', 'MESON_ARGS': '-Dbuildtype=release', 'NM': 'aarch64-conda-linux-gnu-nm', 'OBJCOPY': 'aarch64-conda-linux-gnu-objcopy', 'OBJDUMP': 'aarch64-conda-linux-gnu-objdump', 'PATH': '/workspace/.pixi/envs/default/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin', 'PIXI_CACHE_DIR': '/workspace/.pixi', 'PIXI_COLOR': 'always', 'PIXI_ENVIRONMENT_NAME': 'default', 'PIXI_ENVIRONMENT_PLATFORMS': 'linux-64,linux-aarch64,osx-arm64', 'PIXI_EXE': '/usr/local/bin/pixi', 'PIXI_IN_SHELL': '1', 'PIXI_PROJECT_MANIFEST': '/workspace/pixi.toml', 'PIXI_PROJECT_NAME': 'meson-python', 'PIXI_PROJECT_ROOT': '/workspace', 'PIXI_PROJECT_VERSION': '0.19.0.dev0', 'PIXI_PROMPT': '(meson-python) ', 'PWD': '/workspace/docs', 'RANLIB': 'aarch64-conda-linux-gnu-ranlib', 'READELF': 'aarch64-conda-linux-gnu-readelf', 'SHLVL': '0', 'SIZE': 'aarch64-conda-linux-gnu-size', 'STRINGS': 'aarch64-conda-linux-gnu-strings', 'STRIP': 'aarch64-conda-linux-gnu-strip', 'TERM': 'xterm', '_': '/usr/local/bin/pixi', 'build_alias': 'aarch64-conda-linux-gnu', 'host_alias': 'aarch64-conda-linux-gnu', 'LC_CTYPE': 'C.UTF-8'})
```

The conda-forge `python` we're using in CI was built with that older, much more polluted, set of env vars. If it gets rebuilt, then `sysconfig.get_config_vars` should start missing all `LDFLAGS`/`CFLAGS` content that's in the delta between the above two `os.environ` contents.

The tests would actually fail when `python` gets updated to a build with GCC 15 in conda-forge; they'll likely also fail in Homebrew on Linux. I don't see a good way to do this, so I suggest to make the assertion less strict and drop this utility function.

dnicolodi:
Thanks for the extensive review and testing @rgommers!

> 1\. Whitespace in paths on macOS

I haven't tested this yet, and it should not be difficult to do. In case you already debugged this, can you point to where it break?

> 5\. Duplicate build-only `LC_RPATH` removal: if there's a library like that as build input, it produces repeated identical `-delete_rpath` arguments in one invocation, which `install_name_tool` rejects.

Does this require a specially crafted project or is it possible to bump into this accidentally in a common scenario?

> I prepared 3 test packages that exercise these issues (that seemed like the minimum number of packages that made sense, and I liked it better than 5 packages), see https://github.com/rgommers/meson-python/tree/rpath-test-pkgs. 

Great. I'll pull these tests into this PR.

docs/reference/meson-compatibility.rst
dnicolodi: This comment is based on the fact that now meson-python does not do the removal and no one complained 🙂

rgommers:
I'm about to sign off, so very quick comments: I didn't yet debug the whitespace in paths issue, and for (5) yes I think it requires a bit of effort to create a situation with a duplicate LC_RPATH, so I think it'd be fine to drop that case - it certainly won't be common. It's kinda hard to reason through such cases and I might be missing an easier way to set up that situation, but if you look at the test packages, that's the only case that feels artificial (I suspect other tools could produce such situations, but I couldn't find a simpler way).

docs/how-to-guides/shared-libraries.rst
dnicolodi: Happy to change this, but this was the existing wording.

mesonpy/_rpath.py
dnicolodi: This cannot really happen because `set_rpath` is only called if `rapth` differs from `old`. But adding the check does not cost much.

dnicolodi: I understood now what you meant: because here we do not take order into consideration, it may be that there is no argument passed to `install_name_tool` when the old and new RAPTH differ only in the order of the elements. I took me a while to grok your comment, sorry.

Anyhow, we should really fix the ordering. However, `install_name_tool` has a really weird interface, thus coming up with a set of operation that transforms a set of RPATHs into another is not trivial. On the other hand, Meson itself does not guarantee ordering of the RPATH entries, thus I am not sure how important this is in practice.

rgommers: Yeah I'm also not sure how important it is in practice, but given that the guard is very cheap (and you already applied the fix), I suggest we don't lose sleep over it:)

dnicolodi: I was referring to the ordering of entries in RPATH. This is not solved.

dnicolodi: Now it is 🙂

mesonpy/_rpath.py
dnicolodi: Fixed

dnicolodi:
> 5\. Duplicate build-only `LC_RPATH` removal: if there's a library like that as build input, it produces repeated identical `-delete_rpath` arguments in one invocation, which `install_name_tool` rejects.

I tried to come up with ways to produce a binary with duplicated `LC_RPATH` entries on macOS. Other than the binary patching used in the rest @rgommers proposes, I only found the use of `install_name_tool` to replace one entry with another identical to one already present. I think the first is evil and if someone does that in their build they deserve the consequent build failure. We are guilty of the second in meson-python main, but this PR fixes the underlaying issue. Still, I don't think that any sensible build should do, thus I am tempted to say that we should not go out of our way to support.

rgommers:
I'm fine with dropping that, and only revisiting in case someone comes up with a reasonable-looking real-world problem.

dnicolodi:
> I prepared 3 test packages that exercise these issues (that seemed like the minimum number of packages that made sense, and I liked it better than 5 packages), see https://github.com/rgommers/meson-python/tree/rpath-test-pkgs. Details:
> 
> * `special-library-paths`: exercises the three macOS-specific issues at once (items 1, 3, 5 above). The test for (5) requires most work, so if we defer that case, we can omit the duplicate-preparation step.

I implemented the tests for correct manipulation of RPATHs with synthetic tests in `test_rpath.py`. It is much easier this way to cover many cases. These tests can be made to cover duplicated entries too, but this case seems artificial and I would leave this complication out for now.

> * `same-name-sharedlibs`: exercises issue 2 above

I pulled this in and updated it to work on macOS too. However, I think there is a problem with the test. We have:
```
link_args: '-Wl,-rpath,$ORIGIN/competing'
install_rpath: '$ORIGIN/private'
build_rpath: '$ORIGING/private' # from ``link_with: private``
```
Meson adds linker flags to add the build rpath and then the user link args. This results in an the extension module has an RPATH entry `$ORIGIN/private:$ORIGIN/competing`. The build rpath is added to the introspection data as rpath entries that should be removed on install, and the install_rpath is added to the introspection as rpath entries that should be appended. When meson-python obliges, the resulting RPATH is `$ORIGIN/competing:$ORIGIN/private`, but the test expects them to be in the reverse order. I think the expectation is wrong, unless we want install_rpath to be prepended to RPATH.

The test works form Meson versions prior to 1.9 because there we do not have Meson metadata to remove build rpaths, thus the final RPATH becomes `$ORIGIN/private:$ORIGIN/competing:$ORIGIN/private` which gets deduplicated into `$ORIGIN/competing:$ORIGIN/private`.

TL;DR: rpath entries from `link_args` always come before `install_rpath` entries.

dnicolodi:
> TL;DR: rpath entries from `link_args` always come before `install_rpath` entries.

At a closer look, I found that on install, Meson prepends `install_rpath` entries to existing entries. Obviously meson-python should do the same. This fixes the issue!

dnicolodi:
> At a closer look, I found that on install, Meson prepends `install_rpath` entries to existing entries. Obviously meson-python should do the same. This fixes the issue!

Fixed

dnicolodi:
> 4\. Transitive RPATH lookup (extmodule -> middle -> leaf): the test also fails on main, so this is not a demonstrated regression for this package. The PR converts `RPATH` to `RUNPATH`, which breaks indirect dependency lookup on glibc. This could affect real packages, and we’ve had issues with the distinction before, but I’m happy for this to be left for a follow-up if it turns out to not be a simple fix.

This is the only thing discussed here that remains to be addressed. I pulled in the test, which fails as expected. However, I do not understand what the expectations are, nor how it would be possible to support this use case.

dnicolodi:
> 4\. Transitive RPATH lookup (extmodule -> middle -> leaf): the test also fails on main, so this is not a demonstrated regression for this package. The PR converts `RPATH` to `RUNPATH`, which breaks indirect dependency lookup on glibc. This could affect real packages, and we’ve had issues with the distinction before, but I’m happy for this to be left for a follow-up if it turns out to not be a simple fix.

I spent a bit more time on this and I have a better understanding. The test package builds a `chain` extension module that links with a `middle` shared library, is installed in `$platlib/chainhead/` and has install rpath set to `$ORIGIN/lib` and linker arguments forcing the use of `DT_RPATH` instead of `DT_RUNPATH`. The `middle` shared library links with a `leaf` shared library and is installed in `$platlib/chainhead/lib/`. The `leaf` shared library is also installed in `$platlib/chainhead/lib/`.

The use of `DT_RPATH` in the `chain` extension module should propagate the shared library search path along the chain, thus the `middle` shared library does not need to set an RPATH to find the `leaf` shared library.

This breaks because `patchelf` converts `DT_RPATH` entries into `DT_RUNPATH` entries and the latter do not propagate.

`patchelf` has `--force-rpath` command line flag that forces the use of `DT_RPATH` instead of `DT_RUNPATH` https://manpages.debian.org/unstable/patchelf/patchelf.1.en.html#force-rpath. However, this is not what we need because converting `DT_RUNPATH` to `DT_RPATH` is not recommended either. We need a way to preserve the type of tag already present. I don't think there is a way to do this with current `patchelf`.

Relevant `patchelf` code:
https://github.com/NixOS/patchelf/blob/7688b17c18d16f67fa8d5a82a2404c2e3a18648d/src/patchelf.cc#L3124-L3137 and https://github.com/NixOS/patchelf/blob/7688b17c18d16f67fa8d5a82a2404c2e3a18648d/src/patchelf.cc#L1835-L1845

dnicolodi:
Moved TST: exercise transitive RPATH lookup with a shared library chain to #905.

@rgommers I think all issues and comments have been addressed and this is ready to be merged.

PR #794, opened 2025-09-21
DOC: add instructions for cibuildwheel to support Windows wheels

tobiasdiez:
I had trouble with building wheels ("Python executable could not be found") and fixed it by specifying the `vsenv` option for cibuildwheel. A similar flag is used by numpy as well, so this is probably indeed needed.

### DIFF
--- a/docs/how-to-guides/meson-args.rst
+++ b/docs/how-to-guides/meson-args.rst
@@ -190,6 +190,14 @@ permanently in the project's ``pyproject.toml``:
    [tool.meson-python.args]
    setup = ['--vsenv']
 
+And in case you don't want to set ``--vsenv`` as the default, but do want to
+build Windows wheels with MSVC in CI using cibuildwheel:
+
+.. code-block:: toml
+
+   [tool.cibuildwheel.windows]
+   config-settings = { setup-args = ["--vsenv"] }
+
 To set this option temporarily at build-time:
 
 .. tab-set::


docs/how-to-guides/meson-args.rst
```diff
@@ -190,6 +190,10 @@ permanently in the project's ``pyproject.toml``:
    [tool.meson-python.args]
    setup = ['--vsenv']
 
+   # In case you want to build Windows wheels using cibuildwheel
+   [tool.cibuildwheel.windows]
+   config-settings = { setup-args = ["--vsenv"] }
```
moi15moi, line 195: I think this won't properly work when you are building the wheel on cygwin/msys2

tobiasdiez: That is true. But since the whole section was about MSVC, I saw little risk of someone adding this option while actually building the wheel on msys2. Do you think this needs further clarification, eg in the comment above the snippet?

rgommers (commented):
Thanks @tobiasdiez. This is helpful, I've seen other people need this. I suggest one tweak.

docs/how-to-guides/meson-args.rst
```diff
@@ -190,6 +190,10 @@ permanently in the project's ``pyproject.toml``:
    [tool.meson-python.args]
    setup = ['--vsenv']
 
+   # In case you want to build Windows wheels using cibuildwheel
```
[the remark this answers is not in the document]
rgommers: I applied this change.

PR #797, opened 2025-10-18
ENH: warn on unhandled meson install arguments

dnicolodi:


dnicolodi:
`meson-python` actually runs `meson {setup,compile,dist}` thus if the passed arguments are not valid, the build fails. However, `meson install` is never executed and `meson-python` directly interprets the arguments. `meson-python` does not implement all the arguments supported by `meson install`. So far `meson python` ignored the arguments it does not recognize, but this is problematic for two reasons: it makes spotting typos in supported arguments much harder, and there is no easy way to discover unsupported arguments. We can error our on unrecognized arguments, but it may cause forward compatibility issues for project that have been passing invalid arguments unnoticed.

rgommers (commented):
Of course, that's a good rationale. In it goes, thanks @dnicolodi!

PR #798, opened 2025-10-18
DOC: minor tweaks

dnicolodi:


docs/how-to-guides/config-settings.rst
```diff
@@ -67,9 +67,9 @@ Using a persistent build directory
 By default, ``meson-python`` uses a temporary build directory which is
 deleted when the build terminates. A persistent build directory allows
 faster incremental builds and to access build logs and intermediate
-build artifacts. The ``build-dir`` :ref:`config setting
-<reference-config-settings>` instructs ``meson-python`` to use a
-user-specified build directory which will not be deleted. For example:
+build artifacts. The :option:`build-dir` config setting can be used to
+build in an user-specified build directory which will not be
```
rgommers: nit: it should be "a user-specified", the original was correct here

docs/how-to-guides/config-settings.rst
```diff
@@ -114,7 +114,7 @@ shows the detailed Meson setup log when the build fails:
     - name: Build the package
       run: python -m build --wheel -Cbuild-dir=build
     - name: Show meson-log.txt
-      if: failure()
+      if: ${{ failure() }}
```
rgommers: I think both versions work, but GHA docs now consistently use the `${{ }}` syntax so this is a good change.

rgommers (approved):
Thanks @dnicolodi. One minor grammar thing, other than that this LGTM

PR #799, opened 2025-10-18
Bump required Python version to 3.9 and switch from extra dependencies to dependency groups

dnicolodi:


dnicolodi:
Dependency groups are supported starting with pip 25.1 which dropped support for Python 3.8, thus this will have to wait till we drop support for Python 3.8 or start using uv to manage the test environments.

rgommers:
Python 3.8 has been EOL for over a year, and Python 3.9 is also EOL now (see https://devguide.python.org/versions/). So I think dropping 3.8 now is perfectly fine.

dnicolodi:
@rgommers I opted for dropping Python 3.8 from the CI jobs but not bump the required Python version. Should we bump `requires-python` to `>= 3.9`?

rgommers:
> Should we bump `requires-python` to `>= 3.9`?

I think so, yes. 3.8 support will not go anywhere for 0.18.0, and almost no one needs it anymore so I doubt future bug fixes are going to matter. So if we're dropping CI jobs, bumping `requires-python` is preferred imho.

eli-schwartz:
> I think so, yes. 3.8 support will not go anywhere for 0.18.0, and almost no one needs it anymore so I doubt future bug fixes are going to matter. So if we're dropping CI jobs, bumping `requires-python` is preferred imho.

:+1: to explicitly "breaking it" if you decline to test it. Meson does support 3.7 though I suppose most meson-python users are less concerned about backporting to LTS distros.

rgommers (approved):
All LGTM, in it goes.

PR #803, opened 2025-10-25
DOC: document how to use installed data files with editable installs

dnicolodi:
Replaces #786.


docs/how-to-guides/editable-installs.rst
```diff
@@ -108,6 +108,74 @@ An alternative build directory can be specified using the
...
+when the package is installed in editable mode.  For this reason, data
+files need to be accessed using :mod:`importlib.resources`. The code
+above should be replaced with:
+
+.. code-block:: python
+
+   import importlib.resources
+
+   data = importlib.resources.files().joinpath('data.txt').read_text()
+   uuid = importlib.resources.files().joinpath('uuid.txt').read_text()
```
rgommers: Using `files()` without an argument actually doesn't work in Python <3.12 (see [change/deprecation note [here](https://docs.python.org/3.14/library/importlib.resources.html#importlib.resources.files)). It needs a package reference as the argument - that, or use the functional `importlib.resources.read_text`.

tests/packages/install-data/__init__.py
```diff
@@ -0,0 +1,13 @@
+# SPDX-FileCopyrightText: 2025 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+import importlib.resources
+import pathlib
+
+
+data = pathlib.Path(__file__).parent.joinpath('data.txt').read_text()
+uuid = pathlib.Path(__file__).parent.joinpath('uuid.txt').read_text() # WRONG!
```
rgommers: Remove these two lines?

rgommers (commented):
Thanks @dnicolodi. The added text about how to use data files LGTM (modulo the minor open comments) and is useful.

It doesn't quite address the "can't use the NumPy C API from another package when NumPy itself is installed in editable mode". It can be added to this PR, probably as a `.. warning::` note at the end, or left for a separate follow-up.

docs/how-to-guides/editable-installs.rst
dnicolodi: The functional `importlib.resources.read_text()` does not accept an empty anchor either. I think the easiest way is to depend on `importlib_resources` on older Pythons.

dnicolodi: If there is a way to make `importlib.resources` work for this in Python 3.9, I don't see it. If having the example depend on the `importlib-resources` backport is not an option, the only thing to do is to note that the example does not work for old Python versions and suggest the backport as a solution. I've done this in the last iteration.

rgommers: A code comment stating that seems fine to me - Python 3.9 is end-of-life already and most packages have dropped support already or will do so soon.

rgommers:
> Anything more specific would be more discoverable in the NumPy documentation or in a new "Papercuts" section of the documentation (maybe started by renaming the current "Limitations" section).

Fair enough. And I like the Papercuts name. I can have a look at it, also seeing if/where it needs a warning in the NumPy docs.

docs/how-to-guides/editable-installs.rst
dnicolodi: I added a `.. warning`. Do you prefer that to be just a code comment?

rgommers:
I’d personally go that way yes. But don’t feel super strongly about it either Sent from my iPhoneOn 26 Oct 2025, at 10:21 PM, Daniele Nicolodi ***@***.***> wrote:﻿
@dnicolodi commented on this pull request.



In docs/how-to-guides/editable-installs.rst:
> +   uuid = pathlib.Path(__file__).parent.joinpath('uuid.txt').read_text()  # WRONG!
+
+However, this does not work when modules are not loaded from a package
+installed in the Python library path in the filesystem but with a
+special module loader, as used to implement editable installs in
+``meson-python``. In the example above, the second read would fail
+when the package is installed in editable mode.  For this reason, data
+files need to be accessed using :mod:`importlib.resources`. The code
+above should be replaced with:
+
+.. code-block:: python
+
+   import importlib.resources
+
+   data = importlib.resources.files().joinpath('data.txt').read_text()
+   uuid = importlib.resources.files().joinpath('uuid.txt').read_text()

I added a .. warning. Do you prefer that to be just a code comment?

PR #806, opened 2025-10-28
DOC: remove example

dnicolodi:
It is not linked from anywhere, it is simple but implemented in a quite convoluted way, it is not tested, and we have many similar ones in the test suite.


rgommers (commented):
The example wasn't used already in the rendered docs, so this removal is fine from that perspective. It is used in the test suite though, so CI is unhappy right now. I don't see anything special about this test package, so just removing `test_spam` seems like the way to go?

PR #815, opened 2025-10-31
MAINT: do not require the pytest-cov toml extra

dnicolodi:
AFAICT it never existed.

This avoids a warning when installing the test dependencies.


rgommers (approved):
Good catch

PR #818, opened 2025-11-01
CI: switch to supported macOS VM image on Cirrus CI

dnicolodi:
Apparently the image previously in use has been retired and the jobs were automatically upgraded to this image. Switch it to avoid possible future breakage.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -84,6 +84,6 @@ freebsd_task:
 
 macos-arm64_task:
   macos_instance:
-    image: ghcr.io/cirruslabs/macos-sonoma-xcode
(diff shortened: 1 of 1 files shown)


rgommers (approved):
LGTM. Job hasn't run yet so I won't hit the merge button straight away, but this should be right.

dnicolodi:
Yeah, I went looking at what's up with the macOS arm64 runner because the jobs are taking a very long time to be scheduled. I thought that the use of the deprecated VM image may have something to do with it, but it seems it does not.

dnicolodi:
CI is green. Merging.

PR #824, opened 2025-12-10
Add support for targeting Android

### LINKED ISSUE #810: Feature Request: Add support for building Android wheels
As per [PEP 738](https://peps.python.org/pep-0738/), Android is an officially supported platform for Python. Python 3.14 releases includes [official android embeddable packages](https://www.python.org/downloads/release/python-3140/).

It would be really appreciated if `meson-python` adds support for Android too, just as iOS support added recently.

mhsmith:
Using the [same approach as on iOS](https://github.com/mesonbuild/meson-python/pull/731).

* Closes gh-810
 
This PR branch is temporarily used by several other PRs: see the cross-references below.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -744,6 +744,38 @@ class Project():
                     self._meson_cross_file.write_text(cross_file_data, encoding='utf-8')
                     self._meson_args['setup'].extend(('--cross-file', os.fspath(self._meson_cross_file)))
 
+        # Simplify cross-compilation for Android with cibuildwheel: detect the
+        # cross-compilation environment set up by cibuildwheel and synthesize an
+        # appropriate cross file.
+        elif sysconfig.get_platform().startswith('android-') and 'CIBUILDWHEEL' in os.environ:
+            cpu = platform.machine()
+            cpu_family = 'x86' if cpu == 'i686' else 'arm' if cpu.startswith('arm') else cpu
+
+            # cibuildwheel uses environment variables to point to the cross
+            # compilation toolchain binaries thus the [binaries] section of
+            # the cross file is not needed.
+            cross_file_data = textwrap.dedent(f'''
+                [host_machine]
+                system = 'android'
+                subsystem = 'android'
+                kernel = 'linux'
+                cpu_family = {cpu_family!r}
+                cpu = {cpu!r}
+                endian = {sys.byteorder!r}
+
+                [properties]
+                # To obtain the correct wheel tags when cross-compiling
+                # cibuildwheel monkey-patches platform.system() and
+                # platform.machine() to simulate running on Android.
+                # This makes Meson believe that host and build machines
+                # match, and thus that host binaries can be run on the
+                # build machine when this is not actually the case.
+                # Override the auto-detection.
+                needs_exe_wrapper = true
+            ''')
+            self._meson_cross_file.write_text(cross_file_data, encoding='utf-8')
+            self._meson_args['setup'].extend(('--cross-file', os.fspath(self._meson_cross_file)))
+
         # Support iOS targets. iOS does not have native build tools and always
         # requires cross compilation: synthesize the appropriate cross file.
         elif sysconfig.get_platform().startswith('ios-'):
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -378,6 +378,53 @@ def test_archflags_envvar_parsing_invalid(package_purelib_and_platlib, monkeypat
         os.environ.pop('_PYTHON_HOST_PLATFORM', None)
 
 
+@pytest.mark.parametrize(
+    ('cpu', 'cpu_family'),
+    [
+        ('aarch64', 'aarch64'),
+        ('armv7l', 'arm'),
+        ('armv8l', 'arm'),
+        ('i686', 'x86'),
+        ('x86_64', 'x86_64'),
+    ],
+)
+@pytest.mark.parametrize('cross', [True, False])
+def test_android_project(package_simple, monkeypatch, tmp_path, cpu, cpu_family, cross):
+    # Mock being on Android
+    monkeypatch.setattr(sys, 'platform', 'android')
+    monkeypatch.setattr(sys, 'byteorder', 'little')
+    monkeypatch.setattr(platform, 'system', Mock(return_value='Android'))
+    monkeypatch.setattr(platform, 'machine', Mock(return_value=cpu))
+    monkeypatch.setattr(sysconfig, 'get_platform', Mock(return_value='android-24'))
+    if cross:
+        monkeypatch.setenv('CIBUILDWHEEL', '1')
+
+        # Meson may require some tools to be configured when fatal warnings are enabled.
+        # Set the same set of variables as cibuildwheel.
+        for name in ['ar', 'as', 'cc', 'cxx', 'ld', 'nm', 'ranlib', 'readelf', 'strip']:
+            monkeypatch.setenv(name.upper(), f'/path/to/{name}')
+
+    # Create a project.
+    project = mesonpy.Project(source_dir=package_simple, build_dir=tmp_path)
+
+    # When cross-compiling, a cross file should be generated and used.
+    setup_args = project._meson_args['setup']
+    cross_path = tmp_path / 'meson-python-cross-file.ini'
+    if cross:
+        assert setup_args[-2:] == ['--cross-file', str(cross_path)]
+        cross_config = cross_path.read_text().splitlines()
+        assert "system = 'android'" in cross_config
+        assert "subsystem = 'android'" in cross_config
+        assert "kernel = 'linux'" in cross_config
+        assert f"cpu_family = '{cpu_family}'" in cross_config
+        assert f"cpu = '{cpu}'" in cross_config
+        assert "endian = 'little'" in cross_config
+        assert 'needs_exe_wrapper = true' in cross_config
+    else:
+        assert '--cross-file' not in setup_args
+        assert not cross_path.exists()
+
+
 @pytest.mark.skipif(sys.version_info < (3, 13), reason='requires Python 3.13 or higher')
 @pytest.mark.parametrize('multiarch', [
     'arm64-iphoneos',


mhsmith:
@rgommers: This PR is ready for review. Compared to the iOS PR, this one is a bit simpler:

* No updates were required to Meson itself.
* No special handling was required for Android versions or wheel tags, since this is all done by cibuildwheel.

rgommers (commented):
Thanks @mhsmith. The main question I have now is: it looks like you're assuming that only `cibuildwheel` will be supported, and not regular cross compilation to Android. Is that correct? If so, I'm not sure that's right - and the iOS support doesn't do that as far as I can tell (or it's more hidden and missing code comments?).

Other questions that came up for me, like what platforms Android supports (e.g., `armv7`) and whether that's supported when using `cpu_family = {platform.machine()!r}` kinda depend on that first question.

mhsmith:
> it looks like you're assuming that only `cibuildwheel` will be supported, and not regular cross compilation to Android

That's a good point: Android packages could be native-compiled in an environment like Termux. I've changed the code to only use cross-compilation mode when it detects it's running within cibuildwheel.

As far as I know, there's no native compilation option on iOS, because the platform is too restrictive, e.g. apps cannot create subprocesses.

> Other questions that came up for me, like what platforms Android supports (e.g., `armv7`) and whether that's supported when using `cpu_family = {platform.machine()!r}`

Officially [Python only supports Android on aarch64 and x86_64](https://peps.python.org/pep-0011/). However, it's used unofficially on other architectures, so I've added coverage for all the known values of `platform.machine` that Android can return.

eli-schwartz:
> That's a good point: Android packages could be native-compiled in an environment like Termux. I've changed the code to only use cross-compilation mode when it detects it's running within cibuildwheel.

Anecdotally, Termux for me shows python 3.12 with a platform value of "Linux". :)

mhsmith:
Yes, that was changed in Python 3.13 when we added official Android support. cibuildwheel will only support Android on Python 3.13 or later, so the cross file should use the new value.

rgommers:
We get the occasional Termux-related bug report in NumPy, so that does appear to be used indeed.

The approach you took now looks better to me. `CIBW_HOST_TRIPLET` doesn't seem to be a public environment variable for `cibuildwheel` though (or is it just missing from the docs?), so it might be safer to pick a variable that is public?

mhsmith:
I'm not aware of any public way to detect that a build is running within cibuildwheel. The environment variables in their documentation are all used as configuration inputs to cibuildwheel, rather than outputs that you can rely on being set within the build.

So I've changed it to use `sys.cross_compiling` instead, which has the advantage of not depending on any particular build tool.

rgommers:
Thanks for the update.

> So I've changed it to use `sys.cross_compiling` instead, which has the advantage of not depending on any particular build tool.

It does depend on tools, right? I guess all we care about here is that `cibuildwheel` sets that. Proper cross-compiling (i.e., provide a cross file, not with hacks like `crossenv`) will not have this set, but as long as we document that it's specific to `cibuildwheel` and `crossenv` I can live with this code, given that there indeed doesn't appear to be an officially supported way to special-case `cibuildwheel`.

mhsmith:
Python doesn't really provide any "proper cross-compiling" mechanism, so the crossenv-like approach is the only feasible option for cross-compiling Android and iOS wheels at the moment.

> as long as we document that it's specific to `cibuildwheel` and `crossenv` I can live with this code

I've mentioned that in the comment on line 749.

rgommers:
Thanks @mhsmith. For context: Python itself doesn't, and probably won't for at least a couple more years, but both Meson and CMake have solid cross-compilation support, and with Meson cross files + PEP 739 (`build-details.json`) cross compilation can start to look pretty clean. E.g., see [this branch for `scipy`](https://github.com/scipy/scipy/compare/main...rgommers:scipy:ci-cross-314-builddetails-json), it comes down to setting up the build and host envs, and then:

```
python -m build -wnx -Csetup-args=--cross-file=$PWD/ci/cross_aarch64.ini -Csetup-args=-Dpython.build_config=$HOST_ENV/lib/python3.14/build-details.json
```

Support in Meson for PEP 739 has landed, and support in `meson-python` will follow (hopefully soon). All that "support in Python" should then mean is "how do we point at the target interpreter (i.e., `$HOST_ENV/lib/python3.14/build-details.json`) in a more ergonomic fashion?".

tl;dr we should not build in implicit assumptions around `crossenv` anywhere in the ecosystem, because that's a pretty bad pile of hacks under the hood, and not supportable properly long-term, or even today by distros.

mesonpy/__init__.py
```diff
@@ -744,6 +744,38 @@ def __init__(
                     self._meson_cross_file.write_text(cross_file_data, encoding='utf-8')
                     self._meson_args['setup'].extend(('--cross-file', os.fspath(self._meson_cross_file)))
 
+        # Android may be native-compiled (e.g. with Termux), or cross-compiled (e.g.
+        # with cibuildwheel), in which case we should generate a cross file.
+        # sys.cross_compiling isn't an official Python API, but it originated from
+        # crossenv and has been followed by other cross-compiling tools.
```
rgommers: Hence I suggest to change the comment along these lines:

```suggestion
Cross compilation can be done the regular Meson way with separate build and host envs,
or with `cibuildwheel` which is `crossenv`-based. Here we try to detect whether
cibuildwheel/crossenv are used, through the `sys.cross_compiling` variable which they
define, and we synthesize a cross file in that case to make that work out of the box.
```

I think the idea and code are fine.

mesonpy/__init__.py
```diff
@@ -744,6 +744,38 @@ def __init__(
...
+        elif (
+            sysconfig.get_platform().startswith('android-')
```
dnicolodi: Why not simply
```suggestion
            sys.platform == 'android'
```
?

mesonpy/__init__.py
```diff
@@ -744,6 +744,38 @@ def __init__(
...
+            and getattr(sys, 'cross_compiling', False)
+        ):
+            cpu = platform.machine()
+            cpu_family = (
+                'arm' if cpu.startswith('arm') else 'x86' if cpu.endswith('86') else cpu
+            )
+
+            cross_file_data = textwrap.dedent(f'''
+                # Binaries are controlled by environment variables, so they don't need
+                # to be repeated here.
```
dnicolodi: I don't think there is any value in adding this comment to the generated file.

mesonpy/__init__.py
```diff
@@ -744,6 +744,38 @@ def __init__(
...
+        # sys.cross_compiling isn't an official Python API, but it originated from
+        # crossenv and has been followed by other cross-compiling tools.
+        elif (
+            sysconfig.get_platform().startswith('android-')
+            and getattr(sys, 'cross_compiling', False)
+        ):
+            cpu = platform.machine()
+            cpu_family = (
+                'arm' if cpu.startswith('arm') else 'x86' if cpu.endswith('86') else cpu
+            )
```
dnicolodi:
```suggestion
            cpu_family =  'arm' if cpu.startswith('arm') else 'x86' if cpu.endswith('86') else cpu
```
This is ugly, but please don't use the `black` code uglifier to make it uglier.

mesonpy/__init__.py
```diff
@@ -744,6 +744,38 @@ def __init__(
...
                     self._meson_args['setup'].extend(('--cross-file', os.fspath(self._meson_cross_file)))
 
+        # Android may be native-compiled (e.g. with Termux), or cross-compiled (e.g.
+        # with cibuildwheel), in which case we should generate a cross file.
+        # sys.cross_compiling isn't an official Python API, but it originated from
+        # crossenv and has been followed by other cross-compiling tools.
+        elif (
+            sysconfig.get_platform().startswith('android-')
+            and getattr(sys, 'cross_compiling', False)
+        ):
```
dnicolodi:
```suggestion
        elif sys.platform == 'android' and getattr(sys, 'cross_compiling', False):
```
Please don't use the `black` code uglifier.

mesonpy/__init__.py
```diff
@@ -744,6 +744,38 @@ def __init__(
...
+                subsystem = 'android'
+                kernel = 'linux'
+                cpu_family = {cpu_family!r}
+                cpu = {cpu!r}
+                endian = {sys.byteorder!r}
+
+                [properties]
+                # cibuildwheel's cross virtual environment will make Meson believe it's
+                # running on Android when it's actually running on Linux or macOS.
+                needs_exe_wrapper = true
```
dnicolodi: I don't get how the comment (which I don't think should be in the generated file) goes with the implemented setting. How does cibuildwheel makes Meson believe it is running on Android? And why does it do so? It is clearly not the case. Also, what it the exe wrapper that is needed? This seems incomplete at best.

tests/test_project.py
```diff
@@ -378,6 +378,48 @@ def test_archflags_envvar_parsing_invalid(package_purelib_and_platlib, monkeypat
...
+@pytest.mark.parametrize('cross', [True, False])
+def test_android_project(package_simple, monkeypatch, tmp_path, cpu, cpu_family, cross):
+    # Mock being on Android
+    monkeypatch.setattr(sys, 'platform', 'android')
+    monkeypatch.setattr(sys, 'byteorder', 'little')
+    monkeypatch.setattr(platform, 'machine', Mock(return_value=cpu))
+    monkeypatch.setattr(sysconfig, 'get_platform', Mock(return_value='android-24'))
+    if cross:
+        monkeypatch.setattr(sys, 'cross_compiling', True, raising=False)
+        monkeypatch.setenv('STRIP', '/path/to/strip')
```
dnicolodi: Why is setting this environment variable needed?

dnicolodi (changes requested):
I am not sure I understand what this PR implements. The commit message states that this implements support for Android. However, it seems to implement support for *cross-compiling* for Android. I think native compilation for Android works already as I don't have evidence of the contrary and this PR does not affect native compilation.

Furthermore, this PR seems to only address the case of a cross compilation environment setup by `cibuildwheel`. However, the code and the comments therein seem to indicate that this is somehow more generic.

I think that supporting `cibuildwheel` covers an important use case, thus this PR is good to have. On the other hand, it seems that we are designing support for new systems with the same ugly hacks that were used in the past to work around limitations it the build tools. Concretely, i don't understand why we need to rely on `sys.cross_compiling`. If this is designed to support only `cibuildwheel` I am sure we can use some other indicator to detect the cross compilation environment, especially if we already rely on environment variables to setup the compilation environment.

mhsmith:
Thanks for the comments. I have some business travel and vacation coming up, so it'll probably take me a couple of weeks to reply.

rgommers:
@dnicolodi I think your main review comment above covers the same ground that my comments did? What this implements is the same as for iOS. And yes I already pointed out that regular cross-compiling is a thing and this is `cibuildwheel`-specific, and hence the code comments need a tweak.

> Concretely, i don't understand why we need to rely on `sys.cross_compiling`. If this is designed to support only `cibuildwheel` I am sure we can use some other indicator to detect the cross compilation environment,

Initial reply from @mhsmith in [this comment above](https://github.com/mesonbuild/meson-python/pull/824#issuecomment-3846654379). I'd also like a cleaner way to do this. I looked through the `cibuildwheel` docs again, and it does guarantee that the `CIBUILDWHEEL` environment variable is set to `1`: https://cibuildwheel.pypa.io/en/stable/faq/#optional-extensions. So that may be the cleaner, supported way to implement this?

dnicolodi:
I haven't read the conversation in detail, and my questions are maybe due to my complete ignorance about how you go about compiling something for Android. I saw that you planned to merge this unless someone spoke up, thus I assumed you are happy with the status of the PR. I noticed the disconnect between the commit message (that is the base for our changelog entries, especially when reporting about features not implemented by one of the maintainers) and I thought that some clarification is necessary.

It is fine to merge this to support the `cibuildwheel` use case. However, it would be nice to make sure that this will not make it harder to build in a different development environment where the Android development tools have been installed. The `cibuildwheel` documentation that I have seen does not say much about how the cross-compilation is setup and I haven't had time to look at the code. Thus, gating this on the `CIBUILDWHEEL` env var seems a good idea.

mesonpy/__init__.py
mhsmith: I was following the pattern of the existing branches of the `if`. I agree all three of them would be simpler as `sys.platform`, but that would add noise to this PR, so would better be done in a separate one.

mesonpy/__init__.py
mhsmith: I've updated the comment based on the changed approach to use the CIBUILDWHEEL variable.

mesonpy/__init__.py
mhsmith: Removed.

mesonpy/__init__.py
mhsmith: Fixed.

mesonpy/__init__.py
mhsmith: Fixed.

mesonpy/__init__.py
mhsmith: I've updated the comment to explain this in more detail. I think it's worth keeping the comment in the generated file to explain the setting to anyone who examines that file without being familiar with meson-python internals.

tests/test_project.py
mhsmith: Without it, the tests fail with this error:
```
WARNING: Cross file does not specify strip binary, result will not be stripped.

ERROR: Fatal warnings enabled, aborting
```

However, I guess Meson could just as easily do this with any other build tool, so I've updated the tests to include them all. This more closely simulates what cibuildwheel does.

mhsmith:
> I'd also like a cleaner way to do this. I looked through the `cibuildwheel` docs again, and it does guarantee that the `CIBUILDWHEEL` environment variable is set to `1`: https://cibuildwheel.pypa.io/en/stable/faq/#optional-extensions. So that may be the cleaner, supported way to implement this?

My intention was to avoid privileging cibuildwheel over any similar tools, but I guess we are still making some cibuildwheel-specific assumptions, so this makes sense. I'm not aware of any other active Android Python cross-compilation tools, but if they appear in the future, we can revisit this.

mhsmith:
> The commit message states that this implements support for Android. However, it seems to implement support for _cross-compiling_ for Android.

I've updated the commit message.

mesonpy/__init__.py
```diff
@@ -744,6 +744,35 @@ def __init__(
                     self._meson_cross_file.write_text(cross_file_data, encoding='utf-8')
                     self._meson_args['setup'].extend(('--cross-file', os.fspath(self._meson_cross_file)))
 
+        # Android may be native-compiled (e.g. with Termux), cross-compiled with manual
+        # Meson arguments, or cross-compiled with cibuildwheel, which uses a similar
+        # approach to crossenv. In the latter case, we synthesize a cross file to make
+        # that work out of the box.
+        elif sysconfig.get_platform().startswith('android-') and 'CIBUILDWHEEL' in os.environ:
+            cpu = platform.machine()
+            cpu_family = 'arm' if cpu.startswith('arm') else 'x86' if cpu.endswith('86') else cpu
```
dnicolodi: Can we have a comment that explains what this normalization does? I find the `'x86' if cpu.endswith('86')` part particularly puzzling. What does `platform.machine()` return other than `x86`? Judging from the test, this simply maps `i686` to `x86`. If this is the case, the more explicit `'x86' if cpu == 'i686'` would be better.

mesonpy/__init__.py
```diff
@@ -744,6 +744,35 @@ def __init__(
                     self._meson_cross_file.write_text(cross_file_data, encoding='utf-8')
                     self._meson_args['setup'].extend(('--cross-file', os.fspath(self._meson_cross_file)))
 
+        # Android may be native-compiled (e.g. with Termux), cross-compiled with manual
+        # Meson arguments, or cross-compiled with cibuildwheel, which uses a similar
+        # approach to crossenv. In the latter case, we synthesize a cross file to make
+        # that work out of the box.
```
dnicolodi: I think
```suggestion
        # Simplify cross-compilation for Android with cibuildwheel: detect the
        # cross-compilation environment set up by cibuildwheel and synthesize an 
        # appropriate cross file.
```
is more to the point and describes what the following code block is really about.

mesonpy/__init__.py
```diff
@@ -744,6 +744,35 @@ def __init__(
...
+                cpu = {cpu!r}
+                endian = {sys.byteorder!r}
+
+                [properties]
+                # cibuildwheel monkey-patches platform.system and platform.machine to
+                # simulate running on Android, when it's actually running on Linux or
+                # macOS. So the host and build machines will appear to match, but host
+                # binaries cannot actually be run on the build machine. As described at
+                # https://mesonbuild.com/Cross-compilation.html, we indicate this by
+                # setting needs_exe_wrapper = true, but NOT setting exe_wrapper.
```
dnicolodi: Mostly to make things easier for me to understand when I will go back to read this one year from now:
```suggestion
                # Due to Python lacking proper cross-compilation support, for the build
                # to produce the correct wheel tags when cross-compiling for Android,
                # cibuildwheel monkey-patches platform.system() and platform.machine()
                #  to simulate running on Android. This makes Meson believe that host and
                # build machines match and thus that host binaries can be run on the build
                # machine, when this is not actually the case. Override the auto-detection.
```

dnicolodi (changes requested):
Thanks for working on this. Just a couple of minor things, mostly to make the whole easier for me to understand when I'll go back to it in the future.

mesonpy/__init__.py
mhsmith: This was actually based on some code from Meson itself. I've explained this in a comment, and included some additional normalization code which might be useful.

mesonpy/__init__.py
```diff
@@ -744,6 +744,52 @@ def __init__(
...
+        # cross-compilation environment set up by cibuildwheel and synthesize an
+        # appropriate cross file.
+        elif sysconfig.get_platform().startswith('android-') and 'CIBUILDWHEEL' in os.environ:
+            cpu = platform.machine().lower()
+
+            # Python is inconsistent in its platform module, e.g. for x86 it might
+            # return 'x86', 'i686' or some such. Convert to a canonical name recognized
+            # by Meson (https://mesonbuild.com/Reference-tables.html#cpu-families).
+            # Based on mesonbuild.envconfig.detect_cpu_family, but only including
+            # architectures used by Android.
```
dnicolodi: Now I am confused. We enter this block only if we detect execution in the cibuildwheel cross-compilation setup. IIUC, cibuildwheel patches `platform.machine()` to return the host platform for the Android compilation. Therefore, the only values we should map here are the ones used by cibuildwheel. Aren't they? Looking at the cibuildwheel documentation, the only supported architectures are `arm64_v8a` and `x86_64`, thus those should be the only values that should be mapped here. Am I missing something?

mhsmith: cibuildwheel may support Android on 32-bit x86 and ARM in the future, so it's worth accommodating that. However, if that ever happens, the patched environment will still return a `platform.machine` value consistent with a real Android device, and there are only 5 of those (listed in the tests). So yes, this allows the implementation to be much simpler. Fixed.

dnicolodi: Thanks. But this is not perfect yet.

mesonpy/__init__.py
```diff
@@ -744,6 +744,44 @@ def __init__(
                     self._meson_cross_file.write_text(cross_file_data, encoding='utf-8')
                     self._meson_args['setup'].extend(('--cross-file', os.fspath(self._meson_cross_file)))
 
+        # Simplify cross-compilation for Android with cibuildwheel: detect the
+        # cross-compilation environment set up by cibuildwheel and synthesize an
+        # appropriate cross file.
+        elif sysconfig.get_platform().startswith('android-') and 'CIBUILDWHEEL' in os.environ:
+            cpu = platform.machine().lower()
```
dnicolodi: This makes the `cpu` value used in the cross file lowercase. Is this intended or the `.lower()` was intended to be used only for the comparison below? If it is the latter, I have the impression that it is not necessary.

mhsmith: Fixed.

dnicolodi:
Thanks. Merged with a couple of small tweaks.

mhsmith:
Thanks for all of your help with this PR. It's been 3 months since it was merged; do you have any idea when the next release will be? It's one of the main blockers for building NumPy for Android (https://github.com/numpy/numpy/pull/30412).

rgommers:
I think we can just cut a release right now, it's pretty easy to do, `main` is in good shape, and other open PRs don't seem critical to include. @dnicolodi does that sound fine to you?

dnicolodi:
Sure, that's good. I don't think I'll have time soon to look in depth at the other open PRs.

rgommers:
I opened gh-849 to do the release, planning to merge that tomorrow and push the tag.

PR #829, opened 2026-01-08
ENH: implement support for build-details.json (PEP 739) (continued)

mgorny:
This a continuation of #779. For a start, I just did a dumb rebase to rerun the tests.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -340,6 +340,7 @@ class _WheelBuilder():
     _manifest: Dict[str, List[_Entry]]
     _limited_api: bool
     _allow_windows_shared_libs: bool
+    _build_details: mesonpy._tags.BuildDetails
 
     @property
     def _has_internal_libs(self) -> bool:
@@ -364,14 +365,14 @@ class _WheelBuilder():
     def tag(self) -> mesonpy._tags.Tag:
         """Wheel tags."""
         if self._pure:
-            return mesonpy._tags.Tag('py3', 'none', 'any')
+            return mesonpy._tags.Tag('py3', 'none', 'any', build_details=self._build_details)
         if not self._has_extension_modules:
             # The wheel has platform dependent code (is not pure) but
             # does not contain any extension module (does not
             # distribute any file in {platlib}) thus use generic
             # implementation and ABI tags.
-            return mesonpy._tags.Tag('py3', 'none', None)
-        return mesonpy._tags.Tag(None, self._stable_abi, None)
+            return mesonpy._tags.Tag('py3', 'none', None, build_details=self._build_details)
+        return mesonpy._tags.Tag(None, self._stable_abi, None, build_details=self._build_details)
 
     @property
     def name(self) -> str:
@@ -844,6 +845,27 @@ class Project():
         ''')
         self._meson_native_file.write_text(native_file_data, encoding='utf-8')
 
+        # Starting with version 1.10, Meson can consume a `build-details.json`
+        # file following the specification is PEP 739 to obtain required
+        # information to build extension modules without having to run the
+        # interpreter. The path to the `build-details.json` can be specified
+        # passing the with the `-Dpython.build_config=` option to `meson
+        # setup`. Extract the value passed to this option and use the details
+        # in the `build-details.json` file to compute the wheel tag.
+        self._build_details: mesonpy._tags.BuildDetails = mesonpy._tags.introspect_build_details()
+        parser = argparse.ArgumentParser(add_help=False)
+        parser.add_argument('-D', action='append', default=[])
+        args, _ = parser.parse_known_args(self._meson_args['setup'])
+        for arg in reversed(args.D):
+            name, value = arg.split('=', 1)
+            if name == 'python.build_config':
+                try:
+                    with open(value, 'r', encoding='utf8') as f:
+                        self._build_details = json.load(f)
+                except OSError as err:
+                    raise ConfigError(f'The file specified as "python.build_config" cannot be opened: {err}') from err
+                break
+
         # reconfigure if we have a valid Meson build directory. Meson
         # uses the presence of the 'meson-private/coredata.dat' file
         # in the build directory as indication that the build
@@ -1161,13 +1183,15 @@ class Project():
     def wheel(self, directory: Path) -> pathlib.Path:
         """Generates a wheel in the specified directory."""
         self.build()
-        builder = _WheelBuilder(self._metadata, self._manifest, self._limited_api, self._allow_windows_shared_libs)
+        builder = _WheelBuilder(
+            self._metadata, self._manifest, self._limited_api, self._allow_windows_shared_libs, self._build_details)
         return builder.build(directory)
 
     def editable(self, directory: Path) -> pathlib.Path:
         """Generates an editable wheel in the specified directory."""
         self.build()
-        builder = _EditableWheelBuilder(self._metadata, self._manifest, self._limited_api, self._allow_windows_shared_libs)
+        builder = _EditableWheelBuilder(
+            self._metadata, self._manifest, self._limited_api, self._allow_windows_shared_libs, self._build_details)
         return builder.build(directory, self._source_dir, self._build_dir, self._build_command, self._editable_verbose)
 
 
@@ -1333,7 +1357,7 @@ def build_editable(
     if not config_settings:
         config_settings = {}
     if 'build-dir' not in config_settings and 'builddir' not in config_settings:
-        config_settings['build-dir'] = 'build/' + mesonpy._tags.get_abi_tag()
+        config_settings['build-dir'] = 'build/' + mesonpy._tags.get_abi_tag(mesonpy._tags.introspect_build_details())
 
     out = pathlib.Path(wheel_directory)
     with _project(config_settings) as project:
--- a/mesonpy/_tags.py
+++ b/mesonpy/_tags.py
@@ -9,6 +9,27 @@ import platform
 import struct
 import sys
 import sysconfig
+import typing
+
+
+if typing.TYPE_CHECKING:  # pragma: no cover
+    from typing import TypedDict
+
+    class _Abi(TypedDict):
+        extension_suffix: str
+
+    class _ImplementationVersion(TypedDict):
+        major: int
+        minor: int
+
+    class _Implementation(TypedDict):
+        name: str
+        version: _ImplementationVersion
+
+    class BuildDetails(TypedDict):
+        abi: _Abi
+        implementation: _Implementation
+        platform: str
 
 
 # https://peps.python.org/pep-0425/#python-tag
@@ -24,14 +45,91 @@ INTERPRETERS = {
 _32_BIT_INTERPRETER = struct.calcsize('P') == 4
 
 
-def get_interpreter_tag() -> str:
-    name = sys.implementation.name
+def _get_macosx_platform() -> str:
+    ver, _, arch = platform.mac_ver()
+    major, minor = map(int, ver.split('.')[:2])
+
+    # Python built with older macOS SDK on macOS 11, reports an
+    # nonexistent macOS 10.16 version instead of the real version.
+    #
+    # The packaging module introduced a workaround
+    # https://github.com/pypa/packaging/commit/67c4a2820c549070bbfc4bfbf5e2a250075048da
+    #
+    # This results in packaging versions up to 21.3 generating
+    # platform tags like "macosx_10_16_x86_64" and later versions
+    # generating "macosx_11_0_x86_64".  Using the latter would be more
+    # correct but prevents the resulting wheel from being installed on
+    # systems using packaging 21.3 or earlier (pip 22.3 or earlier).
+    #
+    # Fortunately packaging versions carrying the workaround still
+    # accepts "macosx_10_16_x86_64" as a compatible platform tag.  We
+    # can therefore ignore the issue and generate the slightly
+    # incorrect tag.
+
+    if _32_BIT_INTERPRETER:
+        # 32-bit Python running on a 64-bit kernel.
+        if arch == 'ppc64':
+            arch = 'ppc'
+        if arch == 'x86_64':
+            arch = 'i386'
+
+    return f'macosx-{major}.{minor}-{arch}'
+
+
+def _get_ios_platform() -> str:
+    ver = platform.ios_ver().release  # type: ignore[attr-defined]
+    major, minor = map(int, ver.split('.')[:2])
+
+    # Although _multiarch is an internal implementation detail, it's a core part
+    # of how CPython is implemented on iOS; this attribute is also relied upon
+    # by `packaging` as part of tag determination.
+    multiarch = sys.implementation._multiarch.replace('-', '_')
+
+    return f'ios-{major}.{minor}-{multiarch}'
+
+
+def introspect_build_details() -> BuildDetails:
+    platform = sysconfig.get_platform()
+    if platform.startswith('macosx'):
+        platform = _get_macosx_platform()
+    elif platform.startswith('ios'):
+        platform = _get_ios_platform()
+    elif _32_BIT_INTERPRETER:
+        # 32-bit Python running on a 64-bit kernel.
+        if platform == 'linux-x86_64':
+            platform = 'linux_i686'
+        if platform == 'linux-aarch64':
+            platform = 'linux_armv7l'
+
+    return {
+        'abi': {
+            # PyPy reports a $SOABI that does not agree with $EXT_SUFFIX.
+            # Using $EXT_SUFFIX will not break when PyPy will fix this.
+            # See https://foss.heptapod.net/pypy/pypy/-/issues/3816 and
+            # https://github.com/pypa/packaging/pull/607.
+            'extension_suffix': str(sysconfig.get_config_var('EXT_SUFFIX')),
+        },
+        'implementation': {
+            'name': sys.implementation.name,
+            'version': {
+                'major': sys.version_info.major,
+                'minor': sys.version_info.minor,
+            },
+        },
+        'platform': platform,
+    }
+
+
+def get_interpreter_tag(build_details: BuildDetails) -> str:
+    name = build_details['implementation']['name']
+    _v = build_details['implementation']['version']
+    major = _v['major']
+    minor = _v['minor']
     name = INTERPRETERS.get(name, name)
-    version = sys.version_info
-    return f'{name}{version[0]}{version[1]}'
+    return f'{name}{major}{minor}'
 
 
-def get_abi_tag() -> str:
+def get_abi_tag(build_details: BuildDetails) -> str:
     # The best solution to obtain the Python ABI is to parse the
     # $SOABI or $EXT_SUFFIX sysconfig variables as defined in PEP-314.
 
@@ -39,7 +137,8 @@ def get_abi_tag() -> str:
     # Using $EXT_SUFFIX will not break when PyPy will fix this.
     # See https://foss.heptapod.net/pypy/pypy/-/issues/3816 and
     # https://github.com/pypa/packaging/pull/607.
-    empty, abi, ext = str(sysconfig.get_config_var('EXT_SUFFIX')).split('.')
+    ext_suffix = build_details['abi']['extension_suffix']
+    empty, abi, ext = ext_suffix.split('.')
 
     # The packaging module initially based his understanding of the
     # $SOABI variable on the inconsistent value reported by PyPy, and
@@ -60,8 +159,9 @@ def get_abi_tag() -> str:
     return abi.replace('.', '_').replace('-', '_')
 
 
-def _get_macosx_platform_tag() -> str:
-    ver, _, arch = platform.mac_ver()
+def _get_macosx_platform_tag(platform: str) -> str:
+    name, ver, arch = platform.split('-', 2)
+    assert name == 'macosx'
 
     # Override the architecture with the one provided in the
     # _PYTHON_HOST_PLATFORM environment variable.  This environment
@@ -81,24 +181,7 @@ def _get_macosx_platform_tag() -> str:
         parts = os.environ.get('MACOSX_DEPLOYMENT_TARGET', '').split('.')[:2]
         version = tuple(map(int, parts + ['0'] * (2 - len(parts))))
     except ValueError:
-        version = tuple(map(int, ver.split('.')))[:2]
-
-    # Python built with older macOS SDK on macOS 11, reports an
-    # nonexistent macOS 10.16 version instead of the real version.
-    #
-    # The packaging module introduced a workaround
-    # https://github.com/pypa/packaging/commit/67c4a2820c549070bbfc4bfbf5e2a250075048da
-    #
-    # This results in packaging versions up to 21.3 generating
-    # platform tags like "macosx_10_16_x86_64" and later versions
-    # generating "macosx_11_0_x86_64".  Using the latter would be more
-    # correct but prevents the resulting wheel from being installed on
-    # systems using packaging 21.3 or earlier (pip 22.3 or earlier).
-    #
-    # Fortunately packaging versions carrying the workaround still
-    # accepts "macosx_10_16_x86_64" as a compatible platform tag.  We
-    # can therefore ignore the issue and generate the slightly
-    # incorrect tag.
+        version = tuple(map(int, ver.split('.')[:2]))
 
     # The minimum macOS ABI version on arm64 is 11.0.  The macOS SDK
     # on arm64 silently bumps any compatibility version specified via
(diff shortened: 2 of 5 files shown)


rgommers:
I closed gh-779. Copying over the main comment from that PR discussion:

Good point. Yes, there isn't any reason that `build-details.json` cannot be used for a native build. So we should support that as well. That said, there isn't a benefit from the user passing in such a file explicitly; we can start reading it automatically for Python >=3.14, but I'm a bit hesitant to do that right away since it's so new. We can start doing that at any point in the future if it helps simplify our internal code, since it's a backwards-compatible change.

So for now in this PR we can distinguish three cases, with the user passing:

1. Only a cross file (or multiple cross files)
2. Only a `build-details.json` file
3. Both a cross file and a `build-details.json` file

I focused on (3) first and have a TODO in here already to add (1); I should add (2) as well.

mesonpy/__init__.py
```diff
@@ -332,11 +332,15 @@ def __init__(
...
         limited_api: bool,
         allow_windows_shared_libs: bool,
+        is_cross: bool,
+        build_details: Optional[mesonpy._tags.BuildDetailsDict] = None,
     ) -> None:
         self._metadata = metadata
         self._manifest = manifest
         self._limited_api = limited_api
         self._allow_windows_shared_libs = allow_windows_shared_libs
+        self._is_cross = is_cross
```
dnicolodi: Why does the `_WheelBuilder` class need to know if this is a cross build? This attribute looks unused.

mesonpy/__init__.py
```diff
@@ -367,8 +371,8 @@ def tag(self) -> mesonpy._tags.Tag:
             # does not contain any extension module (does not
             # distribute any file in {platlib}) thus use generic
             # implementation and ABI tags.
-            return mesonpy._tags.Tag('py3', 'none', None)
-        return mesonpy._tags.Tag(None, self._stable_abi, None)
+            return mesonpy._tags.Tag('py3', 'none', None, self._build_details)
+        return mesonpy._tags.Tag(None, self._stable_abi, None, self._build_details)
```
dnicolodi: I think a `mesonpy._tags.Tag.from_build_details()` API would work better. This would avoid to have to pass the build details object to a bunch of functions that do close to nothing when the Python interpreter description is available from `build-details.json`.

BTW: `build-details.json` is really a ugly name.

mesonpy/__init__.py
```diff
@@ -779,6 +783,21 @@ def __init__(
         ''')
         self._meson_native_file.write_text(native_file_data, encoding='utf-8')
 
+        # Handle cross compilation
+        self._is_cross = any(s.startswith('--cross-file') for s in self._meson_args['setup'])
```
dnicolodi: Iif i read the code correctly, `_is_cross` is made an attribute of this class only to pass it to the `WheelBuilder` constructor... which does nothing with it. The other only use is an empty conditional with a TODO item. Maybe we don't need it?

mesonpy/__init__.py
```diff
@@ -779,6 +783,21 @@ def __init__(
...
+        # Use build-details.json (PEP 739) to determine
+        # platform/interpreter/abi tags, if given.
+        for setup_arg in reversed(self._meson_args['setup']):
+            if setup_arg.startswith('-Dpython.build_config='):
+                with open(setup_arg.split('=', 1)[1]) as f:
+                    self._build_details = json.load(f)
+                break
+        else:
+            if self._is_cross:
+                # TODO: warn that interpreter details may be wrong. Get platform from cross file.
```
dnicolodi: I think that cross builds are not something someone sets up by accident. If someone does, we should assume they know what they are doing. Thus I don't think a warning here would be appropriate.

mesonpy/__init__.py
```diff
@@ -779,6 +783,21 @@ def __init__(
         ''')
         self._meson_native_file.write_text(native_file_data, encoding='utf-8')
 
+        # Handle cross compilation
+        self._is_cross = any(s.startswith('--cross-file') for s in self._meson_args['setup'])
+        self._build_details = None
+        # Use build-details.json (PEP 739) to determine
+        # platform/interpreter/abi tags, if given.
+        for setup_arg in reversed(self._meson_args['setup']):
+            if setup_arg.startswith('-Dpython.build_config='):
```
dnicolodi: Extracting arguments like this is extremely fragile. `-Dpython.build_config=foo` can be also spelled `-D python.build_config=foo` or `-Dpython.build_config foo` or a few other variants. The easiest way to do it properly is to use `argparse`. See other places in this file where we do something similar.

dnicolodi (changes requested):
I had a quick look at this. Found some possible improvements.

mesonpy/__init__.py
mgorny: I think the goal here was that if you pass a cross file *without* `python.build_config`, we want to warn that using `python.build_config` is preferable. Ofc if you don't think that's a good idea, I can remove it entirely.

mgorny:
I'm going to rebase it now, and make changes later when I find more time.

mesonpy/__init__.py
rgommers: Looks like that was my WIP code, and I can't even remember the reason for it or if the warning was meant to be temporary. I agree it's best to not implement that and simply remove this comment.

mesonpy/__init__.py
eli-schwartz: > or `-Dpython.build_config foo` or a few other variants

```
meson setup: error: The argument for option '-D' must be in OPTION=VALUE format.
```

mesonpy/__init__.py
mgorny: Ok, removing the comment and the variable.

mesonpy/__init__.py
mgorny: I don't understand this point. The JSON file does not provide a "ready" wheel tag, and we still need all of the "postprocessing" these functions do.

mgorny:
Updated and rebased now.

mesonpy/__init__.py
```diff
@@ -834,6 +835,19 @@ def __init__(
...
+        self._build_details = None
+        # Use build-details.json (PEP 739) to determine
+        # platform/interpreter/abi tags, if given.
+        parser = argparse.ArgumentParser(add_help=False)
+        parser.add_argument('-D', action='append', default=[])
+        args, _ = parser.parse_known_args(self._meson_args['setup'])
+        for setup_arg in reversed(args.D):
+            if setup_arg.startswith('python.build_config='):
+                with open(setup_arg.split('=', 1)[1]) as f:
+                    self._build_details = json.load(f)
```
dnicolodi:
```suggestion
        for arg in reversed(args.D):
            name, value = arg.split('=', 1)
            if name == 'python.build_config':
                with open(value, 'r', encoding='utf8') as f:
                    self._build_details = json.load(f)
```

tests/test_tags.py
```diff
@@ -149,3 +160,8 @@ def test_tag_mixed_abi():
     }, pure=False, limited_api=True)
     with pytest.raises(mesonpy.BuildError, match='The package declares compatibility with Python limited API but '):
         assert str(builder.tag) == f'{INTERPRETER}-abi3-{PLATFORM}'
+
+
+@pytest.mark.skipif(BUILD_DETAILS_JSON is None, reason='No build-details.json for this interpreter')
+def test_system_build_details():
+    assert str(mesonpy._tags.Tag()) == str(mesonpy._tags.Tag(build_details=BUILD_DETAILS_JSON))
```
dnicolodi:
```suggestion
def test_build_details():
    try:
        with open(os.path.join(sysconfig.get_path('stdlib')), 'build-details.json')) as f:
            build_details = json.load(f)
    except FileNotFoundError:
        return pytest.skip('build-details.json not found')
    assert str(mesonpy._tags.Tag()) == str(mesonpy._tags.Tag(build_details=build_details))
```

tests/test_tags.py
```diff
@@ -39,8 +40,18 @@ def get_abi3_suffix():
...
+    try:
+        with open(pathlib.Path(sysconfig.get_path('stdlib')) / 'build-details.json') as f:
+            return json.load(f)
+    except FileNotFoundError:
+        return None
+
+
 SUFFIX = sysconfig.get_config_var('EXT_SUFFIX')
 ABI3SUFFIX = get_abi3_suffix()
+BUILD_DETAILS_JSON = get_build_details_json()
```
dnicolodi: This is used in just one test. I don't see a reason for declaring a module constant.

mesonpy/__init__.py
```diff
@@ -834,6 +835,19 @@ def __init__(
         ''')
         self._meson_native_file.write_text(native_file_data, encoding='utf-8')
 
+        # Handle cross compilation
```
dnicolodi:
```suggestion
```

mesonpy/__init__.py
```diff
@@ -834,6 +835,19 @@ def __init__(
...
+        self._build_details = None
+        # Use build-details.json (PEP 739) to determine
+        # platform/interpreter/abi tags, if given.
```
dnicolodi:
```suggestion
        # Starting with version 1.10, Meson can consume a `build-detail.json`
        # file following the specification is PEP 739 to obtain required
        # information to build extension modules without having to run the
        # interpreter. The path to the `build-details.json` can be specified
        # passing the with the `-Dpython.build_config=` option to `meson
        # setup`. Extract the value passed to this option and use the details
        # in the `build-details.json` file to compute the wheel tag.
```

dnicolodi (changes requested):

mesonpy/__init__.py
dnicolodi: That's right, but `-D python.build_config=foo` is valid.

dnicolodi:
I tied up some loose ends. I think this is ready to be merged now. @rgommers, do you want to take a look?

It would be nice to have an integration test for this, but I don't know how hard it is to get a suitable environment setup on GitHub Actions.

mgorny:
Actually, I think it was in Draft because we wanted to make an integration test, but we were blocked on `build-details.json` problems in conda packages. I need to recheck if we resolved them all now.

rgommers:
Sounds great! Indeed, that integration test would be nice. IIRC https://github.com/conda-forge/python-feedstock/pull/858 should have solved the one issue in conda-forge that was blocking for a cross-compile test.

mesonpy/_tags.py
dnicolodi: This still uses `_32_BIT_INTERPRETER` as determined from the running interpreter and not from `build-details.toml`.

dnicolodi (changes requested):

mesonpy/_tags.py
mgorny: Ok, disabled it when using build details. I suppose it's fine to assume that if you're configuring cross via `build-details.json`, you'd be targeting the bitness from the file.

I've also adjusted the logic for macosx, though I have to say that the logic there still relies on the running interpreter; I really don't know much about macOS, and I don't understand the idea behind versions there, but on the test box I have access to `mac_ver()` returns 26.x, while `platform` is `macosx-11.0-arm64`; and I don't really see a way of getting that `26` out of `build-details.json`.

dnicolodi: > I've also adjusted the logic for macosx, though I have to say that the logic there still relies on the running interpreter

In this case, it is better to emit an error stating that `build-details.json` is not supported on macOS rathe than silently emitting wrong wheel tags. The latter would be difficult to debug.

dnicolodi: > I really don't know much about macOS, and I don't understand the idea behind versions there, but on the test box I have access to `mac_ver()` returns 26.x, while `platform` is `macosx-11.0-arm64`; and I don't really see a way of getting that `26` out of `build-details.json`.

macOS and iOS may introduce ABI incompatible changes at each release. However, a specific ABI can be requested setting some environment variables during compilation. Therefore, the platform tag contains a system ABI version part that is derived from the current OS version, unless the relevant environment variable is set. I am not sure what is the best way to map this to the use of `build-details.json`. When compiling a redistributable wheel on macOS it is customary to set the env var to have the toolchain compile for the macOS ABI supported by the relevant Python interpreter, however, it is not a requirement. I think that returning an error if the detected macOS ABI version and the one in `build-details.json` differ may be a good starting point. For the unit test, you can just set the env variable to what the given Python version expects.

rgommers: There is nothing in `build-details.json` that has a relationship with the minimum macOS version, it's orthogonal. 

> For the unit test, you can just set the env variable to what the given Python version expects.

This should always be the right answer for tests. And at runtime, nothing needs to change to `MACOSX_DEPLOYMENT_TARGET` logic. Only the `platform` part of the tag (e.g., `x86_64` or `arm64`) can come from `build-details.json`.

dnicolodi: > There is nothing in `build-details.json` that has a relationship with the minimum macOS version, it's orthogonal.

I have experience only with cross architecture compilation on macOS, not cross system. However, IIUC, compiling for iOS on macOS is a thing (and we added support for it to meson-python a while back). IIUC, nowadays this requires some compiler wrappers distributed with CPython. However, I can see a future where the same can be accomplished with a Meson cross file and pointing to a correct `build-details.json`. In this case, using the build system OS version as the minimum required ABI is clearly wrong. Where should the ABI version come from in this case?

> Only the `platform` part of the tag (e.g., `x86_64` or `arm64`) can come from `build-details.json`.

Assuming cross system compilation macOS to iOS is a thing, where should the ABI version come from?

rgommers: For iOS: from the `IPHONEOS_DEPLOYMENT_TARGET` env var, as we already do:

https://github.com/mesonbuild/meson-python/blob/47ea86f0cefa3a5285ae804ff5980b3a178bec43/mesonpy/_tags.py#L137

Linux is no different: the cross file will say you're targeting `linux` and `x86-64`, but it will _not_ say what the minimum libc version will be (that's where `auditwheel` still comes in, or possibly some other override like an env var, or running the target interpreter under QEMU).

It's just a somewhat arbitrary historical choice I believe, what ended up in cross files and what didn't. There's no real reason that minimum glibc version or iOS version couldn't be in a cross, they just aren't. It's largely stemming from cross compilers accepting target triples but not glibc version.

dnicolodi: `IPHONEOS_DEPLOYMENT_TARGET` does not need to be defined. It is an override for `platform.ios_ver()`. If setting `IPHONEOS_DEPLOYMENT_TARGET` is a requirement when cross compiling with `build-details.json`, it should be documented.

rgommers: Maybe it should, but it looks like that's unrelated to this PR. It's already required today, because this try-except doesn't look like it will work:

https://github.com/mesonbuild/meson-python/blob/47ea86f0cefa3a5285ae804ff5980b3a178bec43/mesonpy/_tags.py#L136-L140

Gives:

```python
>>> import platform
>>> platform.ios_ver()
IOSVersionInfo(system='', release='', model='', is_simulator=False)
>>> version = tuple(map(int, platform.ios_ver().release.split('.')))[:2]
  Cell In[10], line 1
    version = tuple(map(int, platform.ios_ver().release.split('.')))[:2]
ValueError: invalid literal for int() with base 10: ''
```

dnicolodi: I think that is expected, unless you ran that on iOS. So fat the expectation for cross compilation for all targets is that the system is run with some hack that makes the Python that is running the compilation to behave like the cross compilation target.

Anyhow, I have the impression that this discussion is mostly academic.

rgommers: Agreed

thesamesam (commented):
This looks reasonable to me but I don't touch the meson-python side of things often at all.

mesonpy/__init__.py
dnicolodi: I have been looking at this a bit more, and I still don't like how we need to handle the case where `build_details` are available vs when we need to introspect the interpreter. One practical thing, is that it is not clear at a first glance that all the information used in deriving the tag really comes from `build_details` and not from the current platform. Mixing the two is obviously wrong.

Therefore, I propose to turn this around: have a `mesonpy._tags.BuildDetails` dictionary populated with introspection data when the required info does not come from a `build-details.json` file, pass that around, and remove all conditional handling.

mgorny: Working on that. Also, uh, I just discovered a direct `mesonpy._tags.get_abi_tag()` in `build_editable()` that didn't account for `build-details.json`. I suppose it's not really critical since it only determines the build directory, but still.

dnicolodi: > I just discovered a direct `mesonpy._tags.get_abi_tag()` in `build_editable()` that didn't account for `build-details.json`. I suppose it's not really critical since it only determines the build directory, but still.

cross-built editable wheels are not a thing: they wheel is installed in the same environment where the build is running without ever been exposed as a build artifact. Thus that does not need to be changed.

mgorny: Okay, done that. I've split the macosx/ios logic into two functions: one that performs tasks specific to getting current platform, and another that performs postprocessing using the data from `BuildDetails` (which may be the current platform or `build-details.json`). This also implies that passing `BuildDetails` to the functions is required.

I have left `build_editable()`'s `build-dir` use current system, given that implementing `build-details.json` support there would involve quite a bit of duplication, and it doesn't seem a major issue to solve anyway.

mesonpy/__init__.py
```diff
@@ -834,6 +835,26 @@ def __init__(
...
         self._meson_native_file.write_text(native_file_data, encoding='utf-8')
 
+        # Starting with version 1.10, Meson can consume a `build-details.json`
+        # file following the specification is PEP 739 to obtain required
+        # information to build extension modules without having to run the
+        # interpreter. The path to the `build-details.json` can be specified
+        # passing the with the `-Dpython.build_config=` option to `meson
+        # setup`. Extract the value passed to this option and use the details
+        # in the `build-details.json` file to compute the wheel tag.
+        self._build_details: mesonpy._tags.BuildDetails | None = None
```
dnicolodi: This annotation is now wrong: `self._build_details` cannot be `None`. Also, now that we use this attribute to always store information on the interpreter we can consider to name it in a less confusing way (now, at a first look it could seem that this attribute contains information about the build of the Meson project). Maybe `self._info` is not too bad of a name.

mesonpy/__init__.py
```diff
@@ -834,6 +835,26 @@ def __init__(
...
+        parser = argparse.ArgumentParser(add_help=False)
+        parser.add_argument('-D', action='append', default=[])
+        args, _ = parser.parse_known_args(self._meson_args['setup'])
+        for arg in reversed(args.D):
+            name, value = arg.split('=', 1)
+            if name == 'python.build_config':
+                with open(value, 'r', encoding='utf8') as f:
```
dnicolodi: I think that, if the user specifies a path to a file that does not exist, this results in a traceback here: Meson at this point hasn't validate the command line parameters yet.

mesonpy/__init__.py
```diff
@@ -1323,7 +1346,8 @@ def build_editable(
     if not config_settings:
         config_settings = {}
     if 'build-dir' not in config_settings and 'builddir' not in config_settings:
-        config_settings['build-dir'] = 'build/' + mesonpy._tags.get_abi_tag()
+        # TODO: account for build-details.json for the abi tag?
```
dnicolodi: This is not necessary, IMO.

mesonpy/_tags.py
```diff
@@ -24,22 +44,100 @@
...
+    # This results in packaging versions up to 21.3 generating
+    # platform tags like "macosx_10_16_x86_64" and later versions
+    # generating "macosx_11_0_x86_64".  Using the latter would be more
+    # correct but prevents the resulting wheel from being installed on
+    # systems using packaging 21.3 or earlier (pip 22.3 or earlier).
+    #
+    # Fortunately packaging versions carrying the workaround still
+    # accepts "macosx_10_16_x86_64" as a compatible platform tag.  We
+    # can therefore ignore the issue and generate the slightly
+    # incorrect tag.
```
dnicolodi, line 67: What about the code this comments refers to?

mesonpy/_tags.py
```diff
@@ -81,24 +180,7 @@ def _get_macosx_platform_tag() -> str:
...
-    # platform tags like "macosx_10_16_x86_64" and later versions
-    # generating "macosx_11_0_x86_64".  Using the latter would be more
-    # correct but prevents the resulting wheel from being installed on
-    # systems using packaging 21.3 or earlier (pip 22.3 or earlier).
-    #
-    # Fortunately packaging versions carrying the workaround still
-    # accepts "macosx_10_16_x86_64" as a compatible platform tag.  We
-    # can therefore ignore the issue and generate the slightly
-    # incorrect tag.
+        version = tuple(version.split('.')[:2])
```
dnicolodi: Why is the `tuple(map(int` part gone? The comparison below now compares ints to strings...

mesonpy/_tags.py
```diff
@@ -120,53 +202,39 @@ def _get_macosx_platform_tag() -> str:
...
     return platform.replace('-', '_').replace('.', '_').lower()
 
 
 class Tag:
-    def __init__(self, interpreter: str | None = None, abi: str | None = None, platform: str | None = None):
-        self.interpreter = interpreter or get_interpreter_tag()
-        self.abi = abi or get_abi_tag()
-        self.platform = platform or get_platform_tag()
+    def __init__(self, interpreter: str | None = None, abi: str | None = None, platform: str | None = None,
+                 build_details: BuildDetails | None = None):
```
dnicolodi: Build details can become a required argument. Are there any instances where `interpreter` is overridden?

mesonpy/_tags.py
```diff
@@ -24,22 +45,100 @@
...
+
+    # Although _multiarch is an internal implementation detail, it's a core part
+    # of how CPython is implemented on iOS; this attribute is also relied upon
+    # by `packaging` as part of tag determination.
+    multiarch = sys.implementation._multiarch.replace('-', '_')
+
+    return f'ios-{major}.{minor}-{multiarch}'
+
+
+def get_build_details_for_system() -> BuildDetails:
```
dnicolodi: I would name this `introspect_build_details()`

mesonpy/_tags.py
```diff
@@ -24,22 +45,100 @@
...
+            platform = 'linux_i686'
+        if platform == 'linux-aarch64':
+            platform = 'linux_armv7l'
+
+    return {
+        'abi': {
+            # PyPy reports a $SOABI that does not agree with $EXT_SUFFIX.
+            # Using $EXT_SUFFIX will not break when PyPy will fix this.
+            # See https://foss.heptapod.net/pypy/pypy/-/issues/3816 and
+            # https://github.com/pypa/packaging/pull/607.
```
dnicolodi, line 109: I don't think this comment belongs here.

mesonpy/_tags.py
```diff
@@ -120,53 +202,39 @@ def _get_macosx_platform_tag() -> str:
...
-        if arch == 'ppc64':
-            arch = 'ppc'
-        if arch == 'x86_64':
-            arch = 'i386'
+    return f'{platform_os}_{major}_{minor}_{arch}'
 
-    return f'macosx_{major}_{minor}_{arch}'
 
+def _get_ios_platform_tag(platform: str) -> str:
+    platform_os, version, multiarch = platform.split('-', 2)
```
dnicolodi:
```suggestion
    name, version, multiarch = platform.split('-', 2)
    assert name == 'ios'
```

mesonpy/_tags.py
```diff
@@ -120,53 +202,39 @@ def _get_macosx_platform_tag() -> str:
...
-        version = tuple(map(int, platform.ios_ver().release.split('.')))[:2]  # type: ignore[attr-defined]
-
-    # Although _multiarch is an internal implementation detail, it's a core part
-    # of how CPython is implemented on iOS; this attribute is also relied upon
-    # by `packaging` as part of tag determination.
-    multiarch = sys.implementation._multiarch.replace('-', '_')
+        pass
 
-    return f'ios_{version[0]}_{version[1]}_{multiarch}'
+    return f'{platform_os}_{version.replace('.', '_')}_{multiarch.replace('-', '_')}'
```
dnicolodi:
```suggestion
    return f'ios_{version.replace('.', '_')}_{multiarch.replace('-', '_')}'
```

mesonpy/_tags.py
```diff
@@ -60,8 +159,8 @@ def get_abi_tag() -> str:
     return abi.replace('.', '_').replace('-', '_')
 
 
-def _get_macosx_platform_tag() -> str:
-    ver, _, arch = platform.mac_ver()
+def _get_macosx_platform_tag(platform: str) -> str:
+    platform_os, version, arch = platform.split('-', 2)
```
dnicolodi:
```suggestion
    name, version, arch = platform.split('-', 2)
    assert name == 'macos'
```

mesonpy/_tags.py
```diff
@@ -120,53 +202,39 @@ def _get_macosx_platform_tag() -> str:
         # the patch level.  Reset the patch level to zero.
         minor = 0
 
-    if _32_BIT_INTERPRETER:
-        # 32-bit Python running on a 64-bit kernel.
-        if arch == 'ppc64':
-            arch = 'ppc'
-        if arch == 'x86_64':
-            arch = 'i386'
+    return f'{platform_os}_{major}_{minor}_{arch}'
```
dnicolodi:
```suggestion
    return f'macos_{major}_{minor}_{arch}'
```

mesonpy/_tags.py
dnicolodi: Also, why join the version with `.` to replace it with `_` here? Overall, I liked how this was written before much more, among other things because it was symmetric to what is done for macos.

dnicolodi (changes requested):

mesonpy/_tags.py
mgorny: My understanding is that this comment is saying that we don't need any code to handle that case. It was followed by empty line, then another comment. I've figured out it's better to move it where we get the platform, since it applied to what the system gives us.

mesonpy/_tags.py
mgorny: Yeah, sorry. I'm still fixing that part.

mesonpy/_tags.py
mgorny: It's been here:

https://github.com/mesonbuild/meson-python/blob/7c78da354abacda6b39aa78c2dd2eec7bff44cf1/mesonpy/_tags.py#L38-L42

and it seems related to why we're using `EXT_SUFFIX`. Where else should I put it?

mesonpy/_tags.py
mgorny: I've written it under the assumption that we should have the same input whether it's taken from the platform or from `build-details.json`. Since the latter uses dots, I've used the same here. Ofc, it practically doesn't matter, so I can go with underscores, but the conversion will still be necessary for `build-details.json`.

mgorny:
When I'm done addressing feedback, I'm going to add more tests for the tag logic.

mesonpy/__init__.py
mgorny: It is `None` temporarily. Unless you want me to undo the optimization and set the default value unconditionally, then override if `python_build_config` arg is provided.

mesonpy/_tags.py
mgorny: Yes, for pure Python wheels: https://github.com/mesonbuild/meson-python/blob/7c78da354abacda6b39aa78c2dd2eec7bff44cf1/mesonpy/__init__.py#L366-L373

mgorny: Of course, there's no harm in passing them there as well.

mesonpy/__init__.py
mgorny: I don't mind renaming it, though `info` sounds a bit unclear. Maybe `tag_info`? Should I also rename the classes and the argument elsewhere?

mgorny:
I'm going to continue fixing the code tomorrow.

mgorny:
I've implemented most of the requested changes, replied where I wasn't sure. I've also fixed all the test failures, and added some more tests to make sure that the tags are correct for various `build-details.json` cases. The code changed a fair bit, so I'd appreciate another look.

dnicolodi:
I was looking at this once more and I found a problem related to issue #890 that we just fixes. In #890 we found that there are Python distributions that do no implement PEP 3149, namely they do not add the SOABI into EXT_SUFFIX. The issue is that the only thing that `build-details.json` exposes to be used for deriving the ABI tag is the `abi.extension_suffix` field, which I assume is populated with the same value exposed as the `EXT_SUFFIX` var in `sysconfig`. Therefore, for interpreters that do no implement PEP 3149, `build-details.json` cannot be used to derive an ABI tag.

rgommers:
> which I assume is populated with the same value exposed as the `EXT_SUFFIX` var in `sysconfig`. Therefore, for interpreters that do no implement PEP 3149, `build-details.json` cannot be used to derive an ABI tag.

That doesn't sound quite right to me. If a distro patches CPython to change `EXT_SUFFIX`, then either `abi.extension_suffix` will come out with a matching change, or the distro should _also_ patch the `build-details.json` generation change in CPython. I suspect it's the former.

Unlike for PEP 3149, there's no opt-out in PEP 739: `build-details.json` must be correct, or it's a bug in the interpreter in question.

dnicolodi:
I think we are saying the same thing. What I meant is that `abi.extension_suffix` is `build-detail.json` needs to match what `syconfig.get_variable('EXT_SUFFIX')` returns. If the interpreter does not implement PEP 3149, neither will contain the information required to get the wheel ABI tag.

Fortunately, it is a bit more complicated, but it seems that it is possible to put together an ABI tag from other fields in `build-details.json`.

dnicolodi:
Furthermore, reading PEP 739, `abi.extension_suffix` is not a required field, thus it may very well not even be there https://peps.python.org/pep-0739/#abi-extension-suffix

rgommers:
Ah you're talking about these two lines:
```python
    ext_suffix = build_details['abi']['extension_suffix']
    empty, abi, ext = ext_suffix.split('.')
```
Yes, for PkgSrc that will give a plain `.so`, and that will need handling by checking `if len(ext_suffix.split('.')) < 3:` and then constructing `abi` another way. 

> Furthermore, reading PEP 739, `abi.extension_suffix` is not a required field, thus it may very well not even be there https://peps.python.org/pep-0739/#abi-extension-suffix

Please ignore that. Way too many fields are optional there, for obscure reasons. It should be present for CPython on all normal platforms, which we can require. A reason why it may not be present is a case like "interpreter + all extension modules are shipped as a single binary, no separate .so files exist". That's not supportable. We require all fields we need, as does `meson`.

PR #836, opened 2026-01-20
Support installing files in `{prefix}/share/`

### LINKED ISSUE #835: meson-python doesn't support files in {prefix}/share
Currently, meson-python doesn't support files under `{prefix}/share`. For some files (those that explicitly go to `{datadir}`, which is usually `{prefix}/share`) it will not crash, but those file will go to `{prefix}`, not `{prefix}/share`, in the python installation. For most files, it will just refuse to build the wheel.

This currently prevents wheel installation of Meld.

To reproduce:

```
git clone --depth=1 https://gitlab.gnome.org/GNOME/meld.git
python3 -m venv /tmp/venv
/tmp/venv/bin/pip install meld/
...
meson-python: error: Could not map installation path to an equivalent wheel directory: '{prefix}/share/help/cs/meld/command-line.page'
```

Note that if you replace `meson-python>=0.5.0` with `meson-python==0.5.0` in `meld/pyproject.toml` the installation seems to succeed, but the files that should be in `{prefix}/share` are under `{prefix}`, for example, you'll find `/tmp/venv/icons` and `/tmp/venv/meld` which should have been `/tmp/venv/share/icons` and `/tmp/venv/share/meld`.

noamraph:
This resolves https://github.com/mesonbuild/meson-python/issues/835.

The current logic looks at the first part of the installation path. This is usually something like `{bindir}`, `{libdir}` etc. There's a hardcoded map from those strings to paths in the wheel. `{datadir}` is mapped to `data` which I think is a mistake, since `data` goes directly to the installation prefix, and `{datadir}` is usually `{prefix}/share`.

This means that installation fails for many paths, such as `{prefix}/share` and such as `{localedir}`, which is usually `{prefix}/share/locale`.

This PR updates the behavior so that if the first part is not a known "magic", it uses the `intro-buildoptions` meson file to replace it with the actual path. If the result is a relative path which starts with `share`, it goes to `data/share`.

Another possibility is to allow all files under `{prefix}`, not only the files under `{prefix}/share`. I don't know what's better. Currently in this PR I raised a BuildError if the path is not under `{prefix}/share`, but the code would actually be simpler if everything under `{prefix}` would be allowed. For Meld, which is what I'm interested it, all files not in the other paths are in `share`, so it's good enough.

If this PR is accepted, it would allow Meld to be installed by using `pip install`, which would be very nice.

Thanks!
Noam

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -106,7 +106,6 @@ _INSTALLATION_PATH_MAP = {
     '{py_platlib}': 'platlib',
     '{moduledir_shared}': 'platlib',
     '{includedir}': 'headers',
-    '{datadir}': 'data',
     # custom location
     '{libdir}': 'mesonpy-libs',
     '{libdir_shared}': 'mesonpy-libs',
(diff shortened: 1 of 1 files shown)


noamraph:
Closing because the underlying issue was resolved. Thanks @dnicolodi !

Issue #820, opened 2025-11-05
Recompilation in an editable install hides all compiler output


Cadair:
This line:

https://github.com/mesonbuild/meson-python/blob/8e985984f9044a467b226772486c140cf9b3b467/mesonpy/_editable.py#L346

hides all build output, so if the build fails due to missing build requirements or some other reason, you don't get any indication as to why.

I would expect the compilation output to be shown at import time.

I found this while trying to do an editable install of numpy, and because the initial install of numpy used build isolation the import time recompile did not have the build requirements (cython).

rgommers:
Does https://mesonbuild.com/meson-python/how-to-guides/editable-installs.html#verbose-mode help?

There might be an improvement to make there, showing the build output by default only if the build fails.

orlitzky:
We've received a lot of complaints about this since switching to meson-python (with an editable install by default) in SageMath. Sage is a big project, so touching the wrong file can lead to a mysterious hour-long delay when you go to test your changes.

dimpase:
As far as I understand, it's currently not possible to pass `-Ceditable-verbose=yes` via `meson.options`.
Although it would be useful...

eli-schwartz:
> As far as I understand, it's currently not possible to pass `-Ceditable-verbose=yes` via `meson.options`. Although it would be useful...

That file is solely used to define new options (and their default values) that don't previously exist; it cannot pass values for any existing options. The project `default_options` can be used to pass values for any built-in meson option (but not pep517 build backend options owned by meson-python).

Meson-python as it stands today is primarily configured via pyproject.toml

(Editable-verbose is not listed as available to be set via https://mesonbuild.com/meson-python/reference/pyproject-settings.html however.)

dimpase:
> Editable-verbose is not listed as available

why can't it be set in the project (meta)data, be it in `pyproject.toml`, or in `menson.build`, or in `meson.options` ?

Looks like a bug to me. We don't want to tell the users to normally/always use a non-default option, or always have an environment variable set. Normal operations should be working out of the box.

dnicolodi:
No one ever asked for it. But it can be added. See #838

rgommers:
A setting in pyproject.toml to change the default behavior is no merged (gh-838) and will be released very soon in 0.21.0. A mode to only show the relevant part of a build error on failure by default, and be silent otherwise, still has an open PR: gh-750.

dnicolodi:
[...]

PR #750, opened 2025-05-16
Print meson output in case of error during editable rebuild

tobiasdiez:
In case there is an error during the call of meson build, it currently only prints
```
Traceback (most recent call last):
  File "/opt/conda/envs/sage-dev/lib/python3.11/site-packages/_sagemath_editable_loader.py", line 345, in _rebuild
    result = subprocess.run(
             ^^^^^^^^^^^^^^^
  File "/opt/conda/envs/sage-dev/lib/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/opt/conda/envs/sage-dev/bin/ninja']' returned non-zero exit status 1.
```
in the console. This is not very helpful to diagonize the problem. With this PR, the full output of meson is printed.

EDIT: addresses a part of gh-820

PR #838, opened 2026-02-07
ENH: allow to specify editable-verbose in pyproject.toml

dnicolodi:
Build config settings passed via the build front-end take precedence over the value configured via the `tool.meson-python.editable-verbose` setting in `pyproject.toml`.

For this to work, the `editable-verbose` config setting is extended to take an optional boolean argument: "true" or "1" are true values, "false" or "0" are false, anything else is invalid. For backward compatibility, the empty string (no argument passed to the config setting) is treated as true.

Addresses a part of gh-820

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -610,6 +610,7 @@ def _validate_pyproject_config(pyproject: Dict[str, Any]) -> Dict[str, Any]:
             'exclude': _strings,
             'include': _strings,
         }),
+        'editable-verbose': _bool,
     })
 
     table = pyproject.get('tool', {}).get('meson-python', {})
@@ -624,8 +625,18 @@ def _validate_config_settings(config_settings: Dict[str, Any]) -> Dict[str, Any]
             raise ConfigError(f'Only one value for "{name}" can be specified')
         return value
 
-    def _bool(value: Any, name: str) -> bool:
-        return True
+    def _empty_or_bool(value: Any, name: str) -> bool:
+        if isinstance(value, bool):
+            return value
+        if isinstance(value, str):
+            if value == 'false':
+                return False
+            if value == 'true':
+                return True
+            # For backward compatibility, treat a missing value as True.
+            if value == '':
+                return True
+        raise ConfigError(f'Invalid value for "{name}": {value!r}')
 
     def _string_or_strings(value: Any, name: str) -> List[str]:
         return list([value,] if isinstance(value, str) else value)
@@ -633,7 +644,7 @@ def _validate_config_settings(config_settings: Dict[str, Any]) -> Dict[str, Any]
     options = {
         'builddir': _string,
         'build-dir': _string,
-        'editable-verbose': _bool,
+        'editable-verbose': _empty_or_bool,
         'dist-args': _string_or_strings,
         'setup-args': _string_or_strings,
         'compile-args': _string_or_strings,
@@ -674,11 +685,10 @@ class Project():
         source_dir: Path,
         build_dir: Path,
         meson_args: Optional[MesonArgs] = None,
-        editable_verbose: bool = False,
+        editable_verbose: Optional[bool] = None,
     ) -> None:
         self._source_dir = pathlib.Path(source_dir).absolute()
         self._build_dir = pathlib.Path(build_dir).absolute()
-        self._editable_verbose = editable_verbose
         self._meson_native_file = self._build_dir / 'meson-python-native-file.ini'
         self._meson_cross_file = self._build_dir / 'meson-python-cross-file.ini'
         self._meson_args: MesonArgs = collections.defaultdict(list)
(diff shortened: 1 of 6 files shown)


tests/test_project.py
```diff
@@ -410,3 +412,55 @@ def test_ios_project(package_simple, monkeypatch, multiarch, tmp_path):
...
+    project = mesonpy.Project(tmp_path, tmp_path / 'build')
+    assert project._editable_verbose == (verbose == 'true')
+
+
+@pytest.mark.parametrize('settings,expected', [
+    ({}, True),
+    ({'editable-verbose': ''}, True),
+    ({'editable-verbose': 'true'}, True),
+    ({'editable-verbose': 'false'}, False),
+], ids=['', '-Ceditable-verbose', '-Ceditbale-verbose=true', '-Ceditable-verbose=false'])
```
orlitzky: typo? (editbale)

dnicolodi:
@rgommers What do you think of this approach?

dnicolodi:
@rgommers gentle ping

mesonpy/__init__.py
```diff
@@ -617,16 +618,24 @@ def _string(value: Any, name: str) -> str:
...
 
-    def _bool(value: Any, name: str) -> bool:
-        return True
+    def _empty_or_bool(value: Any, name: str) -> bool:
+        if isinstance(value, bool):
+            return value
+        if isinstance(value, str):
+            if value == 'false' or value == '0':
+                return False
+            # For bakward compatibility, treat a missing value as True.
```
rgommers: typo: bakward -> backward

mesonpy/__init__.py
```diff
@@ -617,16 +618,24 @@ def _string(value: Any, name: str) -> str:
             raise ConfigError(f'Only one value for "{name}" can be specified')
         return value
 
-    def _bool(value: Any, name: str) -> bool:
-        return True
+    def _empty_or_bool(value: Any, name: str) -> bool:
```
rgommers, line 628: This name is slightly awkward. Not a big deal, I'm fine if you prefer this one - but perhaps consider `_optional_bool`?

rgommers (commented):
Thanks @dnicolodi, and apologies for the delay - I've been pretty swamped (now on a short holiday, so trying to catch up here).

Adding a pyproject.toml setting seems like a reasonable and clean approach. I'm not sure about is explicitly accepting `1`/`0` (which are untested) but not `yes`/`no`/`on`/`off`. I did a quick search on GitHub and could only found `true` being used; in gh-820 there's one usage of `yes` in a question. Either being more strict or more loose might be appropriate here? 

Technically this could break someone who uses now `yes` or some random string, however that doesn't seem worth worrying about.

This section should probably be expanded a little: https://mesonbuild.com/meson-python/how-to-guides/editable-installs.html#verbose-mode

The docs should probably have a warning about the "install scripts" problem you explain at https://github.com/mesonbuild/meson-python/issues/820#issuecomment-3515535980. 

Question: did you consider what we discussed there, buffering to a file and only displaying on a build failure? That would probably be cleaner and wouldn't have to come with a disclaimer, but perhaps more complex to reliably implement?

rgommers:
Thinking about that some more, an `only-on-error` mode is orthogonal to this, and could be added either later or instead of this. And that could be a change in default behavior, it wouldn't need an explicit setting I'd think, you always want to show the output if there's a build error.

The change to this line:

https://github.com/mesonbuild/meson-python/blob/e5ba90fc2c0bad36055e7a393013facb221bb21e/mesonpy/_editable.py#L346

would have to be something like this (replace `DEVNULL` with `PIPE`):
```python
proc = subprocess.run(self._build_cmd, ..., stdout=subprocess.PIPE)
if proc.returncode != 0:
    sys.stderr.buffer.write(proc.stdout)
    raise ImportError(...)
```

rgommers:
Ah, I rediscovered gh-750, which deals with the `only-on-error` case, so we should revisit that, and not reinvent that wheel here. Let's focus on the `pyproject.toml` settings only here.

mesonpy/__init__.py
dnicolodi: I choose this name to try to convey that the empty string is mapped to `True`. I would expect `_optional_bool()` to return `None` for the empty string.

dnicolodi:
> I'm not sure about is explicitly accepting `1`/`0` (which are untested) but not `yes`/`no`/`on`/`off`. I did a quick search on GitHub and could only found `true` being used; in [gh-820](https://github.com/mesonbuild/meson-python/issues/820) there's one usage of `yes` in a question. Either being more strict or more loose might be appropriate here?

Maybe accepting only `true` and `false` is more appropriate.

mesonpy/__init__.py
rgommers: Okay fair reason, resolving this comment.

rgommers:
> Maybe accepting only `true` and `false` is more appropriate.

Sounds good to me.

dnicolodi:
Updated with the discussed changes.

rgommers (approved):
LGTM, approving. pre-commit is still unhappy with some trailing whitespace.

dnicolodi:
> pre-commit is still unhappy with some trailing whitespace.

Fixed

PR #843, opened 2026-04-24
Add support PEP 770 (SBOMs)

### LINKED ISSUE #763: Support for PEP 770 (SBOMs)
[PEP 770](https://peps.python.org/pep-0770/) is accepted, and specifies how wheels can start incorporating SBOMs as metadata under `.dist-info/sboms/`. 

PEP 770 does not provide metadata in the `[project]` table for SBOMs, for reasons discussed in the PEP (static + dynamic metadata is expected to be common). An earlier version did use `[project]`, and https://github.com/pypa/pyproject-metadata/pull/225 prototyped support for that in `pyproject-metadata`. That part is no longer needed, but support in a build backend will be since `.dist-info` is generated by a build backend.

Technically it's possible to do something hacky like this today in a `meson.build` file:
```meson
install_data(
    'sboms/an_sbom_filename.spdx.json',
    install_dir: py.get_install_dir() / meson.project_name() + '-' + meson.project_version() + '.dist-info' / 'sboms',
)
```
But obviously that isn't recommended.

What we need instead is this in `pyproject.toml`:
```ini
# static SBOM files that go into all wheels
[tool.meson-python.sboms]
sbom-files = [
    "sboms/component1.spdx.json",
    "sboms/component2.spdx.json",
]
```
That's the basic support. There are also cases possible where a vendored component only gets included in wheels for say one platform, or if a particular build option is given. That's a lot harder to deal with, and could be done either in `[tool.meson-python]` or through some mechanism with data files in `meson.build` files (e.g., install_data(..., install_tag: 'sbom')` - lots of options and more limited needs, so let's leave that for the future I'd say.

It'd be nice to align this with other backends, so the mechanism looks similar.
- `scikit-build-core`: I don't see an issue yet, Cc @henryiii for thoughts
- `maturin`: open feature request at https://github.com/PyO3/maturin/issues/2554

fangchenli:
closes #763

### DIFF
--- a/docs/how-to-guides/sboms.rst
+++ b/docs/how-to-guides/sboms.rst
@@ -1,0 +1,108 @@
+.. SPDX-FileCopyrightText: 2026 The meson-python developers
+..
+.. SPDX-License-Identifier: MIT
+
+.. _how-to-guides-sboms:
+
+***********************************
+Including SBOMs in wheels (PEP 770)
+***********************************
+
+`PEP 770`_ defines a standard location for Software Bill of Materials
+(SBOM) documents describing the contents of Python packages: the
+``sboms/`` subdirectory of the wheel's ``.dist-info/`` directory.
+
+``meson-python`` moves files installed by the Meson project at a
+location matching one of the glob patterns in the
+:option:`tool.meson-python.sbom-files` setting to the
+``.dist-info/sboms/`` directory in the wheel. The default pattern is
+``{datadir}/{name}/sboms/*``, with ``{name}`` the normalized project
+name: files installed in the ``sboms/`` subdirectory of the project's
+data directory are treated as SBOM files. When the Meson project is
+installed directly, rather than through a Python build front-end, the
+SBOM files are installed in the data directory, alongside the
+project's other data files.
+
+.. _PEP 770: https://peps.python.org/pep-0770/
+
+Static SBOM files
+=================
+
+SBOM documents checked into the source tree, typically describing
+source-vendored components, only need to be installed to the location
+matched by the default pattern:
+
+.. code-block:: meson
+
+   project('my-project', 'c', version: '1.0.0')
+
+   install_data(
+     'sboms/component1.cdx.json',
+     'sboms/component2.cdx.json',
+     install_dir: get_option('datadir') / meson.project_name() / 'sboms',
+   )
+
+The files end up in the wheel at
+``my_project-1.0.0.dist-info/sboms/component1.cdx.json`` and
+``component2.cdx.json``.
+
+Dynamically generated SBOMs
+===========================
+
+When the SBOM is generated at build time, use a ``custom_target`` that
+writes the file and installs it to the same location:
+
+.. code-block:: meson
+
+   py = import('python').find_installation()
+
+   custom_target('vendored-sbom',
+     output: 'vendored.cdx.json',
+     command: [py, files('scripts/generate_sbom.py'), '@OUTPUT@',
+               '--version', meson.project_version()],
+     install: true,
+     install_dir: get_option('datadir') / meson.project_name() / 'sboms',
+   )
+
+The generator is provided by the project; ``meson-python`` does not
+ship one. It can be a script checked into the source tree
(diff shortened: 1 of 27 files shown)


rgommers:
Thanks @fangchenli, this looks pretty comprehensive. I'll have to think about this a bit longer, in particular whether the targeting of the `.dist-info` directory is foolproof. The way you did this now is more general than adding a `[tool.meson-python.sboms]` setting that was also discussed in gh-763 I think, since it also deals with the "generate SBOM files on the fly with a script". It's just a little more verbose for the package author, but if you need it then it's hard to see a more ergonomic alternative.

I'll cross-link https://github.com/mesonbuild/meson/pull/15738 for the idea of adding a custom Meson install tag, which is still in the commit history here but you dropped (and that decision looks correct to me).

rgommers:
I'll also explicitly cross-link [this comment](https://github.com/mesonbuild/meson/pull/15738#issuecomment-4310509678) from @eli-schwartz (_"It may be separately interesting to teach meson about SBOM files, ..."_. That does make sense, but nothing in this PR is dependent on that or adds something public that would need removing if Meson adds SBOM support in the future.

docs/how-to-guides/sboms.rst
```diff
@@ -0,0 +1,114 @@
...
+
+***********************************
+Including SBOMs in wheels (PEP 770)
+***********************************
+
+`PEP 770`_ defines a location for Software Bill of Materials (SBOM)
+files inside the wheel's ``.dist-info/sboms/`` directory.
+``meson-python`` routes any file installed under
+``{py_purelib}/<name>-<version>.dist-info/<subdir>/...`` or
+``{py_platlib}/<name>-<version>.dist-info/<subdir>/...`` into the
```
rgommers: I suggest removing purelib/platlib mentions from the docs and the changelog. That's really not a relevant concept for the end user, they only need to be using `py.get_install_dir()` and that just does the right thing based on the `pure` keyword.

dnicolodi:
I had a quick look at this.

First, please squash the commits into a set of logical changes, we are not interested in how we arrived at the current form of the code and separate commits in a PR should only be used to separate changes in self contained bites easier to review. 

Second, I am not sure I like the shape of this. The main objection is that this works for when producing a wheel. However, there are many projects that use meson (without meson-python) to directly install. In this case there would be an undesirable `.dist-info` folder installed alongside the Python package. The only way for the users to get rid of it would be to make the installation of these files conditional on a user build option and to pass that option when building the wheel. This does not seem a very ergonomic interface to me.

A better solution that comes to mind is to have a `pyptoject.toml` setting
```toml
[project]
name = 'foo'

[tool.meson-python]
sboms = ['{datadir}/foo/sboms/*']
```
which mimics `project.license-files`. This would then pick any file matching the specified pattern (in the example anything installed in `{datadir}/foo/sboms/`) and place it in `.dist-info/sboms/`. The default for `tool.meson-python.sboms` can be `{datadir}/{package-name}/sboms/*` but keeping the setting allow to deactivate special handling for the files contained in that directory. Using `{datadir}` has the advantage that the files installed there would not look odd when installed on a system when the meson project is used outside a wheel. If there is a convention for installing SBOMs system wide, we should use that instead. Another advantage is that this scheme can be extended to be used with license files declared in `meson.build`.

This scheme can maybe further extended to pick files from the source directory when the glob pattern specified in `tool.meson-python.sboms` does not contain an installation directory anchor.

dnicolodi:
Maybe the setting name should be `tool.meson-python.sbom-files` to mirror `project.license-files`.

eli-schwartz:
> The only way for the users to get rid of it would be to make the installation of these files conditional on a user build option and to pass that option when building the wheel. This does not seem a very ergonomic interface to me.

Well, this does remind me that there are several projects which produce C libraries and also (optionally?) python bindings, and not the ctypes variety.

For wheels, it's potentially desirable to statically link and/or not waste time building stuff that won't ever go into a wheel. I added a -Dwheel-only=true option to https://github.com/dgibson/dtc/blob/main/pyproject.toml#L24

PR #845, opened 2026-04-29
DOC/TST: add a how-to page and test packages for dynamic versioning

### LINKED ISSUE #159: Grab version from __version__ attribute

rgommers:
Closes gh-159

Most of our most prominent users (NumPy, pandas, Matplotlib) figured this out, and other packages copy from there if they want dynamic versioning. It's quite a common question (see incoming links to gh-159), so it's past time to document this.

***

I do feel like the tradeoffs of `versioneer`/`setuptools-scm` are quite poor, and this stuff really isn't so hard that it needs separate dependencies (4 in the case of setuptools-scm?!?!) and thousands of lines of code. So there may be space for a follow-up where we either improve Meson's `vcs_tag` (or an alternative to it) so that works inside `project()`, or we provide something simple to vendor or a CLI script, a la what NumPy does (that just works in O(100) lines of code). But for now, this should help.

### DIFF
--- a/docs/how-to-guides/dynamic-versioning.rst
+++ b/docs/how-to-guides/dynamic-versioning.rst
@@ -1,0 +1,179 @@
+.. SPDX-FileCopyrightText: 2023 The meson-python developers
+..
+.. SPDX-License-Identifier: MIT
+
+.. _how-to-guides-dynamic-versioning:
+
+******************
+Dynamic versioning
+******************
+
+The most common approach to versioning is to keep a static version number in
+``pyproject.toml`` only, and update it before a new release in a regular commit.
+This is simple and robust. However, sometimes a package author may want more
+from versioning, and hence reach for dynamic versioning. E.g.:
+
+1. Use the package version in a ``meson.build`` file without duplicating the version string between ``pyproject.toml`` and ``meson.build``.
+2. Use the hash of the current commit in the package version, or store it in a configuration file.
+3. Derive the version from the most recent git tag rather than maintain it in the code.
+
+.. note::
+
+    Each of these things has a cost - keeping all metadata static and not
+    running ``git`` or introspecting the ``.git`` directory as part of the build
+    avoids running extra build steps in some cases, extra build dependencies or
+    custom scripts, and potential issues with shallow checkouts in CI where the
+    ``.git`` directory may not be present. Only use these dynamic features if
+    you have a good reason to do so!
+
+Single-sourcing the version string
+----------------------------------
+
+When you want to define your project's version string in a single place,
+``meson-python`` knows how to extract the version number from the ``project()``
+call in the top-level ``meson.build``; in ``pyproject.toml`` it can be declared
+as dynamic:
+
+.. code-block:: toml
+
+    [project]
+    dynamic = ['version']
+
+Then in ``meson.build``, define the version:
+
+.. code-block:: meson
+
+   project('my-project', 'c', version: '1.2.3')
+
+It can also be done the other way around - this requires a bit more code,
+however it has the advantage that all metadata remains static in
+``pyproject.toml``, which can in some cases avoid triggering a build
+when an installer needs to obtain the version. To implement this,
+set the version in ``pyproject.toml``:
+
+.. code-block:: toml
+
+    [project]
+    version = '1.2.3'
+
+And in ``meson.build``, run a helper script as part of the project call
+(again, this is only needed if you actually use the version string inside a
+``meson.build`` file):
+
+.. code-block:: meson
+
+    project('my-project',
+        'c',
+        version: run_command(
+            ['get_version.py'],
+            check: true
+        ).stdout().strip(),
+    )
+
+With that ``get_version.py`` script retrieving the version from
+``pyproject.toml``:
+
+.. code-block:: python
+
+    #!/usr/bin/env python3
+    import os
+
+    def get_version():
+        pyproject_toml = os.path.join(os.path.dirname(__file__), 'pyproject.toml')
+        with open(pyproject_toml) as f:
+            data = f.readlines()
+
+        version_line = next(
+            line for line in data if line.startswith('version =')
+        )
+        version = version_line.strip().split(' = ')[1]
+        return version.replace('"', '').replace("'", '')
+
+
+    if __name__ == "__main__":
+        print(get_version())
+
+
+Storing the git commit hash inside your package
+-----------------------------------------------
+
+Capturing the git commit hash alongside the version can be useful for bug
+reports and reproducibility: the user can print ``pkgname.__version__`` and
+``pkgname.__git_hash__`` to identify exactly which commit they are running. The
+commit hash is not part of ``pyproject.toml`` and cannot be derived from a
+source distribution after the fact, so it has to be written into the package at
+build time.
+
+A pattern to achieve this, which used by NumPy for example, is a single helper
+script that does double duty: it prints the version when called from
+``project()``, and it writes a generated ``_version.py`` file containing both
+the version and the git hash when called from a build step. The
+script is wired up via ``custom_target()`` for normal builds and via
+``meson.add_dist_script()`` so that the generated file is also included
+in source distributions.
+
+In ``meson.build``:
+
+.. literalinclude:: ../../tests/packages/dynamic-version-from-script/meson.build
+   :language: meson
+   :lines: 5-
+
+The ``build_always_stale: true`` flag ensures that the recorded hash is
+refreshed every time the project is rebuilt, rather than being cached
+from a previous build.
+
+The helper script reads the version from ``pyproject.toml`` (so the
+version still has a single source of truth) and resolves the git hash
+via ``git rev-parse``, falling back to ``'unknown'`` when called outside
+a checkout — for example when building from an extracted source
+distribution:
+
+.. literalinclude:: ../../tests/packages/dynamic-version-from-script/generate_version.py
+   :language: python
+   :lines: 1,5-
+
+The ``MESON_DIST_ROOT`` branch ensures that when the script is invoked
+as a dist script, it writes the generated file into the staging
+directory ``meson dist`` is preparing, so it is included in the source
+distribution. See :ref:`sdist` for the surrounding context.
+
+The package's ``__init__.py`` re-exports the generated symbols:
+
+.. code-block:: python
+
+    from pkgname._version import __git_hash__, __version__
+
+A complete worked example lives at ``tests/packages/dynamic-version-from-script``
+in the meson-python source tree.
+
+
+Derive version from latest git tag
+----------------------------------
+
+When the version is encoded in git tags rather than in source files, the
+build system has to query git at configure time. There are a number of
+packages that provide this functionality - e.g., ``setuptools-scm`` as the most
+popular one - and that can be used together with ``meson-python``. The
+integration principle is the same as above: use a ``run_command()`` call inside
+``project()`` (either directly or through a small wrapper script like
+``get_version.py`` higher up) that prints the version and (optionally) writes
+out a file to disk that can be included in the sdist.
+
+The example below uses ``setuptools-scm``; the same approach applies
+to the other tools - only the wrapper script differs. Declare it as a build
+requirement in ``pyproject.toml``:
+
+
+.. literalinclude:: ../../tests/packages/version-setuptools-scm/pyproject.toml
+   :language: toml
+   :lines: 5-12
+
+In ``meson.build``, invoke ``setuptools-scm`` to compute the version:
+
+.. literalinclude:: ../../tests/packages/version-setuptools-scm/meson.build
+   :language: meson
+   :lines: 5-
+
+That's it. You can use ``setuptools-scm`` config options as explained in its docs.
+If you do want to store a generated file ``.py`` file with versioning metadata,
+use ``meson.add_dist_script()`` as explained higher up.
--- a/tests/packages/dynamic-version-from-script/generate_version.py
+++ b/tests/packages/dynamic-version-from-script/generate_version.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: 2026 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+import argparse
+import os
+import subprocess
+
+
+try:
+    import tomllib
+except ModuleNotFoundError:
+    import tomli as tomllib
+
+
+def get_version_from_pyproject():
+    here = os.path.dirname(os.path.abspath(__file__))
+    pyproject_toml = os.path.join(here, 'pyproject.toml')
+    with open(pyproject_toml, 'rb') as f:
+        return tomllib.load(f)['project']['version']
+
+
+def get_git_hash():
+    here = os.path.dirname(os.path.abspath(__file__))
+    try:
+        result = subprocess.run(
+            ['git', 'rev-parse', 'HEAD'],
+            cwd=here,
+            capture_output=True,
+            check=True,
+            text=True,
+        )
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        return 'unknown'
+    return result.stdout.strip()
+
+
+def write_version_file(outfile, version, git_hash):
+    if 'MESON_DIST_ROOT' in os.environ:
+        outfile = os.path.join(os.environ['MESON_DIST_ROOT'], outfile)
+    with open(outfile, 'w') as f:
+        f.write(f"__version__ = '{version}'\n")
+        f.write(f"__git_hash__ = '{git_hash}'\n")
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('outfile', nargs='?')
+    args = parser.parse_args()
+
+    version = get_version_from_pyproject()
+    if args.outfile is None:
+        print(version)
+    else:
+        write_version_file(args.outfile, version, get_git_hash())
+
+
+if __name__ == '__main__':
+    main()
--- a/tests/packages/dynamic-version-from-script/meson.build
+++ b/tests/packages/dynamic-version-from-script/meson.build
@@ -1,0 +1,36 @@
+# SPDX-FileCopyrightText: 2026 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+project(
+    'dynamic-version-from-script',
+    version: run_command(['generate_version.py'], check: true).stdout().strip(),
+)
+
+fs = import('fs')
+py = import('python').find_installation()
+
+py.install_sources(
+    'dynamic_version_from_script/__init__.py',
+    subdir: 'dynamic_version_from_script',
+)
+
+meson.add_dist_script('generate_version.py', 'dynamic_version_from_script/_version.py')
+
+if not fs.exists('dynamic_version_from_script/_version.py')
+    custom_target(
+        'write_version_file',
+        output: '_version.py',
+        command: ['generate_version.py', '@OUTPUT@'],
+        build_by_default: true,
+        build_always_stale: true,
+        install: true,
+        install_dir: py.get_install_dir() / 'dynamic_version_from_script',
+    )
+else
+    # When building from sdist, _version.py exists and should be included
+    py.install_sources(
+        ['dynamic_version_from_script/_version.py'],
+        subdir: 'dynamic_version_from_script',
+    )
+endif
--- a/tests/packages/version-setuptools-scm/meson.build
+++ b/tests/packages/version-setuptools-scm/meson.build
@@ -1,0 +1,11 @@
+# SPDX-FileCopyrightText: 2026 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+project(
+    'version-setuptools-scm',
+    version: run_command(
+        ['python3', '-m', 'setuptools_scm'],
+        check: true,
+    ).stdout().strip(),
+)
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -83,6 +83,7 @@ the use of ``meson-python`` and Meson for Python packaging.
    how-to-guides/meson-args
    how-to-guides/debug-builds
    how-to-guides/shared-libraries
+   how-to-guides/dynamic-versioning
    reference/limitations
    projects-using-meson-python
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,7 @@ test = [
   'pytest-cov',
   'pytest-mock',
   'cython >= 3.0.3', # required for Python 3.12 support
+  'setuptools-scm', # used by the version-setuptools-scm test package
   'wheel',
   'typing-extensions >= 3.7.4; python_version < "3.11"',
 ]
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -84,6 +84,7 @@ def in_git_repo_context(path=os.path.curdir):
         subprocess.run(['git', 'config', 'user.name', 'A U Thor'], cwd=path, check=True)
         subprocess.run(['git', 'add', '*'], cwd=path, check=True)
         subprocess.run(['git', 'commit', '-q', '-m', 'Test'], cwd=path, check=True)
+        subprocess.run(['git', 'tag', 'v1.2.3'], cwd=path, check=True)
         yield
     finally:
         # PermissionError raised on Windows.
--- a/tests/packages/dynamic-version-from-script/dynamic_version_from_script/__init__.py
+++ b/tests/packages/dynamic-version-from-script/dynamic_version_from_script/__init__.py
@@ -1,0 +1,8 @@
+# SPDX-FileCopyrightText: 2026 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+from dynamic_version_from_script._version import __git_hash__, __version__
+
+
+__all__ = ['__git_hash__', '__version__']
--- a/tests/packages/dynamic-version-from-script/pyproject.toml
+++ b/tests/packages/dynamic-version-from-script/pyproject.toml
@@ -1,0 +1,11 @@
+# SPDX-FileCopyrightText: 2026 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+[build-system]
+build-backend = 'mesonpy'
+requires = ['meson-python']
+
+[project]
+name = 'dynamic-version-from-script'
+version = '1.2.3'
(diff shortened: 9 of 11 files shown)


rgommers:
Cirrus says: _Cirrus CI shuts down on June 1st 2026._ - guess we'll have to deal with that soon.

eli-schwartz:
> I do feel like the tradeoffs of `versioneer`/`setuptools-scm` are quite poor, and this stuff really isn't so hard that it needs separate dependencies (4 in the case of setuptools-scm?!?!)

setuptools-scm has the advantage of being the only one that is actually e.g. redistributor friendly. So ideally I would like to see information about versioneer be buried forever so that nobody ever uses it again. I don't even want its existence acknowledged. As a downstream builder and repackager of software seeing a project use versioneer is a red flag that I'm in for detrimental experiences and suffering, and that I should think carefully about whether I want to use the project at all, if that project has viable competition.

I'd also like to correct you about setuptools-scm's dependencies:

Recent work on setuptools-scm has involved splitting out its core logic into "vcs-versioning", which has only one dependency (assuming recent python) and contains the bits you need to parse versions from git without the integration into the setuptools build backend. The setuptools-scm entrypoint and `__main__.py` for those versions post split, are just a compatibility shim that imports and executes `vcs_versioning._cli.main()`.

I won't deny that the packaging module isn't strictly necessary for our use cases, but it's still pretty lightweight compared to what you feared (and the packaging module is pretty much ubiquitous as a dependency for any package which is at all pyproject adjacent and doesn't actively have a policy forbidding the use of dependencies).

> So there may be space for a follow-up where we either improve Meson's `vcs_tag` (or an alternative to it) so that works inside `project()`,

This is compatible with my long term desires. I want meson to directly support extracting versions from git, and Meson can't have dependencies so this would anyways require building it ourselves. I think that setuptools-scm has lots of cool knobs that... few people really need. :D So reimplementing it won't be a bad thing. Setuptools-scm also has to support a lot of legacy burden in its UX, sooo...

Modern setuptools-scm prefers using .git_archival.txt where possible (this required a fix to git itself that allows using git describe in gitattributes export-subst, I'm the one who pushed for that in git upstream; old projects will still have depended on other, legacy workflows in setuptools-scm). My feeling is that meson doesn't *need* to support anything other than this.

So what I'd probably want to see is
```meson
project(..., version: files('.git_archival.txt'))
```

with special detection of archival information, and then vcs_tag() could "know" when this was used and redetect the version.

People could still use custom scripts if they want to do something more complex than this, as they can today.  It for many common use cases things can become a lot simpler.

rgommers:
> setuptools-scm has the advantage of being the only one that is actually e.g. redistributor friendly. So ideally I would like to see information about versioneer be buried forever

I agree `setuptools-scm` is the best of the bunch, so I'm fine with rewording as "setuptools-scm and other similar tools" or some such thing to not mention `versioneer` explicitly. We used to use `versioneer` in NumPy, and I agree it's a bit of a mess behavior and code-wise (although we never had distro packager complaints I believe). It does have the advantage of being designed for vendoring; which at least addresses one of the main downsides of setuptools-scm.



> Recent work on setuptools-scm has involved splitting out its core logic into "vcs-versioning", which has only one dependency (assuming recent python) and contains the bits you need to parse versions from git without the integration into the setuptools build backend

I did look at that, and it seems much better in the long term, but as of now it has no docs of its own, and the setuptools-scm docs only say it's for building other packages on top of (https://setuptools-scm.readthedocs.io/en/latest/integrators/) plus they warn explicitly about using setuptools-scm functions like `get_version` directly. So as of now, I don't think we can be pointing at it.



> So what I'd probably want to see is
> 
> ```meson
> project(..., version: files('.git_archival.txt'))
> ```
> 
> with special detection of archival information,

With that last statement, do you mean some special magic to make that work for direct installs from vcs (`pip install .` & co) so that `files('.git_archival.txt')` doesn't choke when that file doesn't exist? Might work, but a bit confusing. A new function like `vcs_version()` that uses `.git_archival.txt` under the hood might be cleaner?

eli-schwartz:
The file always exists. The way it works is you commit the file with special tokens into git, and then `git archive` replaces the tokens.

rgommers:
Thanks, I missed that. The point remains though: `git archive` is not run for direct installs. So shouldn't it be something like `version: vcs_version()`, with some awareness of both `.git` and `.git-archival.txt` (or perhaps with `.git-archival.txt` as an input argument)? That also keeps typing easier, with `version:` only accepting a string.

eli-schwartz:
Version already accepts a files() object today -- it is currently expected to have a single line containing the version number.

It wouldn't be that hard to detect when a multi-line file has replacement tokens instead of the processed metadata and run `git archive` internally to process it.

I will admit that mostly what exists at the moment is shower thoughts in my head, so don't necessarily treat this as immutable requirements. ;)

rgommers:
Okay, I did a quick check: `git archive --format=tar -o scipy.tar.gz HEAD` takes about 20 ms on an M5 Macbook, which is much faster than I expected (and similar to invoking something like `git -c log.showSignature=false log -1`). Generating an 82 MB tarball in a tmpdir isn't complete ideal, but also not a showstopper since it's only done once per build. For very large repos it might become a problem though.

eli-schwartz:
To be clear, on a 1.6 gb `git archive` of the Linux kernel, it takes 0.182s -- that's because `git archive HEAD COPYING` does essentially no work. It takes 6 seconds if I want to write the entire 1.6gb out to a pipe for `| wc -c`.

20 seconds to write the whole thing to a file is still only 0.189s to do just COPYING. But a single text file wrapped in a tar envelope is trivially small enough to do all operations in a python bytes buffer anyway.

I don't like wasted work and I don't like slow-as-heck tools that do tons of work for "convenience". Trust me, this was on my mind. :D

tests/packages/dynamic-version-from-script/generate_version.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def get_version_from_pyproject():
+    here = os.path.dirname(os.path.abspath(__file__))
+    pyproject_toml = os.path.join(here, 'pyproject.toml')
+    with open(pyproject_toml) as f:
+        for line in f:
+            if line.startswith('version ='):
+                return line.split('=', 1)[1].strip().strip('\'"')
+    raise RuntimeError('version not found in pyproject.toml')
```
dnicolodi: Is there any reason to do not use `tomllib` and fall back to `tomli` on older Python? `tomli` is a meson-python dependency that is not going away till we drop Python < 3.11 support, thus it is guaranteed to be present at build time. Because there are good chanced of this being copied in other projects, I think there is value in keeping this as straight forward and as future proof as possible.

tests/packages/dynamic-version-from-script/generate_version.py
```diff
@@ -0,0 +1,59 @@
...
+        f.write(f"__version__ = '{version}'\n")
+        f.write(f"__git_hash__ = '{git_hash}'\n")
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    group = parser.add_mutually_exclusive_group(required=True)
+    group.add_argument('--print-version', action='store_true')
+    group.add_argument('-o', '--outfile')
+    args = parser.parse_args()
```
dnicolodi, line 50: For keeping this as straight forward as possible, why do not simply write to stdout, unless a file path is specified as first argument?

tests/packages/dynamic-version-from-script/generate_version.py
rgommers: I adapted it from another project that doesn't depend on `tomli`, and in general this will be rare. So to make this an easy copy-paste for other projects, I avoided it. Can change to `tomllib` though, because very few projects will still be supporting Python 3.10 at this point - will update.

tests/packages/dynamic-version-from-script/generate_version.py
rgommers: sure, will update (EDIT: done)

tests/packages/dynamic-version-from-script/generate_version.py
dnicolodi: What I meant is that `tomli` is a dependence of meson-python, and this being an example of how to implement something with meson-python, `tomli` is guaranteed to be there during the build, and this is not supposed to be run at runtime.

rgommers: Ah, good point, that's true even for non-isolated builds.

rgommers:
All comments addressed now

rgommers:
This has been ready for a while @dnicolodi, shall we get this one in too?

tiran:
`vcs-versioning>=2.3.2` now handles `PKG-INFO` in source dists correctly. `setuptools-scm` is no longer needed Ronny has released a new version with a fix a few hours ago. This minimal example works for me on Linux, Windows, and macOS:

```toml
# pyproject.toml
[build-system]
requires = ["meson-python>=0.17", "meson>=1.6", "vcs-versioning>=2.3.2"]
build-backend = "mesonpy"

[project]
dynamic = ["version"]
...
```

```meson
# root meson.build
project(
    ...,
    version: run_command(
        'python', '-m', 'vcs_versioning',
        check: true,
    ).stdout().strip(),
)
```

See https://github.com/tiran/pycxxfilt/pull/19

tests/packages/version-setuptools-scm/meson.build
```diff
@@ -0,0 +1,11 @@
+# SPDX-FileCopyrightText: 2026 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+project(
+    'version-setuptools-scm',
+    version: run_command(
+        ['python3', '-m', 'setuptools_scm'],
```
tiran, line 8: `python3` didn't work for me on Windows. I guess a Windows virtual env does not have a `python3.exe`?

```suggestion
        ['python', '-m', 'setuptools_scm'],
```

rgommers:
Great, thanks @tiran! I'm still hoping for `vcs-versioning` docs so it's clear it is the thing to use/recommend. But I'll have a look at switching to it already - if docs are the only blocker, maybe I can contribute them. I'd be happy to switch over.

tests/packages/version-setuptools-scm/meson.build
rgommers: Hmm, that is annoying. We have Windows CI and this package passes, so I need to figure out what config is missing. Probably install in venv, then kick off pytest. How did you install Python exactly?

In NumPy et al. we use it via a script rather than invoking `python3`; Meson guarantees that the shebang of the script is handled on every platform:

https://github.com/numpy/numpy/blob/6e9bc3714bfefe3c3c31e4f70fb010b41b60b9dd/meson.build#L4

A little annoying if that'd be the solution though, because it requires committing a script file. Can't change it to plain `python` I think, that goes wrong elsewhere - macOS only has `python3`, and probably some Linux distros as well.

tiran: Here is a reproducer, https://github.com/tiran/pycxxfilt/pull/20 . The test case is using `uv pip install`:

```
Run uv pip install --system ".[test]"
Using Python 3.14.7 environment at: C:\hostedtoolcache\windows\Python\3.14.7\x64
   Building pycxxfilt @ file:///D:/a/pycxxfilt/pycxxfilt
  × Failed to build `pycxxfilt @ file:///D:/a/pycxxfilt/pycxxfilt`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `mesonpy.build_wheel` failed (exit code: 1)

      [stdout]
      + meson setup D:\a\pycxxfilt\pycxxfilt
      D:\a\pycxxfilt\pycxxfilt\.mesonpy-o8iw1x_f
      -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md
      --native-file=D:\a\pycxxfilt\pycxxfilt\.mesonpy-o8iw1x_f\meson-python-native-file.ini
      The Meson build system
      Version: 1.12.0
      Source dir: D:\a\pycxxfilt\pycxxfilt
      Build dir: D:\a\pycxxfilt\pycxxfilt\.mesonpy-o8iw1x_f
      Build type: native build

      ..\meson.build:10:13: ERROR: Command
      `C:\hostedtoolcache\windows\Python\3.14.7\x64\python3.exe -m
      vcs_versioning` failed with status 1.

      A full log can be found at
      D:\a\pycxxfilt\pycxxfilt\.mesonpy-o8iw1x_f\meson-logs\meson-log.txt
==== CI platform detected, click here for meson-log.txt
        contents. ====
        Build started at 2026-09-02T12:10:33.901755
        Main binary:
        D:\a\_temp\setup-uv-cache\builds-v0\.tmpznY1kJ\Scripts\python.exe
        Build Options: -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md
        '--native-file=D:\a\pycxxfilt\pycxxfilt\.mesonpy-o8iw1x_f\meson-python-native-file.ini'
        Python system: Windows
        The Meson build system
        Version: 1.12.0
        Source dir: D:\a\pycxxfilt\pycxxfilt
        Build dir: D:\a\pycxxfilt\pycxxfilt\.mesonpy-o8iw1x_f
        Build type: native build
        Running command: C:\hostedtoolcache\windows\Python\3.14.7\x64\python3.exe
        -m vcs_versioning
        --- stdout ---
  
        --- stderr ---
        C:\hostedtoolcache\windows\Python\3.14.7\x64\python3.exe: No module
        named vcs_versioning
  
  
  
        ..\meson.build:10:13: ERROR: Command
        `C:\hostedtoolcache\windows\Python\3.14.7\x64\python3.exe -m
        vcs_versioning` failed with status 1.
```

tiran: cibuildwheel also fails with `python3 -m vcs_versioning`:

```
    + python -m build 'C:\Users\runneradmin\AppData\Local\Temp\cibw-sdist-yw473z_0\pycxxfilt-1.0.1.dev3+g3a52413fe' --wheel '--outdir=C:\Users\runneradmin\AppData\Local\Temp\cibw-run-e91uomyi\cp311-win_amd64\built_wheel' -Csetup-args=-Dpython.allow_limited_api=true
    * Creating isolated environment: venv+pip...
    * Installing packages in isolated environment:
      - meson-python>=0.17
      - meson>=1.6
      - vcs-versioning>=2.3.2
    * Getting build dependencies for wheel...
    * Building wheel...
    + meson setup C:\Users\runneradmin\AppData\Local\Temp\cibw-sdist-yw473z_0\pycxxfilt-1.0.1.dev3+g3a52413fe C:\Users\runneradmin\AppData\Local\Temp\cibw-sdist-yw473z_0\pycxxfilt-1.0.1.dev3+g3a52413fe\.mesonpy-_vv00v2s -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md -Dpython.allow_limited_api=true --native-file=C:\Users\runneradmin\AppData\Local\Temp\cibw-sdist-yw473z_0\pycxxfilt-1.0.1.dev3+g3a52413fe\.mesonpy-_vv00v2s\meson-python-native-file.ini
    The Meson build system
    Version: 1.12.0
    Source dir: C:\Users\runneradmin\AppData\Local\Temp\cibw-sdist-yw473z_0\pycxxfilt-1.0.1.dev3+g3a52413fe
    Build dir: C:\Users\runneradmin\AppData\Local\Temp\cibw-sdist-yw473z_0\pycxxfilt-1.0.1.dev3+g3a52413fe\.mesonpy-_vv00v2s
    Build type: native build
    
    ..\meson.build:10:13: ERROR: Command `C:\Users\runneradmin\AppData\Local\Temp\cibw-run-e91uomyi\cp311-win_amd64\build\venv\Scripts\python3.exe -m vcs_versioning` failed with status 1.
    
    A full log can be found at C:\Users\runneradmin\AppData\Local\Temp\cibw-sdist-yw473z_0\pycxxfilt-1.0.1.dev3+g3a52413fe\.mesonpy-_vv00v2s\meson-logs\meson-log.txt
==== CI platform detected, click here for meson-log.txt contents. ====
    Build started at 2026-09-02T12:10:45.273167
    Main binary: C:\Users\runneradmin\AppData\Local\Temp\build-env-kbz25du8\Scripts\python.exe
    Build Options: -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md -Dpython.allow_limited_api=true '--native-file=C:\Users\runneradmin\AppData\Local\Temp\cibw-sdist-yw473z_0\pycxxfilt-1.0.1.dev3+g3a52413fe\.mesonpy-_vv00v2s\meson-python-native-file.ini'
    Python system: Windows
    The Meson build system
    Version: 1.12.0
    Source dir: C:\Users\runneradmin\AppData\Local\Temp\cibw-sdist-yw473z_0\pycxxfilt-1.0.1.dev3+g3a52413fe
    Build dir: C:\Users\runneradmin\AppData\Local\Temp\cibw-sdist-yw473z_0\pycxxfilt-1.0.1.dev3+g3a52413fe\.mesonpy-_vv00v2s
    Build type: native build
    Running command: C:\Users\runneradmin\AppData\Local\Temp\cibw-run-e91uomyi\cp311-win_amd64\build\venv\Scripts\python3.exe -m vcs_versioning
    --- stdout ---
    
    --- stderr ---
    C:\Users\runneradmin\AppData\Local\Temp\cibw-run-e91uomyi\cp311-win_amd64\build\venv\Scripts\python3.exe: No module named vcs_versioning
    
    
    
    ..\meson.build:10:13: ERROR: Command `C:\Users\runneradmin\AppData\Local\Temp\cibw-run-e91uomyi\cp311-win_amd64\build\venv\Scripts\python3.exe -m vcs_versioning` failed with status 1.
```

rgommers: I had a look at the `setuptools-scm` docs, and the recommendation there is `python -m setuptools_scm`. There is no entrypoint for it. Then I had the brighter idea to look for other projects that already use `setuptools-scm`, and Matplotlib does (and has the solution):

https://github.com/matplotlib/matplotlib/blob/374b2da71ccbfeb9edc955a2420180fed58364d3/meson.build#L4-L7

tiran: That works! I have updated my PR to use `find_program('python3', 'python', version: '>= 3.11'),` and Windows builds are passing.

eli-schwartz: Meson guarantees that `find_program('python3')` will always find a python interpreter because if it cannot find that program name in PATH it will fall back to sys.executable for meson itself (which it knows has to be a 3.x interpreter).

The problems with this are that:
- Meson might not be in the same virtualenv as setuptools-scm / vcs-versioning. :) For Linux distro builds, all build dependencies will be installed to the same environment, of course, and for pip install if Meson isn't installed, it is installed to the isolated environment.
- if python3 doesn't exist in the isolated env, but does exist outside of it / from a different python installation, the version outside always wins and there's no real way to detect the difference.

This is a mess given that different OSes have different ideas about which the "correct" name that must always be present is: python or python3.

To make matters even worse, the BSDs will only have "python3.14", and lack both "python" and "python3".

You basically cannot win.

tiran: The curse of py3k persists... :/

tiran: I have tried `find_program('python3')` in https://github.com/tiran/pycxxfilt/pull/21 . It fails on Windows. `find_program('python3', 'python', version: '>= 3.11')` is passing fine.

eli-schwartz: Sure. The error message there is:

```
    Build started at 2026-09-02T13:12:37.204127
    Main binary: C:\Users\runneradmin\AppData\Local\Temp\build-env-1s9gao16\Scripts\python.exe
    Build Options: -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md -Dpython.allow_limited_api=true '--native-file=C:\Users\runneradmin\AppData\Local\Temp\cibw-sdist-3t1qu6ic\pycxxfilt-1.0.1.dev5+g814e39458\.mesonpy-f6u4fdik\meson-python-native-file.ini'
    Python system: Windows
    The Meson build system
    
[...]

    Program python3 found: YES (C:\Users\runneradmin\AppData\Local\Temp\cibw-run-0di4fykp\cp311-win_amd64\build\venv\Scripts\python3.exe)
    Running command: C:\Users\runneradmin\AppData\Local\Temp\cibw-run-0di4fykp\cp311-win_amd64\build\venv\Scripts\python3.exe -m vcs_versioning
    --- stdout ---
    
    --- stderr ---
    C:\Users\runneradmin\AppData\Local\Temp\cibw-run-0di4fykp\cp311-win_amd64\build\venv\Scripts\python3.exe: No module named vcs_versioning
```

From "Main binary" you can see what the fallback python3 is (the python.exe from `python -m build` is powering Meson and reporting itself as the task manager executable etc.) and also that Meson discovered "Program python3 found: YES" and it is the python3.exe from cibw-run.

That falls under rule 2 above:

- if python3 doesn't exist in the isolated env, but does exist outside of it / from a different python installation, the version outside always wins and there's no real way to detect the difference.

Consequently, the vcs_versioning command succeeded at running python (that is all find_program can really "guarantee") but failed at finding vcs_versioning.

What I said above was maybe a bit confusingly worded. Meson guarantees that `find_program('python3')` will always find *some interpreter, at least* -- even if nowhere on PATH is a "python3" command -- because as a last ditch fallback it can use sys.executable even on platforms where "python3" is (alarmingly) never available at all. However it prefers to find one on PATH, if available, because adding a new python install to the front of PATH usually means one wishes to use it. That messes up here because python -m build doesn't create python3, but cibuildwheel does.

rgommers:
I have adopted the `find_program('python3', 'python')` fix, and rebased this PR. After attempting to add a CI job that would have caught that problem, I decided that that belongs in a follow-up PR, because it's a larger CI change. The problem shows up with doing isolated builds of test packages, which is a CI-only thing (test suites should not download dependencies by default) and turned into a broader update that caught a few more small issues in other test packages.

I also added a commit to switch over from `setuptools-scm` to `vcs-versioning`. It's a like-for-like replacement, no other tweaks needed, as the diff shows.

@dnicolodi I think this is good to go, PTAL.

docs/how-to-guides/dynamic-versioning.rst
```diff
@@ -0,0 +1,179 @@
...
+Then in ``meson.build``, define the version:
+
+.. code-block:: meson
+
+   project('my-project', 'c', version: '1.2.3')
+
+It can also be done the other way around - this requires a bit more code,
+however it has the advantage that all metadata remains static in
+``pyproject.toml``, which can in some cases avoid triggering a build
+when an installer needs to obtain the version. To implement this,
```
dnicolodi, line 51: In which cases an installer needs to determine a package version from `pyproject.toml` and it build the package if the version is declared as a dynamic field? meson-python always resolves the version before writing the package metadata.

docs/how-to-guides/dynamic-versioning.rst
```diff
@@ -0,0 +1,179 @@
...
+    def get_version():
+        pyproject_toml = os.path.join(os.path.dirname(__file__), 'pyproject.toml')
+        with open(pyproject_toml) as f:
+            data = f.readlines()
+
+        version_line = next(
+            line for line in data if line.startswith('version =')
+        )
+        version = version_line.strip().split(' = ')[1]
+        return version.replace('"', '').replace("'", '')
```
dnicolodi, line 90: I would use `tomllib` here too as in the other script.
```suggestion
        return tomllib.load(f)['project']['version']
```

docs/how-to-guides/dynamic-versioning.rst
```diff
@@ -0,0 +1,179 @@
...
+        ).stdout().strip(),
+    )
+
+With that ``get_version.py`` script retrieving the version from
+``pyproject.toml``:
+
+.. code-block:: python
+
+    #!/usr/bin/env python3
+    import os
```
dnicolodi, line 79:
```suggestion
    import os
    try:
        import tomllib
    except ModuleNotFoundError:
        import tomli as tomllib
```

tests/packages/dynamic-version-from-script/meson.build
```diff
@@ -0,0 +1,36 @@
...
+        command: ['generate_version.py', '@OUTPUT@'],
+        build_by_default: true,
+        build_always_stale: true,
+        install: true,
+        install_dir: py.get_install_dir() / 'dynamic_version_from_script',
+    )
+else
+    # When building from sdist, _version.py exists and should be included
+    py.install_sources(
+        ['dynamic_version_from_script/_version.py'],
```
dnicolodi, line 33: If you touch this up for other reasons:
```suggestion
        'dynamic_version_from_script/_version.py',
```
It is a bit strange to use a list here and not in the `py.install_sources()` call a few lines above.

docs/how-to-guides/dynamic-versioning.rst
```diff
@@ -0,0 +1,179 @@
...
+    [project]
+    dynamic = ['version']
+
+Then in ``meson.build``, define the version:
+
+.. code-block:: meson
+
+   project('my-project', 'c', version: '1.2.3')
+
+It can also be done the other way around - this requires a bit more code,
```
dnicolodi, line 48: Here I would add that having the package version in `meson.build` it is not necessary, unless the build process needs it for some reason, which do not exist in the vast majority of cases. I would go as far as recommending this option. Maybe the opening paragraph of this section should be:

```
In many cases the build process does not need to know the package version. Thus
the version string can be left unspecified in ``meson.build`` and stored exclusively 
in ``pyproject.toml`` 

.. code-block:: toml

    [project]
    name = 'my-project'
    version = '1.2.3'

leaving it out from the ``project()`` call in the top-level ``meson.build``:

.. code-block:: meson

   project('my-project', 'c')

This is the simplest and recommended solution. In the less common case where the build
process requires knowledge of the package version, the version string can be specified
via the ``project()`` call in the top-level ``meson.build``:

.. code-block:: meson

   project('my-project', 'c', version: '1.2.3')

``meson-python`` knows how to retrieve the version number from the ``project()``
call when ``version`` is declared to be a dynamic field in ``pyproject.toml``:

.. code-block:: toml

    [project]
    name = 'my-project'
    dynamic = ['version']

It can also be done the other way around: record the version string in
``pyproject.toml`` and retrieve it at build time from ``meson.build``.  This
requires a bit more code, but has the advantage that all metadata remains
static in ``pyproject.toml``.

dnicolodi (commented):
Thanks @rgommers. LGTM. I would just add one clarification. The others are just minor comments that can be addressed in follow ups, if necessary. I haven't built the HTML to check that everything renders as intended, but I trust that this is the case 🙂.

PR #852, opened 2026-06-15
BUG: do no pass default options when reconfiguring

### LINKED ISSUE #851: pip install overrides existing build configuration
When using `meson-python` with editable installs, running `pip install -e` a second time, reusing the build directory, but without repeating the `buildtype` argument overwrites the existing build configuration.

## Reproduction

```bash
#!/usr/bin/env bash

mkdir repro

# Create files for meson-python

cat <<EOF > repro/pyproject.toml
[project]
name = "foo"
version = "0.0.1"

[build-system]
build-backend = "mesonpy"
requires = ["meson-python"]
EOF

cat <<EOF > repro/meson.build
project('foo', 'c', version: '0.1.0')
EOF

python -m pip install meson-python

# Create a debug build
pip install --no-build-isolation -e repro -Csetup-args=-Dbuildtype=debug -Cbuild-dir=debug

grep buildtype repro/debug/meson-logs/meson-setup.txt
# buildtype   : debug

# If I decide to reuse this build directory,
# and I call `pip install` without `-Csetup-args=-Dbuildtype=debug`
# it reconfigures to have `buildtype=release` instead of reusing the current configuration.
pip install --no-build-isolation -e repro -Cbuild-dir=debug

grep buildtype repro/debug/meson-logs/meson-setup.txt
# buildtype   : release
```

Alvaro-Kothe:
This ensures build options specified by the user when the build was first initialized are preserved, unless explicitly overridden.

Fixes #851.

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -899,13 +899,26 @@ class Project():
 
     def _configure(self, reconfigure: bool = False) -> None:
         """Configure Meson project."""
+        # Override Meson default build options to adjust the build
+        # process to the task of building Python wheels. These
+        # defaults can be overridden by user options. To preserve
+        # user specified build options, do not pass these defaults
+        # when reconfiguring an existing build directory.
+        defaults = (
+            [
+                '-Dbuildtype=release',
+                '-Db_ndebug=if-release',
+                '-Db_vscrt=md',
+            ]
+            if not reconfigure
+            else []
+        )
+
         setup_args = [
             os.fspath(self._source_dir),
             os.fspath(self._build_dir),
             # default build options
-            '-Dbuildtype=release',
-            '-Db_ndebug=if-release',
-            '-Db_vscrt=md',
+            *defaults,
             # user build options
             *self._meson_args['setup'],
             # pass native file last to have it override the python
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -332,6 +332,35 @@ def test_invalid_build_dir(package_pure, tmp_path, mocker):
     assert '--reconfigure' not in meson.call_args_list[0].args[1]
     project.build()
 
+def test_reconfigure_preserves_config(package_simple, tmp_path):
+    # initial configuration
+    project = mesonpy.Project(package_simple, tmp_path, meson_args={
+        'setup': ['-Dbuildtype=debug', '-Dwarning_level=2'],
+    })
+
+    initial_options = {opt['name']: opt['value'] for opt in project._info('intro-buildoptions')}
+
+    assert initial_options['buildtype'] == 'debug'
+    assert initial_options['debug'] is True
+    assert initial_options['warning_level'] == '2'
+
+    # reconfiguration with a new option
+    project = mesonpy.Project(package_simple, tmp_path, meson_args={
+        'setup': ['-Doptimization=2'],
+    })
+
+    reconfigure_options = {opt['name']: opt['value'] for opt in project._info('intro-buildoptions')}
+
+    assert reconfigure_options['buildtype'] == 'debug'
+    assert reconfigure_options['debug'] is True
+    assert reconfigure_options['warning_level'] == '2'
+
+    assert initial_options != reconfigure_options
+
+    initial_options['optimization'] = '2'
+    assert initial_options == reconfigure_options
+
+
 
 @pytest.mark.skipif(not os.getenv('CI') or sys.platform != 'win32', reason='requires MSVC')
 def test_compiler(venv, package_detect_compiler, tmp_path):
--- a/tests/test_options.py
+++ b/tests/test_options.py
@@ -35,3 +35,23 @@ def test_ndebug(package_purelib_and_platlib, tmp_path, args, expected):
             ['ninja', '-C', os.fspath(project._build_dir), '-t', 'commands', '../plat.c^'],
             stdout=subprocess.PIPE, check=True).stdout
         assert (b'-DNDEBUG' in command) == expected
+
+
+def test_default_options(package_purelib_and_platlib, tmp_path, mocker):
+    meson = mocker.spy(mesonpy.Project, '_run')
+
+    project = mesonpy.Project(package_purelib_and_platlib, tmp_path)
+
+    mesonpy_default_options = {'-Dbuildtype=release', '-Db_ndebug=if-release', '-Db_vscrt=md'}
+    # Check if default options were passed
+    assert mesonpy_default_options.issubset(meson.call_args_list[0][0][1])
+
+    options = {opt['name']: opt['value'] for opt in project._info('intro-buildoptions')}
+
+    # Check if default options have taken effect
+    assert options['buildtype'] == 'release'
+    assert options['b_ndebug'] == 'if-release'
+
+    compilers = project._info('intro-compilers')
+    if compilers['build']['c']['id'] in {'msvc', 'clang-cl', 'intel-cl'}:
+        assert options['b_vscrt'] == 'md'


mesonpy/__init__.py
```diff
@@ -899,23 +899,33 @@ def _run(self, cmd: Sequence[str]) -> None:
...
-            # pass native file last to have it override the python
-            # interpreter path that may have been specified in user
-            # provided native files
-            f'--native-file={os.fspath(self._meson_native_file)}',
         ]
-        if reconfigure:
-            setup_args.insert(0, '--reconfigure')
-        self._run(self._meson + ['setup', *setup_args])
+
+        if not reconfigure:
```
rgommers: This is the only change, the rest is just a reshuffle. It would be useful to expand the code comment, something like:
```python
# default build options, overriding Meson's defaults for them.
# on a reconfigure, the values will persist unless the user explicitly
# passes new values for them.
```

tests/test_project.py
```diff
@@ -332,6 +332,27 @@ def test_invalid_build_dir(package_pure, tmp_path, mocker):
...
+    # initial configuration
+    project = mesonpy.Project(package_simple, tmp_path, meson_args={
+        'setup': ['-Dbuildtype=debug', '-Dwarning_level=2'],
+    })
+
+    initial_options = {opt['name']: opt['value'] for opt in project._info('intro-buildoptions')}
+
+    # reconfiguration with a new option
+    project = mesonpy.Project(package_simple, tmp_path, meson_args={
+        'setup': ['-Doptimization=2'],
```
rgommers, line 349: This test is correct but a little obscure, given that it doesn't directly checks default/updated values of any of the build options that meson-python actually overrides. Perhaps also assert that the buildtype is and remains `debug`. 

While we're at it here, it looks like there is no test that checks that the default `buildtype` is `release` nor the default for `b_vscrt`. There is an test for `ndebug` in `test_options.py`. Would you mind adding a new test there checking that the defaults are as expected on a clean build?

rgommers (commented):
Thanks @Alvaro-Kothe, this PR is helpful, it looks like a clean solution. It shows that this can be done without anything special, and it makes these build options behave like all other options on a reconfigure, so they're less special while keeping their intended defaults. 

This is a behavior change, but I _think_ we can get away with it. The one relevant situation we had before is building invoking `meson setup` separately to pass build options, before a subsequent `pip install` command. That was necessary in some cases in the early days, because IIRC there were some limitations on what could be passed directly via `pip` or `pypa/build`, especially relevant for cross compilation. Here is an example I remember:

- https://github.com/conda-forge/scipy-feedstock/blob/06d0da3ce0e753836a6cce9cc6db2f515c41ae74/recipe/build.sh
- https://github.com/conda-forge/scipy-feedstock/pull/205#issuecomment-1406439247

It's possible that on less actively maintained distros, such recipes are still in place. If they don't pass `-Dbuildtype=release` and the other options themselves (`-Dn_debug=if-release` is likely to be missing), then that will be broken by this PR, possibly silently at build time.

I think the change is a good one, but we'll have to do some more digging to find cases like the one above that could be affected, and see if they can be updated.

rgommers:
> I think the change is a good one, but we'll have to do some more digging to find cases like the one above that could be affected, and see if they can be updated.

I checked these distros: conda-forge, Spack, Nix, OpenEmbedded/Yocto, Buildroot, Arch, Alpine, Void, Gentoo, Debian, Fedora. None would be affected.

There are some usages in individual packages and end user dev recipes with the `meson setup && pip install` pattern, e.g.:

https://github.com/veethahavya-CU-cz/gwswex/blob/b07177a0e9c4cd6992782f287f5d9379a2ad7c8d/pixi.toml#L65

https://github.com/numpy/numpy-user-dtypes/blob/5942234b1b88bd360e56eb6f3f5f520ffaa19890/stringdtype/reinstall.sh#L11-L13

https://github.com/scipy/scipy/blob/2b736e79c05f4df3a9ab0ae0f7f368dbfecfe117/.github/workflows/wheels.yml#L214-L220

That all uses patterns that are no longer needed, but used to be - so if we want to make this change, we should probably at a minimum emit clear warnings for one release cycle.

mesonpy/__init__.py
```diff
@@ -899,23 +899,33 @@ def _run(self, cmd: Sequence[str]) -> None:
...
+
+        # user build options
+        setup_cmd += self._meson_args['setup']
+
+        # pass native file last to have it override the python
+        # interpreter path that may have been specified in user
+        # provided native files
+        setup_cmd.append(f'--native-file={os.fspath(self._meson_native_file)}')
+
+        self._run(setup_cmd)
```
dnicolodi: I haven't thought much about the implications of this change. However, the implementation can be much more linear:
```python
# Override Meson default build options to adjust the build
# process to the task of building Python wheels. These 
# defaults can be overridden by user options. To preserve
# user specified build options, do not pass these defaults
# when reconfiguring an existing build directory.
defaults = [
    '-Dbuildtype=release',
    '-Db_ndebug=if-release',
    '-Db_vscrt=md',
] if not reconfigure else []

setup_args = [
    os.fspath(self._source_dir),
    os.fspath(self._build_dir),
    # default build options
    *defaults
    # user build options
    *self._meson_args['setup'],
    # pass native file last to have it override the python
    # interpreter path that may have been specified in user
    # provided native files
    f'--native-file={os.fspath(self._meson_native_file)}',
]
```
and the rest remains the same.

dnicolodi (changes requested):

dnicolodi:
I don't think this satisfies the user requirement in all cases.

Let's assume that the `pyproject.toml` has a setting like this:
```toml
[tool.meson-python.args]
setup = ['-Dbuildtype=release']
```
If the user installs the package like this `pip install . -Csetup-args=-Dbuildtype=debug -Cbuild-dir=debug` they get a debug build. However, when they re-install it like this `pip install . -Cbuild-dir=debug` they get a release build because the user options specified in `pyproject.toml` are still applied.

rgommers:
I think that's a different requirement, and I'm not sure how valid it is (it's certainly more niche). I'd expect that package authors only hardcode build options in `[tool.meson-python.args]` if they always want them - and for `buildtype` specifically, I'm not sure it makes sense to do that. 

It could be addressed, but wouldn't that require keeping track of state explicitly? Right now everything is stateless, aside from what's in `meson-info/intro-buildoptions.json`.

mesonpy/__init__.py
```diff
@@ -899,23 +899,33 @@ def _run(self, cmd: Sequence[str]) -> None:
```
dnicolodi, line 899: There is also an imprecision in the commit message: the `reconfigure` argument is passed when the specified build directory is detected to contain a Meson build dir, not when it merely exist. And, as @rgommers already pointed out (IIRC) `buildtype` is not the only default argument passed. I would reword the commit message like this
```
BUG: do no pass default options when reconfiguring

This ensures build options specified by the user when the build was
first initialized are preserved, unless explicitly overridden.

Fixes #851.
```

dnicolodi (changes requested):

tests/test_project.py
Alvaro-Kothe: > Perhaps also assert that the buildtype is and remains debug.

I added some assertions in the initial configuration and in the reconfiguration, but I feel like most of them were redundant, considering the last assertion `initial_options == reconfigure_options`

> There is an test for ndebug in test_options.py. Would you mind adding a new test there checking that the defaults are as expected on a clean build?

Created a new test in `test_options.py`, but I am not sure if the `vscrt` assertion is correct, since I am not on Windows and couldn't test it.

mesonpy/__init__.py
Alvaro-Kothe: Changed. I opted to not use the ternary, since the list is a little big.

mesonpy/__init__.py
Alvaro-Kothe: Updated. Do I need to change the PR description too?

Alvaro-Kothe:
From https://github.com/mesonbuild/meson-python/pull/852#issuecomment-4718634515

> It could be addressed, but wouldn't that require keeping track of state explicitly?

`_meson_args` combine the CLI and `pyproject.toml` arguments, but seems to be possible to pass the cli arguments directly to `_configure`.

---

From https://github.com/mesonbuild/meson-python/pull/852#issuecomment-4716636054

> That all uses patterns that are no longer needed, but used to be - so if we want to make this change, we should probably at a minimum emit clear warnings for one release cycle.

How do I emit warnings? I tried to use `warnings.warn` and `_log` and they don't show up when I use `pip install` unless I add ` --verbose`.

Also, should the warning be a deprecation warning, i.e., create another PR/commit just for the warning, or should it warn on the current behaviour change?

The first link seems a little bit problematic with this change... It seems that they will end up creating `debug` builds instead of `release` builds because they do `meson setup` without setting the `buildtype` option.

mesonpy/__init__.py
rgommers: Thanks. Updating the PR description as well may be useful - at least for me, this captured the essence of the change, and I didn't quite get it from the initial issue or PR description.

rgommers:
> How do I emit warnings?

Let's first check with @dnicolodi whether he agrees with a one-release warning cycle or not. One could also make a case about that probably not reaching the people who need to see it most.

> they don't show up when I use `pip install` unless I add ` --verbose`.

That's unavoidable I think, since it swallows all output.

mesonpy/__init__.py
dnicolodi: Please use the code block exactly as proposed (minus line breaks, I haven't typed it with the correct indentation to fix those), including the comment.

dnicolodi:
> Let's first check with @dnicolodi whether he agrees with a one-release warning cycle or not. One could also make a case about that probably not reaching the people who need to see it most.

This patch seems to make life easier in one case, while introducing an hard to communicate behavior change. It is not clear at all to me that this brings a net benefit. As rediscovered here, there isn't a way to emit a user visible warning, and even so, the vast majority of uses of `meson setup` followed by `pip install` (or equivalent) are in dev or build scripts, where the output is likely not monitored, unless things break.

mesonpy/__init__.py
Alvaro-Kothe: `ruff format` added parenthesis around the ternary if that's okay.

https://play.ruff.rs/acb43456-ea6d-40e5-92d6-447ee66619f0?secondary=Format

PR #853, opened 2026-06-17
Test Python 3.15

hugovk:
Python 3.15 is now in beta, so it's time to start testing:

> We ***strongly encourage*** maintainers of third-party Python projects to ***test with 3.15*** during the beta phase and report issues found to [the Python bug tracker](https://github.com/python/cpython/issues) as soon as possible. While the release is planned to be feature-complete entering the beta phase, it is possible that features may be modified or, in rare cases, removed up until the start of the release candidate phase (2026-08-04). Our goal is to have ***no ABI changes*** after beta 4 and as few code changes as possible after the first release candidate. To achieve that, it will be ***extremely important*** to get as much exposure for 3.15 as possible during the beta phase.

https://discuss.python.org/t/python-3-15-0-beta-2-is-here/107610

This is especially important for foundational projects such as meson-python, so libraries know they can use it for their own 3.15 testing. 

---

Also bump GitHub Actions.

PR #856, opened 2026-06-23
ENH: add support for PEP 803 `abi3t` with Python 3.15.0b2+

mgorny:
Add tag/extension support and initial tests for `abi3t` as specified in PEP 803 and implemented in Python 3.15.0b2.  This is currently limited to building `abi3t` extensions from freethreading builds of Python, since Meson does not support forcing `abi3t` builds explicitly right now.  See https://github.com/mesonbuild/meson/issues/15637 for the relevant discussion.

The additional test case utilizes the new API introduced in PEP 793 and PEP 820, and therefore requires Python 3.15.0b2.  The relevant test tests wheel correctness both with GIL-enabled and free-threading builds of Python 3.15.0b2.

The behavior for older versions of Python remains unchanged.


mgorny:
I didn't touch the test matrix, since testing on 3.15 is already being addressed by #853.

mesonpy/__init__.py
```diff
@@ -874,9 +881,9 @@ def __init__(
             if not allow_limited_api:
                 self._limited_api = False
 
-        if self._limited_api and bool(sysconfig.get_config_var('Py_GIL_DISABLED')):
+        if self._limited_api and bool(sysconfig.get_config_var('Py_GIL_DISABLED')) and sys.version_info < (3, 15, 0, 'beta', 2):
```
mgorny: What's your preferred style for wrapping this line?

dnicolodi: Put it in a function 🙂 The same quite verbose check appears twice now and I think it would be clearer if a private function with a short docstring is added. Finding a good name for such function is left as an exercise to the reader 😄

mgorny: Good idea. Note that there were two variants (`<=` and `>`), so I've moved the version check into `_is_abi3t_capable()`, and while at it also replaced all the repetitions of free-threaded checks with `_is_free_threaded()`. Let me know if you prefer it another way.

tests/packages/limited-api-free-threaded/meson.build
```diff
@@ -0,0 +1,14 @@
+# SPDX-FileCopyrightText: 2023-2026 The meson-python developers
```
dnicolodi: What is this test testing? I think it verifies that abi3t API is available. However, I have the impression that anything that would make this test fail is not under the control of meson-python. In other words, I think the existing `limited-api` test package covers everything this new test package covers.

Integration tests like this are slow to run and having many of them it is hard to disentangle what each one tests. I would like more tests converted to unit tests, rather than more integration tests added.

mesonpy/__init__.py
```diff
@@ -426,6 +436,13 @@ def _stable_abi(self) -> Optional[str]:
         # not use the stable ABI filename suffix and wheels should not
         # be tagged with the abi3 tag.
         if self._limited_api and '__pypy__' not in sys.builtin_module_names:
+            # On free-threaded Python 3.15.0b2+, we expect to be
+            # building 'abi3t' wheels for the time being. In the future
+            # we will want an option to target 'abi3t' from GIL-enabled
+            # Python too.
+            is_abi3t = _is_free_threaded() and _is_abi3t_capable()
```
dnicolodi: I don't love this variable name. I would expect this boolean to tell me that something is abi3t but it is not clear what being abi3t means nor what is the thing that has this property. Maybe calling this simply `abi3t` is better.

mesonpy/__init__.py
```diff
@@ -874,9 +891,9 @@ def __init__(
             if not allow_limited_api:
                 self._limited_api = False
 
-        if self._limited_api and bool(sysconfig.get_config_var('Py_GIL_DISABLED')):
+        if self._limited_api and _is_free_threaded() and not _is_abi3t_capable():
             raise BuildError(
-                'The package targets Python\'s Limited API, which is not supported by free-threaded CPython. '
+                'The package targets Python\'s Limited API, which is not supported by free-threaded CPython < 3.15.0b2. '
```
dnicolodi: This is an user visible error. I would avoid mentioning the exact beta release in which `abi3t` has been introduced:
```suggestion
                'The package targets Python\'s Limited API, which is not supported by free-threaded CPython before version 3.15. '
```

tests/test_editable.py
```diff
@@ -75,7 +75,7 @@ def test_collect(package_complex):
     assert tree['complex']['more']['__init__.py'] == os.path.join(root, 'complex', 'more', '__init__.py')
 
 
-@pytest.mark.skipif(NOGIL_BUILD and CYTHON_VERSION < (3, 1, 0),
+@pytest.mark.skipif(mesonpy._is_free_threaded() and CYTHON_VERSION < (3, 1, 0),
```
dnicolodi: I think the `NOGIL_BUILD` was more self explanatory. Reading the function call `mesonpy._is_free_threaded()` I would imagine that what has the property of being free threaded is `mesonpy`, which does not make much sense.

I would keep the old constant.

dnicolodi (changes requested):

tests/packages/limited-api-free-threaded/meson.build
mgorny: Strictly speaking, it's testing extension using the new PEP 820 API. The existing `limited-api` test cannot compile for `abi3t`. I suppose I could `#if` the existing `limited-api` test instead if you like.

tests/test_editable.py
mgorny: Do you think it would be better to put it as a constant in `mesonpy`? I was torn between using a function and a constant, given this is fixed value.

mesonpy/__init__.py
```diff
@@ -874,9 +886,10 @@ def __init__(
             if not allow_limited_api:
                 self._limited_api = False
 
-        if self._limited_api and bool(sysconfig.get_config_var('Py_GIL_DISABLED')):
+        if self._limited_api and bool(sysconfig.get_config_var('Py_GIL_DISABLED')) and not _is_abi3t_capable():
             raise BuildError(
-                'The package targets Python\'s Limited API, which is not supported by free-threaded CPython. '
+                'The package targets Python\'s Limited API, which is not supported by free-threaded CPython before '
+                'version 3.15. '
                 'The "python.allow_limited_api" Meson build option may be used to override the package default.')
```
dnicolodi: Can you please rewrap these lines in a nicer way?

mgorny: Done. I wasn't sure if they weren't deliberately split over sentences.

tests/packages/limited-api-free-threaded/meson.build
dnicolodi: Sorry, I admit that I haven't followed the `abi3t` discussion closely. But I don't understand. The existing limited API test needs to be amended because, when compiled on Python 3.15 with free threading enabled, the produce wheel ABI tag is `api3.api3t`. I understand this as to mean that the wheel is compatible with `abi3` AND `abi3t`. Here you say that that code cannot be compiled for `abi3t`. Wouldn't this mean that the produced ABI tag is wrong?

I understand that the test cannot use the PEP 820 API, but the fact that the API can be used when the correct limited API version is requested via the dedicated preprocessor macro, is not something that depends on how things are wired in meson-python, thus I don't think we should test for it.

On the other hand, it can be valuable to test how projects really integrate support for `abi3t` and PEP 820. Both for understanding the ergonomics of the solutions we implement, and to provide an example to users. Therefore, it may be worth extending the `limited-api` example to do it.

mesonpy/__init__.py
```diff
@@ -329,6 +329,11 @@ def _is_native(file: Path) -> bool:
             return f.read(4) == b'\x7fELF'  # ELF
 
 
+def _is_abi3t_capable() -> bool:
+    """Return true if Python has abi3t support."""
+    return sys.version_info >= (3, 15, 0, 'beta', 2)
```
dnicolodi: I think that so far all our version tests didn't bother to specify the exact development release in which a feature has been added or a bug has been fixed. I am thinking whether simplifying this to check for simply `(3, 15, 0)` may be appropriate, at least for consistency.

@rgommers What do you think?

tests/packages/limited-api-free-threaded/meson.build
mgorny: Don't worry, it's been changing so much it's hard to follow anyway.

Basically, `abi3t` and the freethreading limited API requires PEP 793 + PEP 820, so the test case needs to be update to use that (and declare support for freethreading). If you try to compile pre-PEP 820 code for limited API on freethreading Python (i.e. the existing limited API test case), it will fail to compile. However, PEP 793 + PEP 820 code does compile fine both for plain `abi3` and `abi3t` (but it requires Python 3.15+, since older Python versions don't support PEP 820).

Or to put it otherwise:
- current `limited api` test is fine for `abi3` but will fail to compile for `abi3t`
- new test is fine for both `abi3` and `abi3t` but requires Python 3.15+

If we don't want two tests for this, I can change the `limited api` test to use PEP 793 + PEP 820 code for Python 3.15+, and old code for <3.15. Then we'd have good coverage as well, except we won't be testing the old extension code for 3.15+ (which I think is fine).

rgommers: I think we should have two test cases here, one for old-style and one for new-style, because otherwise there are Python versions where the test case won't run at all. The diff here is quite small, so I'm not concerned about this being too heavy.

I had a draft PR adding such a second test case already: gh-833. I'll close that in favor of this PR.

mesonpy/__init__.py
rgommers: I agree. We don't really care about testing on 3.15.0b1 and alphas at this point, so (3, 15) working and all (3, 14) being rejected seems fine to me.

rgommers:
Downstream testing, and releasing after rc1 in early August, is currently blocked on this, so I added a v0.21.0 tag - we should do that next month I think.

tests/test_editable.py
rgommers: Constant seems fine to me. Slight preference for `FREE_THREADED_BUILD` over `NOGIL_BUILD` (the latter name will fade over time I think), but I don't care that much either way.

dnicolodi: @rgommers IIRC it was you that introduced the `NOGIL_BUILD` constant 🙂

mesonpy/__init__.py
dnicolodi: Checking just `(3, 15)` is my preference as well. Maybe with this simplification the check is short and explanatory enough that we can inline it again. Sorry for the churn!

tests/packages/limited-api-free-threaded/meson.build
dnicolodi: > Basically, `abi3t` and the freethreading limited API requires PEP 793 + PEP 820, so the test case needs to be update to use that (and declare support for freethreading). If you try to compile pre-PEP 820 code for limited API on freethreading Python (i.e. the existing limited API test case), it will fail to compile.

Why do we need to change the existing `limited-api` test to expect an `abi3.abi3t` wheel tag then? IIUC, that code should fail to compile on platforms where we emit this tag.

tests/test_editable.py
rgommers: Guilty:) I guess that was early days, before the Steering Council pronounced that that was a negative name and they wanted everyone to use free-threading/free-threaded everywhere.

tests/packages/limited-api-free-threaded/meson.build
rgommers: We don't, the existing test remains `abi3` (i.e., GIL-enabled).

mesonpy/__init__.py
mgorny: Ok, done.

tests/test_editable.py
mgorny: For now, I've focused on the changes relevant to the PR. Do you want me to change the constant name, and possibly move it to `mesonpy/__init__.py` here, or separately?

rgommers: > For now, I've focused on the changes relevant to the PR.

That sounds fine to me - small diffs are nice

tests/packages/limited-api-ft/module.c
```diff
@@ -0,0 +1,33 @@
...
+
+static PyMethodDef methods[] = {
+    {"add", add, METH_VARARGS, NULL},
+    {NULL, NULL, 0, NULL},
+};
+
+PyABIInfo_VAR(abi_info);
+
+static PySlot module_slots[] = {
+    PySlot_STATIC_DATA(Py_mod_name, "plat"),
```
rgommers: `"plat"` should be `"module"` right, to match the filename and the `PyModExport_module` below?

tests/packages/limited-api-ft/module.c
```diff
@@ -0,0 +1,33 @@
...
+    PySlot_STATIC_DATA(Py_mod_methods, methods),
+    PySlot_STATIC_DATA(Py_mod_gil, Py_MOD_GIL_NOT_USED),
```
rgommers: `Py_MOD_GIL_NOT_USED` should use `PySlot_DATA` rather than `PySlot_STATIC_DATA`. It's not actually static, and checking upstream in CPython shows it's used like that, e.g. https://github.com/python/cpython/blob/fa81cd976ad88e022839a37501d75d8716e22e3b/Modules/xxlimited.c#L639. The migration guide too: https://docs.python.org/3.15/howto/abi3t-migration.html

rgommers (changes requested):
A couple of small issues in the test extension. The rest looks right; I refreshed my memory on `Py_TARGET_ABI3T` and that should be added automatically in `Python.h` for a free-threaded interpreter if `Py_LIMITED_API` is defined - so all good there.

tests/packages/limited-api-free-threaded/meson.build
dnicolodi: Why is the resulting wheel tagged `abi3.abi3t` then? I think I don't understand what the `abi3.abi3t` means.

dnicolodi: Oh. Scratch all this. I have been reading the commit diff wrong all the time. Sorry for the noise.

rgommers: No worries - resolving.

tests/test_tags.py
```diff
@@ -130,7 +131,12 @@ def test_tag_stable_abi():
         'platlib': [f'extension{ABI3SUFFIX}'],
     }, limited_api=True)
     # PyPy does not support the stable ABI.
-    abi = 'abi3' if '__pypy__' not in sys.builtin_module_names else ABI
+    if '__pypy__' in sys.builtin_module_names:
+        abi = ABI
+    elif NOGIL_BUILD and sys.version_info >= (3, 15):
+        abi = 'abi3.abi3t'
```
dnicolodi, line 140: I think that, if this works, it does by chance. The fake data passed to the wheel builder uses `ABI3SUFFIX` which is defined at the top of this module. That that resolves to `abi3t` because of the orders in which the suffixes are listed in `importlib.machinery.EXTENSION_SUFFIXES`. Is the order guaranteed or should we try to make this more robust? If the order is guaranteed, I would still add a comment in the `get_abi3_suffix()` explaining what it is expected and why it works.

rgommers: A comment would be great. My understanding is that it is guaranteed - and Meson relies on it as well, which came up in the PEP 803 discussions. (that said, future-proofing might be wise).

tests/packages/limited-api-ft/module.c
mgorny: To be honest, I've copied the name from the `limited-api` test. I have no clue why it's `"plat"` there, so I've assumed there must be a reason :-).

tests/test_tags.py
mgorny: Yes, it is guaranteed, CPython is driving module loading on it. For GIL-enabled, it's:

```python
>>> importlib.machinery.EXTENSION_SUFFIXES
['.cpython-315-x86_64-linux-gnu.so', '.abi3.so', '.abi3t.so', '.so']
```

It's ordered from most specific (i.e. presumably most performant) to least specific, to `.so` legacy fallback. I'll add a comment.

tests/packages/limited-api-ft/module.c
mgorny: Changed. I suspect I'm going to follow this up with "misc. fixes" PR xD.

rgommers: You changed it in the old test, but not here yet

mgorny: Uh, sorry about that. Should be correct now.

rgommers:
This is looking pretty good now. Are you expecting it to be squash-merged, or do you want to rewrite history? (we usually do the latter, and rebase-merge)

tests/test_tags.py
dnicolodi: Doesn't this mean that `get_abi3_suffix()` on Python 3.15t returns `.abi3.so` and this test fails because when building on Python 3.15t the extension modules should be suffixed with `.abi3t.so`?

ngoldbaum: > Doesn't this mean that get_abi3_suffix() on Python 3.15t returns .abi3.so and this test fails because when building on Python 3.15t the extension modules should be suffixed with .abi3t.so?

Nope, `EXTENSION_SUFFIXES` is different on 3.15t:

```
Python 3.15.0b2 free-threading build (main, Jun  4 2026, 14:23:26) [Clang 21.0.0 (clang-2100.1.1.101)] on darwin
>>> import importlib.machinery
>>> importlib.machinery.EXTENSION_SUFFIXES
['.cpython-315t-darwin.so', '.abi3t.so', '.so']
```

This is actually explicitly specified in PEP 803: https://peps.python.org/pep-0803/#the-abi3t-wheel-and-filename-tags

ngoldbaum: And yes, `abi3.so` is included on 3.14t and 3.13t and that's probably a bug but it was only noticed in the discussion around PEP 803 after 3.14t came out. Arguably we probably should have fixed this in the lead-up to 3.13t but no one noticed. I have no idea if removing `.abi3.so` from EXTENSION_SUFFIXES on 3.14t would have fallout.

mgorny:
Rewrote it now. I've kept (and reworded) the commit changing `limited-api` test separate.

rgommers (approved):
LGTM now, thanks @mgorny.

@dnicolodi WDYT about releasing this quickly as 0.21.0? Getting this supported would be nice, makes it more likely package authors will experiment with it before the ABI gets frozen in 5-6 weeks from now. I had a look through the rest of the PR queue, and nothing else is critical, although PR 838 is close and we could aim to get that in as well.

tests/packages/limited-api-free-threaded/meson.build
dnicolodi: I would prefer if this test case would get a more descriptive name. I'm sure that before tomorrow I'm going to forget what ft stands for. If spelling it out as `limited-api-free-threaded` is too long, maybe `limited-api-pep803` or `limited-api-abi3t`?

dnicolodi:
LGTM too, with the minor exception of the test package name: I am sure that I will promptly forget what ft stands for, thus a longer more descriptive name would help there.

> I had a look through the rest of the PR queue, and nothing else is critical, although PR 838 is close and we could aim to get that in as well.

I can try to get that in shape this evening. I would still like to get the RPATH handling in order sooner rather than later, but I didn't find the time to finish that work. I guess we can go another release without it.

tests/packages/limited-api-free-threaded/meson.build
rgommers: I'd prefer `limited-api-abi3t`, avoids the "don't use PEP numbers for features" police:)

dnicolodi: What I dislike about that is mixing API and ABI concepts. Technically, the test case exercises the API and the build exercises the ABI (but we don't try to compile with one Python version and run the obtained extension with another version, thus technically we are never testing the stable ABI...). I think I would simply go with the longish `limited-api-free-threaded`. We have to type that just a couple of times.

rgommers: Okay, sure, fine with me

mgorny: Done.

dnicolodi: Thanks!

dnicolodi:
```suggestion
# SPDX-FileCopyrightText: 2026 The meson-python developers
```

tests/packages/limited-api-free-threaded/module.c
```diff
@@ -0,0 +1,33 @@
+// SPDX-FileCopyrightText: 2023-2026 The meson-python developers
```
dnicolodi:
```suggestion
// SPDX-FileCopyrightText: 2026 The meson-python developers
```

tests/packages/limited-api-free-threaded/meson.build
```diff
@@ -0,0 +1,14 @@
...
+project('limited-api-free-thraded', 'c', version: '1.0.0', meson_version: '>= 1.3')
+
+py = import('python').find_installation(pure: false)
+
+py.extension_module(
+  'module',
+  'module.c',
+  limited_api: '3.15',
+  install: true,
+)
```
dnicolodi, line 14: We use 4 spaces indentation in `meson.build` everywhere else. It would be nice to be consistent here.

tests/packages/limited-api-free-threaded/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+# SPDX-FileCopyrightText: 2023-2026 The meson-python developers
```
dnicolodi:
```suggestion
# SPDX-FileCopyrightText: 2026 The meson-python developers
```
or the appropriate year if this was copied from some other test package.

dnicolodi (changes requested):
Sorry, I have a couple more nits before merging this. Let me know if you prefer me to fix these small things up myself.

tests/test_editable.py
dnicolodi: Opened #857 about the renaming. Marking this as resolved.

tests/packages/limited-api-free-threaded/pyproject.toml
mgorny: Yeah, this was all copied from the `limited-api` test and modified. That's why I preserved the dates.

tests/packages/limited-api-free-threaded/meson.build
mgorny: Yeah, this one was copied too xP.

mgorny:
Fixed the indentation. Lemme know if you want to me reset the dates too, though the files have more original content from `limited-api` test than my modifications (or just change them in place).

tests/packages/limited-api-free-threaded/pyproject.toml
dnicolodi: We do not use year ranges in meson-python, but just the date in which a file was created. Thus, if these were copied from the other test, I think it is appropriate to keep just 2023 as the year here.

And, anyway, even if we would use year ranges in meson-python, and you just copied these files around, that would not be reason to bump the copyright year.

mgorny: Ok, done.

dnicolodi (approved):

dnicolodi:
Thanks! Merged.

mgorny:
Thanks!

rgommers:
> Downstream testing, and releasing after rc1 in early August, is currently blocked on this, so I added a v0.21.0 tag - we should do that next month I think.

Note that the reason I haven't put up a PR for a next release yet is that there is a still a request to the steering council to change the extension on POSIX platforms from `.abi3t.so` to `.abi3t-SOABI_PLATFORM.so` (lots of threads, e.g. [this one](https://discuss.python.org/t/can-we-consider-an-uncoditional-change-of-abi3t-so-to-abi3t-soabi-platform-so-for-3-15-last-minute/107919)). That should not require any changes for `meson-python`, but I'd prefer if this was settled so no one uploads wheels with extension module names that might not be final (even though they should remain working).

dnicolodi:
Aren't `abi3t` wheels a Python 3.15 only thing, anyway? That would mean that `abi3t` wheels should be rejected by PyPI at least until 3.15rc1 is out, and by then the extension module file name extension should be settled. There are other package indexes than PyPI, but PyPI should be the only relevant one here.

Anyhow, I don't see strong downsides in delaying a release for a little while longer.

rgommers:
They are >=3.15 yes

> That would mean that `abi3t` wheels should be rejected by PyPI at least until 3.15rc1 is out

That's not how it works, you can upload wheels now. And while that seems unhealthy, I have even seen some CPython core encourage that after beta 1 or beta 2 (they like early signal and the ABI is "probably stable, and otherwise just do a new release"). I don't think many package authors will do that, but it's possible.

dnicolodi:
I stand corrected. Maybe I remember a proposal for blocking wheel updates till rc1, or this is how it worked in the past, or it is how I wish this would be handled. It does not seem a good idea to me to allow wheels target for an ABI that may still change to be uploaded. It opens the possibility for wheels tagged with the wrong ABI to arrive to users, causing very hard to diagnose bugs.

PR #859, opened 2026-06-28
ENH: capture and report build errors when rebuilding editable wheels

### LINKED ISSUE #820: Recompilation in an editable install hides all compiler output

dnicolodi:
When editable-verbose is not enabled, redirect the build output to a file. When the build fails, parse this file to look for the build error and append it to the ImportError exception message.

Fixes #820.

Replaces #750.

### DIFF
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -345,7 +345,7 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
                     subprocess.run(self._build_cmd, cwd=self._build_path, env=env, check=True,
                                    stderr=subprocess.STDOUT, stdout=log)
         except subprocess.CalledProcessError as err:
-            msg = f're-building the {self._name} meson-python editable wheel package failed'
+            msg = f'rebuilding the "{self._name}" editable package failed'
             if log_path:
                 with open(log_path, 'r', encoding='utf8') as log:
                     # Skip to the error.
--- a/mesonpy/_editable.py
+++ b/mesonpy/_editable.py
@@ -319,7 +319,7 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
             dry_run_build_cmd = self._build_cmd + ['-n']
         # Check adapted from
         # https://github.com/mesonbuild/meson/blob/a35d4d368a21f4b70afa3195da4d6292a649cb4c/mesonbuild/mtest.py#L1635-L1636
-        p = subprocess.run(dry_run_build_cmd, cwd=self._build_path, env=env, capture_output=True)
+        p = subprocess.run(dry_run_build_cmd, cwd=self._build_path, env=env, check=False, capture_output=True)
         return b'ninja: no work to do.' not in p.stdout and b'samu: nothing to do' not in p.stdout
 
     @functools.lru_cache(maxsize=1)
@@ -332,15 +332,43 @@ class MesonpyMetaFinder(importlib.abc.MetaPathFinder):
             env[MARKER] = os.pathsep.join((env.get(MARKER, ''), self._build_path))
 
             if self._verbose or bool(env.get(VERBOSE, '')):
+                log_path = None
                 # We want to show some output only if there is some work to do.
                 if self._work_to_do(env):
                     build_command = ' '.join(self._build_cmd)
                     print(f'meson-python: building {self._name}: {build_command}', flush=True)
                     subprocess.run(self._build_cmd, cwd=self._build_path, env=env, check=True)
             else:
-                subprocess.run(self._build_cmd, cwd=self._build_path, env=env, stdout=subprocess.DEVNULL, check=True)
-        except subprocess.CalledProcessError as exc:
-            raise ImportError(f're-building the {self._name} meson-python editable wheel package failed') from exc
+                # Redirect build log to file.
+                log_path = os.path.join(self._build_path, 'meson-logs', 'meson-python-build-log.txt')
+                with open(log_path, 'w') as log:
+                    subprocess.run(self._build_cmd, cwd=self._build_path, env=env, check=True,
+                                   stderr=subprocess.STDOUT, stdout=log)
+        except subprocess.CalledProcessError as err:
+            msg = f're-building the {self._name} meson-python editable wheel package failed'
+            if log_path:
+                with open(log_path, 'r', encoding='utf8') as log:
+                    # Skip to the error.
+                    for line in log:
+                        if line.startswith('FAILED: '):
+                            break
+                    else:
+                        # When no `FAILED: ` line is found, rewind to the
+                        # beginning of the log.
+                        log.seek(0)
+                    if line.strip().endswith(' build.ninja'):
+                        # When the error occurred when rebuilding `ninja.build`,
+                        # the meson output appears before the `FAILED: ` line.
+                        # Rewind the build log to the beginning to report the
+                        # error.
+                        log.seek(0)
+                    error = log.read()
+                if sys.version_info >= (3, 11):
+                    exc = ImportError(msg)
+                    exc.add_note(error)
+                    raise exc from err
+                msg = f'{msg}:\n{error}'
+            raise ImportError(msg) from err
 
         install_plan_path = os.path.join(self._build_path, 'meson-info', 'intro-install_plan.json')
         with open(install_plan_path, 'r', encoding='utf8') as f:
--- a/tests/test_editable.py
+++ b/tests/test_editable.py
@@ -335,7 +335,7 @@ def test_editable_rebuild_error(package_purelib_and_platlib, tmp_path, verbose):
             # Import module and trigger rebuild: the build fails and ImportErrror is raised
             stdout = io.StringIO()
             with redirect_stdout(stdout):
-                with pytest.raises(ImportError, match='re-building the purelib-and-platlib ') as exc:
+                with pytest.raises(ImportError, match='rebuilding the "purelib-and-platlib" ') as exc:
                     import plat  # noqa: F401
             assert not verbose or stdout.getvalue().startswith('meson-python: building ')
 
@@ -373,7 +373,7 @@ def test_editable_reconfigure_error(package_purelib_and_platlib, tmp_path):
             # Import module and trigger rebuild: the build fails and ImportErrror is raised
             stdout = io.StringIO()
             with redirect_stdout(stdout):
-                with pytest.raises(ImportError, match='re-building the purelib-and-platlib ') as exc:
+                with pytest.raises(ImportError, match='rebuilding the "purelib-and-platlib" ') as exc:
                     import plat  # noqa: F401
             msg = exc.value.__notes__[0] if sys.version_info >= (3, 11) else exc.value.msg
             assert 'ERROR: Problem encountered: injected error' in msg
--- a/tests/test_editable.py
+++ b/tests/test_editable.py
@@ -332,7 +332,7 @@ def test_editable_rebuild_error(package_purelib_and_platlib, tmp_path, verbose):
             # Insert invalid code in the extension module source code
             path.write_text('return')
 
-            # Import module and trigger rebuild: the build fails and ImportErrror is raised
+            # Import module and trigger rebuild: the build fails and ImportError is raised
             stdout = io.StringIO()
             with redirect_stdout(stdout):
                 with pytest.raises(ImportError, match='rebuilding the "purelib-and-platlib" ') as exc:
--- a/tests/test_editable.py
+++ b/tests/test_editable.py
@@ -335,9 +335,49 @@ def test_editable_rebuild_error(package_purelib_and_platlib, tmp_path, verbose):
             # Import module and trigger rebuild: the build fails and ImportErrror is raised
             stdout = io.StringIO()
             with redirect_stdout(stdout):
-                with pytest.raises(ImportError, match='re-building the purelib-and-platlib '):
+                with pytest.raises(ImportError, match='re-building the purelib-and-platlib ') as exc:
                     import plat  # noqa: F401
             assert not verbose or stdout.getvalue().startswith('meson-python: building ')
+
+            if sys.version_info >= (3, 11):
+                assert verbose or hasattr(exc.value, '__notes__')
+                assert verbose or 'ninja: build stopped: subcommand failed.' in exc.value.__notes__[0]
+            else:
+                assert verbose or 'ninja: build stopped: subcommand failed.' in exc.value.msg
+
+        finally:
+            del sys.meta_path[0]
+            sys.modules.pop('pure', None)
+            path.write_text(code)
+
+
+def test_editable_reconfigure_error(package_purelib_and_platlib, tmp_path):
+    with mesonpy._project({'builddir': os.fspath(tmp_path)}) as project:
+
+        finder = _editable.MesonpyMetaFinder(
+            project._metadata.name, {'plat', 'pure'},
+            os.fspath(tmp_path), project._build_command,
+            verbose=False,
+        )
+        path = package_purelib_and_platlib / 'meson.build'
+        code = path.read_text()
+
+        try:
+            # Install editable hooks
+            sys.meta_path.insert(0, finder)
+
+            # Emit an error (with unicode characters) during reconfigure
+            with open(path, 'a', encoding='utf8') as f:
+                f.write('\n\nerror(\'injected error \N{BOMB}\')\n')
+
+            # Import module and trigger rebuild: the build fails and ImportErrror is raised
+            stdout = io.StringIO()
+            with redirect_stdout(stdout):
+                with pytest.raises(ImportError, match='re-building the purelib-and-platlib ') as exc:
+                    import plat  # noqa: F401
+            msg = exc.value.__notes__[0] if sys.version_info >= (3, 11) else exc.value.msg
+            assert 'ERROR: Problem encountered: injected error' in msg
+            assert 'ninja: error: rebuilding \'build.ninja\': subcommand failed' in msg
 
         finally:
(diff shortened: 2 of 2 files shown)


mesonpy/_editable.py
```diff
@@ -332,15 +332,28 @@ def _rebuild(self) -> Node:
...
-            raise ImportError(f're-building the {self._name} meson-python editable wheel package failed') from exc
+            msg = f're-building the {self._name} meson-python editable wheel package failed'
+            if log_path:
+                with open(log_path, 'r') as log:
+                    # Skip to the error.
+                    for line in log:
+                        if line.startswith('FAILED: '):
+                            break
+                    error = log.read()
+                msg = f'{msg}:\n{error}'
```
tobiasdiez, line 370: What do you think about also adding the path to the complete log file. The additional context might be helpful to diagnose an issue more easily.

dnicolodi: I'm not opposed, but the whole log adds very little information: it contains only the ninja summary of the compilation and warnings, for example:
```
ninja: Entering directory `.../Src/meson-python/tests/packages/purelib-and-platlib/_build'
[1/2] Compiling C object plat.cp314-win_amd64.pyd.p/plat.c.obj
[2/2] Linking target plat.cp314-win_amd64.pyd
````

File: mesonpy/_editable.py [hunk 2]
```diff
@@ -332,15 +332,28 @@ def _rebuild(self) -> Node:
...
+                log_path = None
                 # We want to show some output only if there is some work to do.
                 if self._work_to_do(env):
                     build_command = ' '.join(self._build_cmd)
                     print(f'meson-python: building {self._name}: {build_command}', flush=True)
                     subprocess.run(self._build_cmd, cwd=self._build_path, env=env, check=True)
             else:
-                subprocess.run(self._build_cmd, cwd=self._build_path, env=env, stdout=subprocess.DEVNULL, check=True)
+                # Redirect build log to file.
+                log_path = os.path.join(self._build_path, 'meson-info', 'meson-python-build-log.txt')
```
dnicolodi: I'm not sure placing the log into the `meson-info` directory is a good idea. It may be see as a namespacing violation. Maybe putting it directly into the build directory is better.

rgommers: Agreed. In the build directory seems perfectly fine to me. Let's leave `meson-*` directories to `meson`.

dnicolodi: Changed

mesonpy/_editable.py
```diff
@@ -332,15 +332,28 @@ def _rebuild(self) -> Node:
...
                 if self._work_to_do(env):
                     build_command = ' '.join(self._build_cmd)
                     print(f'meson-python: building {self._name}: {build_command}', flush=True)
                     subprocess.run(self._build_cmd, cwd=self._build_path, env=env, check=True)
             else:
-                subprocess.run(self._build_cmd, cwd=self._build_path, env=env, stdout=subprocess.DEVNULL, check=True)
+                # Redirect build log to file.
+                log_path = os.path.join(self._build_path, 'meson-python-build-log.txt')
+                with open(log_path, 'w') as log:
+                    subprocess.run(self._build_cmd, cwd=self._build_path, env=env, stdout=log, check=True)
```
rgommers: This was already an issue potentially: `stderr` doesn't seem to be handled. Compilation failures would go to stdout, but there are things that could go to `stderr` I think: `ninja` errors, meson reconfigures.

rgommers: This may also cause an empty message if there are no lines that start with `FAILED`, so `if line.startswith('FAILED: '):` never gets hit.

mesonpy/_editable.py
```diff
@@ -332,15 +332,28 @@ def _rebuild(self) -> Node:
...
         except subprocess.CalledProcessError as exc:
-            raise ImportError(f're-building the {self._name} meson-python editable wheel package failed') from exc
+            msg = f're-building the {self._name} meson-python editable wheel package failed'
+            if log_path:
+                with open(log_path, 'r') as log:
+                    # Skip to the error.
+                    for line in log:
+                        if line.startswith('FAILED: '):
+                            break
+                    error = log.read()
```
rgommers, line 365: The first `line` that the `break` triggers on won't get printed here, right? Would be nice to capture that line as well.

mesonpy/_editable.py
```diff
@@ -332,15 +332,28 @@ def _rebuild(self) -> Node:
...
-                subprocess.run(self._build_cmd, cwd=self._build_path, env=env, stdout=subprocess.DEVNULL, check=True)
+                # Redirect build log to file.
+                log_path = os.path.join(self._build_path, 'meson-python-build-log.txt')
+                with open(log_path, 'w') as log:
+                    subprocess.run(self._build_cmd, cwd=self._build_path, env=env, stdout=log, check=True)
         except subprocess.CalledProcessError as exc:
-            raise ImportError(f're-building the {self._name} meson-python editable wheel package failed') from exc
+            msg = f're-building the {self._name} meson-python editable wheel package failed'
+            if log_path:
+                with open(log_path, 'r') as log:
```
rgommers: I think this should use `encoding='utf-8'`, that's how we read other files - and error messages can contain UTF-8 characters.

rgommers (commented):
I like the idea, and the implementation is nice and concise. A few comments around robustness (encoding, ninja errors, etc.)

mesonpy/_editable.py
dnicolodi: > This was already an issue potentially: `stderr` doesn't seem to be handled. Compilation failures would go to stdout, but there are things that could go to `stderr` I think: `ninja` errors, meson reconfigures.

AFAIK ninja emits all output on standard output, including output that the invoked tools emit on standard error.

The only thing that can go to standard error are ninja errors. AFAIK these are possible only if there is an error in `build.ninja`. If we want to cover this possibility, we need to decide what to do with them. The easy solution is to redirect stderr to stdout and decide that interleaving is not going to be an issue.

> This may also cause an empty message if there are no lines that start with `FAILED`, so `if line.startswith('FAILED: '):` never gets hit.

This would be possible only if running `ninja` fails all together. This happens only if there is a syntax error in `ninja.build`. I think this case should not be common enough to warrant special treatment.

mesonpy/_editable.py
dnicolodi: Leaving it out was a conscious decision. That line reads:
```
FAILED: [code={exitcode}] {target_name}
```
and I don't think it adds much information: the exit code is not relevant in the vast majority of the cases, and the target name is more often than not completely opaque to the user.

However, if you insist, I can include this line.

mesonpy/_editable.py
dnicolodi: One interesting corner case is a failure in reconfiguring the project. In this case, the meson output is emitted before the `FAILED: ` line. For this, a failure in rebuilding `ninja.build` can be detected and the whole compilation output can be attached to the exception message.

dnicolodi: All this is now addressed. PTAL

mesonpy/_editable.py
dnicolodi: Also added a test for this.

mesonpy/_editable.py
eli-schwartz: `meson-info/` contains json introspection data, and feels like a conceptually weird place to store a logfile when `meson-logs/` exists.

I have no particular problem with meson-python using the latter, personally. As long as it has a name denoting "meson-python" it won't clash, whereas a meson.build file can theoretically create all sorts of crazy files if they aren't names meson reserves for itself.

dnicolodi: Thanks @eli-schwartz I somehow forgot about the existence of `meson-logs`. I changed the path of the log file to `$build/meson-logs/meson-python-build-log.txt`.

dnicolodi:
This should be ready to be merged, thus tentatively marking this to be included in the next release.

dnicolodi:
@rgommers do you want to have another look at this?

mesonpy/_editable.py
```diff
@@ -332,15 +332,39 @@ def _rebuild(self) -> Node:
...
+                        # beginning of the log.
+                        log.seek(0)
+                    if line.strip().endswith(' build.ninja'):
+                        # When the error occureed when rebuilding `ninja.build`,
+                        # the meson output appears before the `FAILED: ` line.
+                        # Rewind the build log to the beginning to report the
+                        # error.
+                        log.seek(0)
+                    error = log.read()
+                msg = f'{msg}:\n{error}'
```
dnicolodi, line 370: Thinking more about this: should the compilation error be added as an exception note (at least on Python versions that support it, namely Python 3.11 and later)?

lesteve:
Hi there, thanks a lot @dnicolodi for the PR, this would be very helpful for scientific Python maintainers who are not necessarily meson-python or meson expert (my self included :sweat_smile:).

I gave it a go with scikit-learn and I am a bit confused because I don't get the Cython complication error in the output. I am probably missing something and I will probably try again in a few days and possibly debug what is happening.

Here is what I am doing in the mean time:

- build scikit-learn in editable mode (without editable verbose) `pip install --no-build-isolation --editable .`
- Introduce a Cython compilation error in one of our `.pyx`
- then do `python -c 'import sklearn'`. I am expecting a rebuild which fails and the Cython compilation error to show up in the output but it doesn't somehow:
```
❯ python -c 'import sklearn'                            
  File "/home/lesteve/micromamba/envs/scikit-learn-dev/lib/python3.14t/site-packages/_scikit_learn_editable_loader.py", line 346, in _rebuild
    subprocess.run(self._build_cmd, cwd=self._build_path, env=env, stdout=subprocess.DEVNULL, check=True)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lesteve/micromamba/envs/scikit-learn-dev/lib/python3.14t/subprocess.py", line 577, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/home/lesteve/micromamba/envs/scikit-learn-dev/bin/ninja']' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:

    import sklearn
  File "/home/lesteve/micromamba/envs/scikit-learn-dev/lib/python3.14t/site-packages/_scikit_learn_editable_loader.py", line 312, in find_spec
    tree = self._rebuild()
  File "/home/lesteve/micromamba/envs/scikit-learn-dev/lib/python3.14t/site-packages/_scikit_learn_editable_loader.py", line 348, in _rebuild
    raise ImportError(f're-building the {self._name} meson-python editable wheel package failed') from exc
ImportError: re-building the scikit-learn meson-python editable wheel package failed
```
- my understanding is that `build/cp314t/meson-logs/meson-python-build-log.txt` should contain the Cython error but this file doesn't exist.

And yes, in case you are wondering, I did triple-check that I was using the dev meson-python in this PR branch :wink:.

dnicolodi:
Thanks for testing @lesteve 

> And yes, in case you are wondering, I did triple-check that I was using the dev meson-python in this PR branch 😉.

And yet, you are not 😉 

> Traceback (most recent call last):
>   File "/home/lesteve/micromamba/envs/scikit-learn-dev/lib/python3.14t/site-packages/_scikit_learn_editable_loader.py", line 346, in _rebuild
>     subprocess.run(self._build_cmd, cwd=self._build_path, env=env, stdout=subprocess.DEVNULL, check=True)

With this PR applied, line 346 should not look like this, actually the `subprocess.run()` function call should have its arguments split onto two lines. This code is copied into the editable loader for the installed editable package, thus updating meson-python is not enough. The editable package must be re-installed. I don't know if a package manager, when invoked with `--no-build-isolation`, may fail to detect that one of the build dependencies has been updated and install a cached version of the editable package when no edits have been made to the sources.

dnicolodi:
I've tested with a package containing Cython code and Cython errors are reported correctly:
```
>>> import complex.test
  File "/Users/daniele/src/meson-python/.venv/lib/python3.14/site-packages/_complex_editable_loader.py", line 345, in _rebuild
    subprocess.run(self._build_cmd, cwd=self._build_path, env=env, check=True,
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                   stderr=subprocess.STDOUT, stdout=log)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/subprocess.py", line 578, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/opt/homebrew/bin/ninja']' returned non-zero exit status 1.

  File "<python-input-0>", line 1, in <module>
    import complex.test
  File "/Users/daniele/src/meson-python/.venv/lib/python3.14/site-packages/_complex_editable_loader.py", line 307, in find_spec
    tree = self._rebuild()
  File "/Users/daniele/src/meson-python/.venv/lib/python3.14/site-packages/_complex_editable_loader.py", line 369, in _rebuild
    raise exc
ImportError: re-building the complex meson-python editable wheel package failed
cython -M --fast-fail -3 /Users/daniele/src/meson-python/tests/packages/complex/test.pyx -o test.cpython-314-darwin.so.p/test.pyx.c

Error compiling Cython file:
------------------------------------------------------------
...
# SPDX-License-Identifier: MIT

def answer():
    return 42

baz
^
------------------------------------------------------------
/Users/daniele/src/meson-python/tests/packages/complex/test.pyx:8:0: undeclared name not builtin: baz
ninja: build stopped: subcommand failed.
```

mesonpy/_editable.py
dnicolodi: I moved the compilation error to an exception note. It looks nicer, especially on Python 3.14 where the traceback is colorized.

lesteve:
> And yet, you are not 😉

It's kind of :sob: and :tada: at the same time. I got it to work eventually and this looks great, thanks :pray:!

No idea what was going on (and I almost don't want to understand it :roll_eyes:) I decided to remove the file that the traceback was pointing to (`/home/lesteve/micromamba/envs/scikit-learn-dev/lib/python3.14t/site-packages/_scikit_learn_editable_loader.py`), rebuild and it works, I get the Cython compilation error when the rebuilding fails.

mesonpy/_editable.py
lesteve: To give an idea in a scikit-learn context, this line would look like this:
```
FAILED: [code=1] sklearn/tree/_splitter.cpython-314t-x86_64-linux-gnu.so.p/_splitter.c 
```

I am fine with not having it, since one of the next line points directly at the Cython file (which is more useful than the ninja target I think).

```
/home/lesteve/dev/scikit-learn/sklearn/tree/_splitter.pyx:846:13: Syntax error in simple statement list
```

lesteve (commented):
I don't have the authority to approve this but here is a gentle nudge, which hopefully helps getting this merged: 

- the functionality is super useful and solves a genuine caveat for scientific Python maintainers, who like me are not necessarily advanced meson and meson-python users.
- a few rounds of feed-back already happened and Ralf was happy with the implementation and only had a few questions, which I think have been addressed.
- I had a look at the diff, and as far as I can see this looks good.

rgommers:
I had started having a look, just keep getting distracted. Sorry for the delay, and thanks for the nudges! Agreed it's super useful and we should include it in the next release, which is soon.

rgommers:
@dnicolodi I had another look at this. Overall it looks great. The only thing I could see that might be off is non-UTF-8 error content. E.g., a tool call output or source code line in a traceback displays in whatever the encoding of the tool or file is. 

Reasoning about encodings is always hard, but I _think_ that this would fix it:
- For writing the log file - `open(log_path, 'w')` - change the mode from `'w'` to `'wb'`
- For reading it back - `open(log_path, 'r', encoding='utf8')` - add `errors='backslashreplace'` or `errors=sys.stderr.errors`

dnicolodi:
I have been thinking about the encoding problem a bit more.

> Reasoning about encodings is always hard, but I _think_ that this would fix it:

Indeed, and sometimes Python adds its own complications...

> For writing the log file - `open(log_path, 'w')` - change the mode from `'w'` to `'wb'`

The documentation says that the mode in which the file is opened does not matter (and some local tests verify that this is the case, at least on Unixy platforms). The code behind `subprocess.run()` just grabs the file descriptor from the file object passed as value for the `stdout` argument.  

> For reading it back - `open(log_path, 'r', encoding='utf8')` - add `errors='backslashreplace'` or `errors=sys.stderr.errors`

Here is where things get complicated. I expect tools to write log messages in the default environment encoding. On most Unixy systems this is UTF8, but there are Windows and legacy systems. This would suggest that using `locale.getencoding()` could be a better idea.  `errors='backslashreplace'` should be used as a failsafe anyway.

rgommers:
Yes, that all sounds right. `locale.getencoding` was only added in Python 3.11 and we currently support >=3.9, so on 3.9/3.10 it needs `'utf-8'` or `locale.getpreferredencoding(False)`.

dnicolodi:
See #887.

Issue #861, opened 2026-07-01
Claude Fable review of meson-python


henryiii:
Trying to make use of the one week 50% access to Fable! :) Here's a review of meson-python, following what I've done in other projects I maintain: https://iscinumpy.dev/post/claude-code-reviews/, such as scikit-build-core: https://github.com/scikit-build/scikit-build-core/issues/1317 - I've been really happy with the results, the FP rate is nearly zero. Happy to generate fixes for things if other maintainers want me to!

(Also, congrats, no high-severity bugs. :) )

:robot: _AI text below_ :robot:

Review complete — I read all six source modules in full and verified the suspicious findings against git history. Here's what I found, ordered by how much I'd prioritize each.

## Bugs

1. **`_MESON_REQUIRED_VERSION` is out of sync with `pyproject.toml`** (`mesonpy/__init__.py:95`). The constant says `0.63.3` (unchanged since the runtime check was added in cc2e326), but `pyproject.toml` requires `meson >= 0.64.0`, and `>= 1.2.3` on Python 3.12+. The comment on that line — and CLAUDE.md — say to keep them in sync. Beyond bumping it, the runtime check should probably be conditional on the Python version like the requirement is, otherwise meson 1.0 passes the check on Python 3.12 where it's known broken. Relatedly, on meson < 1.3.0 the `python.allow_limited_api` introspection at `__init__.py:896-899` comes back `None` and silently disables the user's `limited-api = true` — worth confirming that's intended.

2. **No-op `str.replace` in the editable path hook** (`mesonpy/_editable.py:352`). `path.replace(os.altsep, os.sep)` discards its result; it needs `path = path.replace(...)`. On POSIX `altsep` is `None` so nothing is affected, but on Windows the intended separator normalization never happens before `path.rpartition(os.sep)`.

3. **Missing `check=True` on RPath reads** (`mesonpy/_rpath.py:27` and `:75`). The macOS `otool -l` and ELF `patchelf --print-rpath` invocations don't check the return code. If the tool fails (missing patchelf, unreadable file), `_get_rpath` returns nothing, `fix_rpath` becomes a silent no-op, and the wheel ships with an RPATH still pointing at the build directory — a broken wheel with no error. The Solaris variant already passes `check=True`; the other two should match.

[...]

PR #862, opened 2026-07-02
MAINT: synchronize meson version requirements

dnicolodi:
Make the runtime version check the same as the dependency declaration in pyproject.toml and add a test to verify the equivalence.

Requiring 0.64 over 0.63.3 was mostly a cosmetic change to allow to write `import('python').find_installation(pure: false)` in the test packages, thus the runtime version check was not update in sync to allow to use the older meson for a while longer in special settings. The old meson version should have been retired by now, and keeping the version checks consistent is less confusing.

xref gh-861


henryiii (approved):
Nice test. :)

PR #863, opened 2026-07-02
BUG: fix typo in regular expression for wheel file names

dnicolodi:
The bug was inconsequential because the regular expression is never used to match wheel file names containing a build specifier.

xref gh-861


henryiii (approved):

rgommers (approved):
Thanks both!

Issue #787, opened 2025-08-07
Allow the user to specify `limited-api` from the command line


arcondello:
I think this is a more specific version of https://github.com/mesonbuild/meson-python/issues/639.

Currently, so far as I know, the only way to tell `meson-python` to tag a wheel as ABI compatible is with a [project setting](https://mesonbuild.com/meson-python/reference/pyproject-settings.html#project-settings) via the `pyproject.toml`
```toml
[tool.meson-python]
limited-api = true
```

However, for my project I would like to be able to build abi3 wheels for Python 3.12+, and not for lower versions. So it would be nice to be able to instead specify it via a [build config setting](https://mesonbuild.com/meson-python/reference/config-settings.html). Something like
```bash
python -m build --config-setting=limited-api=true
```

I'd be happy to take a crack at a PR if this seems reasonable?

oscarbenjamin:
The way that this is configured is a bit confusing. For the benefit of anyone else who gets here:

If you just want to try out building with the limited API / stable ABI then add the argument for all extension modules:
```meson
limited_api_version = '3.12' # or make a project option

py.extension_module(
   'foo',
   ...
   limited_api: limited_api_version,
)
```
That is what gives extension modules like `foo.abi3.so`.

These can then all be disabled globally by default with
```meson
project('myproject', 'c',
    default_options: ['python.allow_limited_api=false'],
)
```
Now a limited API build is off by default but can be enabled with
```
meson setup build -Dpython.allow_limited_api=true
```
However now meson-python would give a wheel containing abi3 extension modules but without the abi3 tag in the filename:
```
# gives a cp312-cp312 wheel even though the extension modules are abi3:
pip wheel . -Csetup-args=-Dpython.allow_limited_api=true
```
You need to tell meson-python to set the abi3 tag in the wheel filename:
```
[tool.meson-python]
limited-api = true
```
Now it might seem like this means you are always going to use the limited API or that it will be used by default but it does not mean either of those things. The default is controlled by the `default_options` in `meson.build` which you can set to true or false and the non-default case can still be configured with e.g.
```
pip wheel . -Csetup-args=-Dpython.allow_limited_api=true
```
So you should always set `limited-api = true` in pyproject.toml. What this setting means is: "if we use the limited API then meson-python should set the correct abi3 tag in the wheel filename" but that is always what you want so there is no point in ever setting `limited-api = false`.

Unless I am missing something the `limited-api` configuration is not very useful and should just be always true.

[...]

PR #876, opened 2026-08-05
BUG: adapt to platform-specific abi3 and abi3t filename suffixes

dnicolodi:
Python 3.15.0rc1 introduced platform-specific filename suffixes for stable ABI extension modules alongside the generic ".abi3" and ".abi3t".

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -99,7 +99,7 @@ _MESON_REQUIRED_VERSION = '0.64.0' if sys.version_info < (3, 12) else '1.2.3'
 _MESON_ARGS_KEYS = ['dist', 'setup', 'compile', 'install']
 
 _SUFFIXES = importlib.machinery.all_suffixes()
-_EXTENSION_SUFFIX_REGEX = re.compile(r'^[^.]+\.(?:(?P<abi>[^.]+)\.)?(?:so|pyd|dll)$')
+_EXTENSION_SUFFIX_REGEX = re.compile(r'^[^.]+\.(?:(?P<abi>[^.-]+)(?:-[^.]+)?\.)?(?:so|pyd|dll)$')
 assert all(re.match(_EXTENSION_SUFFIX_REGEX, f'foo{x}') for x in importlib.machinery.EXTENSION_SUFFIXES)
 
 # Map Meson installation path placeholders to wheel installation paths.
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -429,7 +429,7 @@ class _WheelBuilder():
         # not use the stable ABI filename suffix and wheels should not
         # be tagged with the abi3 tag.
         if self._limited_api and '__pypy__' not in sys.builtin_module_names:
-            # On free-threaded Python 3.15.0b2+, we expect to be
+            # On free-threaded Python 3.15 or later, we expect to be
             # building 'abi3t' wheels for the time being. In the future
             # we will want an option to target 'abi3t' from GIL-enabled
             # Python too.
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -105,6 +105,14 @@ jobs:
           - os: windows-latest
             python: '3.14'
             meson: '@git+https://github.com/mesonbuild/meson.git'
+          # Test free-threaded builds to verify the correct handling of the
+          # 'abi3t' stable ABI, see PEP 803. Test on Linux and Windows to cover
+          # the case where stable ABI extension modules use a dedicated
+          # filename suffix and the case where they do not.
+          - os: ubuntu-latest
+            python: '3.15t'
+          - os: windows-latest
+            python: '3.15t'
 
     steps:
       - name: Checkout
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -29,14 +29,27 @@ PLATFORM = adjust_packaging_platform_tag(tag.platform)
 
 
 def get_abi3_suffix():
-    # EXTENSION_SUFFIXES are ordered in preference order, and more specific ABI is preferred.
-    # On free-threaded 3.15+, this will match ".abi3t" as the only supported stable ABI.
-    # On GIL-enabled 3.15+, it will match plain ".abi3" first, since that ABI is more specific.
+    """Get the filename suffix identifying stable ABI extension modules."""
+
+    # EXTENSION_SUFFIXES is ordered from the more to the less specific ABI.
+    # On Python 3.15 or later, this resolves to ``.abi3t-${arch}`` for
+    # free-threaded builds and to ``.abi3-${arch}`` on regular builds.
+    #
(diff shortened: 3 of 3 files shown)


mesonpy/__init__.py
```diff
@@ -332,6 +332,29 @@ def _is_native(file: Path) -> bool:
...
+    """Get the filename suffix identifying stable ABI extension modules."""
+
+    # EXTENSION_SUFFIXES is ordered from the more to the less specific ABI.
+    # On Python 3.15 or later, this resolves to ".abi3t-${platform}" for
+    # free-threaded builds and to ".abi3-${platform}" on regular builds.
+    # Python 3.15 and later also accepts the generic ".abi3t" or ".abi3"
+    # suffixes, but we ignore this for the moment.
+    #
+    # Older Python versions do not support a stable ABI for free-threaded
+    # build, and do not support a platform-specific filename suffix, and this
```
rgommers: trivial textual issue with "build": I suggest "for free-threaded builds" or "for free-threading"

mesonpy/__init__.py
```diff
@@ -429,12 +452,12 @@ def _stable_abi(self) -> Optional[str]:
...
         if self._limited_api and '__pypy__' not in sys.builtin_module_names:
-            # On free-threaded Python 3.15.0b2+, we expect to be
+            # On free-threaded Python 3.15 and later, we expect to be
             # building 'abi3t' wheels for the time being. In the future
             # we will want an option to target 'abi3t' from GIL-enabled
             # Python too.
-            abi3t = bool(sysconfig.get_config_var('Py_GIL_DISABLED')) and sys.version_info >= (3, 15)
-            expected_abi = 'abi3t' if abi3t else 'abi3'
+            abi3 = _get_abi3_suffix()
+            assert isinstance(abi3, str)
```
rgommers: I think there's an issue here with cross-compilation: Meson will now generate `abi3.cpython3.xx-<arch>.so` instead of `abi3.so`, and the current implementation gets the suffix from the build rather than the host interpreter, so what matched before will now stop matching. The previous `expected_abi` looks like a more robust approach - both with and without suffix are valid after all, and trying to assert the suffix here is a backwards-incompatible change (and not necessary I believe).

rgommers (changes requested):
Thanks @dnicolodi! It looks like we lost CI coverage for free-threading with the demise of Cirrus. I tried adding 3.15t jobs for Linux and Windows; Linux is happy, but Windows shows two test failures: https://github.com/rgommers/meson-python/actions/runs/31088854382/job/92574711478.

The changes to `_WheelBuilder` look too strict, we need to be careful there until we land a cross-compilation CI job.

dnicolodi:
> Windows shows two test failures

I see why this fails. The fact that stable ABI extension modules do not have a dedicated filename suffix on Windows is a pain.

mesonpy/__init__.py
dnicolodi: > I think there's an issue here with cross-compilation: Meson will now generate `abi3.cpython3.xx-<arch>.so` instead of `abi3.so`, 

The suffix is something like `.abi3-x86-64-linux-gnu.so`.

> and the current implementation gets the suffix from the build rather than the host interpreter, so what matched before will now stop matching.

I understand where the objection is coming from. However, the code here uses a very similar method to derive the expected filename suffix as the one used by Meson, see https://github.com/mesonbuild/meson/blob/f004d10e758073a5aa15b195e7047f5c662f5581/mesonbuild/scripts/python_info.py#L95-L101 If anything, the method used here is a bit more robust, as it does not depend on the exact position of the stable ABI filename suffix in `importlib.machinery.EXTENSION_SUFFIXES`. Therefore, if there is a problem with this way of determining the extension suffix, its roots are deeper than this check.

> The previous `expected_abi` looks like a more robust approach - both with and without suffix are valid after all, and trying to assert the suffix here is a backwards-incompatible change (and not necessary I believe).

I switched to check the whole suffix because I found it easier to do this than to discard the platform part of the filename in the regexp the matches the filename and I didn't see any drawback in doing so. However, we can make the regulkar expression more complicated instead.

rgommers: > However, the code here uses a very similar method to derive the expected filename suffix as the one used by Meson

There's several ways this gets handled today, from patching `python` itself with `crossenv` hacks (which may or may not work for meson-python, hard to predict) to using `build-details.json` instead of the script you linked:

https://github.com/mesonbuild/meson/blob/f004d10e758073a5aa15b195e7047f5c662f5581/mesonbuild/modules/python.py#L122-L130

> we can make the regulkar expression more complicated instead.

Either that, or make `abi != expected_abi:` more forgiving, something like `not abi.startwith(expected_abi):`?

rgommers:
> The fact that stable ABI extension modules do not have a dedicated filename suffix on Windows is a pain.

💯. Doesn't seem like there's any interest upstream in fixing that, unfortunately. Maybe once they finally replace their build systems with something more modern.

eli-schwartz:
> > The fact that stable ABI extension modules do not have a dedicated filename suffix on Windows is a pain.
> 
> 💯. Doesn't seem like there's any interest upstream in fixing that, unfortunately. Maybe once they finally replace their build systems with something more modern.

This was previously discussed around https://github.com/mesonbuild/meson-python/issues/787#issuecomment-3470324145 and @zooba basically said it would be easy to do but nobody thought of it "because the original motivation for adding any kind of data there, was for a Linux specific reason and nobody really reevaluated the assumption since then".

mesonpy/__init__.py
dnicolodi: > > we can make the regulkar expression more complicated instead.
> 
> Either that, or make `abi != expected_abi:` more forgiving, something like `not abi.startwith(expected_abi):`?

This does not work for distinguishing between `abi3` and `abi3t`. Right now this is not a concern, but it may become one in the future when building for `abi3` from a free-threaded build (or vice-versa) will be a thing.

dnicolodi: I solved this making the regular expression a tiny bit more complex.

dnicolodi:
@rgommers I think I addressed your comments. Please take a look.

dnicolodi:
@rgommers Unless you have more comments, I'll be merging this soon

rgommers:
Thanks @dnicolodi! I had an event the week before last, so I had a hard time keeping up with OSS responsibilities. I did some post-hoc review here, this looks great now. I opened gh-882 as a follow-up with a couple of test improvements.

dnicolodi:
Thanks @rgommers. I didn't want to force matters but I think it would be nice to have a release out soon to allow `abi3t` build of projects using meson-python, thus when I didn't hear from you for a while I went ahead as I felt that this was ready.

rgommers:
Yeah no worries - you did ping me and announced your intention to merge, so all good. I was just so busy I didn't even read the ping in time (which is rare, luckily).

Also agreed that we need to get a release out. Let's try to do that this week?

PR #881, opened 2026-08-23
ENH: implement editable loader with PEP 829 .start files

### LINKED ISSUE #846: Update editable install support for PEP 829 (`.start` files)
[PEP 829 – Package Startup Configuration Files](https://peps.python.org/pep-0829/) is deprecating `import` statements in `.pth` files, and replaces them with entrypoints in `.pth` files and `.start` files for Python 3.15. It looks like editable installs were considered in the discussion and the replacement should be a small diff. However, no one tested that it looks like, so it's hard to be sure - we should try this soon to see if there are any unexpected issues. The implementation landed two days ago in https://github.com/python/cpython/pull/149109, so should be available in 3.15.0b1

dnicolodi:
on Python 3.15 and later. Editable wheels are Python version specific, thus there is no need to write both a .pth and a .start file for Python versions that support the latter.

Editable wheels implemented with .pth files just imported the implementation module and relied on the execution of the module body on import for the installation of the module loader.  PEP 829 .start files need to specify a callable that takes no arguments. To keep the difference between .pth and .start file implementations to a minimum, define such callable in the implementation module and switch the .pth file to explicitly call it.

Fixes #846

### DIFF
--- a/mesonpy/__init__.py
+++ b/mesonpy/__init__.py
@@ -553,18 +553,28 @@ class _EditableWheelBuilder(_WheelBuilder):
             whl.writestr(
                 f'{loader_module_name}.py',
                 importlib.resources.files('mesonpy').joinpath('_editable.py').read_bytes() + textwrap.dedent(f'''
-                   install(
-                       {self._metadata.name!r},
-                       {self._top_level_modules!r},
-                       {os.fspath(build_dir)!r},
-                       {build_command!r},
-                       {verbose!r},
-                   )''').encode('utf-8'))
+                   def init():
+                       finder = MesonpyMetaFinder(
+                           {self._metadata.name!r},
+                           {self._top_level_modules!r},
+                           {os.fspath(build_dir)!r},
+                           {build_command!r},
+                           {verbose!r},
+                       )
+                       sys.meta_path.insert(0, finder)
+                       sys.path_hooks.insert(0, finder._path_hook)
+                   ''').encode('utf-8'))
 
-            # install .pth file
-            whl.writestr(
-                f'{self._metadata.canonical_name}-editable.pth',
-                f'import {loader_module_name}'.encode('utf-8'))
+            if sys.version_info >= (3, 15):
+                # install .start file
+                whl.writestr(
+                    f'{self._metadata.canonical_name}-editable.start',
+                    f'{loader_module_name}:init'.encode('utf-8'))
+            else:
+                # install .pth file
+                whl.writestr(
+                    f'{self._metadata.canonical_name}-editable.pth',
+                    f'import {loader_module_name}; {loader_module_name}.init()'.encode('utf-8'))
 
         return wheel_file
 
--- a/tests/test_editable.py
+++ b/tests/test_editable.py
@@ -13,6 +13,7 @@ import sys
 from contextlib import redirect_stdout
 
 import pytest
+import wheel.wheelfile
 
 import mesonpy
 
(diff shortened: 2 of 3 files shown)


mesonpy/__init__.py
```diff
@@ -553,18 +553,25 @@ def build(self, directory: Path, source_dir: pathlib.Path, build_dir: pathlib.Pa
...
-                       {build_command!r},
-                       {verbose!r},
+                   def init():
+                       install(
+                           {self._metadata.name!r},
+                           {self._top_level_modules!r},
+                           {os.fspath(build_dir)!r},
+                           {build_command!r},
+                           {verbose!r},
                    )''').encode('utf-8'))
```
rgommers: nit: this bracket is de-indented too much in the generated file. I'm seeing:

```python
def init():
    install(
        'meson-python',
        {'mesonpy'},
        '/Users/rgommers/code/pixi-dev-scipystack/meson-python/meson-python/build/cp313',
        ['/Users/rgommers/code/pixi-dev-scipystack/meson-python/.pixi/envs/default/bin/ninja'],
        False,
)
```

Bumping it by four spaces would fix that.

rgommers (approved):
Thanks @dnicolodi. This LGTM, and it seems to work as advertised. One tiny stylistic comment. 

I considered the "ship both" vs. "gate on >=3.15" - both are defensible, so happy to go with your preference here.

I also considered whether this needs more test coverage. A small test that actually checks whether `.start` is present and `.pth` is not could be added, however the current editable tests cover perfectly well that things are working and the chance that we'll regress to shipping a `.pth` file seems quite low, so it's fine as is.

Should be good to merge as is, or after addressing the small style comment.

dnicolodi:
> A small test that actually checks whether `.start` is present and `.pth` is not could be added

I hesitated adding this test when I implemented the PR. Given that you had the same instinct, I added it.

> Should be good to merge as is, or after addressing the small style comment.

I went for a small refactoring that removed the indirection through the `install()` function. PTAL

tests/test_editable.py
```diff
@@ -190,6 +191,21 @@ def test_editable_install(venv, editable_simple):
     assert venv.python('-c', 'import simple; print(simple.data())').strip() == 'ABC'
 
 
+def test_editable_contents(editable_simple):
+    artifact = wheel.wheelfile.WheelFile(editable_simple)
+
+    impl = 'start' if sys.version_info > (3, 15) else 'pth'
```
rgommers: `> (3, 15)` is actually correct, but I had to do a double take here; `>= (3, 15)` behaves identically and matches the implementation and intuition better, so I'd change it to that.

rgommers (approved):
Refactoring LGTM. One other very minor comment around readability. The commit message of the second commit is incomplete, so it needs squashing.

tests/test_editable.py
eli-schwartz: It is practically correct but semantically incorrect.

Given that there is never actually a 3.15 release, only a 3.15.0 release, comparing this as a tuple of ints instead of as an "abstract version number" will always see the side with trailing bits as greater. However technically it is wrong to assume that you're checking for "at least patchlevel greater than `<omitted argument>`" (even NULL is sorta ironically not a good fit to model this case), quite aside from the intuition aspect.

dnicolodi:
> One other very minor comment around readability. 

That was actually a typo.

> The commit message of the second commit is incomplete, so it needs squashing.

And I fat fingered something to get the silly commit message, the intention was fixup the previous commit.

Thanks for the review and for merging.

PR #882, opened 2026-08-24
TST: fix a small abi3t tag test issue, and add a new test for multiarch filename

rgommers:
This is a follow-up to gh-876. The new multiarch test addresses a gap in test coverage that is important - we've stumbled over this case a few times over the past years.

### DIFF
--- a/tests/test_tags.py
+++ b/tests/test_tags.py
@@ -54,6 +54,7 @@ def get_abi3_suffix():
 
 SUFFIX = sysconfig.get_config_var('EXT_SUFFIX')
 ABI3SUFFIX = get_abi3_suffix()
+STABLE_ABI_KIND = 'abi3t' if FREE_THREADED_BUILD and sys.version_info >= (3, 15) else 'abi3'
 
 
 def test_wheel_tag():
@@ -162,3 +163,16 @@ def test_tag_mixed_abi():
     }, pure=False, limited_api=True)
     with pytest.raises(mesonpy.BuildError, match='The package declares compatibility with Python limited API but '):
         assert str(builder.tag) == f'{INTERPRETER}-abi3-{PLATFORM}'
+
+
+@pytest.mark.skipif(not ABI3SUFFIX, reason='interpreter does not support the stable ABI')
+def test_tag_stable_abi_multiarch():
+    # Python 3.15 and later append the multiarch tuple to the stable ABI
+    # filename suffix. Verify that it is ignored rather than causing the
+    # module to be rejected: when cross compiling it differs from the one
+    # of the build interpreter.
+    builder = wheel_builder_test_factory({
+        'platlib': [f'extension.{STABLE_ABI_KIND}-aarch64-linux-gnu.so'],
+    }, pure=False, limited_api=True)
+    abi = 'abi3.abi3t' if STABLE_ABI_KIND == 'abi3t' else 'abi3'
+    assert str(builder.tag) == f'{INTERPRETER}-{abi}-{PLATFORM}'
--- a/tests/test_wheel.py
+++ b/tests/test_wheel.py
@@ -411,5 +411,5 @@ def test_limited_api_free_threaded(wheel_limited_api_free_threaded):
     artifact = wheel.wheelfile.WheelFile(wheel_limited_api_free_threaded)
     name = artifact.parsed_filename
     assert name.group('pyver') == INTERPRETER
-    assert name.group('abi') == 'abi3.abi3t' if FREE_THREADED_BUILD else 'abi3'
+    assert name.group('abi') == ('abi3.abi3t' if FREE_THREADED_BUILD else 'abi3')
     assert name.group('plat') == PLATFORM


tests/test_tags.py
```diff
@@ -162,3 +163,16 @@ def test_tag_mixed_abi():
...
     with pytest.raises(mesonpy.BuildError, match='The package declares compatibility with Python limited API but '):
         assert str(builder.tag) == f'{INTERPRETER}-abi3-{PLATFORM}'
+
+
+@pytest.mark.skipif(not ABI3SUFFIX, reason='interpreter does not support the stable ABI')
+def test_tag_stable_abi_multiarch():
+    # Python 3.15 and later append the multiarch tuple to the stable ABI
+    # filename suffix. Verify that it is ignored rather than causing the
+    # module to be rejected: when cross compiling it differs from the one
+    # of the build interpreter.
```
dnicolodi, line 173: I had this kind of issue on my mind when thinking about the support for `build-detail.json` in #829. We have this check right now https://github.com/mesonbuild/meson-python/blob/7c78da354abacda6b39aa78c2dd2eec7bff44cf1/mesonpy/__init__.py#L436-L437 which makes it impossible to cross-compile between a Python platform that uses abi3t and one that does not.

tests/test_wheel.py
```diff
@@ -411,5 +411,5 @@ def test_limited_api_free_threaded(wheel_limited_api_free_threaded):
     artifact = wheel.wheelfile.WheelFile(wheel_limited_api_free_threaded)
     name = artifact.parsed_filename
     assert name.group('pyver') == INTERPRETER
-    assert name.group('abi') == 'abi3.abi3t' if FREE_THREADED_BUILD else 'abi3'
+    assert name.group('abi') == ('abi3.abi3t' if FREE_THREADED_BUILD else 'abi3')
```
dnicolodi, line 414: Ups! Thanks for catching this.

tests/test_tags.py
dnicolodi: The regular expression used to extract the ABI from the filename matches both `f'extension.{STABLE_ABI_KIND}-aarch64-linux-gnu.so'` and `f'extension.{STABLE_ABI_KIND}.so'` and any other valid extension module filename, and this is verified at runtime. One test more is better than one test less, thus wer should include this one, but I would like to add some proper cross-compilation tests.

rgommers: >  I would like to add some proper cross-compilation tests.

💯 me too. However, that's for the `build-details.json` PR or a follow-up to that I think. This test is cheap (just regex/strings, no building), and would have caught an issue.

rgommers: > which makes it impossible to cross-compile between a Python platform that uses abi3t and one that does not.

I think we can fix that up by extending support to more combinations once gh-829 is in? In practice there are lots of combinations of build/host-platform that are difficult. The most common one uses the same Python version and just changes the host platform. That's the test case I'd add first. Then extend from there.

dnicolodi: > This test is cheap (just regex/strings, no building), and would have caught an issue.

Out of curiosity, which issue would this test have caught that other tests did have not?

dnicolodi: > I think we can fix that up by extending support to more combinations once [gh-829](https://github.com/mesonbuild/meson-python/pull/829) is in?

Sure. It is just matter to replace checks against the values obtained from the current interpreter with checks against what is in `build-details.json`. Should we try to get gh-829 in for this release?

rgommers: My review comment on gh-876: https://github.com/mesonbuild/meson-python/pull/876#discussion_r3727922919

Which I think was a recurrence of the thing fixed in gh-322.

rgommers: > Should we try to get [gh-829](https://github.com/mesonbuild/meson-python/pull/829) in for this release?

Yes, that'd be great.

dnicolodi: > > Should we try to get [gh-829](https://github.com/mesonbuild/meson-python/pull/829) in for this release?
> 
> Yes, that'd be great.

I can try to find time to finish the PR. However, I would be hesitant to merge it without an integration test. Ideally the test would compiler an extension module, thus it requires a C cross compiler and a `build-details.json` for the target foreign architecture. What's the easiest way to get these?

rgommers: We can also get the release out, merge it right after, and do another release. To take it off of the critical path (we shouldn't drop it for months though).

>  However, I would be hesitant to merge it without an integration test. Ideally the test would compiler an extension module, thus it requires a C cross compiler and a build-details.json for the target foreign architecture. What's the easiest way to get these?

Conda/pixi is the easiest way to get a cross compilation toolchain. I already have an integration test for `scipy`; would need to scope that down to running on a test package (or set of packages) and then adding it in a new CI job.

Let's move the CI coverage part to gh-829 though? I think we can discuss the release plan here; this PR is ready to go, and we can probably push to get 1-2 more PRs in, but then ship 0.21.0

PR #884, opened 2026-08-25
CI: add a pixi job - conda-forge python/deps, linux aarch64

### LINKED ISSUE #847: Cirrus CI shuts down on June 1st 2026
Acquired by OpenAI https://cirruslabs.org/...

We need to find another solution for our CI.

rgommers:
I decided to run this job on a Linux aarch64 runner, so we get some more variety for free here - we weren't testing on that platform at all, other jobs all use `ubuntu-latest`.

One minor annoyance: this requires adding a `REUSE.toml`, because by design https://codeberg.org/fsfe/reuse-tool doesn't support ignoring files in a simple way: https://reuse.software/faq/#exclude-file

Closes gh-847


.github/workflows/pixi.toml
```diff
@@ -0,0 +1,33 @@
+# SPDX-FileCopyrightText: 2025 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+[workspace]
+name = "meson-python"
+version = "0.22.0.dev0"
+description = "Meson PEP 517 Python build backend"
```
dnicolodi: This remembers me that maybe it is time to change the package description everywhere to do not mention PEP 517 anymore. It seems something only of historical relevance now.

.github/workflows/pixi.toml
```diff
@@ -0,0 +1,33 @@
+# SPDX-FileCopyrightText: 2025 The meson-python developers
+#
+# SPDX-License-Identifier: MIT
+
+[workspace]
+name = "meson-python"
+version = "0.22.0.dev0"
```
dnicolodi: Can we omit this? It seems another place where we encode the version number that we may forget to update. If a version number is required, we should probably extend the test that checks the consistency of the version numbers to check this location too.

.github/workflows/pixi.toml
```diff
@@ -0,0 +1,33 @@
...
+description = "Meson PEP 517 Python build backend"
+license = "MIT"
+documentation = "https://mesonbuild.com/meson-python"
+repository = "https://github.com/mesonbuild/meson-python"
+channels = ["https://prefix.dev/conda-forge"]
+platforms = ["linux-aarch64"]
+
+[dependencies]
+meson = ">=1.12.0,<2"
```
dnicolodi: Is there any reason to pin meson to this versions range?

.github/workflows/pixi.toml
```diff
@@ -0,0 +1,33 @@
...
+c-compiler = "*"
+pip = "*"
+python-build = "*"
+cython = "*"
+pytest = "*"
+pytest-mock = "*"
+wheel = "*"
+
+[tasks]
+build = { cmd = "pip install -e . --no-build-isolation", cwd = "../.." }
```
dnicolodi: What's the reason to prefer an editable installation in CI jobs? The other CI jobs use a regular installation and the install is done only used as the easiest way to install all  get the dependencies, but here the dependencies are already added to the workspace. I would probably remove the `build` step.

REUSE.toml
```diff
@@ -0,0 +1,11 @@
+# SPDX-FileCopyrightText: 2026 The meson-python developers
+
+# SPDX-License-Identifier: MIT
+
+version = 1
+
+[[annotations]]
```
dnicolodi, line 7: For other files were we cannot or do not want t have license headers we use `.license` companion files. I would keep this consistent. Although I think I like this solution better.

dnicolodi (changes requested):
Thanks @rgommers. I have a few comments.

.github/workflows/pixi.toml
rgommers: Good point, let's fix that now. How about "The Python build backend for Meson projects" ?

.github/workflows/pixi.toml
rgommers: Not really, we can replace it with `*`, we'll get the latest version in the lock file either way, unless the resolver does something very strange. This is more documentation than anything.

REUSE.toml
rgommers: Ah right, we already have three:
```
docs/spelling_wordlist.txt.license
tests/packages/install-data/data.txt.license	2025
tests/packages/simple/data.txt.license	2021
```

I do like this solution better as well. How about I remove those `.license` files in this PR?

.github/workflows/pixi.toml
rgommers: Hmm good question, I didn't consider not installing at all. But sure, why not - saves a few lines and a few seconds.

.github/workflows/pixi.toml
rgommers: Looks like it is allowed now: https://pixi.prefix.dev/latest/reference/pixi_manifest/#version-optional. So yes, let's drop it.

.github/workflows/pixi.toml
dnicolodi: Sounds good. The "The" at the beginning may sound a bit pretentious as AFAIK there (or at least there were) other projects with the same scope, but meson-python is by far the most used (I believe).

.github/workflows/pixi.toml
dnicolodi: I also thought that this is more documentation than anything, thus it would need to be more nuanced that this. Maybe simply `*` is best.

eli-schwartz: The `<2` is even more interesting than requiring a minimum of the just-released version. :)

Meson currently has no plans for a semver major 2 release, in the sense that we will undoubtedly want to make one someday but it is quite unlikely to be any time soon. It will surely drop deprecated functionality leading to breakage, but only for projects that don't deal with deprecation warnings for years...

In general this kind of rolling "you get a few years to handle deprecations" approach to compatibility is something I think is nicer than pinning to major versions (even if I *also* like SemVer :P). It would be slightly disappointing to to see `<2` locking start happening in general -- is this something the tooling automatically defaulted to, or ...?

REUSE.toml
dnicolodi: > How about I remove those `.license` files in this PR?

Seems like a good idea. I think `docs/spelling_wordlist.txt` can go completely.

.github/workflows/pixi.toml
dnicolodi: > is this something the tooling automatically defaulted to, or ...?

I don't know for this specific case, but `pixi` defaults to this kind of version ceiling when adding a package to a workspace. I don't like it either, but it is a very popular thing to do.

.github/workflows/pixi.toml
eli-schwartz: AFAIK the only other extant pep517 build backend directly using meson was `mesonpep517` and that is obsolescent (last developed in January 2022, still uses `[tool.*]` to specify all metadata).

There are some other build backends that just take python functions as configuration, and people use that to do `subprocess.run(['meson', ...])`, then copy over files from the build directory. The author of poetry did that for a while (pendulum) before rewriting in rust, for example. And then even the rust rewrite used `build.py` to run maturin, then unzipped the resulting wheel and extracted the .so files for itself. :P

dnicolodi: A while back, someone forked meson-python to turn it into an hatchling plugin (and add features that we think are not a great idea) https://github.com/virtuald/hatch-meson I don't know if the result of that is used by any project.

dnicolodi: I am happy with the "The" 🙂

eli-schwartz: ... well, if it's a hatchling plugin, it's not a build backend. :D 

I would probably view that as "a library to abstract the subprocess.run approach".

.github/workflows/pixi.toml
rgommers: Yes, it's the default behavior you get from `pixi`. It pins minor versions for 0.x version numbers, and major versions for >=1 numbers. This kind of choice is always controversial no matter what one chooses exactly. Given that `pixi` is language-independent and handles compilers, C libraries, etc. it seems like a reasonable choice to me to not blindly go over major version boundaries when a user runs `pixi update`.

For Python packages I agree, not a fan. But there we don't typically worry about ABIs etc., nor is SemVer all that common.

> In general this kind of rolling "you get a few years to handle deprecations" approach to compatibility is something I think is nicer than pinning to major versions

✅

REUSE.toml
rgommers: Updated. I considered letting them all have separate entries rather than a single list of files in `REUSE.toml`, for the purpose of keeping the year of the copyright snippet different between them. However, these are all completely trivial or generated files anyway, so there seems little point in making things more verbose for that. 

I avoided using CC-0 instead of MIT for the generated file for the same reason - keep it simple, and keep a single license. If we added CC-0 in REUSE.toml, some over-diligent packager would probably come report that our package-level SPDX expression is wrong.

.github/workflows/pixi.toml
rgommers: Removed `build` task, and CI is still happy, so resolving.

.github/workflows/pixi.toml
dnicolodi: This could be
```
meson = ">= 1.2.3"
```
to match a simplified version of what is in `pyproject.toml` but I am fine with the `*` too.

rgommers: Yeah I thought about it, but then it might goes out of sync again, so I'm happy with `*`.

dnicolodi (approved):

PR #885, opened 2026-08-26
DOC: improve docs for limited api builds

### LINKED ISSUE #624: Document setting limited-api via build option better

ngoldbaum:
Adds a new howto guide on targeting the CPython limited API and producing stable ABI wheels.

Fixes #624

### DIFF
--- a/docs/how-to-guides/limited-api.rst
+++ b/docs/how-to-guides/limited-api.rst
@@ -1,0 +1,176 @@
+.. SPDX-FileCopyrightText: 2026 The meson-python developers
+..
+.. SPDX-License-Identifier: MIT
+
+.. _howto-limited-api:
+
+***********************************
+Targeting the CPython Limited C API
+***********************************
+
+This page describes how to configure your package to build against the
+CPython `limited API`_ and build `stable ABI`_ wheels. Limited API builds
+target a minimum interpreter version and produce a single wheel that
+can support newer interpreter versions.
+
+You can enable limited API builds by default or make them opt-in.
+We recommend making them opt-in to avoid failures or special handling
+when building for interpreters that do not support the stable ABI, such
+as free-threaded CPython 3.14. Using the version-specific C API can also
+`improve performance`_ for source installations and distribution packages
+that do not need the stable ABI.
+
+Compile your extension modules for the limited API by specifying the
+``limited_api`` argument to the ``extension_module()`` function in the
+Meson Python module. The version passed to ``limited_api`` selects
+the limited API your extension uses. Meson converts this version to
+the corresponding value of the ``Py_LIMITED_API`` preprocessor macro.
+
+The wheel's :ref:`Python compatibility tag <limited-api-wheel-tags>`
+still corresponds to the interpreter used for the build, regardless of
+the ``limited_api`` version you select.
+
+Enabling the
+:option:`tool.meson-python.limited-api` setting declares that all the
+extension modules in your package target the limited API and adjusts the
+wheel filename ABI tag accordingly:
+
+.. code-block:: meson
+
+   py = import('python').find_installation(pure: false)
+
+   py.extension_module(
+       '_core',
+       '_core.c',
+       limited_api: '3.10',
+       install: true,
+       subdir: 'example',
+   )
+
+.. code-block:: toml
+
+   [tool.meson-python]
+   limited-api = true
+
+
+For limited API builds, on platforms where Python extension modules
+targeting the stable ABI use a dedicated filename suffix, ``meson-python``
+verifies that all the extension modules included in the wheel use the
+stable ABI filename suffix and fails the build otherwise.
+
+
+.. _opt-in-limited-api-example:
+
+Set up Limited API support as opt-in
+------------------------------------
+
+To make limited API builds opt-in, add
+``python.allow_limited_api=false`` to the ``default_options`` in your
+existing ``project()`` call in ``meson.build``:
+
+.. code-block:: meson
+
+   project(
+       'example',
+       'c',
+       default_options: ['python.allow_limited_api=false'],
+   )
+
+Keep ``limited-api = true`` in ``pyproject.toml`` and the ``limited_api``
+argument to ``extension_module()`` as shown above. When
+``python.allow_limited_api`` is false, Meson ignores the ``limited_api``
+argument, compiles extension modules for the ABI of the interpreter
+used for the build, and produces a wheel tagged as compatible with
+the build interpreter.
+
+To build a stable ABI wheel, explicitly enable limited API builds:
+
+.. tab-set::
+
+   .. tab-item:: pypa/build
+      :sync: key_pypa_build
+
+      .. code-block:: console
+
+         $ python -m build --wheel -Csetup-args="-Dpython.allow_limited_api=true" .
+
+   .. tab-item:: pip
+      :sync: key_pip
+
+      .. code-block:: console
+
+         $ python -m pip wheel -Csetup-args="-Dpython.allow_limited_api=true" .
+
+
+.. _limited-api-wheel-tags:
+
+Limited API compatibility
+-------------------------
+
+``meson-python`` conservatively chooses the wheel's Python compatibility
+tag based only on the interpreter used for the build, regardless of the
+target ``limited_api`` version.  For example, a wheel built with CPython
+3.12 for the 3.10 limited API is tagged ``cp312-abi3`` and can be
+installed only on GIL-enabled CPython 3.12 and later. When building
+stable ABI wheels for distribution, for example on PyPI, build with the
+oldest CPython version you want the wheel to support, as `recommended by
+CPython`_.  If you are installing from source into a particular
+environment, use that environment's interpreter; you do not need to
+build with the oldest supported version.
+
+PyPy supports the limited API, but does not implement a stable ABI, thus
+the :option:`tool.meson-python.limited-api` setting has no effect on the
+wheel tags when building with PyPy.
+
+
+The ``abi3t`` stable ABI
+------------------------
+
+CPython 3.15 introduces the ``abi3t`` stable ABI. Extension modules built
+for ``abi3t`` can be loaded by both the GIL-enabled and the free-threaded
+builds of CPython 3.15 and later: a single wheel tagged ``abi3.abi3t``
+supports all CPython interpreters from version 3.15 on.
+
+``abi3t`` extension modules require limited API version 3.15 or later
+and, currently, a free-threaded interpreter for the build: compiling
+for ``abi3t`` with a GIL-enabled interpreter is not supported yet.
+Once you have made the source changes described in the
+`abi3t migration guide`_, no additional build configuration is required:
+the CPython headers select ``abi3t`` when ``Py_LIMITED_API`` is defined
+while compiling for a free-threaded interpreter.
+
+Meson refuses a ``limited_api`` version newer than the interpreter
+used for the build.  If you build ``abi3`` wheels for older
+CPython versions and ``abi3.abi3t`` wheels for CPython 3.15 and later,
+you can select the limited API version by querying the ``Py_GIL_DISABLED``
+sysconfig variable, which is 1 for free-threaded builds:
+
+.. code-block:: meson
+
+   py = import('python').find_installation(pure: false)
+
+   limited_api = '3.10'
+   if py.language_version().version_compare('>=3.15')
+       if py.get_variable('Py_GIL_DISABLED') == 1
+           limited_api = '3.15'
+       endif
+   endif
+
+   py.extension_module(
+       '_core',
+       '_core.c',
+       limited_api: limited_api,
+       install: true,
+       subdir: 'example',
+   )
+
+Free-threaded CPython 3.13 and 3.14 do not support the limited API:
+disable limited API builds when using these interpreters.
+The :ref:`opt-in configuration above <opt-in-limited-api-example>`
+disables them by default.
+
+.. _limited API: https://docs.python.org/3/c-api/stable.html#limited-c-api
+.. _stable ABI: https://docs.python.org/3/c-api/stable.html#stable-application-binary-interface
+.. _recommended by CPython: https://docs.python.org/3/c-api/stable.html#limited-api-caveats
+.. _improve performance: https://docs.python.org/3/c-api/stable.html#limited-api-scope-and-performance
+.. _abi3t migration guide: https://docs.python.org/3.15/howto/abi3t-migration.html
--- a/docs/reference/pyproject-settings.rst
+++ b/docs/reference/pyproject-settings.rst
@@ -33,11 +33,12 @@ use them and examples.
    modules can be compiled for the Python limited API specifying the
    ``limited_api`` argument to the |extension_module()|_ function
    in the Meson Python module.  When this setting is set to true, the
-   value ``abi3`` is used for the Python wheel filename ABI tag.
-
-   This setting is automatically reverted to false when the
+   Python wheel filename ABI tag is set to ``abi3``, or ``abi3.abi3t``
+   when building with free-threaded CPython 3.15 or later.  This
+   setting is automatically reverted to false when the
    ``-Dpython.allow_limited_api=false`` option is passed to ``meson
-   setup``.
+   setup``. See :ref:`howto-limited-api` for more information on limited
+   API builds.
 
 .. option:: tool.meson-python.meson
 
@@ -91,7 +92,7 @@ use them and examples.
    glob pattern is useful exclusively to limit the effect of an exclude
    pattern that matches too many files.
 
-.. _python limited api: https://docs.python.org/3/c-api/stable.html?highlight=limited%20api#stable-application-binary-interface
+.. _python limited api: https://docs.python.org/3/c-api/stable.html#stable-application-binary-interface
 .. _extension_module(): `https://mesonbuild.com/Python-module.html#extension_module
 .. _meson introspection data: https://mesonbuild.com/IDE-integration.html#install-plan
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -83,6 +83,7 @@ the use of ``meson-python`` and Meson for Python packaging.
    how-to-guides/meson-args
    how-to-guides/debug-builds
    how-to-guides/shared-libraries
+   how-to-guides/limited-api
    reference/limitations
    projects-using-meson-python
 


docs/reference/pyproject-settings.rst
```diff
@@ -26,25 +26,70 @@ use them and examples.
...
+   CPython 3.15 introduces the ``abi3t`` stable ABI for the
+   free-threaded build, see :pep:`803`.  Extension modules targeting it
+   can be loaded by both the GIL-enabled and the free-threaded builds
+   of CPython 3.15 and later.  The CPython headers select ``abi3t``
+   when ``Py_LIMITED_API`` is defined while compiling for a
+   free-threaded interpreter, thus the ``limited_api`` argument is
+   sufficient to compile the extension modules for ``abi3t`` when
+   building with free-threaded CPython 3.15 or later. Building
+   ``abi3t`` extension modules with a GIL-enabled interpreter is not
+   supported, see `Meson issue 15637`_.
```
rgommers: Let's just say "not supported yet". We don't link issues from the docs, and there isn't much of interest to see anyway in that issue for the average user.

docs/reference/pyproject-settings.rst
```diff
@@ -26,25 +26,70 @@ use them and examples.
...
 
    A boolean indicating whether the extension modules contained in the
    Python package target the `Python limited API`_.  Extension
    modules can be compiled for the Python limited API specifying the
    ``limited_api`` argument to the |extension_module()|_ function
    in the Meson Python module.  When this setting is set to true, the
-   value ``abi3`` is used for the Python wheel filename ABI tag.
+   Python wheel filename ABI tag is set to the stable ABI tag:
+   ``abi3`` when building with GIL-enabled CPython, and ``abi3.abi3t``
+   when building with free-threaded CPython 3.15 or later, see below.
```
rgommers: We want to keep the reference guide content very short and to the point. The above three sentences are enough on this page I'd say. Then reference the detailed docs for more info. That's where the below content should move.

rgommers (changes requested):
Thanks @ngoldbaum for writing docs for the Limited API! Overall this content looks like a great start. My main comment right now is that it's not in the right places: it can't be in the reference guide, and the examples section of the "Passing arguments to Meson" page also feels wrong. I think we need a new page, as a how-to under "Using shared libraries" called something like _"Using the Limited C API"_ or some such title. 

This content, put together in a single page, will be a great start. I have a few things I'd like to add, like how to set it up as opt-in (e.g., as in https://github.com/PyWavelets/pywt/pull/828).

We can then also address a couple of open issues: gh-624, some parts of gh-787.

ngoldbaum:
I rearranged with a new howto guide page. Let me know what you think.

docs/reference/pyproject-settings.rst
```diff
@@ -26,25 +26,19 @@ use them and examples.
...
-   A boolean indicating whether :ref:`verbose mode
-   <how-to-guides-editable-installs-verbose>` should be enabled when building
-   an :ref:`editable install <how-to-guides-editable-installs>`. This setting
-   can be overridden via the config option of the same name.
-
 .. option:: tool.meson-python.limited-api
 
    A boolean indicating whether the extension modules contained in the
-   Python package target the `Python limited API`_.  Extension
-   modules can be compiled for the Python limited API specifying the
```
dnicolodi, line 40: This paragraph should be kept.

docs/reference/pyproject-settings.rst
```diff
@@ -26,25 +26,19 @@ use them and examples.
...
-   in the Meson Python module.  When this setting is set to true, the
-   value ``abi3`` is used for the Python wheel filename ABI tag.
-
-   This setting is automatically reverted to false when the
-   ``-Dpython.allow_limited_api=false`` option is passed to ``meson
-   setup``.
+   Python package target the `Python limited API`_.  Extension modules
+   can be compiled for the Python limited API specifying the
+   ``limited_api`` argument to the |extension_module()|_ function in the
+   Meson Python module.  When this setting is set to true, the Python
```
dnicolodi: Please avoid whitespace only changes.

docs/reference/pyproject-settings.rst
```diff
@@ -26,25 +26,19 @@ use them and examples.
...
+   wheel filename ABI tag is set to the stable ABI tag: ``abi3`` when
+   building with GIL-enabled CPython, and ``abi3.abi3t`` when building
+   with free-threaded CPython 3.15 or later. Free-threaded CPython before
+   3.15 does not support the limited API, so this setting must be
+   disabled at build time for those interpreter versions. See
+   :ref:`howto-limited-api` for more details and examples.
```
dnicolodi: I think this is too detailed: the description of options should be as concise as possible. 
```suggestion
   wheel filename ABI tag is set to ``abi3`` or ``abi3.abi3t`` when
   the build targets a Python interpreter supporting the free-threaded
   stable ABI. See :ref:`howto-limited-api` for details.
```

docs/how-to-guides/limited-api.rst
```diff
@@ -0,0 +1,118 @@
...
+Building against the limited API produces a wheel with a Python version
+compatibility tag derived from the interpreter used for the build, not
+from the ``limited_api`` version: a wheel built with CPython 3.12 for
+the 3.10 limited API is tagged ``cp312-abi3`` and can be installed only
+on CPython 3.12 and later.  Build with the oldest supported CPython
+version to produce a wheel that can be installed on all of them.
+``meson-python`` verifies that all the extension modules included in the
+wheel use the stable ABI filename suffix and fails the build
+otherwise. This setting has no effect on the wheel tag when building
+with PyPy.
```
dnicolodi:
```suggestion
otherwise. PyPy does not support the limited API thus this setting has
no effect when building with PyPy.
```

docs/how-to-guides/limited-api.rst
```diff
@@ -0,0 +1,118 @@
...
+build, see :pep:`803` and the `abi3t migration guide`_.  Extension
+modules targeting it can be loaded by both the GIL-enabled and the
+free-threaded builds of CPython 3.15 and later.  The CPython headers
+select ``abi3t`` when ``Py_LIMITED_API`` is defined while compiling for a
+free-threaded interpreter, thus the ``limited_api`` argument is
+sufficient to compile the extension modules for ``abi3t`` when building
+with free-threaded CPython 3.15 or later. Building ``abi3t`` extension
+modules with a GIL-enabled interpreter is not supported yet.
+
+``abi3t`` requires targeting limited API version 3.15 or later,
```
dnicolodi: This is now stated in three different ways. Please find a way to make the content easier to follow.

dnicolodi (changes requested):
Just a few quick remarks. I find the content in the new howto page hard to read, mostly because there are things repeated several times with different language and because the content is not presented at increasing level of detail but zooming in and out constantly. The other are minor comments. Please squash the commits into just one commit.

ngoldbaum:
@dnicolodi thanks! I've applied your suggestions. I also rearranged the howto page and I think it's clearer now.

docs/reference/pyproject-settings.rst
```diff
@@ -26,25 +26,16 @@ use them and examples.
...
-   A boolean indicating whether :ref:`verbose mode
-   <how-to-guides-editable-installs-verbose>` should be enabled when building
-   an :ref:`editable install <how-to-guides-editable-installs>`. This setting
-   can be overridden via the config option of the same name.
-
 .. option:: tool.meson-python.limited-api
 
    A boolean indicating whether the extension modules contained in the
    Python package target the `Python limited API`_.  Extension
    modules can be compiled for the Python limited API specifying the
```
dnicolodi, line 40: Please keep this sentence.

ngoldbaum: Done.

docs/how-to-guides/limited-api.rst
```diff
@@ -0,0 +1,125 @@
...
+       install: true,
+       subdir: 'example',
+   )
+
+.. code-block:: toml
+
+   [tool.meson-python]
+   limited-api = true
+
+Build with the minimum CPython interpreter version you wish to
```
rgommers: This should say "if you're building wheels to distribute, e.g. on PyPI". Otherwise, you build with whatever interpreter you're targeting - and then you'll get the `py3xx` tag matching that interpreter in the wheel filename (the text should explain that).

I'd actually make this a separate paragraph, because it deserves a longer explanation, and it's logically distinct from the verification part in the next sentence here.

docs/how-to-guides/limited-api.rst
```diff
@@ -0,0 +1,125 @@
...
+         $ python -m build --wheel -Csetup-args="-Dpython.allow_limited_api=false" .
+
+   .. tab-item:: pip
+      :sync: key_pip
+
+      .. code-block:: console
+
+         $ python -m pip wheel -Csetup-args="-Dpython.allow_limited_api=false" .
+
+To set this option only when building wheels for free-threaded CPython
```
rgommers: This is technically correct, but breaks installing from source on 3.14t, so it's probably not ideal to include this.

docs/how-to-guides/limited-api.rst
```diff
@@ -0,0 +1,125 @@
...
+
+   py.extension_module(
+       '_core',
+       '_core.c',
+       limited_api: limited_api,
+       install: true,
+       subdir: 'example',
+   )
+
+Disable the Limited API
```
rgommers: I would actually present this section as "Set up Limited API support as opt-in". And right at the top of this page, explain that there are two ways to set things up and that we recommend the "as opt-in" method, because (1) it's more performant when building from source and for distro packaging that has no need for stable ABI wheels, and (2) it's more portable than always targeting the limited API (e.g. 3.14t and alternative interpreters that don't support the stable ABI will work).

The order of explanation is fine as is, given that this opt-in setting is extra config on top of the basic "always use" flavor.

rgommers (commented):
This is looking pretty good. A couple more suggestions to give the desired recommendations to users.

docs/how-to-guides/limited-api.rst
ngoldbaum: This is now explained in its own paragraph.

docs/how-to-guides/limited-api.rst
ngoldbaum: I restructured so this section now talks about opt-in builds and why you'd want to do it that way.

docs/how-to-guides/limited-api.rst
ngoldbaum: I removed the cibuildwheel-specific example and the opt-in configuration automatically avoids this problem as well.

ngoldbaum:
Thanks for the comments Ralf! I think I addressed them.

docs/how-to-guides/limited-api.rst
```diff
@@ -0,0 +1,153 @@
...
+can support newer interpreter versions.
+
+You can enable limited API builds by default or make them opt-in.
+If your package does not yet use the limited API, we recommend making
+limited API builds opt-in so you can transition to stable ABI wheels
+gradually or distribute stable ABI wheels only on selected platforms.
+
+Compile your extension modules for the limited API by specifying the
+``limited_api`` argument to the ``extension_module()`` function in the
+Meson Python module.  Enabling the
```
dnicolodi: Here, I would explain what the value to be passed to the `limited_api` argument represents. Something vaguely on the lines of

```suggestion
Meson Python module.  The value passed to the ``limited_api`` argument
is used by Meson to derive the value passed to the `Py_LIMITED_API` define.
It corresponds to the version of the limited API the extension modules uses.

Enabling the
```

docs/how-to-guides/limited-api.rst
```diff
@@ -0,0 +1,153 @@
...
+   limited-api = true
+
+Building against the limited API produces a wheel with a Python version
+compatibility tag derived from the interpreter used for the build, not
+from the ``limited_api`` version: a wheel built with CPython 3.12 for
+the 3.10 limited API is tagged ``cp312-abi3`` and can be installed only
+on GIL-enabled CPython 3.12 and later. When building stable ABI wheels
+for distribution, for example on PyPI, build with the oldest CPython
+version you want the wheel to support. For a source installation in a
+particular environment, use that environment's interpreter.
```
dnicolodi:
```suggestion
version you want the wheel to support. This is not relevant when installing
from sources in particular environment. In this case the simplest is to build
with the same interpreter where the wheel will be installed.
```
Expanded a little to make it clearer (maybe).

docs/how-to-guides/limited-api.rst
```diff
@@ -0,0 +1,153 @@
...
+
+For limited API builds, ``meson-python`` verifies that all the extension
+modules included in the wheel use the stable ABI filename suffix and
+fails the build otherwise. PyPy does not support the limited API thus
+the :option:`tool.meson-python.limited-api` setting has no effect when
+building with PyPy.
```
dnicolodi: Some clarifications:
```suggestion
For limited API builds, on platforms where Python extension modules 
targeting the stable ABI use a dedicated filename suffix, ``meson-python``
verifies that all the extension modules included in the wheel use the
stable ABI filename suffix and fails the build otherwise.

PyPy supports the limited API, but does not implement a stable ABI, thus
the :option:`tool.meson-python.limited-api` setting has no effect on the
wheel tags when building with PyPy.
```

docs/how-to-guides/limited-api.rst
```diff
@@ -0,0 +1,153 @@
...
+      .. code-block:: console
+
+         $ python -m pip wheel -Csetup-args="-Dpython.allow_limited_api=true" .
+
+
+The ``abi3t`` stable ABI
+------------------------
+
+CPython 3.15 introduces the ``abi3t`` stable ABI, see :pep:`803` and
+the `abi3t migration guide`_.  Extension modules built for ``abi3t``
```
dnicolodi: I am not sure that linking to the PEP is relevant here. These documents tend to be way more technical than necessary and become outdated very quickly.
```suggestion
CPython 3.15 introduces the ``abi3t`` stable ABI, see
the `abi3t migration guide`_.  Extension modules built for ``abi3t``
```

dnicolodi (dismissed):
Thanks. This looks very nice. A few more comments.

ngoldbaum:
Thanks for all of the detailed suggestions! I think I've applied them. 

I also had a private chat with Ralf where he explained to me that the wheel interpreter compatibility tag being different from the target API is an intentional design choice in meson-python and not just an accident of how support for limited API builds is implemented. I now phrase that as meson-python "conservatively chooses" this to make that clear. I also split out all the compatibility details into their own subsection after the main build configuration discussion.

Now whenever people have questions about that point in the future, you can point to a specific docs page about it!

I also marked the issue as explicitly fixing #624 and updated the PR description to reflect the final state.

docs/how-to-guides/limited-api.rst
```diff
@@ -0,0 +1,174 @@
...
+CPython 3.15 introduces the ``abi3t`` stable ABI, see
+the `abi3t migration guide`_.  Extension modules built for ``abi3t``
+can be loaded by both the GIL-enabled and the free-threaded builds of
+CPython 3.15 and later: a single wheel tagged ``abi3.abi3t`` supports
+all CPython interpreters from version 3.15 on.
+
+``abi3t`` extension modules require limited API version 3.15 or later
+and, currently, a free-threaded interpreter for the build: compiling
+for ``abi3t`` with a GIL-enabled interpreter is not supported yet.
+Nothing else is required: the CPython headers select ``abi3t`` when
```
rgommers: I'd say "Nothing else is required for configuring the build to target ``abi3t``, assuming you've already made the source-level changes that are required". With source-level changes a link that links to the appropriate CPython docs page, if there is such a thing. If not, let's mentioning opaque types and PyModExport in brackets?

docs/how-to-guides/limited-api.rst
```diff
@@ -0,0 +1,174 @@
...
+
+This page describes how to configure your package to build against the
+CPython `limited API`_ and build `stable ABI`_ wheels. Limited API builds
+target a minimum interpreter version and produce a single wheel that
+can support newer interpreter versions.
+
+You can enable limited API builds by default or make them opt-in.
+If your package does not yet use the limited API, we recommend making
+limited API builds opt-in so you can transition to stable ABI wheels
+gradually or distribute stable ABI wheels only on selected platforms.
```
rgommers: I don't think the rationales are right here. I'd remove both. What I would add instead is:

1. Avoid failures or special handling for interpreters that do not support the Stable ABI (e.g., free-threading on 3.14, PyPy, GraalPy)
2. "and keep the performance gains from targeting the Python version-specific C API for users building from source, as well as for distros that do not need the stable ABI".

rgommers (commented):
Thanks Nathan. Very close, two more comments from me.

docs/how-to-guides/limited-api.rst
ngoldbaum: 👍 I moved the abi3t migration guide link down here

docs/how-to-guides/limited-api.rst
ngoldbaum: I didn't use your exact phrasing but I think the gist is there now. I only talk about 3.14t builds because right below this we talk about how pypy limited API builds do actually work and I thought that might be confusing.

rgommers: Yes, that LGTM now, resolving.

rgommers (approved):
This looks great to me now. I checked and resolved all open comments. I've already been pointing people at these docs, so I'll merge this in a few days unless there are more comments.

rgommers:
Thanks again @ngoldbaum, and thanks @dnicolodi for the reviews.

PR #886, opened 2026-09-02
MAINT: ignore the wraplock files from meson

henryiii:
This files have started showing up from meson. I think we should ignore them.


rgommers (approved):
Agreed, I noticed them too recently. Thanks @henryiii. 

I'll update to our commit convention (chore -> MAINT) and merge.

PR #887, opened 2026-09-02
BUG: use the correct encoding to read the compilation log

dnicolodi:
This is a follow-up to the recently-merged gh-859, and in particular the `encoding` comments starting at https://github.com/mesonbuild/meson-python/pull/859#issuecomment-5400960449.


rgommers (approved):
This LGTM, and CI is all green - so in it goes. Thanks @dnicolodi!

PR #892, opened 2026-09-10
BUG: support interpreters not implementing PEP 3149

### LINKED ISSUE #890: 0.21.0: problem with ABI tag handling (regression)
With 0.21.0, I am not able to build [pycairo](https://pycairo.readthedocs.io/en/latest/) nor numpy. I get

```
 File "/usr/pkg/lib/python3.14/site-packages/mesonpy/_tags.py", line 42, in get_abi_tag
   empty, abi, ext = str(sysconfig.get_config_var('EXT_SUFFIX')).split('.')
   ^^^^^^^^^^^^^^^
ValueError: not enough values to unpack (expected 3, got 2)
```

I had to partially revert changes in mesonpy/_tags.py from 0.20.0 to make them build again. Here's my patch:

```
--- mesonpy/_tags.py.orig	2026-09-08 19:43:03.139256199 +0000
+++ mesonpy/_tags.py
@@ -31,6 +31,11 @@ def get_interpreter_tag() -> str:
     return f'{name}{version[0]}{version[1]}'
 
 
+def _get_cpython_abi() -> str:
+    version = sys.version_info
+    return f'cp{version[0]}{version[1]}'
+
+
 def get_abi_tag() -> str:
     # The best solution to obtain the Python ABI is to parse the
     # $SOABI or $EXT_SUFFIX sysconfig variables as defined in PEP-314.
@@ -39,7 +44,12 @@ def get_abi_tag() -> str:
     # Using $EXT_SUFFIX will not break when PyPy will fix this.
     # See https://foss.heptapod.net/pypy/pypy/-/issues/3816 and
     # https://github.com/pypa/packaging/pull/607.
-    empty, abi, ext = str(sysconfig.get_config_var('EXT_SUFFIX')).split('.')
+    try:
+        empty, abi, ext = str(sysconfig.get_config_var('EXT_SUFFIX')).split('.')
+    except ValueError as exc:
+        if sys.implementation.name != 'cpython':
+            raise NotImplementedError from exc
+        return _get_cpython_abi()
 
     # The packaging module initially based his understanding of the
     # $SOABI variable on the inconsistent value reported by PyPy, and
```

dnicolodi:
Some Python distributions (notably pkgsrc, the NetBSD default package manager) patch the Python configure script to remove $SOABI from $EXT_SUFFIX.

Fixes #890.

### DIFF
--- a/mesonpy/_tags.py
+++ b/mesonpy/_tags.py
@@ -31,6 +31,26 @@ def get_interpreter_tag() -> str:
     return f'{name}{version[0]}{version[1]}'
 
 
+def _get_config_var(name: str, default: str | int | None = None) -> str | int | None:
+    value: str | int | None = sysconfig.get_config_var(name)
+    if value is None:
+        return default
+    return value
+
+
+def _get_cpython_abi() -> str:
+    # Fallback for Python distributions not implementing PEP 3149.
+    if sys.implementation.name != 'cpython':
+        raise NotImplementedError
+    version = sys.version_info
+    debug = threading = ''
+    if _get_config_var('Py_DEBUG', hasattr(sys, 'gettotalrefcount')):
+        debug = 'd'
+    if version >= (3, 13) and _get_config_var('Py_GIL_DISABLED'):
+        threading = 't'
+    return f'cp{version[0]}{version[1]}{threading}{debug}'
+
+
 def get_abi_tag() -> str:
     # The best solution to obtain the Python ABI is to parse the
     # $SOABI or $EXT_SUFFIX sysconfig variables as defined in PEP-314.
@@ -39,7 +59,13 @@ def get_abi_tag() -> str:
     # Using $EXT_SUFFIX will not break when PyPy will fix this.
     # See https://foss.heptapod.net/pypy/pypy/-/issues/3816 and
     # https://github.com/pypa/packaging/pull/607.
-    empty, abi, ext = str(sysconfig.get_config_var('EXT_SUFFIX')).split('.')
+    parts = str(sysconfig.get_config_var('EXT_SUFFIX')).split('.')
+
+    # Some Python distributions (notably pkgsrc, the default NetBSD
+    # package manager) patch the Python configure script to remove
+    # $SOABI from $EXT_SUFFIX.
+    if len(parts) < 3:
+        return _get_cpython_abi()
 
     # The packaging module initially based his understanding of the
     # $SOABI variable on the inconsistent value reported by PyPy, and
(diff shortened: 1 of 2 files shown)


dnicolodi:
@rgommers We should probably get a 0.21.1 release out with this fix soon.

rgommers:
> We should probably get a 0.21.1 release out with this fix soon.

Yes, good idea - thanks for this PR, reviewing now. Happy to get a 0.21.1 out asap.

rgommers (commented):
Thanks @dnicolodi. I did some searching, and PkgSrc seems to be the one holdout regarding PEP 3149 implementation. PEP 3149 hedged a bit though: implementation is optional. Hence what PkgSrc does can be argued to be valid.

We should probably have a test case for this, at least if we can manage a small unit test that monkeypatches so `SOABI` is `'cpython-314'` while `EXT_SUFFIX` is `'.so'` (matching PkgSrc).

The fallback in this PR works as written I believe. It's a little verbose though; given that `SOABI` itself should still contain everything we need, it looks to me like this could be a simpler fallback:
```python
        abi = sysconfig.get_config_var('SOABI')
        if not isinstance(abi, str):
            raise NotImplementedError
    else:
        abi = parts[1]
```
WDYT?

dnicolodi:
> The fallback in this PR works as written I believe. It's a little verbose though; given that `SOABI` itself should still contain everything we need, it looks to me like this could be a simpler fallback:

I thought about this possibility, but I was not very motivated to investigate whether there is any Python distribution out there that does something strange with `SOABI`. 

The main reason to write the fallback like this is that `packaging` does something very similar, and it was working before I decided to drop what I thought was only support for old Python versions. On the other hand, maybe `packaging` uses the verbose fallback because it wants to support (although I think it does not explicitly) support Python versions where `SOABI` is not reliable.

I have a slight preference for the code as written in this PR. But I am not opposed to fallback to `SOABI`.

Adding a test is a good idea, though.

rgommers:
> I have a slight preference for the code as written in this PR. But I am not opposed to fallback to `SOABI`.

Okay, fine with me to keep it as is. One is not clearly better than the other, and this code is already written plus if `packaging` does it like that it's a good insurance that it will work portably.

dnicolodi:
Test added.

rgommers:
Looks like the new test needs skipping on non-CPython interpreters, given the PyPy failures and that it tests a CPython-specific code path.

dnicolodi:
Fixed

rgommers (approved):
LGTM now, thanks @dnicolodi

rgommers:
I'll open a PR for 0.21.1 in a minute

PR #894, opened 2026-09-11
MAINT: add 'dev' dependency group, add uv-sync CI check

mwtoews:
This PR helps development with [uv](https://docs.astral.sh/uv/). There is an issue with [uv sync](https://docs.astral.sh/uv/concepts/projects/sync/) in this project:

```
$ uv sync
Using CPython 3.14.2
Creating virtual environment at: .venv
  × No solution found when resolving dependencies for split (markers: python_full_version == '3.9.*' and sys_platform != 'ios'):
  ╰─▶ Because the requested Python version (>=3.9) does not satisfy Python>=3.10 and sphinx>=8.1.0,<=8.1.3 depends on Python>=3.10, we can conclude that
      sphinx>=8.1.0,<=8.1.3 cannot be used.
      And because only the following versions of sphinx are available:
          sphinx<=8.1.0
          sphinx==8.1.1
          sphinx==8.1.2
          sphinx==8.1.3
          sphinx>8.2.dev0
      we can conclude that sphinx>=8.1.0,<8.2.dev0 cannot be used.
      And because meson-python:docs depends on sphinx>=8.1.0,<8.2.dev0 and your project requires meson-python:docs, we can conclude that your project's requirements
      are unsatisfiable.

hint: While the active Python version is 3.14, the resolution failed for other Python versions supported by your project. Consider limiting your project's supported Python versions using `requires-python`.
hint: The `requires-python` value (>=3.9) includes Python versions that are not supported by your dependencies (e.g., sphinx>=8.1.0,<=8.1.3 only supports >=3.10). Consider using a more restrictive `requires-python` value (like >=3.10).
```
The incompatibility with [`sphinx ~= 8.1.0`](https://pypi.org/project/Sphinx/8.1.3/) is resolved by requiring this dependency to require Python >= 3.10. (An alternative fix is to drop Python 3.9 from this project, but I'm not here to suggest that right now). This means that an older version of sphinx is selected for Python 3.9 users that want to build docs (probably not anyone).

This PR also adds a "dev" dependency group, with pre-commit and including everything in the "test" group (but not "docs"). Uv [automatically includes "dev"](https://docs.astral.sh/uv/concepts/projects/sync/#syncing-development-dependencies), which is handy when running `uv sync`. Or with a venv, `pip install --group dev` will install all the development dependencies.

Lastly, ignore all `*.lock` files (e.g. `uv.lock`, `pixi.lock`, and others).


rgommers (commented):
Thanks @mwtoews. First impression: this looks good. Two thoughts:

- We can actually drop Python 3.9 I think, the EOL date for it was October 31, 2025. And we dropped Python 3.8 about a year ago, in gh-799. @dnicolodi do you agree?
- We may want to exercise this in CI to ensure it stays fixed. We don't have to commit the lock file, but just a job that runs `uv sync` would catch this.

Both could be done as follow-ups, this seems mergeable as is.

pyproject.toml
```diff
@@ -59,12 +59,15 @@ test = [
...
+  'sphinx ~= 8.1.0; python_version >= "3.10"',
   'sphinx-copybutton >= 0.5.0',
   'sphinx-design >= 0.1.0',
   'sphinxext-opengraph >= 0.7.0',
 ]
-
+dev = [
+  "pre-commit",
+  {include-group = "test"},
+]
```
dnicolodi, line 70: Why this addition? I don't have `pre-commit` installed in the environment where I work on meson-python and I am happy like this.

pyproject.toml
```diff
@@ -59,12 +59,15 @@ test = [
 ]
 docs = [
   'furo >= 2024.08.06',
-  'sphinx ~= 8.1.0',
+  'sphinx ~= 8.1.0; python_version >= "3.10"',
```
dnicolodi, line 62: If I read it correctly, this means that `sphinx` is required only when the environment has Python 3.10 or later, and it is not required with older Python versions. This may solve the `uv` resolver issue, but it does not make sense to me: we need sphinx to build the documentation.

dnicolodi (changes requested):

pyproject.toml
rgommers: That's a good point, I think an extra group is cleaner, rather than auto-installing it. I also avoid `pre-commit`, I don't trust it on my machine. That's a fairly common sentiment around here, so having it in a dependency group that is opt-in would be better.

pyproject.toml
rgommers: I'm not sure that that matters at all, since no one will actually develop on 3.9 anymore. And we do require that `sphinx` version, it doesn't help to require an older version that we don't test and might not build the docs or misrender it.

That said, if it bothers you, WDYT about dropping 3.9 now to avoid the problem altogether?

pyproject.toml
dnicolodi: What about not having `pre-commit` mentioned anywhere? I think having a `dev` dependency group as an alias for `test` is a good thing. I should have named it `dev` from the beginning but I don't think `uv` automatically installed the `dev` group back then, or I was not aware of it. We could just rename `test` to `dev`, but I don't want to break anyone's setup.

dnicolodi:
> * We can actually drop Python 3.9 I think, the EOL date for it was October 31, 2025. And we dropped Python 3.8 about a year ago, in [gh-799](https://github.com/mesonbuild/meson-python/pull/799). @dnicolodi do you agree?

Meson 1.12 requires Python 3.10. I think we can just drop support for Python 3.9. This would allow to reduce a tiny bit the number of CI jobs we need to run, and to use a more modern version of mypy. The only obstacle I see is our Cygwin CI job using Python 3.9. Until not very long ago, that was the latest Python version available on Cygwin.

dnicolodi:
> Cygwin CI job using Python 3.9. Until not very long ago, that was the latest Python version available on Cygwin.

I just checked: in the time between Cygwin added Python 3.12 packages.

pyproject.toml
dnicolodi: I was not proposing to require an older version for Python 3.9 and earlier. However, the PR description says that the effect of the change is that an older `sphinx` is installed on Python 3.9, and this is not what I believe the patch does.

I am fine dropping Python 3.9 support.

rgommers: Ah okay, so code is fine but you were commenting on this sentence:

> This means that an older version of sphinx is selected for Python 3.9 users that want to build docs (probably not anyone).

That may or may not be correct (probably it isn't), but it's unclear because _some_ version of `sphinx` will be pulled in as a transitive dependency from those other packages.

> I am fine dropping Python 3.9 support.

Okay, let's go that way. Let's do a separate PR and land that first (easier to find back later on, and I look for such PRs on occasion), and then simplify this. I can make the PR, unless one of you wants to.

dnicolodi: I'm working on the Python required version bump.

pyproject.toml
rgommers: > I think having a `dev` dependency group as an alias for `test` is a good thing. I should have named it `dev` from the beginning but I don't think `uv` automatically installed the `dev` group back then, or I was not aware of it. We could just rename `test` to `dev`, but I don't want to break anyone's setup.

SGTM

> What about not having `pre-commit` mentioned anywhere?

I think it's a fair ask, given we use it in CI and quite a few contributors will prefer local usage over iterating in CI. And the 3 hooks we use are from trust-worthy sources.

That said, I'm not a user so I don't actually know if it helps to have it in a separate dependency-group. Is there an advantage over a separate `pre-commit` group you'd need to know about vs. not having that group, @mwtoews?

pyproject.toml
rgommers: Version bump is in (gh-896), so this change can be dropped.

pyproject.toml
mwtoews: Another note on uv is that it can specify [default groups](https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups), e.g.
```toml
[tool.uv]
default-groups = ['test']
```
but I'm going to agree to add "dev" as an alias to "test" since is a somewhat universal group name (i.e. uv has `--dev` and `--no-dev` flags exclusive to that group name).

As for pre-commit, I added this to `dev` because I saw the presence of `pre-commit-config.yaml`, but it can be removed. There is probably no value adding a separate dependency group named `pre-commit` containing only `['pre-commit']`. Users can always `[uv] pip install pre-commit` as needed, maybe this can be a comment in case someone needs it.

pyproject.toml
mwtoews: Thanks! This will resolve much of this PRs purpose.

mwtoews:
> * We may want to exercise this in CI to ensure it stays fixed. We don't have to commit the lock file, but just a job that runs `uv sync` would catch this.

This is a good idea, and would be worthwhile including in this PR while I'm at it. It would effectively do this:
```shell
$ uv sync --group test --dry-run
Using CPython 3.14.2
Would create project environment at: .venv
Resolved 49 packages in 47ms
Would create lockfile at: uv.lock
Would download 1 package
Would install 15 packages
 + build==1.6.1
 + coverage==7.16.0
 + cython==3.3.0
 + iniconfig==2.3.0
 + meson==1.12.0
 + meson-python @ file:///tmp/meson-python
 + packaging==26.3
 + pluggy==1.6.0
 + pygments==2.21.0
 + pyproject-hooks==1.2.0
 + pyproject-metadata==0.12.1
 + pytest==9.1.1
 + pytest-cov==7.1.0
 + pytest-mock==3.15.1
 + wheel==0.48.0
```

pyproject.toml
rgommers: Okay, sounds like we're all on the same page, and the current solution with just a code comment is helpful - documents we've thought about it and prevents the next contributor for trying to add it again.

I'll resolve this thread.

rgommers (changes requested):
Two requests:

- Can you add the `uv sync` test at the bottom of `tests.yml` please? No need for a separate file here, nor a cron job.
- Can you rewrite the commit history so it's one commit per logical change? We prefer using rebase-merging and aim to keep a clean linear history.

.github/workflows/tests.yml
```diff
@@ -343,3 +343,20 @@ jobs:
...
+        with:
+          persist-credentials: false
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@bec219d24cd3e171d82865faccec33120bb574f4 # v10.1.0
+        with:
+          enable-cache: false
+          restore-cache: false
+          save-cache: false
+          version: "latest"
```
dnicolodi: I'd be happier if we pin a specific version.

pyproject.toml
```diff
@@ -64,7 +64,10 @@ docs = [
   'sphinx-design >= 0.1.0',
   'sphinxext-opengraph >= 0.7.0',
 ]
-
+dev = [
+  # 'pre-commit' not included here -- use 'pip install pre-commit' if needed
+  {include-group = 'test'},
+]
```
dnicolodi, line 70:
```suggestion
dev = [
  # 'pre-commit' not included here -- use 'pip install pre-commit' if needed
  {include-group = 'test'},
]

```
Preserve the double empty line between the package metadata and the tools configuration.

pyproject.toml
```diff
@@ -64,7 +64,10 @@ docs = [
   'sphinx-design >= 0.1.0',
   'sphinxext-opengraph >= 0.7.0',
 ]
-
+dev = [
+  # 'pre-commit' not included here -- use 'pip install pre-commit' if needed
+  {include-group = 'test'},
```
dnicolodi, line 69: I find this comment tautological: it is evident that `pre-commit` is not included and `pip install` is the obvious way to install it, if desired. Either explain why we do not include `pre-commit` or leave the comment out. I don't want to not even hint to the fact that the project requires or endorses the use of `pre-commit`.

On the other hand, I would however explain why we have an alias for the `test` dependency group:

```suggestion
# uv includes by default the dev dependency group in the environment
dev = [
    {include-group = 'test'},
```

.github/workflows/tests.yml
```diff
@@ -343,3 +343,20 @@ jobs:
...
+
+      - run: uv sync --all-extras --all-groups --dry-run
```
dnicolodi, line 361: I am soon going to forget what this is a about. Please help me out 🙂 
```suggestion
      # Ensure that dependencies and dependency groups are consistent and resolvable.
      - run: uv sync --all-extras --all-groups --dry-run
```

dnicolodi (changes requested):
Some more comments from my side.

I am also wondering: should we switch to `uv` for all jobs? That would exercise the dependencies installation without having to add a dedicated job. We can do this at a later time.

rgommers:
> I am also wondering: should we switch to `uv` for all jobs? That would exercise the dependencies installation without having to add a dedicated job. We can do this at a later time.

I'd do that later, if we do it. We can't just add `uv` in front of `pip install`, that doesn't help (it won't do locking) - so a separate `uv sync` check seems easier.

.github/workflows/tests.yml
mwtoews: I'll take [0.11.33](https://github.com/astral-sh/uv/releases?page=2#release-0.11.33) from 2026-07-28 because it is the last of the 0.11.x series ("latest" is currently [0.12.13](https://github.com/astral-sh/uv/releases/tag/0.12.13))

pyproject.toml
mwtoews: Sure, I can replace these comments. Perhaps a small comment in `.pre-commit-config.yaml` would be more appropriate. Edits to me wording to this comment are welcome. Or left untouched, I don't mind.

.github/workflows/tests.yml
dnicolodi: Any reason not to use 0.12.13?

mwtoews:
Note that the `uv-sync` workflow takes 5 to 8 seconds to complete, which is pretty good. And yes, `uv pip install` would *not* catch the incompatible package/python versions as initially reported in this PR.

As for `uv` for all other workflows, sure go for it! But keep a few legacy `pip` workflows for sane coverage.

.github/workflows/tests.yml
mwtoews: It could, it's somewhat arbitrary (unless anyone has been closely following `uv` development of micro releases). My thought is that the 0.11.x series is the most mature of the available releases, in a project that hasn't reached 1.x yet. But I do see some enhancements to `uv sync` in the 0.12.x series, so updating to current "latest".

dnicolodi:
Thanks @mwtoews. Merging.

PR #898, opened 2026-09-11
MAINT: use less annoying indentation for function arguments

dnicolodi:
This has been bothering me for a very long time


rgommers (approved):
Sure, CI is happy so in it goes. I don't feel as strongly about it, but I'd write it in this new form as well.

PR #899, opened 2026-09-13
MAINT: fix no-op str.replace() in editable path hook

### LINKED ISSUE #868: No-op `str.replace` in the editable path hook

dnicolodi:
To the best of my knowledge, this is inconsequential. The path passed to the path hook is used only to check whether it is handled by the editable wheel import hook. The only paths that match this criteria come from the __path__ attribute of a module loaded via the editable package import hook and these are computed with the correct path separator.

However, it is hard to rule out anyone doing something funny with these paths, thus be on the safe side and normalize the path separators.

Fixes #868.


rgommers (approved):
LGTM, thanks @dnicolodi

PR #900, opened 2026-09-13
BUG: detect and ignore cyclic symbolic link chains in sdist

### LINKED ISSUE #870: Possible issue with symlink cycles in sdist creation
Symlink cycles hang `sdist()`. The `while member.issym():` resolution loop never terminates if the archive contains a symlink cycle (a -> b, b -> a). Malformed input, but a bounded loop or visited-set would turn a hang into a warning like the other symlink cases.

Originally reported in #861.

Hitting this in real projects is extremely unlikely. I am not even sure whether `git archive` allows to create such symlink cycles. Anyhow, just to be on the safe side we should add protection for this, but given the impact it is very low priority.

dnicolodi:
Fixes #870.


mesonpy/__init__.py
```diff
@@ -1058,8 +1069,15 @@ def sdist(self, directory: Path) -> pathlib.Path:
                 # This makes symbolic links support unusable in most cases,
                 # therefore include the symbolic link targets as regular files
                 # in all cases.
+                visited = set()
                 while member.issym():
                     name = member.name
+                    # Detect symbolic link chains resulting in a cycle.
+                    if name in visited:
+                        warnings.warn(
+                            'symbolic link resultin in a cycle ignored: {name}', stacklevel=1)
```
rgommers: This misses the `f` prefix for an f-string.

rgommers (changes requested):
Thanks @dnicolodi. This fix seems incomplete: it drops all symlink chains now, rather than only cycles. Try adding a `ccc.py` symlink to the test case, so there's a chain with length >1 (`ccc -> bbb -> aaa`). Then `ccc.py` goes missing from the sdist.

Actual file targets that symlinks point to probably need keeping track of separately.

dnicolodi:
Thanks for the review @rgommers. I also found another subtle bug for which symbolic link chains crossing directories where not resolved correctly. All problems should be fixed now. I added one more test file to make sure.

mesonpy/__init__.py
```diff
@@ -1052,24 +1063,31 @@ def sdist(self, directory: Path) -> pathlib.Path:
 
         with tarfile.open(meson_dist_path, 'r:gz') as meson_dist, mesonpy._util.create_targz(sdist_path) as sdist:
             for member in meson_dist.getmembers():
+                # Record the original member name.  The symbolic link
+                # resolution loop will make ``member`` point to the link
+                # target, but it needs to be archived under the original name.
+                # Symbolic ink target resolution must be relative to
```
rgommers: typo here: ink -> link

rgommers (approved):
LGTM now, thanks @dnicolodi. There's one typo you may want to fix before merging, other than that it's all happy it seems.

dnicolodi:
Thanks for the review @rgommers. Typo fixed. Merging.

PR #904, opened 2026-09-19
DOC: require pygments 2.20.0 or later

dnicolodi:
To get support for *, /, and % operators in Meson code. See https://github.com/pygments/pygments/issues/2918. Remove relative work-around.


rgommers (approved):
LGTM, thanks @dnicolodi

